PR #232, opened 2022-03-23
Add tags tests

rtorrero:
In this PR:
 - Improved a bit the return codes/body in the tags endpoints.
 - Added basic tests for tags addition/removal in cluster/host/sap_system controllers
 - Added basic tests for listing in cluster/host/sap_system controllers


fabriziosestito (approved):
:shipit:

PR #240, opened 2022-03-25
Sap system detail UI

nelsonkopliku:
And some Chores.

image

image


dottorblaster (approved):
LGTM

arbulu89 (approved):
Nice!
I would split the feature components as we do in the overview lists, for consistency.

nelsonkopliku:
> Nice! I would split the feature components as we do in the overview lists, for consistency. And use an icon in the health column instead of a text

You're right. I am just too lazy to do it in this PR :stuck_out_tongue: 
Gonna do that next :pray:

PR #241, opened 2022-03-25
Add checks catalog view

arbulu89:
Checks catalog view.

I have changed the catalog json struct. Have a look please. If you like it, I would change the runner to create that struct.
Now, the checks catalog can be updated based on the provider. The data is loaded at initialization, so we wouldn't have real updates if the catalog is changed in the runner.

**I didn't update the package-lock.json file, as I think I have a different npm version, and it changes the whole file...***
**This is just a frontend work. It doesn't do anything in the backend. We will discuss about this afterwards**.
catalog

### DIFF
--- a/assets/js/components/ChecksCatalog/ChecksCatalog.jsx
+++ b/assets/js/components/ChecksCatalog/ChecksCatalog.jsx
@@ -70,7 +70,7 @@ const ChecksCatalog = () => {
                         <Disclosure.Panel className="border-none">
                           <div className="px-8 py-4 sm:px-8">
                             <div className="px-4 py-4 sm:px-4 bg-slate-100 rounded">
-                              <ReactMarkdown remarkPlugins={[remarkGfm]} children={check.remediation} />
+                              <ReactMarkdown className="markdown" remarkPlugins={[remarkGfm]} children={check.remediation} />
                             </div>
                           </div>
                         </Disclosure.Panel>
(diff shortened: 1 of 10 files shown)


stefanotorresi:
@arbulu89 would it be possible to preserve the code blocks around the parameters like it was before?

arbulu89:
> @arbulu89 would it be possible to preserve the code blocks around the parameters in the description, like it was before?

Do you mean in the main description text? Yes, of course

assets/js/components/ChecksCatalog/ChecksCatalog.jsx
```diff
@@ -0,0 +1,105 @@
...
+  useEffect(() => {
+    setSelected(providers[0]);
+  }, [providers[0]]);
+
+  return (
+    <div>
+      <ProviderSelection
+        providers={providers}
+        selected={selected}
+        setSelected={setSelected}
```
dottorblaster, line 25: call this prop `onChange` :wink:

dottorblaster (approved):
LGTM, if you want to address the tiny comment I made... otherwise it's alright :D

assets/js/components/ChecksCatalog/ChecksCatalog.jsx
arbulu89: Thanks!
Done: https://github.com/trento-project/web/pull/241/commits/2204511c4bde44584b72ad4dcd3afbf4d7fcda1d

nelsonkopliku (approved):

PR #243, opened 2022-03-25
Serialize events in json and add DSL

fabriziosestito:
This PR introduces `deftype`, `defcommand` and `defevent` macros/DSL to build, validate and serialize data. 
It changes the serialization of events in the event store from Erlang Term to jsonb

Rationale:
Erlang [External Term Format ](https://www.erlang.org/doc/apps/erts/erl_ext_dist.html) is not human readable, hard to debug, can't be queried and it is not guaranteed to be stable between erlang versions.
On the contrary the `jsonb` data type can be exported in plain json and queried in PostgreSQL natively. 

Ecto solves the problem of validating and mapping data to structs, in other words we can use it  to validate commands, external data (think of json payloads form the API), events serialization/deserialization and so on.

Ecto is a toolkit which contains both data mapping goodies and persistence.
 [It is a common misconception to think Ecto schemas map only to your database tables.](https://hexdocs.pm/ecto/data-mapping-and-validation.html)
By leveraging ecto we are able to produce three macros:
- `defevent` to define an event
- `defcommand` to define a command
- `deftype` to define a generic data type 

Here's an example of defining an event:
```elixir
defmodule Trento.Domain.Events.ApplicationInstanceRegistered do
  @moduledoc """
  This event is emitted when a database application is registered to the SAP system.
  """

  use Trento.Event

  defevent do
    field :sap_system_id, Ecto.UUID
    field :sid, :string
    field :host_id, Ecto.UUID
    field :instance_number, :string
    field :instance_hostname, :string
    field :features, :string
    field :http_port, :integer
    field :https_port, :integer
    field :start_priority, :string
    field :health, Ecto.Enum, values: [:passing, :warning, :critical, :unknown]
  end
end
```

Under the hood, we use an ecto `embedded_schema` to cast and validate incoming data from maps. Therefore libraries such us `vex`, `domo` and `typedstruct` are not needed anymore and they are removed in this PR.

TODO:
some serialization/deserialization test


nelsonkopliku (approved):

arbulu89 (approved):
Didn't review the whole code, but the new logic is very clever. I love it!

fabriziosestito:
> Didn't review the whole code, but the new logic is very clever. I love it!

Thanks, a TL;DR of the PR could  be this commit: https://github.com/trento-project/web/pull/243/commits/4aacec58c9d147adbaafcf356d3da08f46265f6a and this commit: https://github.com/trento-project/web/pull/243/commits/2db117826cf8e634c852b763ad4e02cb4b2f4fb0

since the rest of the PR is basically just modifying the existing events and commands with the new syntax

dottorblaster (approved):
`defevent` FTW

lib/trento/infrastructure/router.ex
```diff
@@ -26,7 +23,7 @@ defmodule Trento.Router do
   }
 
   middleware Enrich
-  middleware Validate
+  # middleware Validate
```
dottorblaster, line 26: Do we wanna remove this?

fabriziosestito: yes we want

fabriziosestito: there is no need to validate since the structs are already validated

PR #250, opened 2022-03-28
Flavor determination in about api

nelsonkopliku:


fabriziosestito:
Mmmh I am not sure about this, we could use atoms instead e.g. :community / :premium and have a function for displaying those in the about page

arbulu89 (approved):

lib/trento_web/controllers/about_controller.ex
```diff
@@ -4,8 +4,8 @@ defmodule TrentoWeb.AboutController do
   alias Trento.Hosts
 
   @version Mix.Project.config()[:version]
-  # TODO determine Flavor
-  @flavor "Community"
+
+  @flavor Trento.Installation.flavor()
```
fabriziosestito, line 8: it is better not to use a module attribute here but just the function call in the info function below

fabriziosestito: module attributes are evaluated at compile time, so if you change the `Trento.Installation` module this will recompile

test/trento/application/integration/telemetry/telemetry_test.exs
```diff
@@ -6,16 +6,17 @@ defmodule Trento.Integration.TelemetryTest do
...
 
   alias Trento.Installation
+  alias Trento.Installation.Flavor
   alias Trento.Integration.Telemetry
 
   setup :verify_on_exit!
 
   setup do
-    Application.put_env(:trento, :flavor, "Premium")
+    Application.put_env(:trento, :flavor, Flavor.premium())
```
fabriziosestito, line 15: why not just an atom :) ?

dottorblaster: Strong upvote, we could just set an atom inside the config

dottorblaster (commented):
Also keep in mind every time you write code that reads from the config to update the `config/{dev,prod,test}.exs` :metal:

lib/trento_web/controllers/about_controller.ex
nelsonkopliku: Good to know, thanks!

test/trento/application/integration/telemetry/telemetry_test.exs
nelsonkopliku: I switched to a really simpler version, keeping everything as it was.
There is some concerns from @fabriziosestito about the possibility to use atoms.

lib/trento_web/controllers/about_controller.ex
```diff
@@ -3,15 +3,11 @@ defmodule TrentoWeb.AboutController do
...
-  # TODO determine Flavor
-  @flavor "Community"
-
   @spec info(Plug.Conn.t(), map) :: Plug.Conn.t()
   def info(conn, _) do
     json(conn, %{
-      flavor: @flavor,
-      version: @version,
+      flavor: Trento.Installation.flavor(),
+      version: Mix.Project.config()[:version],
```
fabriziosestito, line 10: this needs to stay as a module attribute since the release does not have Mix, and we need to inject the version at compile time

fabriziosestito (changes requested):

nelsonkopliku:
Thanks for the feedbacks guys!

I switched to a much simpler version of this PR just by making `Trento.Installation.flavor/0` public and using it in the about controller.

What I tried to do earlier was to reduce some repetition for `"Community"` `"Premium"` strings here and there, yet I realize it is not a crucial thing :smile:

lib/trento_web/controllers/about_controller.ex
nelsonkopliku: Fixed.

fabriziosestito (approved):

dottorblaster (approved):
:metal:

PR #257, opened 2022-03-29
Get catalog from runner

arbulu89:
Implement the backend part of the runner connection to get the catalog.
It includes:
- 2 adapters to get the catalog. The `MockRunner`, so we don't need a real runner to create the catalog, and the real `Runner`, which interacts with the runner. The implement an adapter, so we can use one or another, and even make the testing easier.
- Configuration of the adapter and the `RUNNER_URL` flag
- Update the `catalog.json` to use the new flat list

TODO next:
- Update the frontend to do some error handling and call the new api correctly


config/dev.exs
```diff
@@ -92,7 +92,10 @@ config :trento, Trento.Scheduler,
   ]
 
 config :trento, Trento.Integration.Telemetry, adapter: Trento.Integration.Telemetry.ToLogger
-config :trento, Trento.Integration.Checks, adapter: Trento.Integration.Checks.MockRunner
+
+config :trento, Trento.Integration.Checks,
+  adapter: Trento.Integration.Checks.MockRunner,
+  runner_url: ""
```
fabriziosestito, line 98: the `runner_url` is configuration owned by the `Trento.Integration.Checks.Runner` adapter

config/prod.exs
```diff
@@ -5,5 +5,7 @@ config :trento, TrentoWeb.Endpoint,
   cache_static_manifest: "priv/static/cache_manifest.json",
   server: true
 
+config :trento, Trento.Integration.Checks, runner_url: System.get_env("RUNNER_URL")
+
```
fabriziosestito, line 9: not needed here

lib/trento/application/integration/checks/adapter/runner.ex
```diff
@@ -0,0 +1,52 @@
+defmodule Trento.Integration.Checks.Runner do
+  @moduledoc """
+  Trento runner integration adapter
+  """
+
+  @behaviour Trento.Integration.Checks.Gen
+
+  require Logger
+
```
fabriziosestito, line 9: `@runner_url Application Application.compile_env!(:trento, __MODULE__)[:runner_url]`

config/runtime.exs
```diff
@@ -54,6 +54,15 @@ if config_env() == :prod do
...
     secret_key_base: secret_key_base
 
+  runner_url =
+    System.get_env("RUNNER_URL") ||
+      raise """
+      environment variable RUNNER_URL is missing.
+      For example: http://localhost:8080
+      """
+
+  config :trento, Trento.Integration.Checks, runner_url: runner_url
```
fabriziosestito, line 64:
```suggestion
    config :trento, Trento.Integration.Checks.Runner, runner_url: runner_url
```

lib/trento/application/integration/checks/adapter/gen.ex
```diff
@@ -10,4 +10,10 @@ defmodule Trento.Integration.Checks.Gen do
               selected_checks :: [String.t()]
             ) ::
               :ok | {:error, any}
+
+  @callback get_runner_ready_content(runner_url :: String.t()) ::
+              {:ok, term} | {:error, any}
+
+  @callback get_catalog_content(runner_url :: String.t()) ::
+              {:ok, term} | {:error, any}
```
fabriziosestito, line 18: `@callback get_catalog(runner_url :: String.t()) :: {:ok, map} | {:error, :not_ready} | {:error, any}`

nelsonkopliku (commented):
Nice! I left a couple of comments but @fabriziosestito can help us better :smile: 

Then a little reasoning.

This guy exposes an endpoint `/api/checks/catalog` which actually calls `/api/catalog` on the runner, 

Sort of a proxy, but not even, because it does _something_, namely grouping a flatted catalog received from the Runner, so we could say this is sort of a BFF feature.

runner response
```
[
  {
    "id": "156F64",
    "name": "1.1.1",
    "group": "Corosync",
    "provider": "azure",
    "description": "...",
    "labels": "generic"
  },
  {
    "id": "156F64",
    "name": "1.1.1",
    "group": "Corosync",
    "provider": "aws",
    "description": "...",
    "labels": "generic"
  },
  ...
]
```

web response
```
[
    {
      "groups": [
        {
          "checks": [
            {
                ...
            }
          ],
          "group": "Corosync"
        },
        {
          "checks": [
            {
              ...
            }
          ],
          "group": "Miscellaneous"
        }
      ],
      "provider": "aws"
    },
    {
      "groups": [
        {
          "checks": [
            {
             ...
            }
          ],
          "group": "Corosync"
        },
        {
          "checks": [
            {
              ...
            }
          ],
          "group": "Miscellaneous"
        }
      ],
      "provider": "azure"
    }
  ]
```
This is perfectly fine.

Yet I'd consider the opportunity to actually just have a proxy for the time being (an ingress or a proxy controller on web).

Why did we chose to do it this way? Mainly to make the grouping on elixir side I believe.

lib/trento/application/integration/checks/adapter/runner.ex
arbulu89: Is it not better to have:
```
defp runner_url,
    do: Application.fetch_env!(:trento, __MODULE__)[:runner_url]
```

As this value must be obtained on `runtime` in production from a ENV variable

lib/trento/application/integration/checks/adapter/gen.ex
arbulu89: I will change this to `@callback get_catalog :: {:ok, term} | {:error, :not_ready} | {:error, any}` as the catalog conversion will happen in the main `Checks` module

lib/trento/application/integration/checks/adapter/runner.ex
fabriziosestito: yes my bad, this is the way!

lib/trento/application/integration/checks/adapter/gen.ex
fabriziosestito: nitpick: `{:ok, any}`. It is  actually equivalent to term, but term is more an erlang thing and usually elixir specs use `any`

fabriziosestito: here you could also use `{:ok, map}`

arbulu89:
@fabriziosestito @nelsonkopliku , 
Most of your comments (or all of them) addressed in the last commit: https://github.com/trento-project/web/commit/a5d179ff04f87a7e52b15542134748ab95d11f89

Besides that, I have updated slightly the catalog controller to allow a `flat` param (`api/checks/catalog?flat`) which will return the flat version of the catalog.

config/dev.exs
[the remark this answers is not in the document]
arbulu89: No. Moved to the real runner code only.

lib/trento/application/integration/checks/checks.ex
```diff
@@ -7,6 +9,67 @@ defmodule Trento.Integration.Checks do
   def request_execution(execution_id, cluster_id, hosts, selected_checks),
     do: adapter().request_execution(execution_id, cluster_id, hosts, selected_checks)
 
+  @spec get_catalog ::
+          {:ok, Catalog.t()} | {:error, any}
+  def get_catalog do
+    case is_catalog_ready(runner_url()) do
```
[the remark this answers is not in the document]
arbulu89: Changed a bit the code to look for catalog readiness. Now, if the `api/catalog` returns a `204` status code, we assume that the catalog is not ready yet

arbulu89:
> 
> Why did we chose to do it this way? Mainly to make the grouping on elixir side I believe.
> 
> I believe it can go this way.

Yes, mostly the idea is to keep the runner as dump as possible, mostly, because doing list operations in golang is more costly (and much more in ansible...).
Finally, using a simple ingress wouldn't be that easy, because the runner has some certain intelligence, like some state where the catalog is not ready yet. Finally, I took the chance of doing some error handling and messages customization. In any case, the catalog controller is really simple at this point, so I hope it doesn't require too much maintenance.

config/dev.exs
```diff
@@ -93,6 +93,7 @@ config :trento, Trento.Scheduler,
 
 config :trento, Trento.Integration.Telemetry, adapter: Trento.Integration.Telemetry.ToLogger
 config :trento, Trento.Integration.Checks, adapter: Trento.Integration.Checks.MockRunner
+config :trento, Trento.Integration.Checks.Runner, runner_url: ""
```
fabriziosestito, line 96: we can remove this since in dev we use the mockrunner

lib/trento/application/integration/checks/checks.ex
```diff
@@ -7,6 +12,52 @@ defmodule Trento.Integration.Checks do
...
 
+  @spec get_catalog ::
+          {:ok, FlatCatalog.t()} | {:error, any}
+  def get_catalog do
+    case adapter().get_catalog() do
+      {:ok, catalog} ->
+        FlatCatalog.new(%{checks: catalog})
+
+      {:error, :not_ready} ->
+        {:error, "The catalog is still being built. Try again in some moments"}
```
fabriziosestito, line 23: it is  preferable to not send strings as errors since they can be hardly matched, here we would just bubble up the error and maybe add a log

lib/trento/application/integration/checks/checks.ex
```diff
@@ -7,6 +12,52 @@ defmodule Trento.Integration.Checks do
...
+
+      {:error, reason} ->
+        {:error, reason}
+
+      _ ->
```
fabriziosestito, line 28: there is no need for this match, blackholes hide cases and if we do not receive an :ok or an :error it is fine to break because there is something wrong in the code

lib/trento/application/integration/checks/adapter/gen.ex
```diff
@@ -1,6 +1,6 @@
 defmodule Trento.Integration.Checks.Gen do
   @moduledoc """
-  Behaviour of a telemetry adapter.
+  Behaviour of a runner adapter.
```
fabriziosestito, line 3: the behavior is missing the callback `get_catalog_by_group`

lib/trento/application/integration/checks/adapter/runner.ex
```diff
@@ -0,0 +1,41 @@
...
+      {:error, %HTTPoison.Error{reason: :econnrefused}} ->
+        {:error, "Connection to the runner component on #{runner_url} was refused."}
+
+      {:error, %HTTPoison.Error{reason: :nxdomain}} ->
+        {:error,
+         "Connection url to the runner component host on #{runner_url} could not be resolved."}
+
+      {:error, %HTTPoison.Error{reason: reason}} ->
+        {:error, reason}
+
```
fabriziosestito, line 33: Do we need to match all this cases? just matching with error reason is enough and also we want to return :error, atom tuples and not strings

lib/trento/application/integration/checks/checks.ex
fabriziosestito: also the get_catalog doesnt return anything except :ok tuples and :error tuples so this case will never match

fabriziosestito (changes requested):

fabriziosestito:
@arbulu89 left some commends about styles / error handling in general.
Some doubts about the naming of the DTOs / models but we can iterate on it

lib/trento/application/integration/checks/adapter/runner.ex
arbulu89: And where do we "create" the error messages? In the frontend?
I wouldn't like to show a message like `econnrefused`

lib/trento/application/integration/checks/checks.ex
```diff
@@ -7,6 +12,52 @@ defmodule Trento.Integration.Checks do
   def request_execution(execution_id, cluster_id, hosts, selected_checks),
     do: adapter().request_execution(execution_id, cluster_id, hosts, selected_checks)
 
+  @spec get_catalog ::
+          {:ok, FlatCatalog.t()} | {:error, any}
+  def get_catalog do
+    case adapter().get_catalog() do
+      {:ok, catalog} ->
+        FlatCatalog.new(%{checks: catalog})
```
fabriziosestito, line 20: we should return the `FlatCatalog.t()` directly from the adapter, this way we enforce a contract with the adapter and we check that the contract was respected before consuming it.

lib/trento/application/integration/checks/adapter/runner.ex
fabriziosestito: Does the user need to know if there was an nslookup error or a connection refused error? For the frontend i would just bubble up that something went wrong and log it for debugging purposes.
Also, in general, is not responsability of an adapter to think about the FE since there could be no FE at all.

arbulu89: Well, I personally find it much more user friendly to give specific error messages than "Something went wrong" message, if we know at least which was the error. And this is case, I find it useful that the error contains the runner url, it makes things easier.
Anyway, I don't want to open a battle about this here.
I will remove the messages and we will see what we do with them in the future

fabriziosestito: Keep in mind that when running in kubernetes the runner URL will be always the service name .e.g http://trento-runner-service

lib/trento/application/integration/checks/checks.ex
arbulu89: Perfect!

arbulu89:
@fabriziosestito yet another round of changes hehe
I have removed the "stringtified" error handling and enforced the usage of `FlatCatalog` from the adapter

PR #260, opened 2022-03-29
Close filters when clicking outside

fabriziosestito:


assets/js/components/Table/Filter.jsx
```diff
@@ -110,4 +113,35 @@ const Filter = ({ options, title, value, onChange }) => {
   );
 };
 
+// Hook
+function useOnClickOutside(ref, handler) {
+  useEffect(
```
fabriziosestito, line 118: @dottorblaster where I can move this?

PR #264, opened 2022-03-30
Store tags state

rtorrero:
tags-store

This PR makes use of the redux store to save the state of the UI components, preventing tags from disappearing when the user changes the view.

It should work for the following views:
 - Hosts
 - Clusters
 - SAP systems*
 - Databases**

Note*: There seems to be some issue with the collapsible that activates when interacting with the tags which prevents the tags from working correctly.

Note**: The tags endpoint for the databases overview one is missing so it can't really be tested yet. Working on that next.

### DIFF
--- a/assets/js/components/ClustersList.jsx
+++ b/assets/js/components/ClustersList.jsx
@@ -5,6 +5,7 @@ import { Menu, Transition } from '@headlessui/react';
 import axios from 'axios';
 import Table from './Table';
 import Tags from './Tags';
+import { addTagToCluster, removeTagFromCluster } from '@state/clusters';
 
 import { EOS_EDIT, EOS_RUN_CIRCLE, EOS_LUNCH_DINING } from 'eos-icons-react';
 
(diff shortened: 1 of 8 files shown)


nelsonkopliku (commented):
Nice! Here's some thoughts:

Consider the reducing functions you created

`addTagToHost`
`removeTagFromHost`

`addTagToCluster`
`removeTagFromCluster`

`addTagToDatabase`
`removeTagFromDatabase`

`addTagToSAPSystem`
`removeTagFromSAPSystem`

All the `addTagToSomething` have the same implementation, only differing for the resource being tagged.

All the `removeTagFromSomething` have the same implementation, only differing for the resource being untagged.

Probably we can find a way to reduce boiler.

Then, these functions have been added to the specific state slices for `host`, `cluster`, `database`, `sapSystem`, which should be fine.

There's also the option to have a dedicated state slice for tags and merge information with selectors.
Not saying we need to do this other way, just asking if you considered it :smile: 

Last point: the various `addTag` and `removeTag` functions in the components (the ones that actually call the API) might be cleaned up a bit I believe, as they do the exact same thing, with the difference of the resource being handled and the endpoint that gets hit.
This can very well be a followup.

fabriziosestito:
> There's also the option to have a dedicated state slice for tags and merge information with selectors.
Not saying we need to do this other way, just asking if you considered it smile

Maybe this would be the solution, since i see the duplication here necessary due to the different slices being involved as you pointed out.

fabriziosestito (approved):

dottorblaster (approved):
I like this, we might need a refactoring but we can do that in a follow-up PR. Green light from my side, good job!

nelsonkopliku (approved):

PR #271, opened 2022-03-30
CibLastWritten broadcast

dottorblaster:
Just broadcasting to the frontend the `cib_last_written` field when a `RegisteredClusterHost` command gets dispatched :raised_hands:

### DIFF
--- a/assets/js/components/ClusterDetails/ClusterDetails.jsx
+++ b/assets/js/components/ClusterDetails/ClusterDetails.jsx
@@ -89,6 +89,10 @@ const ClusterDetails = () => {
               content: cluster.details && cluster.details.sr_health_state,
             },
             {
+              title: 'CIB last written',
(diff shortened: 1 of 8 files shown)


nelsonkopliku (approved):
All good to me as long as we have clear that:

- we might be broadcasting this info for a cluster that is not yet actually registered and available
- hence, that info might be available at the second run of the cluster discovery (unless we keep it redux state and merge with the cluster when ready)

nelsonkopliku:
> Just broadcasting to the frontend the `cib_last_written` field when a `RegisteredClusterHost` event comes in :raised_hands: 
>

Almost :stuck_out_tongue: 
I believe what you meant is

>Whenever a `RegisterClusterHost` command gets dispatched

dottorblaster:
Yes :sweat_smile: too much beer yesterday :grimacing:

PR #272, opened 2022-03-31
Rename and clean-up dtos

fabriziosestito:
Cleaned up and moved models to a common dto folder, renamed to make it clear we are referring to DTOs in the code


lib/trento/application/integration/checks/adapter/gen.ex
```diff
@@ -3,7 +3,7 @@ defmodule Trento.Integration.Checks.Gen do
   Behaviour of a runner adapter.
   """
 
-  alias Trento.Integration.Checks.Models.FlatCatalog
+  alias Trento.Integration.Checks.FlatCatalogDto
```
nelsonkopliku, line 6: What about `Trento.Integration.Checks.Dto.FlatCatalog`?

Not a strong opinion tho :smile:

nelsonkopliku (approved):

lib/trento/application/integration/checks/adapter/gen.ex
fabriziosestito: if you do this when you alias you loose the Dto information, making the code a bit funky

arbulu89 (approved):

PR #274, opened 2022-03-31
Add prometheus target discovery endpoint

fabriziosestito:
This PR adds the endpoint needed by prometheus to fetch the target hosts/exporters ports

image


arbulu89 (approved):
Too easy for you @fabriziosestito XD

PR #280, opened 2022-04-01
Paginated tables

dottorblaster:
Peek 01-04-2022 12-07

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -152,8 +152,17 @@ jobs:
             priv/plts
           key: ${{ runner.os }}-${{ env.OTP_VERSION }}-${{ env.ELIXIR_VERSION }}-${{ hashFiles('mix.exs') }}-${{ hashFiles('mix.lock') }}
 
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+
+      - run: cd assets && npm install
       - name: Run test
         run: mix test --color --trace --slowest 10
+
+      - name: Run JS tests
+        run: cd assets && npm test
 
   build-and-push-container-images:
     name: Build and push container images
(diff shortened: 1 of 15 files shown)


arbulu89 (approved):
Smooth!

Some "maybe future todo things":
- It would be nice to have `Items per page` option as we do in old trento
- It looks pretty ugly the the columns width changes depending on the content. Maybe we should have some fixed column widths...
- I don't have any preference on having the `First/Last` options in the pagination options, but we have them in the old trento. Something to comment with @abravosuse maybe

PD: Maybe you already tried, but I would test if the pagination works when filters are applied. It was a pretty big pain in the old implementation hehe

abravosuse:
I think we should not lose the option to select the number of items shown in a Trento overview (hosts, clusters, sap systems, hana databases). As a matter of fact, we got a request from one of our ramp-up customers to make the number of items selected for a particular overview a persistent configuration (see https://jira.suse.com/browse/CFSA-380). We don't need to implement this yet -it's not in the TO-DO list- but giving the user the option to select how many items they want to see in each overview is important. Also the First and Last shortcuts were handy. 

It is important, as @arbulu89 mentions, that pagination still works when filters are applied.

I agree that column widths changing based on content doesn't play nice visually. But my concern with fix widths is how it will behave when the content exceed that size.

dottorblaster:
Wow, tons of stuff to discuss, especially for a PR that was meant to be like a low hanging fruit :sweat_smile: 

### Filtering
Of course filtering works in combo with pagination :smile: 

### First/Last
It surely can be done. Can we note that as a user story for the future?

### Select number of items
This can be done too, but I'd rather note that for the future in another user story.

### Fixed-width columns
> my concern with fix widths is how it will behave when the content exceed that size

That's my concern too so far, and it requires a lot of research so I would postpone this kind of research after GA since we have to inspect how fixed-size layouts play with CSS grid. Clock is ticking... :smile:

fabriziosestito (approved):

nelsonkopliku (approved):

abravosuse:
Agreed to leave the fixed-width columns topic for later. And I don't mind creating new user stories for the First/Last shortcuts and the selection of number of items... as long as we tackle them down before GA? :-) At least the second one, please...

PR #291, opened 2022-04-04
Email alerting

nelsonkopliku:
image

remember: if a database goes critical, and that database is part of a SAP system, the sap system also goes critical and so we'll have 2 alert notifications being sent in this scenario.


stefanotorresi:
Naaaiiissss!!

lib/trento/application/usecases/alerting/alerting.ex
```diff
@@ -0,0 +1,65 @@
...
+  """
+
+  alias Trento.{
+    ClusterReadModel,
+    DatabaseReadModel,
+    HostReadModel,
+    SapSystemReadModel
+  }
+
+  alias Trento.Infrastructure.Alerting.EmailAlert
```
nelsonkopliku, line 13: This breaks the dependency rule :smile: 
From application we're having a hard dependency from infrastructure.

I'd prefer a level of indirection here, yet maybe we can keep it for later.

fabriziosestito: @nelsonkopliku this because emailalert is a presentation component, wrapping a template and so on, not an infrastructural component, so i think this is misplaced in the beginning.
you can move it to the usecase folder i believe

fabriziosestito (approved):

dottorblaster (approved):
I didn't see anything the made me raise an eyebrow.

LGTM

fabriziosestito:
:shipit:

lib/trento/application/usecases/alerting/alerting.ex
nelsonkopliku: I see your point, and considering that it makes sense to have `EmailAlert` in application layer.
Changed like this
image

Good for me.

I want to share my point of view.

I still see emails as an infrastructural thing.
Why? Because to me email is a mean of notifying someone about something.
So I see the email as _how the notification gets delivered_

Some notifications could be email notifications, other ones might SMS, and others might be integration with other systems... and so forth.

So the concepts I see are (forgive the naming :smile: ) 
- `Notification`: what needs to be notified 
- `Transport`: how the notification is delivered

I see how this is **not what we currently need**, just sharing also for future discussions.

Issue #222, opened 2022-03-21
Refactor Tagging


nelsonkopliku:
Now we have several controllers
- `cluster_controller`
- `host_controller`
- `sap_system_controller`
- `database_controller`

defining these two functions
- `create_tag`
- `delete_tag`

```
@spec create_tag(Plug.Conn.t(), map) :: Plug.Conn.t()
def create_tag(conn, %{
      "id" => id,
      "value" => value
    }) do
  case Tags.create_tag(value, id, "host") do
    {:ok, _} ->
      conn
      |> put_status(:accepted)
      |> json(%{})

    {:error, _} ->
      conn
      |> put_status(:bad_request)
      |> json(%{error: "tag creation failed"})
  end
end

@spec delete_tag(Plug.Conn.t(), map) :: Plug.Conn.t()
def delete_tag(conn, %{
      "id" => resource_id,
      "value" => value
    }) do
  case Tags.delete_tag(value, resource_id) do
    :ok ->
      conn
      |> put_status(:accepted)
      |> json(%{})

    :not_found ->
      conn
      |> put_status(:not_found)
      |> json(%{})
  end
end
```

The point is that `delete_tag` is exactly the same for all the controllers while `create_tag` only differs on the third argument of `Tags.create_tag/3`, being the `resource_type` (host, cluster, database, sap_system).

I believe we can improve a bit by:
- having clearer function names like `tag_host`, `tag_sap_system`, etc... and delegate the detail of the resource type
- putting `delete_tag` in a module to reduce duplication :thinking:
- having a `TaggingController` where to route the various `POST /<resourceType>/<resourceId>/tags` and `DELETE /<resourceType>/<resourceId>/tags/<value>`

rtorrero:
I'd also add into this issue:
 - Changes in the tags are gone when you switch to another page. They only reflect again after a reload

rtorrero:
Also, about the naming:
@fabriziosestito and I had a conversation about it and we decided to pick `create_tag` because it reflected better that tags is a separate entity. Would something like `create_tag_host` be good enough? wdyt?

PR #294, opened 2022-04-04
Add database tag endpoint

rtorrero:
This adds the missing bits and pieces to add endpoint(s) to handle database tags. There is soe further code duplication but we will refactor/improve later when we address #222

### DIFF
--- a/lib/trento_web/controllers/sap_system_controller.ex
+++ b/lib/trento_web/controllers/sap_system_controller.ex
@@ -59,7 +59,7 @@ defmodule TrentoWeb.SapSystemController do
   end
 
   # FIXME: refactor tags api, we just need a generic tag API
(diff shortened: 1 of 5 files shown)


fabriziosestito (approved):

lib/trento_web/controllers/sap_system_controller.ex
```diff
@@ -57,4 +57,22 @@ defmodule TrentoWeb.SapSystemController do
         send_resp(conn, :not_found, "")
     end
   end
+
+  # FIXME: refactor tags api, we just need a generic tag API
+  def create_db_tag(conn, %{
```
fabriziosestito, line 62: `create_database_tag`

fabriziosestito (changes requested):
Just a nitpick + test is missing

PR #296, opened 2022-04-04
Add query parameters for catalog controller

arbulu89:
Some small improvements on the catalog controller usage:
- Error handling in the frontend
- Provide the option to filter by provider from the backed (`api/cluster/checks?provider=azure`)
- Use the flatten version of the data in the frontend
- Rename `get_catalog_by_provider` to `get_catalog_grouped_by_provider` to be more meaningful

### DIFF
--- a/assets/js/components/ChecksResults.jsx
+++ b/assets/js/components/ChecksResults.jsx
@@ -34,10 +34,6 @@ const getHostname =
     }, '');
   };
 
-const getCatalogByProvider = (catalog, catalogProvider) => {
-  return catalog.find(({ provider }) => provider === catalogProvider);
-};
-
 const sortChecksResults = (checksResults = [], group) => {
   return checksResults.sort((a, b) => {
     if (a.check_id === b.check_id) {
@@ -76,15 +72,13 @@ const ChecksResults = () => {
   const dispatchUpdateCatalog = () => {
     dispatch({
       type: 'UPDATE_CATALOG',
+      payload: { flat: '', provider: 'azure' }, //FIXME: Get provider properly
     });
   };
 
   const checksResults = getChecksResults(cluster);
 
   const hostname = getHostname(useSelector((state) => state.hostsList.hosts));
-
-  // FIXME: Check the provider by cluster
-  const catalogByProvider = getCatalogByProvider(catalogData, 'azure');
 
   useEffect(() => {
     dispatchUpdateCatalog();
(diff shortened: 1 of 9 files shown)


assets/js/components/ChecksResults.jsx
```diff
@@ -76,16 +72,14 @@ const ChecksResults = () => {
   const dispatchUpdateCatalog = () => {
     dispatch({
       type: 'UPDATE_CATALOG',
+      payload: { flat: '', provider: 'azure' }, //FIXME: Get provider properly
```
nelsonkopliku, line 75: This would translato toa query string like this 
```
/api/checks/catalog?flat=&provider=azure
```
I would make it `flat=true`, I find it a bit more robust.

what do you think?

nelsonkopliku (commented):
Nice! just a tiny thing about a query string parameter.

assets/js/components/ChecksResults.jsx
arbulu89: In the current scenario the queries like `flat&provider=azure` and `flat=&provider=azure`, both would work. I often time think that booleans are a tricky thing, because each language has their own syntax with `true|True|1` etc hehe.

I read that if the query param is a boolean having it without any parameter is pretty normal. But I don't have any issue changing it. What is the most normal thing out there?

fabriziosestito: in elixir we can check the parameter directly since `nil` is a falsy value

nelsonkopliku: > In the current scenario the queries like `flat&provider=azure` and `flat=&provider=azure`, both would work
>
Yes, they would both work. I just find it a bit inconsistent and ambiguous as a contract.

> booleans are a tricky thing, because each language has their own syntax with true|True|1 etc
>
I see, it is also true that URLs should be independent from the language serving or consuming :smile: 
If then there is some language constraints in handling urls, that's another topic :smile: 

That said no big issue for me :+1:, anyway we do not yet have real API consumers to take care of.

arbulu89: @fabriziosestito @nelsonkopliku At the end, what is the conclusion? Should I change it?

> in elixir we can check the parameter directly since nil is a falsy value
I'm not able to decipher your message @fabriziosestito XD

nelsonkopliku (approved):

Issue #146, opened 2022-03-05
Remove Mock -> Use Mox (or Hammox)


fabriziosestito:
http://blog.plataformatec.com.br/2015/10/mocks-and-explicit-contracts/

arbulu89:
Some tests with mock here: https://github.com/trento-project/web/pull/304/commits/046bcd9fac28363d8c3b84a01d16fe8c3065807d

https://github.com/trento-project/web/blob/main/test/trento/application/integration/checks/checks_test.exs

- `"should handle execution completed event properly"`
- `"should handle execution started event properly"`

fabriziosestito:
https://github.com/trento-project/web/pull/409

PR #304, opened 2022-04-06
Handle unreachable checks execution scenario

arbulu89:
Handle the unreachable checks execution scenario.

The implementation was not that simple.

- New table which stores the hosts states for each execution added
- The cluster domain changed to store the checks executions in host basis
- Update the projection to store the new hosts state
- Update the frontend

Initially, I thought of adding a new `NotReached` event or similar (actually, some intermediate commits have this code), but eventually I saw that the code was being duplicated, and the only real difference was the name of the event. I thought that it is not enough the hassle.

Please, don't judge the UI, I didn't have enough time to work on it. But I wanted to open the PR on draft so you can have a view while I'm away.

PD: The tests are still missing

Here how it looks like:
unreachable


arbulu89:
Hey @trento-project/developers , 
Did you have the chance to have a look on this PR?
Should I simply work on the conflicts and tests?

lib/trento/application/read_models/host_checks_results_read_model.ex
```diff
@@ -0,0 +1,25 @@
...
+
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @derive {Jason.Encoder, except: [:__meta__, :__struct__]}
+  @primary_key false
+  schema "hosts_checks_results" do
```
fabriziosestito, line 14: hosts checks results or hosts checks execution?

lib/trento/domain/cluster/cluster.ex
```diff
@@ -438,4 +469,42 @@ defmodule Trento.Domain.Cluster do
       %ClusterHealthChanged{cluster_id: cluster_id, health: new_health}
     end
   end
+
+  defp handle_hosts_execution_data(
```
fabriziosestito, line 473: nitpick: i would vouch against the usage of the word "data" in the domain logic. Had to scroll to the function to see what it was actually about, maybe we can use a name which is more "domain related", such as "emit_host_execution_completed_event".

lib/trento/domain/cluster/cluster.ex
```diff
@@ -438,4 +469,42 @@ defmodule Trento.Domain.Cluster do
...
+         %{host_id: host_id, reachable: reachable, msg: msg},
+         multi,
+         cluster_id,
+         hosts_executions
+       )
+       when reachable == false do
```
fabriziosestito, line 479: no need of using a guard here see above

lib/trento/domain/cluster/cluster.ex
```diff
@@ -438,4 +469,42 @@ defmodule Trento.Domain.Cluster do
       %ClusterHealthChanged{cluster_id: cluster_id, health: new_health}
     end
   end
+
+  defp handle_hosts_execution_data(
+         %{host_id: host_id, reachable: reachable, msg: msg},
+         multi,
```
fabriziosestito, line 475: `%{host_id: host_id, reachable: false, msg: msg},`

lib/trento/domain/cluster/cluster.ex
```diff
@@ -438,4 +469,42 @@ defmodule Trento.Domain.Cluster do
...
+    end)
+  end
+
+  defp handle_hosts_execution_data(
+         %{host_id: host_id, reachable: reachable, msg: msg, checks_results: results},
+         multi,
+         cluster_id,
+         _hosts_executions
+       )
+       when reachable == true do
```
fabriziosestito, line 498: no need to use a guard here see comments abovev

priv/repo/migrations/20220405142819_create_host_checks_results.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Trento.Repo.Migrations.CreateHostChecksResults do
+  use Ecto.Migration
+
+  def change do
+    create table(:hosts_checks_results, primary_key: false) do
```
fabriziosestito, line 5: hosts_checks_executions

test/trento/application/integration/checks/checks_test.exs
```diff
@@ -3,6 +3,7 @@ defmodule Trento.Integration.ChecksTest do
   use Trento.DataCase
 
   import Mox
+  import Mock
```
fabriziosestito, line 6: we want to remove mock in a near future and just rely on mox, could you please add an issue about this test to keep track of it?

test/trento/application/integration/checks/checks_test.exs
```diff
@@ -287,4 +300,114 @@ defmodule Trento.Integration.ChecksTest do
 
     assert {:ok, catalog_by_provider} == Checks.get_catalog_grouped_by_provider()
   end
+
+  test "should handle execution started event properly" do
+    with_mock Trento.Commanded, dispatch: fn _, _ -> :ok end do
```
fabriziosestito, line 305: we want to remove mock library soonish and leverage mox for this kind of things, could we add an issue about this test to keep track of the debt?

dottorblaster:
Woh lots of stuff, great! Could you rewrite the history to make the review easier? Thanks!

lib/trento/application/read_models/host_checks_results_read_model.ex
nelsonkopliku: Between the two I'd prefer `HostsChecksExecution`.

---

Just as a discussion triggering point:
I cannot also think about the possibility of it being at the moment  just `ChecksExecution` :thinking: 
anyway it is bounded to the context of a host, right?

We can refine also on the language and see whether some extra differentiation is needed.

nelsonkopliku (commented):
Great job!

No big thing from my side, @fabriziosestito already pointed out good comments.

Naming is always difficult and good design too :smile:

lib/trento/application/read_models/cluster_read_model.ex
```diff
@@ -24,6 +27,7 @@ defmodule Trento.ClusterReadModel do
     field :details, :map
     field :checks_execution, Ecto.Enum, values: [:not_running, :requested, :running]
 
+    has_many :hosts_executions, HostChecksExecutionsReadModel, foreign_key: :cluster_id
```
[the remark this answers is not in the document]
arbulu89: It was much more hassle to do so, and without big advantages. It forces you to do more hierarchical operations to get the checks (loop through hosts, and them loop through the checks), and as we were already storing the host_id for each check, I thought that it was good enough. In fact, as you commented, it is easier to handle the 2 lists independently in the frontend

test/trento/application/integration/checks/checks_test.exs
arbulu89: At the end added simply a `#TODO` tag. Tech debt issue: https://github.com/trento-project/web/issues/146

lib/trento/application/read_models/cluster_read_model.ex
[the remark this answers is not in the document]
fabriziosestito: Perfect, i was just curious about the rationale. at the end read models are for, well, reading, so it make sense to make them closer to the usecase.

nelsonkopliku (approved):

PR #320, opened 2022-04-07
Health overviews

nelsonkopliku:
Hosts Overview
image

Clusters Overview
image

SAP Systems Overview
image

HANA Databases Overview
image


fabriziosestito (approved):
LGTM, except the size maybe we can have a follow up PR to reduce it

abravosuse:
Arriving late @nelsonkopliku but just wanted to say that looks awesome.

PR #321, opened 2022-04-07
Cluster-e2e-tests

rtorrero:
Follow up to #317:
 - Add more basic tests for the clusters overview
 - Add missing env vars to the login command

### DIFF
--- a/assets/js/components/ClustersList.jsx
+++ b/assets/js/components/ClustersList.jsx
@@ -74,7 +74,9 @@ const ClustersList = () => {
         key: 'name',
         filter: true,
         render: (content, item) => (
-          <ClusterLink cluster={item}>{content}</ClusterLink>
+          <ClusterLink cluster={item}>
+            <span className="tn-clustername">{content}</span>
+          </ClusterLink>
         ),
       },
       {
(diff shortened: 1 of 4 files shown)


assets/js/components/ClustersList.jsx
```diff
@@ -74,7 +74,9 @@ const ClustersList = () => {
         key: 'name',
         filter: true,
         render: (content, item) => (
-          <ClusterLink cluster={item}>{content}</ClusterLink>
+          <ClusterLink cluster={item}>
+            <div className="tn-clustername">{content}</div>
+          </ClusterLink>
```
dottorblaster, line 79: I think we can't do that this way, try adding the `className` prop directly to the `ClusterLink` component, propagating it to the `span` underneath

dottorblaster (changes requested):

assets/js/components/ClustersList.jsx
dottorblaster: ... or you could just convert the `div` to a `span`. Your call.

rtorrero: I had already tried the first approach and it didn't work, going for the second one

dottorblaster (approved):

PR #324, opened 2022-04-08
Change container tag to trento-web

fabriziosestito:
This PR changes the container tag from `trento-web-dev` to `trento-web` overwriting the legacy package


dottorblaster (approved):
LGTM

PR #337, opened 2022-04-11
Added Feedback collector

nelsonkopliku:
Ported but not working because of JIRA :sweat_smile: 

When the form is submitted we get a 404 with this response body `XSRF check failed`.
Doesn't work on trento demo neither.


nelsonkopliku:
Closed as it is not relevant at the moment.

PR #349, opened 2022-04-12
Eula

dottorblaster:
image


fabriziosestito (approved):

lib/trento_web/router.ex
```diff
@@ -100,6 +100,9 @@ defmodule TrentoWeb.Router do
 
     post "/runner/callback", ClusterController, :runner_callback
     get "/prometheus/targets", PrometheusController, :targets
+
+    get "/settings", SettingsController, :settings
```
nelsonkopliku, line 104: These should be `:apikey_authenticated`.

PS rebase on main

nelsonkopliku (approved):

PR #351, opened 2022-04-13
Add api key settings

nelsonkopliku:
api-key-ok
api-key-ko

### DIFF
--- a/assets/js/components/Settings/Settings.jsx
+++ b/assets/js/components/Settings/Settings.jsx
@@ -1,12 +1,10 @@
 import React, { useState, useEffect } from 'react';
-import { get } from 'axios';
-
-import Button from '@components/Button';
-
-import { logError } from '@lib/log';
-
 import { Transition } from '@headlessui/react';
 import classNames from 'classnames';
+
+import { get } from 'axios';
+import Button from '@components/Button';
+import { logError } from '@lib/log';
 
 const Settings = () => {
   const [loading, setLoading] = useState(false);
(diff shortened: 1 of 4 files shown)


fabriziosestito (approved):
this looks just great! thanks!

assets/js/components/Settings/Settings.jsx
```diff
@@ -0,0 +1,159 @@
+import React, { useState, useEffect } from 'react';
+import { get } from 'axios';
+
+import Button from '@components/Button';
+
+import { logError } from '@lib/log';
+
+import { Transition } from '@headlessui/react';
+import classNames from 'classnames';
```
dottorblaster, line 9: Can you move these two imports on top?

assets/js/components/Settings/Settings.jsx
```diff
@@ -0,0 +1,159 @@
...
+        apiKey !== undefined && setApiKey(apiKey);
+        setLoading(false);
+      })
+      .catch((error) => {
+        logError(error);
+        setLoading(false);
+      });
+  }, []);
+
+  const hasApiKey = apiKey !== null;
```
dottorblaster, line 29: Would eslint complain if we put `Boolean(apiKey)` here?

dottorblaster (approved):
Just a couple comments

assets/js/components/Settings/Settings.jsx
nelsonkopliku: Ok. like this?
```
import React, { useState, useEffect } from 'react';
import { Transition } from '@headlessui/react';

import { get } from 'axios';
import Button from '@components/Button';
import { logError } from '@lib/log';
```

assets/js/components/Settings/Settings.jsx
nelsonkopliku: does not complain it seems. Gonna change in one commit with the previous comment.

PR #358, opened 2022-04-14
Moved routes in the correct authenticated pipes

nelsonkopliku:
this goes with agent that sends api key https://github.com/trento-project/agent/pull/12


dottorblaster (approved):
L:b:TM

PR #364, opened 2022-04-15
Disallow empty tags

nelsonkopliku:
Makes the backend a bit more rebust when creating tags.

Frontend couterpart has not been addressed here.


lib/trento_web/controllers/tags_controller.ex
```diff
@@ -0,0 +1,84 @@
...
+  defp resource_type_from_path("hosts"), do: {:ok, :host}
+  defp resource_type_from_path("clusters"), do: {:ok, :cluster}
+  defp resource_type_from_path("sap_systems"), do: {:ok, :sap_system}
+  defp resource_type_from_path("databases"), do: {:ok, :database}
+  defp resource_type_from_path(_), do: {:error, :unrecognized_resource}
+
+  defp empty?(<<" "::binary, rest::binary>>), do: empty?(rest)
+  defp empty?(""), do: true
+  defp empty?(_string), do: false
+
```
fabriziosestito, line 56: not sure what this does? couldn't you check the length of the string?

nelsonkopliku: this guy checks for empty and blank strings :smile:

dottorblaster (approved):
LGTM

PR #369, opened 2022-04-18
Add sap-system/databases linking from host overview

rtorrero:
This PR adds linking to sap-systems and databases overview from the SID column. I've added it as draft as I'm not really sure there is a better way to do this. It gets a bit ugly when we need to apply some logic to find out if the SID should link to a DB or a sap-system.

Thanks to @arbulu89  for the help!

Peek 18-04-2022 16-53

### DIFF
--- a/assets/js/components/HostsList.jsx
+++ b/assets/js/components/HostsList.jsx
@@ -24,14 +24,6 @@ const getHeartbeatIcon = ({ heartbeat }) => {
   }
 };
 
-const getSAPInstanceFromSID = (sid, sapSystems) => {
-  const instance = sapSystems.find(
-    (instance) => instance.instance?.sid === sid
-  );
-
-  return instance;
-};
-
 const getSapSystemsByHost = (sapSystems, hostId) => {
   const appInstances = sapSystems
     .flatMap((sapSystem) => sapSystem.application_instances)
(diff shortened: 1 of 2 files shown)


assets/js/components/HostsList.jsx
```diff
@@ -23,6 +24,27 @@ const getHeartbeatIcon = ({ heartbeat }) => {
...
 
+const getSAPInstanceFromSID = (sid, sapSystems) => {
+  const instance = sapSystems.find(
+    (instance) => instance.instance?.sid === sid
+  );
+
+  return instance;
+};
+
+const getSapSystemsByHost = (sapSystems, hostId) => {
```
nelsonkopliku, line 35: I believe this can be a selector on the state, selecting all the instances on a host.

rtorrero: Ah! nice, I see you used it in HostDetails, going to give it a try

nelsonkopliku:
as a reminder for possible improvements.

I am not completely confident with the choice of _how_ we show the sap systems on a host.
I see a list of sids, but I miss the information about whether it is an application instance or a database instance.

Then are we solely interested in _instances_ or the aggregation `SapSystem` and `Database`?

As an alternative UI I'd consider having sort of a popover if the information needs to be more than the simple sid.

assets/js/components/HostsList.jsx
```diff
@@ -86,16 +110,30 @@ const HostsList = () => {
...
         title: 'SID',
         key: 'sid',
-        filter: true,
-        render: (content, { cluster }) => (
-          <ClusterLink cluster={cluster}>{content}</ClusterLink>
-        ),
+        filter: (filter, key) => (element) =>
+          element[key].some((sid) => filter.includes(sid)),
+        render: (sids, { sap_systems }) => {
+          let sidsArray = sids.map((sid, index) => [
```
arbulu89, line 122: Maybe here, instead of mapping the `sids` element, we could map the `sap_systems` list, which already has the sid in each element. This way, we could space the `getSAPInstanceFromSID` function.

I mean, we already have all the information in the `sap_systems`. It is nice that we can send the list of sids as strings, so we don't need to do hacky things in the filters

arbulu89 (approved):
@rtorrero Nice! Happy to see that we don't need to have hacky things hehe
I added some comments so we can maybe simplify a bit the PR. We can discuss on them if you want

assets/js/components/HostsList.jsx
rtorrero: I'll create a ticket to figure out later how to use this. It seems to work on first load, but breaks after a refresh. I suspect that `useSelector` is not intended to be used each time we iterate on the hosts list to display each entry of the table. The warning itself after it breaks is:
>Warning: React has detected a change in the order of Hooks called by HostsList. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://reactjs.org/link/rules-of-hooks

PR #376, opened 2022-04-19
Flat map sap systems payload

arbulu89:
Based on Agent changes from: https://github.com/trento-project/agent/pull/23

### DIFF
--- a/lib/trento/application/integration/discovery/policies/sap_system_policy.ex
+++ b/lib/trento/application/integration/discovery/policies/sap_system_policy.ex
@@ -49,22 +49,27 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
     Enum.flat_map(databases, fn %{"Database" => tenant} ->
       Enum.map(
         instances,
-        fn {_, instance} ->
-          instance_number = parse_instance_number(instance)
+        fn instance ->
+          case parse_instance_number(instance) do
+            {:ok, instance_number} ->
 
-          RegisterDatabaseInstance.new(%{
-            sap_system_id: UUID.uuid5(@uuid_namespace, id),
-            sid: sid,
-            tenant: tenant,
-            host_id: host_id,
-            instance_number: instance_number,
-            instance_hostname: parse_instance_hostname(instance, instance_number),
-            features: parse_features(instance, instance_number),
-            http_port: parse_http_port(instance, instance_number),
-            https_port: parse_https_port(instance, instance_number),
-            start_priority: parse_start_priority(instance, instance_number),
-            health: parse_instance_health(instance, instance_number)
-          })
+              RegisterDatabaseInstance.new(%{
+                sap_system_id: UUID.uuid5(@uuid_namespace, id),
+                sid: sid,
+                tenant: tenant,
+                host_id: host_id,
+                instance_number: instance_number,
+                instance_hostname: parse_instance_hostname(instance, instance_number),
+                features: parse_features(instance, instance_number),
+                http_port: parse_http_port(instance, instance_number),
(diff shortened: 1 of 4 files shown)


lib/trento/application/integration/discovery/policies/sap_system_policy.ex
```diff
@@ -46,25 +46,30 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
...
+                https_port: parse_https_port(instance, instance_number),
+                start_priority: parse_start_priority(instance, instance_number),
+                health: parse_instance_health(instance, instance_number)
+              })
+
+            {:error, reason} ->
+              {:error, reason}
+          end
         end
       )
```
rtorrero, line 74: The newline for the `)` looks a bit weird, is this the formatter? (it's not doing the same below for some reason :/)

arbulu89: Changed! `mix format` behaves a bit different depending on how you format the function arguments

lib/trento/application/integration/discovery/policies/sap_system_policy.ex
```diff
@@ -166,9 +175,23 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
...
-         "SAPControl" => %{"Properties" => %{"SAPSYSTEM" => %{"value" => instance_number}}}
-       }),
-       do: instance_number
+         "SAPControl" => %{"Properties" => properties}
+       })
+       do
+
+       properties
+       |> Enum.find(fn %{"property" => property} ->
+         property == "SAPSYSTEM"
```
fabriziosestito: we can use `find_value` here:

```elixir
Enum.find_value(fn %{"property" => "SAPSYSTEM", "value"=> value} -> value,
_ -> false end)
```

lib/trento/application/integration/discovery/policies/sap_system_policy.ex
```diff
@@ -166,9 +175,23 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
...
+       end)
+       |> case do
+         %{"value" => value} ->
+           {:ok, value}
+
+         _ ->
+           {:error, :key_not_found}
```
fabriziosestito: by using find value we can return value or nil here instead of the ok/error tuple

lib/trento/application/integration/discovery/policies/sap_system_policy.ex
```diff
@@ -82,22 +87,26 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
...
-        instance_hostname: parse_instance_hostname(instance, instance_number),
-        features: parse_features(instance, instance_number),
-        http_port: parse_http_port(instance, instance_number),
-        https_port: parse_https_port(instance, instance_number),
-        start_priority: parse_start_priority(instance, instance_number),
-        host_id: host_id,
-        health: parse_instance_health(instance, instance_number)
-      })
+    Enum.map(instances, fn instance ->
+      case parse_instance_number(instance) do
```
fabriziosestito, line 88: if we return value or nil from this function, the case is not needed since ecto would guard against blank/nil values (the field is required in the defcommand)

fabriziosestito (changes requested):
overall lgtm, some comments on the parsing functions for a cleaner code

lib/trento/application/integration/discovery/policies/sap_system_policy.ex
arbulu89: Nice!

arbulu89:
Thanks @fabriziosestito . Changes applied!

PR #386, opened 2022-04-20
Use provider checks execution

fabriziosestito:
This PR adds the missing cloud provider to events/aggregate state so we can use it when requesting a check execution.
Also, it adds a provider parameter in the checks service, removing the hardcoded value.


nelsonkopliku (approved):

arbulu89 (approved):
Great!

PR #394, opened 2022-04-21
Add connection settings api

nelsonkopliku:
Adds
```
GET /api/clusters/7965f822-0254-5858-abca-f6e8b4c27714/connection-settings

[
    {
        "host_id": "13e8c25c-3180-5a9a-95c8-51ec38e50cfc",
        "hostname": "vmhdbdev01",
        "user": "pippo"
    },
    {
        "host_id": "0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4",
        "hostname": "vmhdbdev02",
        "user": null
    }
]
```

and
```
POST /api/clusters/7965f822-0254-5858-abca-f6e8b4c27714/connection-settings
```

Request body
```
{
    "settings": [
        {
            "host_id": "13e8c25c-3180-5a9a-95c8-51ec38e50cfc",
            "user": "pippo"
        },
        {
            "host_id": "0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4",
            "user": null
        }
    ]
}
```
Response body (same as when retrieving)
```
[
    {
        "host_id": "13e8c25c-3180-5a9a-95c8-51ec38e50cfc",
        "hostname": "vmhdbdev01",
        "user": "pippo"
    },
    {
        "host_id": "0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4",
        "hostname": "vmhdbdev02",
        "user": null
    }
]
```


lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -51,6 +55,48 @@ defmodule Trento.Clusters do
...
   end
 
+  @spec get_hosts_connection_settings(String.t()) :: [
+          %{
+            host_id: String.t(),
+            hostname: String.t(),
+            user: String.t()
+          }
+        ]
+  def get_hosts_connection_settings(cluster_id) do
```
arbulu89, line 65: We already have a `get_connection_settings` function in the hosts usecase. It takes the host id as param, that's true.
Now I see this, I wonder if we should maybe remove that one, and only use this, as they do more or less the same, and we even have the `cluster_id` in the place where the old one is used:
https://github.com/trento-project/web/blob/main/lib/trento/application/event_handlers/checks_event_handler.ex#L25
https://github.com/trento-project/web/blob/main/lib/trento/application/usecases/hosts/hosts.ex#L42

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -51,6 +55,48 @@ defmodule Trento.Clusters do
...
+    Repo.all(query)
+  end
+
+  @spec save_hosts_connection_settings([
+          %{
+            host_id: String.t(),
+            user: String.t()
+          }
+        ]) :: :ok
+  def save_hosts_connection_settings(settings) do
```
arbulu89, line 83: I think this should go in the `hosts` usecase in any case

lib/trento_web/controllers/cluster_controller.ex
```diff
@@ -55,4 +55,34 @@ defmodule TrentoWeb.ClusterController do
         |> json(%{error: reason})
     end
   end
+
+  @spec get_connection_settings(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def get_connection_settings(conn, %{"cluster_id" => cluster_id}) do
+    settings = Clusters.get_hosts_connection_settings(cluster_id)
```
arbulu89, line 61: Should we have some `non happy path` scenario handling here?

lib/trento_web/controllers/cluster_controller.ex
```diff
@@ -55,4 +55,34 @@ defmodule TrentoWeb.ClusterController do
...
+    settings
+    |> Enum.map(&map_to_struct/1)
+    |> Clusters.save_hosts_connection_settings()
+
+    get_connection_settings(conn, %{"cluster_id" => cluster_id})
+  end
+
+  defp map_to_struct(map) do
+    for {key, val} <- map, into: %{} do
+      {String.to_atom(key), val}
```
arbulu89, line 85: As far as I know this `String.to_atom` is a non go. It create some vulnerability issues:
http://www.moxleystratton.com/when-to-use-elixirs-string-to-atom/

In any case, I don't really understand why we need this `map_to_struct` method

lib/trento_web/router.ex
```diff
@@ -64,6 +64,12 @@ defmodule TrentoWeb.Router do
          ClusterController,
          :request_checks_execution
 
+    get "/clusters/:cluster_id/connection-settings", ClusterController, :get_connection_settings
+
+    post "/clusters/:cluster_id/connection-settings",
```
arbulu89, line 69: Being purist, I see this more like a `put` rather than a `post`, as we do not create new entries most of the cases, we just update them

arbulu89 (commented):
LGTM in general. Added some comments

lib/trento_web/controllers/cluster_controller.ex
nelsonkopliku: the unhappy path here is something really bad with the database, which will raise.
There's no `{:error, something}` path here.

dottorblaster (approved):

lib/trento/application/usecases/clusters/clusters.ex
nelsonkopliku: Yes I saw that, wanted to fix, but it's a kinda out of scope for what I am doing.
Feel free to fix that, otherwise I'll get back to that next round.

lib/trento_web/router.ex
nelsonkopliku: I see the point, changed.

So, are we gonna be purists? Fine with me :trollface:

lib/trento/application/usecases/clusters/clusters.ex
arbulu89: Ok, fair enough. As we don't even use the cluster_id to save, and we only use host related data, I see it more in the host use case. But well, it doesn't really care, right? jeje

lib/trento_web/controllers/cluster_controller.ex
nelsonkopliku: I needed that because what comes in from an http request is a map
```
[
  %{"host_id" => "13e8c25c-3180-5a9a-95c8-51ec38e50cfc", "user" => "pippo"},
  %{"host_id" => "0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4", "user" => nil}
]
```

and I want to the inner layers to have structs
```
[
  %{host_id: "13e8c25c-3180-5a9a-95c8-51ec38e50cfc", user: "pippo"},
  %{host_id: "0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4", user: nil}
]
```

Gonna do that without using `String.to_atom`

nelsonkopliku: changed to 
```
defp map_to_struct(%{"host_id" => host_id, "user" => user}) do
  %{host_id: host_id, user: user}
end
```

arbulu89: Most probably @fabriziosestito has more insights no this in any case hehe

PR #400, opened 2022-04-21
Remove heartbeat command deduplication

fabriziosestito:
Remove premature optimization which was actually causing race conditions in the heartbeat logic (my bad, part of the first POC).
Erlang processes are super-lightweight so there is no need to prevent hitting the host aggregate with the heartbeat. The event is still emitted when the state changes.


dottorblaster (approved):
I love this

PR #409, opened 2022-04-22
Add scheduled clusters checks execution

fabriziosestito:
This PR adds the scheduled checks execution for each cluster. Defaults to 5 min and overridable by `RUNNER_INTERVAL` env variable (in minutes).

NOTE: little debt taken by using Mock instead of Mox in tests, tracking it.


arbulu89 (approved):
Perfect!!
By this 1st iteration is more than enough!!!

PR #418, opened 2022-04-25
Make Dockerfile registry configurable

arbulu89:
Unfortunately, in the public space, we need to put the `registry.suse.com` prefix to find the containers.
In our OBS setup, this is not needed, so I have make the `REGISTRY` option an ARG, so we can set it empty in OBS, while we keep a correct value here


arbulu89:
Closing the PR, as we agreed to have 2 dockerfiles for the 2 use cases

PR #430, opened 2022-04-26
Discover azure admin user

### LINKED ISSUE #419: "cloudadmin" set as default as connection settings user in azure
The `cloudadmin` value is hardcoded for azure, which is wrong (this is a name we use in our team, nothing else).

We can get the default OS user from the cloud discovery data.

https://github.com/trento-project/web/blob/main/lib/trento/application/usecases/clusters/clusters.ex#L116

The needed data is coming in the `cloud_discovery`, at `Metadata/compute/osProfile/adminUserName`

fabriziosestito:
Fixes #419


arbulu89 (approved):
Go!

PR #435, opened 2022-04-26
Selected checks are now reactive

fabriziosestito:
As discussed with @abravosuse, now we send an event to the front-end when selected checks are updated.


nelsonkopliku (approved):

abravosuse (approved):
I approve this feature!

PR #436, opened 2022-04-26
Add notification message for non supported providers for checks

arbulu89:
When we don't have checks for the deployed provider (out of gcp, aws and azure), a notification error is given to the user.

(in this example, gcp is used on purpose, as in the mocked runner, we don't have a catalog for gcp, but in production it would return the catalog for azure,gcp and aws). The 2nd selection is for azure, to show that the catalog is there for known providers (maybe we could put the provider somewhere there anyway)
demo

### DIFF
--- a/assets/js/components/ClusterDetails/ChecksSelection.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelection.jsx
@@ -36,11 +36,14 @@ export const ChecksSelection = ({ clusterId, cluster }) => {
   const isSelected = (check_id) =>
     selectedChecks ? selectedChecks.includes(check_id) : false;
 
-  const [[catalogData], catalogError, loading] = useSelector((state) => [
-    state.catalog.data,
-    state.catalog.error,
-    state.catalog.loading,
-  ]);
+  const [[catalogData], catalogErrorCode, catalogError, loading] = useSelector(
+    (state) => [
+      state.catalog.data,
+      state.catalog.errorCode,
+      state.catalog.error,
+      state.catalog.loading,
+    ]
+  );
 
   const { saving, savingError, savingSuccess } = useSelector(
     (state) => state.clusterChecksSelection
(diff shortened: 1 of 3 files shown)


rtorrero (approved):
LGTM, paired on this

assets/js/components/ClusterDetails/ChecksSelection.jsx
```diff
@@ -102,6 +105,20 @@ export const ChecksSelection = ({ clusterId, cluster }) => {
...
+      <NotificationBox
+        icon={<EOS_ERROR className="m-auto" color="red" size="xl" />}
+        text={
+          <ReactMarkdown
+            className="markdown"
+            remarkPlugins={[remarkGfm]}
+          >{`Provider \`${cluster?.provider}\` does not support checks execution`}</ReactMarkdown>
+        }
+      />
+    );
```
dottorblaster, line 119: Why do we need ReactMarkdown if here we don't have any markdown to render?

nelsonkopliku (approved):

assets/js/components/ClusterDetails/ChecksSelection.jsx
nelsonkopliku: Good point! 
I believe Markdown has been used to render a backticked provider like
`gcp`
`aws`

Yes, sounds a bit overkill to use markdown here.
We could make another ui choice and just style the provider in a simpler way.

bold, red, whatever :smile:

abravosuse:
Looks good to me.

assets/js/components/ClusterDetails/ChecksSelection.jsx
nelsonkopliku: <b>text</b>
<strong>text</strong>
<p className="font-semibold-or-something-in-tailwind-doc">text</p>

:slightly_smiling_face:

arbulu89: Thanks @dottorblaster and @nelsonkopliku . I will revisit this. @rtorrero is a big fan of codeblock, that's why we finally used that, but I can retry simply with bold hehe

dottorblaster: @arbulu89 no problem, you can either retry with bold or have a switch case that basically maps the constant `gcp`, `aws` and so on to plain cloud provider names. No need to display the raw constant!

arbulu89: @dottorblaster The issue with bold is that I need to escape the `<b>` tags, otherwise they are added as simple text.
For that, I would need to import the npm `html-react-parser`. Is it alright?

dottorblaster: I don't think so, shall we merge this so we can iterate in a second moment?

arbulu89: It works as it is. The styling is a minor thing in my opinion at the moment. The code blocs is pretty visible. And we can easily iterate later on

PR #437, opened 2022-04-26
Use real sites instead of placeholders

fabriziosestito:
Removes placeholders in cluster details page and shows the right sites


nelsonkopliku (approved):

abravosuse (commented):
Bravo! and thank you...

PR #448, opened 2022-04-27
Handle aws clusters empty lists

arbulu89:
Handle properly ha cluster messages coming from hosts running on AWS and GCP.

Basically, some fields are coming with `null`, but we expect to have empty list. Generally, these fields are transformed using the "workaround" `to_list` function (which should be removed once we do proper validation).

Besides that, some new handling for the virtual ip parsing and new test scenarios

PD: The gcp test payload is a bit strange, as the cluster was broken when it was obtained, but it should still work


rtorrero (approved):
LGTM, looks like a fair solution for now

dottorblaster (approved):

nelsonkopliku (approved):

PR #450, opened 2022-04-27
Fix double scrollbars

dottorblaster:
Peek 27-04-2022 12-52

### DIFF
--- a/assets/css/app.css
+++ b/assets/css/app.css
@@ -18,3 +18,7 @@
 .table-col-m {
   width: 20rem;
 }
+
+html {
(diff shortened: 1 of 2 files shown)


fabriziosestito:
:shipit:

stefanotorresi:
Is the fixed footer with content flowing under it a deliberate decision? To be honest I don't think it completely works, but if it's not easy to fix, let's leave it at that and follow up during the next iteration.

dottorblaster:
@stefanotorresi you are right, I am reverting this ASAP.

PR #451, opened 2022-04-27
Filter correctly the check results by host

arbulu89:
Filter properly the check results. Otherwise, I will pick the results from the first node, and they will always be the same.

This error is pretty obvious when 1 node is reachable and the other not.

image


nelsonkopliku (approved):

fabriziosestito (approved):
meh, my bad, thanks for looking it up

PR #454, opened 2022-04-27
Fix alerts event handler ArgumentError when disabled

fabriziosestito:
🔢


fabriziosestito:
> Looks good! I can run it in my deployment to confirm if it fixes the issue, if you want, before merging. Let me know

@arbulu89 sure! thanks!

arbulu89:
@fabriziosestito Tested. I don't see that error message any more!

arbulu89:
@fabriziosestito Tested. I don't see that error message any more!

PR #467, opened 2022-04-28
Update the usage of the spinner icon during checks execution

arbulu89:
Use the EOS loading image. I have removed the initial `Spinner` component, as the we can easily use the eos one.
test


abravosuse (approved):
cool

fabriziosestito (changes requested):
Are we sure we want to remove the spinner component?
It seems that now to have a spinner you have to remember which class to apply to the icon and so on, components are a nice way to remove boiler code like this.

assets/js/components/ClusterDetails/ChecksResults.jsx
```diff
@@ -47,24 +47,8 @@ const sortHosts = (hosts = []) => {
...
-      return <EOS_LENS_FILLED className="fill-jungle-green-500" />;
-    case 'warning':
-      return <EOS_LENS_FILLED className="fill-yellow-500" />;
-    case 'critical':
-      return <EOS_LENS_FILLED className="fill-red-500" />;
-    default:
-      return <EOS_LENS_FILLED className="fill-gray-500" />;
-  }
+const getResultIcon = (execution_state, health) => {
+  return <ExecutionIcon health={health} execution_state={execution_state} />;
```
nelsonkopliku, line 51: For what I know usually in the `js` ecosystem `camelCase` is preferred to `snake_case`, and I also prefer it :smile: 

I am aware we are actually using `snake_case` somewhere else in the fronted codebase, and that's mainly because the api responses are snake case (see redux states, sagas and basically whenever we use json data from apis).

If we can use camelCase here, it'd be preferable.

assets/js/components/ClusterDetails/ExecutionIcon.jsx
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+  switch (execution_state) {
+    case 'requested':
+      return (
+        <EOS_SCHEDULE className={computedCssClass('fill-gray-500', centered)} />
+      );
+    case 'running':
+      return (
+        <EOS_LOADING_ANIMATED
```
nelsonkopliku, line 24: I would have kept the Spinner semantic, maybe simply wrapping this Icon as a firs iteration.

arbulu89: Ok them!
I will wrap this piece of code in the Spinner component

PR #472, opened 2022-04-28
Redirect to login on 401 status responses

rtorrero:
This PR takes care of redirecting the user to the login page when any of the API calls to the trento web server returns a 401 response state.

This is particularly useful in the event of a server restart (e.g. for an update) while a user is connected as it prevents the user from interacting with trento and changing it's state just to lose it after he realizes he needs to login again.

To test this, `config :trento, :api_key_authentication, enabled: true` is needed
redirect

### DIFF
--- a/assets/js/lib/network/index.js
+++ b/assets/js/lib/network/index.js
@@ -1,0 +1,62 @@
+import axios from 'axios';
+import { logError, logWarn } from '@lib/log';
+
+const conf = {
+  validateStatus: (status) => {
+    return status < 500;
+  },
+};
+
+export const axiosPost = function (url, data) {
+  return axios
+    .post(url, data, conf)
+    .then(handleResponseStatus)
+    .catch((error) => {
+      logError(error);
+    });
+};
+
+export const axiosDelete = function (url) {
+  return axios
+    .delete(url, conf)
+    .then(handleResponseStatus)
+    .catch((error) => {
+      logError(error);
+    });
+};
+
+export const axiosPut = function (url, data) {
+  return axios
+    .put(url, data, conf)
+    .then(handleResponseStatus)
+    .catch((error) => {
+      logError(error);
+    });
+};
+
+export const axiosGet = function (url) {
+  return axios
+    .get(url, conf)
+    .then(handleResponseStatus)
+    .catch((error) => {
+      logError(error);
+    });
+};
+
+function handleResponseStatus(response) {
+  if (response.status < 400) {
+    return response;
+  }
+  switch (response.status) {
+    case 401:
+    case 403:
+      logWarn('Redirecting to login after status', response.status);
+      window.location.href = '/session/new';
+      break;
+
+    default:
+      logError(response.statusText);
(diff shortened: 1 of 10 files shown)


dottorblaster:
Well this is something potentially good, but we have a ton of repeated code :smile:

We can pursue two approaches here:
- We don't care about calls and let them crash, and we handle this case inside an asynchronous saga task (more on that later)
- We handle every call with a proper then/catch, but we have to make an API client to centralize this kind of error management by writing a library.

Your call :smile: 

The asynchronous saga task is pretty simple, you can see an example [here](https://redux-saga.js.org/docs/advanced/TaskCancellation), basically my view about that is:
- We start a loop and make an HTTP request every minute or two minutes to an endpoint that we know is behind auth
- If we get a 401/403, we redirect the user to the login

If we want to pursue the API client path, instead, we have to make a library out of `axios.get`/`axios.post`, wrapping those promises and returning them, then import the functions all around the app.

abravosuse (commented):
The feature itself looks awesome!

assets/js/lib/network/index.js
```diff
@@ -0,0 +1,86 @@
...
+  validateStatus: (status) => {
+    return status < 500;
+  },
+};
+
+export const axiosPost = function (url, data) {
+  return new Promise((resolve, reject) => {
+    axios
+      .post(url, data, conf)
+      .then((response) => {
```
dottorblaster, line 14: I think you can avoid returning a new promise because `axios.post` already returns a promise, so basically you can `return axios.post(...).then(handleResponseStatus)`.

What do you think?

dottorblaster (changes requested):
I think we are good to go with this, just a tiny comment and another thing, but maybe we can address this in a follow-up PR: referring as these functions as `axiosGet` is a bit weird, I think you can name them just `get`, `post` or whatever :sweat_smile:

assets/js/lib/network/index.js
rtorrero: Sure, returning a new promise is a bit more explicit but it does indeed add unneeded code

arbulu89 (commented):
Hey @rtorrero ,
Did you check if the catalog and checks usage still works fine?
We get the catalog information from the runner, and if this doesn't work (we know this if certain call returns a 400 return code) we should specific notifications messages with the error.

I don't think we should bypass this errors and send them back to the login session.

Could you confirm this?

rtorrero:
> Hey @rtorrero , Did you check if the catalog and checks usage still works fine? We get the catalog information from the runner, and if this doesn't work (we know this if certain call returns a 400 return code) we should specific notifications messages with the error.
> 
> I don't think we should bypass this errors and send them back to the login session.
> 
> Could you confirm this?

It shouldn't be affected by this. Any other status response should behave as before. The only case where it would affect is if the runner did return a 401/403 incorrectly, but it doesn't seem so.

dottorblaster (approved):
I think we are good to merge for now

PR #476, opened 2022-04-29
Include cluster field in the host details view

arbulu89:
Cluster field always was empty:

Linkable cluster:
image

Non linkable cluster:
image

### DIFF
--- a/assets/js/state/selectors/index.js
+++ b/assets/js/state/selectors/index.js
@@ -69,6 +69,11 @@ const enrichInstances = (instances, sapSystemId, state) =>
       };
     });
 
+export const getClusterByHost = (hostId) => (state) => {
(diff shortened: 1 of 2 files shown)


assets/js/state/selectors/index.js
```diff
@@ -69,6 +69,11 @@ const enrichInstances = (instances, sapSystemId, state) =>
       };
     });
 
+export const getClusterOnHost = (hostId) => (state) => {
```
nelsonkopliku, line 72: We don't have a cluster on a host, rather a host part of a cluster.
I believe naming should reflect this.

What about `getHostCluster` or `getClusterByHost`?

nelsonkopliku (changes requested):
Little naming issue :trollface:

nelsonkopliku (approved):

PR #486, opened 2022-05-02
Update dockerfile and add missing nodejs to tools

stefanotorresi:
This PR mainly splits the dependency installation steps of the `Dockerfile` from the building steps, so that changing a source file doesn't invalidate the cached dependency layers of both mix and npm, which will results in much faster image builds.

It also fixes a version mismatch in the first stage, where we were using a SLE repo on top of the leap image, and updates the final stage to BCI 15.4.

The initial stage, called `system`, is now a full-fledged development image which can be used to develop fully within containers.

Some changes to Mix tasks were required to make all of this seamless:
- the start command no longer starts docker-compose, since we want to enable running `mix` from within the development container.
- a new `reset` task was added, which clears all databases. I'm not sure if we should do also something about Grafana, there.
- a new `install` task compunds mix deps and npm deps installation together.

Finally, the missing nodejs local dependency in `.tools-versions` was added.


mix.exs
```diff
@@ -91,14 +91,13 @@ defmodule Trento.MixProject do
   defp aliases do
     [
       start: [
-        "cmd docker-compose up -d",
-        "deps.get",
-        "ecto.create",
-        "ecto.migrate",
-        "event_store.setup",
+        "install",
+        "setup",
```
dottorblaster, line 95: Could you restore the `cmd docker-compose up -d`?

mix.exs
```diff
@@ -91,14 +91,13 @@ defmodule Trento.MixProject do
...
         "phx.server"
       ],
-      setup: ["deps.get", "event_store.setup", "ecto.setup", "init_grafana_dashboards"],
+      install: ["deps.get", "cmd --cd assets npm install"],
```
dottorblaster, line 98: I'm not sure the team wants `npm` to be part of the mixfile, I think we should vote on this :sweat_smile:

Dockerfile
```diff
@@ -1,39 +1,33 @@
...
-ENV MIX_ENV=prod
-RUN mix phx.digest
-RUN mix release
+FROM deps AS full
+COPY . /app
+RUN mix assets.deploy &&\
+    mix release
 
-FROM registry.suse.com/bci/bci-base:15.3 AS trento
+FROM registry.suse.com/bci/bci-base:15.4 AS minimal
```
dottorblaster, line 23: I think `minimal` is not a meaningful name, maybe switching to `final-stage` or something could be a good idea

Dockerfile
```diff
@@ -1,39 +1,33 @@
...
+COPY assets/package*.json /app/assets/
+RUN mix deps.all
 
-FROM elixir-build AS release
-COPY --from=assets-build /build /build
-WORKDIR /build
-ENV MIX_ENV=prod
-RUN mix phx.digest
-RUN mix release
+FROM deps AS full
```
dottorblaster, line 18: `release` was a good name :sweat_smile:

.tool-versions
```diff
@@ -1,2 +1,3 @@
 elixir 1.13.1-otp-24
-erlang 24.3.3
\ No newline at end of file
+erlang 24.3.3
+nodejs 16.14.1
```
dottorblaster, line 3: Thanks for this :raised_hands:

dottorblaster (changes requested):
Some tiny things to change, IMHO

mix.exs
stefanotorresi: Hmm, I thought it was okay since it's already there for the `assets.deploy` command.

If we don't want npm there, what's the highest order task manager? I understand we don't want a Makefile.

mix.exs
stefanotorresi: That's deliberate, as explained above, as it would make impossible to use `mix start` from within the container.

mix.exs
dottorblaster: You're right, keep it that way.

Dockerfile
stefanotorresi: The rationale about `full` and `minimal` is that both could be used as final images. One only has the static binary, the other has the full development toolkit.
For the demo environment, for instance, we will probably want to use the `full` variant, so that we can tinker with `mix` tasks and the code on a live installation.

mix.exs
fabriziosestito: some of us were relying on this, maybe we can wrap it in another task and add the docker-compose up there?

mix.exs
fabriziosestito: +1 to keep npm install, mix should also take care of this kind of stuff

Dockerfile
fabriziosestito: nitpick: i also like release better too, i find the naming proposed a bit confusing

mix.exs
dottorblaster: Well, I strongly advice against developing inside a container mainly because connecting the Erlang debugger to a running containerized instance is nightmarish :/

fabriziosestito (approved):
Thanks for this! LGTM, some minor comments added

mix.exs
dottorblaster: Oh @fabriziosestito didn't think about this, actually that would be a good "goats and cabbages" solution :sweat_smile:

Dockerfile
fabriziosestito: @stefanotorresi gotcha, thanks for clarifing

mix.exs
stefanotorresi: Ok, so what do we tell people? Should they just run `mix start` or should they run `docker-compose up -d` first?

Have a look at https://github.com/trento-project/web/blob/2cca604bdf9d57a3f7e43ad755987abe9aeab280/docs/development/hacking_trento.md#docker-and-docker-compose .

How do you imagine documenting the first steps? Because as it was, we gave people explicit instructions to run compose first, with an explicit command.

Dockerfile
dottorblaster: @stefanotorresi well your rationale was good but I didn't catch it at a first glance, so... :sweat_smile:

dottorblaster (approved):

Dockerfile
stefanotorresi: let me copy the `ENTRYPOINT` and `EXPOSE` statements in there as well, so that we make this more apparent.

PR #489, opened 2022-05-03
Add development docker-compose service

stefanotorresi:
- reinstate mix task starting compose services
- add the 'web' service in docker-compose and further updates the Dockerfile, following up to #486
- tweak docker context
- move everything related to env vars to runtime.exs
- add helper config module to enable falling back to mix env-specific config settings.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,7 @@
+/.github
+/.pgdata/
 /_build/
 /assets/node_modules/
 /deps/
-/doc/
 /priv/static/assets
-/test/
 /tmp/
-/.pgdata/
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -9,13 +9,13 @@ import Trento.Config
 
 config :trento, Trento.Repo,
   url:
-    System.get_env("DATABASE_URL") || "postgres://postgres@localhost:5433/#{db_name("trento")}",
+    System.get_env("DATABASE_URL") || "postgres://postgres@localhost:5432/#{db_name("trento")}",
   pool_size: String.to_integer(System.get_env("DATABASE_POOL_SIZE") || "10")
 
 config :trento, Trento.EventStore,
   url:
     System.get_env("EVENTSTORE_URL") ||
-      "postgres://postgres@localhost:5433/#{db_name("trento_eventstore")}",
+      "postgres://postgres@localhost:5432/#{db_name("trento_eventstore")}",
   pool_size: String.to_integer(System.get_env("EVENTSTORE_POOL_SIZE") || "10")
 
 config :trento, TrentoWeb.Endpoint,
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -18,22 +18,6 @@ config :trento, Trento.EventStore,
       "postgres://postgres@localhost:5433/#{db_name("trento_eventstore")}",
   pool_size: String.to_integer(System.get_env("EVENTSTORE_POOL_SIZE") || "10")
 
-if config_env() == :prod do
-  # The secret key base is used to sign/encrypt cookies and other secrets.
-  # A default value is used in config/dev.exs and config/test.exs but you
-  # want to use a different value for prod and you most likely don't want
-  # to check this value into version control, so we use an environment
-  # variable instead.
-  secret_key_base =
-    System.get_env("SECRET_KEY_BASE") ||
-      raise """
-      environment variable SECRET_KEY_BASE is missing.
-      You can generate one by calling: mix phx.gen.secret
-      """
-
-  config :trento, TrentoWeb.Endpoint, secret_key_base: secret_key_base
-end
-
 config :trento, TrentoWeb.Endpoint,
   http: [
     # Enable IPv6 and bind on all interfaces.
@@ -82,18 +66,13 @@ config :trento, :grafana,
 # See https://hexdocs.pm/swoosh/Swoosh.html#module-installation for details.
 
 config :trento, :alerting,
-  enabled: System.get_env("ENABLE_ALERTING") == "true",
   recipient: System.get_env("ALERT_RECIPIENT") || "admin@trento-project.io"
 
 config :trento, Trento.Mailer,
-  adapter: Swoosh.Adapters.SMTP,
   relay: System.get_env("SMTP_SERVER") || "",
   port: System.get_env("SMTP_PORT") || "",
   username: System.get_env("SMTP_USER") || "",
-  password: System.get_env("SMTP_PASSWORD") || "",
-  auth: :always,
-  ssl: :if_available,
-  tls: :if_available
+  password: System.get_env("SMTP_PASSWORD") || ""
 
 config :trento, Trento.Scheduler,
   jobs: [
@@ -102,3 +81,23 @@ config :trento, Trento.Scheduler,
       schedule: "*/#{System.get_env("RUNNER_INTERVAL", "5")} * * * *"
     ]
   ]
+
+# prod specific semantics
+if config_env() == :prod do
+  # The secret key base is used to sign/encrypt cookies and other secrets.
+  # A default value is used in config/dev.exs and config/test.exs but you
+  # want to use a different value for prod and you most likely don't want
+  # to check this value into version control, so we use an environment
+  # variable instead.
+  secret_key_base =
+    System.get_env("SECRET_KEY_BASE") ||
+      raise """
+      environment variable SECRET_KEY_BASE is missing.
+      You can generate one by calling: mix phx.gen.secret
+      """
+
+  config :trento, TrentoWeb.Endpoint, secret_key_base: secret_key_base
+
+  # alerting is usually enabled by default, but it's the opposite in prod
+  config :trento, :alerting, enabled: System.get_env("ENABLE_ALERTING") == "true"
+end
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -8,11 +8,14 @@ import Trento.Config
 # any compile-time configuration in here, as it won't be applied.
 
 config :trento, Trento.Repo,
-  url: System.get_env("DATABASE_URL") || "postgres://postgres@localhost:5433/#{db_name("trento")}",
+  url:
+    System.get_env("DATABASE_URL") || "postgres://postgres@localhost:5433/#{db_name("trento")}",
   pool_size: String.to_integer(System.get_env("DATABASE_POOL_SIZE") || "10")
 
 config :trento, Trento.EventStore,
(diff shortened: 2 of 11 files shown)


docker-compose.yaml
```diff
@@ -1,21 +1,53 @@
-version: "3"
+version: "3.9"
```
nelsonkopliku, line 1: I am on LEAP 15.3 and probably my docker/compose is older than yours :sweat_smile:
 
```
ERROR: Version in "./docker-compose.yaml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

```
➜  trento git:(docker-compose) docker-compose version
docker-compose version 1.25.1, build a82fef0
docker-py version: 4.2.0
CPython version: 3.6.15
OpenSSL version: OpenSSL 1.1.1d  10 Sep 2019
```

docker-compose.yaml
```diff
@@ -1,21 +1,53 @@
...
+      GRAFANA_PUBLIC_URL: http://grafana:3000
+      GRAFANA_API_URL: http://grafana:3000/api
+      RUNNER_URL: http://runner:8000
+    volumes:
+      - .:/app
+      - build_cache:/app/_build
+    ports:
+      - 4000:4000
+
+  runner:
```
nelsonkopliku, line 24: Usually the mocked runner adapter we have in the codebase is more than enough for me in development.

Locally we work with mocked data, that refer to some hosts that this runner might try to connect to in order to run checks, and it will fail.

What usage of the real runner container am I missing here? :thinking:
What would be the added value?

config/runtime.exs
```diff
@@ -1,36 +1,24 @@
 import Config
+import Trento.Config
 
 # config/runtime.exs is executed for all environments, including
 # during releases. It is executed after compilation and before the
 # system starts, so it is typically used to load production configuration
 # and secrets from environment variables or elsewhere. Do not define
 # any compile-time configuration in here, as it won't be applied.
-# The block below contains prod specific runtime configuration.
-if config_env() == :prod do
```
nelsonkopliku, line 9: I believe the intent here is to raise errors when some critical config is missing when running the application in `:prod`.
And I personally would keep that semantic instead of letting the app use arbitrarily default set by us.
When you run the app in prod mode, that is your complete responsibility to provide the correct config/env imho.

In addition this is the community standard afaiu :smile:

nelsonkopliku (changes requested):
I appreciate the contribution, thanks!

Left some comments.

.dockerignore
```diff
@@ -1,8 +1,7 @@
+/.github
```
rtorrero, line 1: Do we want to ignore this? Isn't this where we have our ci.yaml & co?

docker-compose.yaml
rtorrero: I have: 
``` ~  docker-compose version
Docker Compose version 2.4.1```
@stefanotorresi did we really need this bump for any of the other changes you brought? if so, which?

Comment (nelsonkopliku) [reply docker-compose.yaml:1]: Very nice reference :smile: 
However, what's the proposed solution?

With `version: "3.7"` I am able to make a step forward, but I get this. Some permissions on that repo?
```
➜  trento git:(docker-compose) ✗ docker-compose up -d
Pulling runner (ghcr.io/trento-project/trento-runner:1.0.0)...
ERROR: Head "https://ghcr.io/v2/trento-project/trento-runner/manifests/1.0.0": denied: denied
```

Comment (nelsonkopliku) [reply docker-compose.yaml:1]: So with `version 3.7` I can proceed, but when entering the web container like this 
https://github.com/trento-project/web/blob/f4b684b6be82a03b26596feca0721cde46aa759f/docs/development/hacking_trento.md?plain=1#L53
I have this issue
```
➜  tronto git:(docker-compose) ✗ docker-compose run -it -p 4000:4000 web
Run a one-off command on a service.

For example:

    $ docker-compose run web python manage.py shell

By default, linked services will be started, unless they are already
running. If you do not want to start linked services, use
`docker-compose run --no-deps SERVICE COMMAND [ARGS...]`.

Usage:
    run [options] [-v VOLUME...] [-p PORT...] [-e KEY=VAL...] [-l KEY=VALUE...]
        SERVICE [COMMAND] [ARGS...]

Options:
    -d, --detach          Detached mode: Run container in the background, print
                          new container name.
...
```
The following works instead `docker-compose run -p 4000:4000 web`

Maybe again related to different docker-compose version?

File: docker-compose.yaml (hunk 2 above)
Comment (stefanotorresi) [reply docker-compose.yaml:24]: I don't know how to use the mock runner and it's not documented anywhere, so this was the best thing I could come up with to get started with running the entire thing. 🙂 

Anyway this is not mutually exclusive with the mock, so I think we can keep both ways to run it, can't we?
If you think it's not that useful for your daily work, I will remove the service it from the new `mix start.full` task so that it doesn't bother you.

Comment (nelsonkopliku) [reply docker-compose.yaml:24]: Here's the mock runner 
https://github.com/trento-project/web/blob/main/lib/trento/application/integration/checks/adapter/mock_runner.ex
and it is enabled by default in `dev`.

> I think we can keep both ways to run it

Sure we could, I am trying to understand why, what's adding to DX.

The runner executes checks on targets, so unless we actually have some targets to hit, I miss other usages.
And if, for some reason there's the need to actually hit some targets, that means being working closer on that context, thus on the runner itself, so another service/repo.

Is the reason integration tests?
If so I believe that in the current scenario that kind of integration test is responsibility of the runner itself.
In relation to testing I believe we should consider contract testing here.

With these information, I am for removing it, and I'd be happy to change my mind if there's anything I am missing.

File: config/runtime.exs (hunk 3 above)
Comment (stefanotorresi) [reply config/runtime.exs:9]: The problem `MIX_ENV` is giant switch with a cascading side-effect on many other parameters at once, both at compile-time and at run-time, and as I've always vocally argued that this is a huge operational anti-pattern, very common in many major frameworks (Rails, Django, Symfony, they all share this outdated macro-config-switch-of-doom approach, and I was underwhelmed to see this in Elixir as well, but I guess they were inspired by Rails 😬).

The "environment" as a concept should never be distilled into a single variable; it's like a value object: it's identified by the all the values it's composed by, and it has no other identity than the sum of all those.
IOW, having a "development environment" means having a bunch of runtime configuration values that are used during development, and that's it. There shouldn't be a `dev` environment; in fact having a "development mode single switch" that flips many settings at once is a huge PITA, because it couples configuration values with a compounded semantic that might should instead be independently controllable for each value.

Let me make a practical example: there is nothing about using `localhost` that tells you it's production, development or testing. If people decided to run multiple containers in the same network namespace, because reasons, then they would use `localhost`, and whether that's production or not is not something we can predict.
Assuming the environment from the configuration values is a backwards model that has origin in the age of fully local development, when we could afford assumptions about how people ran things. This is no longer a safe assumption to make: all kinds of environments could be distributed, nowadays. In fact you could even do distributed development and then monolithic deployments!

The real reason of `MIX_ENV` to exist is to determine how the sources are built, rather than how they are ran. This is why `runtime.exs` was only relatively recently [introduced in Elixir](https://elixir-lang.org/blog/2020/10/06/elixir-v1-11-0-released/#configruntimeexs-and-mix-appconfig).


> I believe the intent here is to raise errors when some critical config is missing when running the application in :prod.

That is still the case, but the only critical config now raising due to missing value is `"SECRET_KEY_BASE"`, because it's not secure to provide a default there.
All the other ones are fine: using wrong values will result in an error, so if the defaults are wrong the get an error, as they should be.


> When you run the app in prod mode, that is your complete responsibility to provide the correct config/env imho.

That is still the case, but see the point above about not making assumptions about what is correct or not in any given environment. The validation logic should be the same everywhere.


> In addition this is the community standard afaiu smile

FWIW I did some research but I couldn't find an established convention about how to write configs; the way it's done is entirely dependant on each single project.

A lot of projects only limit themselves to a single `config.exs` file for compile-time settings.
Some people scatter things across files without logic and even repeat the same `System.get_env()` call with the same fallback value in every single file. 🙄 
I took inspiration from https://github.com/plausible/analytics. @dottorblaster agreed it looked like a good way forward.

Maybe @fabriziosestito will shed some light on this.

File: docker-compose.yaml (hunk 1 above)
Comment (stefanotorresi) [reply docker-compose.yaml:1]: Hmm, could be. I'm not sure. By the way, your version of docker-compose is 3 years old!! They have completely rewritten it, so I strongly advise to remove the python version and get the newest one from upstream: https://github.com/docker/compose#linux

**Comment by fabriziosestito** [comment]
@stefanotorresi Thanks for this, overall I embrace the concept of loading the config from env files and making the runtime behave the same way and not depending on the MIX_ENV. 
Interesting conversation as well regarding the config (which is the biggest Elixir quirk of all).

I've read through the comments, keep in mind that whoever runs the container cannot change the MIX_ENV since the release is "statically" built against the prod environment and `mix` itself is not present. The fact that said environments exist is completely opaque for the final user and just a commodity we reserve to ourself to provide the application with "sane" defaults.

What I usually did for other projects is to remove the big `if prod` as you already did, but I kept the `prod/dev/config` defaults in the respective `exs` files. 
Then I added a `Config helper` which was used in the `runtime.exs` to get either the env variable or the default, here is the snippet:

```elixir
defmodule Trento.Config do
  def get_env(path, env_var, required \\ false) do
    value = System.get_env(env_var) || :trento |> Application.get_all_env() |> get_in(path)

    if required and not value do
      raise "Missing required environment variable: #{env_var}"
    end

    value
  end
end
```

This way we can leave the `config/prod/dev` unchanged, with visible defaults (which is handy), but still leverage the `runtime.exs` to load the ENV variables as you already proposed.

config/runtime.exs
nelsonkopliku: Fair enough, interesting perspective, thanks for sharing :rocket:
With this addition and Fabrizio's comment I believe we can find a middle ground to go.

docker-compose.yaml
nelsonkopliku: > your version of docker-compose is 3 years old!
>
:man_facepalming: oh man...

Look, I was anyway thinking of upgrading my workstation soon, so this is just a reason more for me :trollface: 

But seriously, I am not sure if `version 3.7` would be a temporary acceptable tradeoff, and I don't want to force to use _old_ stuff just because of my system.

I'll temporarily continue with the current workflow I have and get up to date soon.

stefanotorresi:
Thanks for the pointer, Fabrizio. In fact, I was already aiming at that direction with a dedicated helper module for config stuff.

I will keep working on it. I don't know what's up with the E2E test failing, though, because on my local workstation it doesn't fail consistently, sometimes it passes, on subsequent runs.

PR #492, opened 2022-05-04
Use get-latest action to avoid issues with git after CVE-2022-24765

arbulu89:
The version of the container is incorrectly found. Fix for that.

Fix for the git CVE-2022-24765. More info in:
https://github.blog/2022-04-12-git-security-vulnerability-announced/
https://github.com/actions/checkout/issues/760

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -314,8 +314,12 @@ jobs:
 
       - name: Checkout
         uses: actions/checkout@v3
+
+      - uses: actions-ecosystem/action-get-latest-tag@v1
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):

stefanotorresi:
Using the last tag as `VERSION` with `osc` is not enough, otherwise we're gonna do multiple OBS commits with the same version. This is why the python script appended a `+dev{N}-SHA` suffix.

arbulu89:
@stefanotorresi The container delivery system works in a different way. We cannot add the commit sha there, as it is not permitted.

PR #495, opened 2022-05-05
Reactive subscription

fabriziosestito:
Adds reactiveness to the subscription section on the host details page, and move it to the bottom of the page.


arbulu89 (approved):
Perfect!
At some point we might decide to hide even more the subscriptions table somehow, as it uses a lot of space and it is not that important. This for the future though hehe

nelsonkopliku (approved):

dottorblaster (approved):

Issue #190, opened 2022-03-17
sap_systems or sap-systems


fabriziosestito:
API and react routes have low dash and dash

choose your destiny

rtorrero:
Done!

PR #496, opened 2022-05-05
Snake case all the routes

rtorrero:
This PR is a proposal to address #190 
Take it also as an opportunity to suggest other changes on the API and react routes if any other change makes sense.


dottorblaster:
Actually it is snake case but it's ok I guess :smile:

nelsonkopliku (approved):

PR #503, opened 2022-05-09
Adjust z-index inside the Modal component

dottorblaster:
Just increasing the z-index because otherwise everything falls behind.


nelsonkopliku (approved):
:shipit:

PR #504, opened 2022-05-09
Include premium field handling

arbulu89:
Include the premium field from the runner catalog in the DTOs, so we can store and visualize this field in the frontend.
This way, we show if a check is premium or not in the checks catalog view.

Besides that, I have included a small improvement in the CI, in the OBS commit task, to update the changes file in a new release. This very well could go in a different PR, but as initially I thought we would need a release yes or yes after this change, I added it, but finally this issue was not a roadblocker.

### DIFF
--- a/test/trento/application/integration/checks/checks_test.exs
+++ b/test/trento/application/integration/checks/checks_test.exs
@@ -70,7 +70,8 @@ defmodule Trento.Integration.ChecksTest do
           labels: "labels",
           name: "test 1",
           provider: :azure,
-          remediation: "remediation 1"
+          remediation: "remediation 1",
+          premium: true
         },
         %FlatCheckDto{
           description: "description 2",
@@ -80,7 +81,8 @@ defmodule Trento.Integration.ChecksTest do
           labels: "labels",
           name: "test 2",
           provider: :azure,
-          remediation: "remediation 2"
+          remediation: "remediation 2",
+          premium: nil
         },
         %FlatCheckDto{
           description: "description 3",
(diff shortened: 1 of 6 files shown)


test/trento/application/integration/checks/checks_test.exs
```diff
@@ -80,7 +81,8 @@ defmodule Trento.Integration.ChecksTest do
           labels: "labels",
           name: "test 2",
           provider: :azure,
-          remediation: "remediation 2"
+          remediation: "remediation 2",
+          premium: nil
```
nelsonkopliku, line 85: Since we defined the property as `:boolean` I would use `false` here.
What is the reason we have in `nil`? :thinking:

nelsonkopliku (commented):
Great!
I left a little comment on `nil` values.

I also see that an e2e test is failing, I am afraid it is flaky, it fails in other apparently unrelated PRs as well.

I'm going to investigate a bit on that.

test/trento/application/integration/checks/checks_test.exs
arbulu89: If we don't send the `premium` field in the incoming json, as it is not required, our code sets it as `nil`, as this is the default behaviour of our code (maybe inherited from ecto). 

I could try to make the false option the default one if the field is not coming though

rtorrero (commented):
LGTM but I agree with @nelsonkopliku comment about using `false` instead of `nil`. As you said, it would make sense to store it as `false` when the field doesn't arrive in the JSON payload.

arbulu89:
@nelsonkopliku @rtorrero Premium flag set to false by default: https://github.com/trento-project/web/pull/504/commits/a6aae1cd5cf8600a73675ec6420ae30180271b57

nelsonkopliku (approved):

PR #505, opened 2022-05-09
Exporter status endpoint

fabriziosestito:
This PR Adds back the prometheus exporters status API


rtorrero:
I'm probably missing lacking some context here, but just in case: is it expected that any host id on `/api/hosts/<id>/exporters_status` returns 200 OK and `{"Node Exporter":"passing"}` ?

fabriziosestito:
> I'm probably missing lacking some context here, but just in case: is it expected that any host id on `/api/hosts/<id>/exporters_status` returns 200 OK and `{"Node Exporter":"passing"}` ?

if you are testing this in dev this is expected since a mock adapter is configured for this environment

nelsonkopliku (approved):

PR #512, opened 2022-05-10
Initial Swagger integration

nelsonkopliku:
This PR is the inception for exposing our APIs version 1.0.0.

This very first iteration aims to:
- introduce the library being used https://hexdocs.pm/open_api_spex/readme.html
- provide a PoC of how a basic integration works
- provide ground for future iterations

As of now only the Hosts List API is documented. 

Here's how that looks like (nothing new :smile:)

https://user-images.githubusercontent.com/8167114/167700728-33b390fc-c292-4634-8e62-cfc7294f3a5a.mp4

Help me please with defining what should be exposed and what not.
Here's a quick idea.

Expose
- [x] GET     /api/hosts
- [ ] GET     /api/clusters
- [ ] GET     /api/sap_systems
- [ ] GET     /api/databases
- [ ] POST    /api/clusters/:cluster_id/checks
- [ ] POST    /api/clusters/:cluster_id/checks/request_execution
- [ ] GET     /api/checks/catalog
- [ ] POST    /api/runner/callback

Not sure
- POST    /api/collect
- GET     /api/installation/api-key
- GET     /api/sap_systems/health
- GET     /api/clusters/:cluster_id/connection_settings
- PUT     /api/clusters/:cluster_id/connection_settings
- POST    /api/hosts/:id/tags
- DELETE  /api/hosts/:id/tags/:value
- POST    /api/clusters/:id/tags
- DELETE  /api/clusters/:id/tags/:value
- POST    /api/sap_systems/:id/tags
- DELETE  /api/sap_systems/:id/tags/:value
- POST    /api/databases/:id/tags
- DELETE  /api/databases/:id/tags/:value  

No Doc
- GET     /api/about
- GET     /api/settings
- POST    /api/accept_eula
- POST    /api/hosts/:id/heartbeat
- GET     /api/prometheus/targets

Extra considerations:
at the moment we're adding schemas only, later we'll add also DTOs on top of readmodels to decouple layers.
We can start by having those DTOs and respective OpenApi schemas next to each other (maybe in the same file) and since we're still not sure how much of a burden it will become to maintain, when and if needed we might want to add some macro magic like `def_openapi_schema` (or any better name) to link indeed DTOs and schemas.

Again this is a first iteration, let's see how that evolves. :v:


lib/trento_web/openapi/schema/host.ex
```diff
@@ -0,0 +1,97 @@
...
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
+      title: "IPv6",
+      type: :string,
+      format: :ipv6
+    })
+  end
+
+  defmodule Item do
```
arbulu89, line 27: Maybe call it `Host` or `HostItem` better?

lib/trento_web/openapi/schema/host.ex
```diff
@@ -0,0 +1,97 @@
...
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
+      title: "Host",
+      description: "A discovered host on the target infrastructure",
+      type: :object,
+      properties: %{
+        id: %Schema{type: :integer, description: "Host ID"},
+        # pattern: ~r/[a-zA-Z][a-zA-Z0-9_]+/
```
arbulu89, line 36: Leftover comment?

arbulu89 (approved):
Hey @nelsonkopliku ,

The functionality is exactly as expected. Perfect!

As you comment, the need of defining all the fields back again here looks pretty annoying. It looks like something where we easily could make mistakes. And again, as you comment, having some sort of logic which takes the data models from the code to create the spec would be the ideal scenario. I don't know having the DTOs would directly enable this, but it looks something interesting to explore.

From me side, we can merge this as initial iteration, and continue exploring the option to "auto" generate the models in next iterations.

lib/trento_web/openapi/schema/host.ex
```diff
@@ -0,0 +1,97 @@
...
+        },
+        provider: %Schema{
+          type: :string,
+          description: "Detected Provider on which the host is running",
+          enum: [:azure, :aws, :gcp, :unknown]
+        },
+        provider_data: TrentoWeb.OpenApi.Schema.Provider.ProviderData,
+        tags: %Schema{
+          title: "Tags",
+          description: "A list of tags attached to a resource",
```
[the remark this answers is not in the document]
nelsonkopliku: We tag several resources next to hosts: clusters, sap systems, databases.
This same schema could be used for those scenarios.

Does it make sense?

lib/trento_web/openapi/schema/host.ex
nelsonkopliku: yes, also a reminder on how to use it :stuck_out_tongue: 
Removing.

fabriziosestito:
@arbulu89 |
> As you comment, the need of defining all the fields back again here looks pretty annoying. It looks like something where we easily could make mistakes.

I understand the concern but keep in mind this is done once for API version for documentation. 
Once we go public with a v1 we will not change the API contract till v2, so I don't think this will be a problem.

lib/trento_web/openapi/schema/host.ex
nelsonkopliku: going with `HostItem` :+1:

dottorblaster (approved):
I think we are ready to merge

PR #513, opened 2022-05-10
Revert "Add development docker-compose service"

arbulu89:
Reverts trento-project/web#489

Reverting due issues found in released version.
Basically, the `Applicationg.get_all_env` in the `fallback` function doesn't work, as the application env are not laoded during the runtime in the `Task.init` task.

After this revert, we need to revert: https://github.com/trento-project/web/pull/486


stefanotorresi:
can't we just revert both PRs at once in the same branch?

PR #515, opened 2022-05-11
Debounce Homepage health overview refresh

nelsonkopliku:
e2e tests in our pipeline are failing on the `dashboard.js` test. Seems to be flaky.

A possible cause might be the fact that the dashboard gets refreshed whenever any of the following things happen
```
HOST_REGISTERED
CLUSTER_REGISTERED
DATABASE_REGISTERED
SAP_SYSTEM_REGISTERED
HEARTBEAT_FAILED
HEARTBEAT_SUCCEDED
DATABASE_HEALTH_CHANGED
SAP_SYSTEM_HEALTH_CHANGED
CLUSTER_HEALTH_CHANGED
```
In addition, when we run photofinish loading the `healthy-27-node-SAP-cluster` a lot of those thing happen, triggering something like ~40 refreshes, hence ~40 api calls that stack one after another.

By adding debouncing in that scenario we reduced the API calls to ~5.

Not sure it fixes, yet might be anyway helpful :smile:

### DIFF
--- a/assets/js/state/sagas/index.js
+++ b/assets/js/state/sagas/index.js
@@ -4,8 +4,8 @@ import {
   all,
   call,
   takeEvery,
-  takeLatest,
   select,
+  debounce,
 } from 'redux-saga/effects';
 import { urlEncode, keysToCamel } from '@lib/serialization';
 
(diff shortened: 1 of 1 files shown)


stefanotorresi (approved):
LGTM.

dottorblaster (approved):
Impressive, LGTM

PR #517, opened 2022-05-11
Correct log replication and log operation mode names

arbulu89:
Fix the `HANA log replication mode` and `HANA log operation mode` names:

image


nelsonkopliku (approved):
On fireeeeee :fire:

dottorblaster (approved):
I just took what we had in the old one :older_man: :sweat_smile: LGTM

abravosuse (approved):
That was fast!

PR #519, opened 2022-05-11
Add checks catalog e2e tests

arbulu89:
Add checks catalog view E2E tests (ported from legacy trento mostly)

PR #520, opened 2022-05-11
Add complete e2e testing for the cluster details view

rtorrero:
This PR adds testing (except for a few minor things that we need to fix first, see `TODO`) for the cluster details view.


rtorrero:
> If the CI is happy, I am happy as well
CI definitely not happy about this :disappointed: ... I'll set this to draft while I figure out why

nelsonkopliku (changes requested):
Nice!.

Would it be possible to port also the `.feature` file?
```gherkin
Feature: HANA cluster details
    The detail page of a HANA cluster

    Background:
        Given an SAP HANA scale-up cluster with name "hana_cluster_3" and id "9c832998801e28cd70ad77380e82a5c0"
        Given two sites with one host each are part of the cluster
        Given checks were performed on the clusters resulting in 2 checks in critical state,
    2 checks in warning state and 2 checks in passing state

    Scenario: HANA cluster details are the expected ones
        When I navigate to the Pacemaker Clusters Overview (/clusters)
        Then the displayed cluster at a glance detauks should be consistent with the state of the cluster

    Scenario: Cluster health count should show the expected checks status
        When I check the health summary of the cluster
        Then the critical count should be 2
        Then the warning count should be 2
        Then the passing count should be 2

    Scenario: Check results modal should show the expected checks
        When I click on the check results button
        Then the check results modal should be displayed
        And the check results checks list and results should show the expected checks

    Scenario: Check results modal filter works as expected
        When I filter the check results by warning or passing status
        And the check results checks should be the ones in warning or passing status

    Scenario: Cluster sites section should have the expected hosts
        When I scroll to the site section
        Then it should show two sites with one host each
        And the hosts should have the expected host "name", "ip", "virtual ip" and "role"

    Scenario: Host details modal should show the expected host details
        When I click on the host details button
        Then the host details modal should be displayed
        And the "Attribute" tab should show the expected host attributes
        And the "Resources" tab should show the expected host resources

    Scenario: Cluster SBD should have the expected devices and status
        When I scroll to the SBD section
        Then it should show the expected SBD devices name
        And they should have the expected health

```

test/e2e/cypress/integration/hana_cluster_details.js
```diff
@@ -0,0 +1,215 @@
...
+    cy.loadScenario("healthy-27-node-SAP-cluster");
+    cy.login();
+
+    cy.visit(`/clusters/${availableHanaCluster.id}`);
+    cy.url().should("include", `/clusters/${availableHanaCluster.id}`);
+  });
+
+  describe("HANA cluster details should be consistent with the state of the cluster", () => {
+    it(`should have name ${availableHanaCluster.name}`, () => {
+      cy.get("div")
```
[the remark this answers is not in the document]
rtorrero: I get your point but I'm not sure I agree on this one; I do think the placement is something worth testing for.

As the UI gets more features, it's quite possible that we will get `.contains()` to match unwanted elements in the UI (it already happened to me). The possibility of using CSS classes is there indeed and would prevent this, but it feels a bit that we are abusing the classes.

Between: 
 - Matching a label and looking for the next item
 - Using CSS classes to target the element we want to test
 
 I'd find the first one the lesser evil :smile:. More opinions on this?

rtorrero:
> ```
> Feature: HANA cluster details
>     The detail page of a HANA cluster
> 
>     Background:
>         Given an SAP HANA scale-up cluster with name "hana_cluster_3" and id "9c832998801e28cd70ad77380e82a5c0"
>         Given two sites with one host each are part of the cluster
>         Given checks were performed on the clusters resulting in 2 checks in critical state,
>     2 checks in warning state and 2 checks in passing state
> 
>     Scenario: HANA cluster details are the expected ones
>         When I navigate to the Pacemaker Clusters Overview (/clusters)
>         Then the displayed cluster at a glance detauks should be consistent with the state of the cluster
> 
>     Scenario: Cluster health count should show the expected checks status
>         When I check the health summary of the cluster
>         Then the critical count should be 2
>         Then the warning count should be 2
>         Then the passing count should be 2
> 
>     Scenario: Check results modal should show the expected checks
>         When I click on the check results button
>         Then the check results modal should be displayed
>         And the check results checks list and results should show the expected checks
> 
>     Scenario: Check results modal filter works as expected
>         When I filter the check results by warning or passing status
>         And the check results checks should be the ones in warning or passing status
> 
>     Scenario: Cluster sites section should have the expected hosts
>         When I scroll to the site section
>         Then it should show two sites with one host each
>         And the hosts should have the expected host "name", "ip", "virtual ip" and "role"
> 
>     Scenario: Host details modal should show the expected host details
>         When I click on the host details button
>         Then the host details modal should be displayed
>         And the "Attribute" tab should show the expected host attributes
>         And the "Resources" tab should show the expected host resources
> 
>     Scenario: Cluster SBD should have the expected devices and status
>         When I scroll to the SBD section
>         Then it should show the expected SBD devices name
>         And they should have the expected health
> ```

Actually, I just noticed we already have it (though based on the old one) under `test/fixtures/scenarios/hana-cluster-details`. Should we port it under this file or create a new directory under the `cypress` directory, as @arbulu89 did in #519 ?

test/e2e/cypress/integration/hana_cluster_details.js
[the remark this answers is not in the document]
nelsonkopliku: >  I do think the placement is something worth testing for.
>
I see your point, and I believe it is a matter of **value**, **perception of value** and **flexibility**.

Here's my argument.
_Sticking on the example, and porting the reasoning to other stuff._
In the current scenario, from my perspective, checking that there in the page a component with the cluster name is more valuable than checking that it is exactly next another one.

Why do I find it more valuable in the current situation? 
Because that would allow us to experiment on the UI in a more efficient way.
We'd still have the value of a test that makes sure a specific information is in the page, with less of a burden when we move things around.

Certainly depends on cases, this is a general recommendation and heuristic.


> It feels a bit that we are abusing the classes.
>
What do you mean? Why do you see it as abusing classes? 
is it because they're not tied to actual styles? :sweat_smile: 
If this is the reason I'd say that:
1. I see that as an acceptable tradeoff considering the previous reasons
2. The alternative would be using [html data attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes), which I see more expensive

> I'd find the first one the lesser evil smile
>
Why? Could you elaborate a bit more?

test/e2e/cypress/support/commands.js
```diff
@@ -56,3 +56,21 @@ Cypress.Commands.add('navigateToItem', (item) => {
   const items = Array.isArray(item) ? item : [item];
   items.forEach((it) => cy.get('.tn-menu-item').contains(it).click());
 });
+
+Cypress.Commands.add('disableAllChecks', (clusterId) => {  
+  cy.visit(`/clusters/${clusterId}/settings`);
+
+  // Find all active switches and disable them
+  cy.get('button[role="switch"]').each(($el) => {
+    if (cy.wrap($el).should('have.class', 'bg-jungle-green-500')) {
```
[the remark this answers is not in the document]
nelsonkopliku: Don't have a working example, you might want to google a bit more on html attributes.

Anyway, I checked and we could use the `aria-checked` attribute on the button.

If you check in the browser inspector on what happens on the switch when you click it you'll see that the value of `aria-checked` toggles from `true` to `false`.

I'd give it a try.

nelsonkopliku:
Bytw, the gherkin might need to be modified to reflect what we have currently

rtorrero:
I've simplified a bit the tests, hope it's enough @nelsonkopliku

test/e2e/cypress.json
```diff
@@ -2,6 +2,7 @@
   "baseUrl": "http://127.0.0.1:4000",
   "viewportWidth": 1366,
   "viewportHeight": 768,
+  "defaultCommandTimeout": 10000,
```
nelsonkopliku, line 5: Let's keep an eye on this. If we can avoid high timeouts, I believe it'd be nice :smile:

test/e2e/cypress/integration/hana_cluster_details.js
```diff
@@ -0,0 +1,133 @@
+import { availableHanaCluster } from '../fixtures/hana-cluster-details/available_hana_cluster';
+
+context('HANA database details', () => {
+  before(() => {
+    //cy.loadScenario('healthy-27-node-SAP-cluster');
```
nelsonkopliku, line 5: What about this? Is it needed or a leftover?

nelsonkopliku (approved):
LGTM

test/e2e/.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
arbulu89, line 1: @rtorrero Why is adding this files as new? They were already in the project.
I think something strange happened during the rebase

.github/workflows/ci.yaml
```diff
@@ -217,7 +217,7 @@ jobs:
         run: mix deps.clean --all && mix deps.get
 
       - name: Mix setup
-        run: mix setup
+        run: mix do ecto.drop, event_store.drop, ecto.create, ecto.migrate, setup
```
arbulu89, line 220: Did we need this `drop` thing?
The database should be coming empty.
Just to avoid not needed steps, as we are aiming to reduce the execution time at some point

PR #526, opened 2022-05-13
Add default value to os_version

fabriziosestito:
In my development machine, the os_version detected by the agent is nil. (I use Arch, btw)
This PR adds a default value to the field, so the discovery does not fail.


nelsonkopliku (approved):
Arch btw :smile:

dottorblaster (approved):

PR #532, opened 2022-05-16
Use ExMachina as test data generation factory for host telemetry

arbulu89:
Here a nitpick example of how [ExMachina](https://github.com/thoughtbot/ex_machina) usage looks like.

I just wanted to show a small change, so we all agree that this is the way to go. If we like it, I can continue modifying the rest of the factories. 

The work seems pretty straightforward.

PD: For people not familiar with ExMachina, I would recommend to have a look in the project README, as I find that it has important "hidden magic" functionalities, like attributes merging

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -54,6 +54,7 @@ defmodule Trento.MixProject do
       {:eventstore, "~> 1.1",
        [env: :prod, git: "https://github.com/commanded/eventstore.git", override: true]},
       {:eventstore_dashboard, github: "commanded/eventstore-dashboard"},
+      {:ex_machina, "~> 2.7.0", only: :test},
       {:faker, "~> 0.17", only: [:dev, :test]},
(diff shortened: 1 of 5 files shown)


fabriziosestito:
LGTM I vote yes to exmachina, also we could split factories in events, commands, readmodels, models

fabriziosestito (approved):

arbulu89:
> LGTM I vote yes to exmachina, also we could split factories in events, commands, readmodels, models (but let's start with a single factory for now fi you agree)

Yes. I would go with a single factory module by now. It looks pretty easy, and we don't have that many factories, so it is not hard to find them or create new ones.

By the way, nice stuff this ExMachina, easy to use and well documented. Basically, it helps to remove a lot of boilerplate we have in our factories

nelsonkopliku:
Late to the party :sweat_smile: 
Nice btw!

PR #533, opened 2022-05-16
Decode host discovery payload with Ecto

dottorblaster:
Just decoding Host Discovery payload through Ecto.


fabriziosestito (approved):
LGTM!

lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
```diff
@@ -0,0 +1,20 @@
+defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
```
arbulu89, line 1: In the checks code we used `DTO` instead of `Payload` to refer to this kind of messages.
Just as a comment

dottorblaster: @fabriziosestito @nelsonkopliku what do you think about this? Shall I rename this to DTO?

PR #534, opened 2022-05-16
Fix validate required for embedded fields

fabriziosestito:
This PR fixes the validate required option for embedded fields and removes ecto warnings when running tests

### DIFF
--- a/lib/trento/support/type.ex
+++ b/lib/trento/support/type.ex
@@ -63,36 +63,46 @@ defmodule Trento.Type do
       validates the required ones and returns a changeset.
       """
       def changeset(struct, params) do
-        {embedded_fields, fields} =
-          :fields
-          |> __MODULE__.__schema__()
-          |> Enum.split_with(fn field ->
-            case __MODULE__.__schema__(:type, field) do
-              {_, Ecto.Embedded, _} ->
-                true
-
-              _ ->
-                false
-            end
-          end)
-
-        changes =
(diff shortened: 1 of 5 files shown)


lib/trento/support/type.ex
```diff
@@ -63,36 +63,46 @@ defmodule Trento.Type do
...
+      def cast_and_validate_required_embed(changeset, field, nil),
+        do: cast_embed(changeset, field)
+
+      def cast_and_validate_required_embed(changeset, field, :all),
+        do: cast_embed(changeset, field, required: true)
+
+      def cast_and_validate_required_embed(changeset, field, required_fields),
+        do: cast_embed(changeset, field, required: field in required_fields)
+
+      defp fields, do: __MODULE__.__schema__(:fields) -- __MODULE__.__schema__(:embeds)
```
arbulu89, line 103: Just curiosity.
What does this double `--` do here?

arbulu89 (approved):
Super!!!

nelsonkopliku (approved):

lib/trento/support/type.ex
fabriziosestito: differences between lists,
`[1, 2, 3, 4] -- [2, 4] == [1, 3]`
basically it is saying give me all the non-embedded fields

dottorblaster (approved):
Mega LGTM

PR #528, opened 2022-05-13
Sap systems and Databases openapi

nelsonkopliku:
- [x] GET     /api/hosts
- [x] GET     /api/clusters
- [x] GET     /api/sap_systems
- [x] GET     /api/databases
- [ ] GET     /api/checks/catalog
- [ ] POST    /api/clusters/:cluster_id/checks
- [ ] POST    /api/clusters/:cluster_id/checks/request_execution
- [ ] POST    /api/runner/callback

For extra information see here 
https://github.com/trento-project/web/pull/512
https://github.com/trento-project/web/pull/522

PR #536, opened 2022-05-16
Checks catalog openapi

nelsonkopliku:
Added Checks Catalog OpenApi documentation.

https://user-images.githubusercontent.com/8167114/168640015-d136e520-1962-4b56-bbc8-d1b4c024db11.mp4

The `flat` option fits a bit strangely in the open api spec.
It might be worth considering making that more standard and actually making it a boolean.

- [x] GET     /api/hosts
- [x] GET     /api/clusters
- [x] GET     /api/sap_systems
- [x] GET     /api/databases
- [x] GET     /api/checks/catalog
- [ ] POST    /api/clusters/:cluster_id/checks
- [ ] POST    /api/clusters/:cluster_id/checks/request_execution
- [ ] POST    /api/runner/callback

Follows up https://github.com/trento-project/web/pull/512 https://github.com/trento-project/web/pull/522 https://github.com/trento-project/web/pull/528

### DIFF
--- a/lib/trento_web/controllers/catalog_controller.ex
+++ b/lib/trento_web/controllers/catalog_controller.ex
@@ -3,6 +3,35 @@ defmodule TrentoWeb.CatalogController do
 
   alias Trento.Integration.Checks
 
+  alias TrentoWeb.OpenApi.Schema.{ChecksCatalog, Provider}
+
+  use OpenApiSpex.ControllerSpecs
+
+  tags(["Checks"])
+
+  operation(:checks_catalog,
+    summary: "Checks Catalog",
+    description:
+      "The list of the available checks that can be configured to run on the target SAP infrastructure",
+    parameters: [
+      flat: [
+        in: :query,
+        type: :string,
+        description:
+          "Whether to output a flat catalog or not. Just provide the flag, not the value. eg /api/checks/catalog?flat"
+      ],
+      provider: [
+        in: :query,
+        type: Provider.FilterableProviders,
+        description: "Whether to filter by a specific provider"
+      ]
+    ],
+    responses: [
+      ok: {"A Collection of the available Checks", "application/json", ChecksCatalog.Catalog},
+      not_found: {"Not found", "application/json", ChecksCatalog.CatalogNotfound}
+    ]
+  )
+
   @spec checks_catalog(Plug.Conn.t(), map) :: Plug.Conn.t()
   def checks_catalog(conn, params) do
     with {:ok, content} <- get_catalog(params),
--- a/lib/trento_web/controllers/catalog_controller.ex
+++ b/lib/trento_web/controllers/catalog_controller.ex
@@ -28,7 +28,8 @@ defmodule TrentoWeb.CatalogController do
     ],
     responses: [
       ok: {"A Collection of the available Checks", "application/json", ChecksCatalog.Catalog},
-      not_found: {"Not found", "application/json", ChecksCatalog.CatalogNotfound}
+      not_found: {"Not found", "application/json", ChecksCatalog.CatalogNotfound},
+      bad_request: {"Bad Request", "application/json", ChecksCatalog.UnableToLoadCatalog}
     ]
   )
 
--- a/lib/trento_web/openapi/schema/checks_catalog.ex
+++ b/lib/trento_web/openapi/schema/checks_catalog.ex
@@ -1,0 +1,178 @@
+defmodule TrentoWeb.OpenApi.Schema.ChecksCatalog do
+  @moduledoc false
+
+  require OpenApiSpex
+  alias OpenApiSpex.Schema
+
+  alias TrentoWeb.OpenApi.Schema.Provider
+
+  defmodule Check do
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
+      title: "Check",
+      description: "An available check to be executed on the target infrastructure",
+      type: :object,
+      properties: %{
+        sap_system_id: %Schema{type: :string, description: "Check ID", format: :uuid},
+        name: %Schema{type: :string, description: "Check Name"},
+        description: %Schema{type: :string, description: "Check Description"},
+        remediation: %Schema{type: :string, description: "Check Remediation"},
+        implementation: %Schema{type: :string, description: "Check Implementation"},
+        labels: %Schema{type: :string, description: "Check Labels"},
+        premium: %Schema{
+          type: :boolean,
+          description: "Indicates whether the current check is a Premium check"
+        },
+        group: %Schema{
+          type: :string,
+          description: "Check Group, available when requiring a Flat Catalog"
+        },
+        provider: Provider.SupportedProviders
+      }
+    })
+  end
+
+  defmodule FlatCatalog do
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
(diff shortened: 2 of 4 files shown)


lib/trento_web/controllers/catalog_controller.ex
```diff
@@ -3,6 +3,35 @@ defmodule TrentoWeb.CatalogController do
...
+          "Whether to output a flat catalog or not. Just provide the flag, not the value. eg /api/checks/catalog?flat"
+      ],
+      provider: [
+        in: :query,
+        type: Provider.FilterableProviders,
+        description: "Whether to filter by a specific provider"
+      ]
+    ],
+    responses: [
+      ok: {"A Collection of the available Checks", "application/json", ChecksCatalog.Catalog},
```
arbulu89, line 30: Are we deliveratetely not putting the `bad_request` here?
For example, we have the `bad_request` with a `not_ready` content for the moment when the catalog is being built.
This would eventually be `%{error: :not_ready}`

lib/trento_web/openapi/schema/checks_catalog.ex
```diff
@@ -0,0 +1,178 @@
...
+
+  defmodule Check do
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
+      title: "Check",
+      description: "An available check to be executed on the target infrastructure",
+      type: :object,
+      properties: %{
+        sap_system_id: %Schema{type: :string, description: "Check ID", format: :uuid},
```
arbulu89, line 17: `sap_system_id` -> `check_id` I guess

lib/trento_web/openapi/schema/checks_catalog.ex
```diff
@@ -0,0 +1,178 @@
...
+
+  defmodule ProviderCatalog do
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
+      title: "ProviderCatalog",
+      description: "A Provider specific Catalog, and respective values",
+      type: :object,
+      properties: %{
+        provider: %Schema{
```
arbulu89, line 69: Cannot we use `Provider.SupportedProviders` here like above?

lib/trento_web/openapi/schema/checks_catalog.ex
```diff
@@ -0,0 +1,178 @@
...
+        labels: %Schema{type: :string, description: "Check Labels"},
+        premium: %Schema{
+          type: :boolean,
+          description: "Indicates whether the current check is a Premium check"
+        },
+        group: %Schema{
+          type: :string,
+          description: "Check Group, available when requiring a Flat Catalog"
+        },
+        provider: Provider.SupportedProviders
```
arbulu89, line 31: Where is this `SupportedProviders` specified?

lib/trento_web/openapi/schema/provider.ex
```diff
@@ -15,6 +15,18 @@ defmodule TrentoWeb.OpenApi.Schema.Provider do
...
   end
 
+  defmodule FilterableProviders do
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
+      title: "FilterableProvider",
+      type: :string,
+      description: "A provider that can be used to filter the Catalog",
+      # default?
```
arbulu89, line 25: I would personally include `default` here, as we actually have the data, but I understand that is not really meaningful.

arbulu89 (changes requested):
Great!
I have noted some small issues to be corrected, together with other doubts.

lib/trento_web/controllers/catalog_controller.ex
nelsonkopliku: Yes, my brain deliberately forgot about that :sweat_smile: 
Thanks!

nelsonkopliku: There it is dbfaff12a4f4844640730a0fa3198442dd5eec09

image

lib/trento_web/openapi/schema/checks_catalog.ex
nelsonkopliku: That's defined in `lib/trento_web/openapi/schema/provider.ex:7`

```elixir
defmodule SupportedProviders do
  @moduledoc false

  OpenApiSpex.schema(%{
    title: "SupportedProviders",
    type: :string,
    description: "Detected Provider where the resource is running",
    enum: [:azure, :aws, :gcp, :unknown]
  })
end
```
And now that I see it I think:
- should we consider `default` part of that list?
- is `:unknown` actually a _SupportedProvider_?

I believe we can iterate more on this details.

lib/trento_web/openapi/schema/checks_catalog.ex
nelsonkopliku: I though about that, too.

But then I wouldn't have been able to provide a custom description for this specific case (or at least I wasn't able to find a way to do so)

```elixir
description:
    "The provider determining the values for the attached checks (azure, aws ...)",
```

I'd be fine with losing the description specificity, that'd be an acceptable tradeoff.
Using `SupportedProviders` as it is now would be a tradeoff as well I believe, because of the `:unknown` value

```elixir
defmodule SupportedProviders do
  @moduledoc false

  OpenApiSpex.schema(%{
    title: "SupportedProviders",
    type: :string,
    description: "Detected Provider where the resource is running",
    enum: [:azure, :aws, :gcp, :unknown]
  })
end
```

I think I will fix what I did
```elixir
provider: %Schema{
  title: "ChecksProvider",
  type: :string,
  description:
    "The provider determining the values for the attached checks (azure, aws ...)",
  # ??
  enum: [:azure, :aws, :gcp, :unknown]
},
```
by removing `unknown` and adding `default`.
Would that make sense?

lib/trento_web/openapi/schema/provider.ex
nelsonkopliku: Fine to me, adding it.

nelsonkopliku:
@arbulu89 Thanks for the review! Another pair of eyes is always beneficial especially with this kind of work :sweat_smile: 

Addressed your comments here f3b1404595d85590cc3e136c4d0a3057a7e6fa68 and here ed5b4def74e11725a556772f4ede6cda442516d2

Should be good enough for now.

lib/trento_web/openapi/schema/checks_catalog.ex
arbulu89: Ok, let's go with this

lib/trento_web/openapi/schema/checks_catalog.ex
arbulu89: `:unknown` is not supported in the checks catalog. We have in the code as an incoming provider in hosts, as we will have hosts that are not running in our known providers, and they will have this value.
So, in this `SupportedProviders` I see it totally right.
`default` is not part of this list, as I don't see any provider option as default

lib/trento_web/openapi/schema/checks_catalog.ex
arbulu89: And in fact, you did right not using `SupportedProviders` as they are different things. The catalog has some already known providers checks, while in the discovered providers, we have the `unknown` entry, as we don't detect anything outside of the big 3 cloud providers

arbulu89 (approved):
Ready to merge from my side!

PR #542, opened 2022-05-17
Dump discarded discovery events

arbulu89:
Add the option to dump the `unaccepted` events.

It can be used running:
```
mix dump_unaccepted_events failed_events -p /tmp
```

Now, with this change, I don't know if we need to start providing some migration tools, as this adds a change in the database, and we cannot ask the users to completely remove their database now on hehe

**I have had many doubts with the naming. I have gone with `accepted` and `unaccepted`. It could be very well something else. Feel free to propose other alternative**

The code has some duplicated code in the mix tasks and in the bash script, but I thought that better to deal with this small duplication rather than adding more complexity to the code.

It includes:
- Code to dump the unaccepted events.
- Update the dump bash script to work with the 2 options. Have a look on the `usage` function. But basically it works like:
```
dump_data_from_k8s.sh scenario --name failover --path /tmp
dump_data_from_k8s.sh unaccepted-events --name failover --path /tmp --event-number 5
```
- Remove error logging in one of the prometheus exporters unit test (1st commit)


fabriziosestito (changes requested):
Thanks for looking into this! This would be an excellent improvement for debugging and auditing purposes.
However, I would propose a slightly different implementation.
I noticed that since we cannot know the shape of the discarded payloads, trying to fit them into the `discovery events table` seems not optimal for our usecase.
When a discovery event with a broken payload is received we need to use a magic UUID as an `agent_id` to store the data, and leverage an extra `accepted` field to filter out the discovery events, which for now is a boolean but in the future there could be a set of reasons.
This is coupling a bit the concept of "current scenario events" and "discarded/dead lettered payloads".

A common pattern to solve this is to leave the `discovery_events` table unchanged and add a [dead letters](https://en.wikipedia.org/wiki/Dead_letter_queue) table where to store malformed/unacceptable payloads. I image the table to have just one field for now containing the whole event, rather than try to store the payload and the metadata (agent_id, etc) in different payloads. This would allow us to keep things separated by concern and also to extend the dead letter with different reasons, levels, and retry in the future.
I would also slightly modify the dump script to pull the discovery events + 1 JSON `dead_letters.json` containing all the discarded payloads, so we provide just one command to the users which is going to output both the current scenario and the discarded messages.

arbulu89:
Thanks @fabriziosestito ,
Yep, good proposals!
I will work on them

arbulu89:
Hey @fabriziosestito ,
I have implemented the suggested changes.
1. Add a new `discarded_events` table, with the event payload and reason of the failure as columns. The `discovered_events` table continues untouched
2. I have included the dump of the discarded events in the same `dump_scenario` task, and included the needed flags in the script. It is pretty much the same. Now, it will include the discarded events in the `discarded_events.txt` file with the next format (including the whole payload of course):

```
Discarded event: 5

Reason:

%{details: %{system_replication_operation_mode: ["can't be blank"]}, host_id: ["is invalid"]}

Payload:

{"agent_id":"0d1e3b0b-b9df-5e98-976a-21f92835136","discovery_type":"ha_cluster_discovery","payload":{"Cib":{"Configuration":{"Constraints":{"RscLocations":null},"CrmConfig":{"ClusterProperties":[{"Id":"cib-bootstrap-options-have-watchdog","Name":"have-watchdog","Value":"false"},{"Id":"cib-bootstrap-options-dc-version","Name":"dc-version","Value":"2.0.1+20190417.13d370ca9-3.24.1-2.0.1+20190417.13d370ca9"},{"Id":"cib-bootstrap-options-cluster-infrastructure","Name":"cluste

...

Discarded event: 4

Reason:

%{details: %{system_replication_operation_mode: ["can't be blank"]}, host_id: ["is invalid"]}

Payload:

{"agent_id":"0d1e3b0b-b9df-5e98-976a-21f92835136","discovery_type":"ha_cluster_discovery","payload":{"Cib":{"Configuration":{"Constraints":{"RscLocations":null},"CrmConfig":{"ClusterProperties":[{"Id":"cib-bootstrap-options-have-watchdog","Name":"have-watchdog","Value":"false"},{"Id":"cib-bootstrap-options-dc-version","Name":"dc-version","Value":"2.0.1+20190417.13d370ca9-3.24.1-2.0.1+20190417.13d370ca9"},{"Id":"cib-bootstrap-options-cluster-infrastructure","Name":"cluster-infrastructure","Value":"corosync"},{"Id":"cib-bootstrap-options-stonith-enabled","Name":"stonith-enabled","Value":"true"},{"Id":"cib-bootstrap-options-stonith-action","Name
```

This way, we have all the discarded events payloads and the reasons at hand. In order to use them with photofinish, we would need to create individual `json` files, but it doesn't look a big effort

nelsonkopliku (approved):
@arbulu89 this is great progress! Thanks!

As long as we follow up by iterating on improving the `dump_scenario` task I believe this is very good as a first step.

Let's hear what the rest of the team thinks.

lib/trento/application/integration/discovery/discarded_event.ex
```diff
@@ -0,0 +1,17 @@
...
+  This module contains the schema used to store an append log of the discarded events,
+  for debugging and auditing purposes.
+  No changeset is defined here, since the schema is used to store append-only data.
+  """
+
+  use Ecto.Schema
+
+  @type t :: %__MODULE__{}
+
+  schema "discarded_events" do
```
fabriziosestito, line 12: I would rename this `discovery_events_dead_letters`.
it is a bit of a nitpick, but the pattern is broadly known, and also "discarded_events" sounds a bit generic because it is not clear to which events we are referring. For instance, we also have domain events, and in the future, we could add even more integration events.

fabriziosestito (changes requested):

lib/trento/support/mix/tasks/dump_scenario.ex
```diff
@@ -40,29 +54,47 @@ defmodule Mix.Tasks.DumpScenario do
...
+                     id: id,
+                     payload: payload,
+                     reason: reason
+                   } ->
+      data = Jason.encode!(payload)
+
+      discarded_events_file
+      |> File.write!(
+        "Discarded event: #{id}\n\nReason:\n\n#{reason}\n\nPayload:\n\n#{data}\n\n",
+        [:append]
```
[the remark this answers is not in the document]
fabriziosestito: e.g.
```json
[
{
"reason": "reason,
"payload": ....
},
{
"reason": "reason,
"payload": ....
},
]
```

lib/trento/application/integration/discovery/discarded_event.ex
arbulu89: Done: https://github.com/trento-project/web/pull/542/commits/e7a279a8027c9b5229f62f8371f6d5d6e4f645a4

lib/trento/support/mix/tasks/dump_scenario.ex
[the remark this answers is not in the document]
arbulu89: Done: https://github.com/trento-project/web/pull/542/commits/a1c89b4808dbeed07c1ad434bc0b988f9c07dcde

I have included the `id` and `inserted_at`. They might be helpful for debugging and communicating with others.

Example:
```
[
  {
    "id": 3,
    "inserted_at": "2022-05-20T08:01:19.000000Z",
    "payload": {
      "agent_id": "0d1e3b0b-b9df-5e98-976a-21f92835136",
      "discovery_type": "ha_cluster_discovery",
      "payload": {
        "Cib": {
          "Configuration": {
            "Constraints": {
...
},
    "reason": "%{details: %{system_replication_operation_mode: [\"can't be blank\"]}, host_id: [\"is invalid\"]}"
  },
  {
    "id": 2,
    "inserted_at": "2022-05-20T07:43:39.000000Z",
    "payload": {
      "agent_id": "0d1e3b0b-b9df-5e98-976a-21f92835136",
      "discovery_type": "ha_cluster_discovery",
      "payload": {
        "Cib": {
          "Configuration": {
            "Constraints": {
              "RscLocations": null
            },
...
```

hack/dump_scenario_from_k8.sh
```diff
@@ -8,15 +8,16 @@ usage() {
     cat <<-EOF
     usage: dump_scenario_from_k8s.sh options
 
-    Dump the current scenario from a running trento-server installation on the k8s cluster
+    Dump the current scenario and discarded discovery events from a running trento-server installation on the k8s cluster
 
     OPTIONS:
         -n, --name            The name to use for the scenario. Defaults to "current".
         -p, --path            The path where the scenario should be saved. Defaults to the current directory.
+        -d, --event-number    Then number of discarded events to dump. Default to 10.
```
fabriziosestito, line 16: why this? couldn't we dump just all the discarded events?

rtorrero (approved):
Nice addition @arbulu89, I was wondering about the need of something like this recently.

hack/dump_scenario_from_k8.sh
arbulu89: Considering that this action is for debugging purposes, and that most commonly, the discarded event will be repeated often, I thought that limiting it would be a good idea.

Example:
We have a discarded message sent by the cluster discovery, which by default is sent every 10 seconds. There are possibilities that the 2 nodes are sending an invalid message.

If the landscape has been working just 1 hour. We would already have a file with 720 events.
I don't think we want this hehe More if we start getting dumps with big landscape working days or weeks.

I think having the flexibility to limit the messages is a good feature

fabriziosestito: Alright, I see the rationale and I agree with keeping it, just a couple of points that I would tackle before merging:

-  increase the default 10 to 100 so it's still easy to search for interesting events but we have more of them in case 
- rename event-number to something less generic such as discarded-events-number
- add the discarded events in the pruning job so they get deleted and the table does not grow forever

fabriziosestito (approved):

PR #525, opened 2022-05-12
Migrate sap overview e2e

arbulu89:
Import SAP System overview page E2E tests, coming from the legacy trento project.

- Test view layout
- Test tagging filtering
- Test the health changes on events
- Test unknown sap instances like the diagnostic service (DAA) are not displayed

PR #545, opened 2022-05-18
About page e2e

arbulu89:
Improve the e2e tests for the about page.
I have included the version test. The version identifier is coming from a `mix` task. I'm not sure if this is a good approach. What do you think @fabriziosestito ?

Otherwise, we will need to have other option to get the version or simply remove this test.

### DIFF
--- a/lib/trento/support/mix/tasks/print_version.ex
+++ b/lib/trento/support/mix/tasks/print_version.ex
@@ -1,0 +1,10 @@
+defmodule Mix.Tasks.PrintVersion do
+  @moduledoc "The hello mix task: `mix help hello`"
+
+  use Mix.Task
+
+  @shortdoc "Print application version"
+  def run(_args) do
+    IO.puts(Mix.Project.config()[:version])
+  end
+end
--- a/lib/trento/support/mix/tasks/version.ex
+++ b/lib/trento/support/mix/tasks/version.ex
@@ -1,4 +1,4 @@
-defmodule Mix.Tasks.PrintVersion do
+defmodule Mix.Tasks.Version do
   @moduledoc "The hello mix task: `mix help hello`"
 
   use Mix.Task
--- a/test/e2e/cypress/fixtures/about/about.feature
+++ b/test/e2e/cypress/fixtures/about/about.feature
@@ -1,0 +1,22 @@
+Feature: About
+    Show Trento info in the about page
+
+    Background:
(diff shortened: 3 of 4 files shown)


rtorrero (commented):
LGTM! Just a minor comment: @nelsonkopliku was suggesting to use new CSS classes to target the views elements rather than relying on their location (e.g. using `.next()`) which is what we are using now mostly in the clusters details view

arbulu89:
> LGTM! Just a minor comment: @nelsonkopliku was suggesting to use new CSS classes to target the views elements rather than relying on their location (e.g. using `.next()`) which is what we are using now mostly in the clusters details view

Personally, I wouldn't over use the specific class for everything. In this particular case, the `next()` option checks if the `header` and `content` are properly located (one after the another), not only the containing value. This was the use case.

nelsonkopliku:
I shared my thoughts and reasoning about this topic here https://github.com/trento-project/web/pull/525#pullrequestreview-971789811 and here https://github.com/trento-project/web/pull/520#discussion_r870440310.

It is also ok if we disagree, anyway for me this is good enough to go, I am not going to be a blocker :smile:

nelsonkopliku (commented):
Overall looks good.
Only a doubt on the mix task to get the version. I am wondering if a static expected value in tests would have helped as well :thinking:

arbulu89:
> Overall looks good. Only a doubt on the mix task to get the version. I am wondering if a static expected value in tests would have helped as well thinking

Well, of course it would work, but we would need to update the constant in every release. Something that we would most probably forget at the beginning hehe.
But hey, if the mix task is not a good option we hardcode it or we remove the test.

lib/trento/support/mix/tasks/print_version.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Mix.Tasks.PrintVersion do
```
fabriziosestito, line 1: Great! This is the way! I would just rename the command to `mix version` to be consistent whit what we already have generated in the release commands.

fabriziosestito (changes requested):

lib/trento/support/mix/tasks/print_version.ex
arbulu89: Renamed in: https://github.com/trento-project/web/pull/545/commits/9e4a9f51c02625470458962f28b6317f80063923

fabriziosestito (approved):

nelsonkopliku (approved):

PR #555, opened 2022-05-19
Cluster health when no checks are selected

arbulu89:
When there is not any check selected, the cluster health is being still built aggregating the discovered health and the checks health, which is `uknown` in this case. This creates an unexpected cluster health, as it would be unknown.

With this change, when no checks are selected, the checks health is not used in the cluster health computation.

Besides that, I have improved the dashboards E2E tests to disable all checks in order to make the test more robuts

### DIFF
--- a/lib/trento/domain/cluster/cluster.ex
+++ b/lib/trento/domain/cluster/cluster.ex
@@ -158,11 +158,10 @@ defmodule Trento.Domain.Cluster do
   def execute(
         %Cluster{
           cluster_id: cluster_id,
-          selected_checks: selected_checks
+          selected_checks: []
         },
         %RequestChecksExecution{cluster_id: cluster_id}
-      )
-      when selected_checks == [],
+      ),
       do: nil
 
   def execute(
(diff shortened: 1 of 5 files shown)


lib/trento/domain/cluster/cluster.ex
```diff
@@ -136,20 +136,35 @@ defmodule Trento.Domain.Cluster do
...
 
   # Request checks execution
+  def execute(
+        %Cluster{
+          cluster_id: cluster_id,
+          selected_checks: selected_checks
+        },
+        %RequestChecksExecution{cluster_id: cluster_id}
+      )
+      when selected_checks == [],
```
fabriziosestito, line 165: the guard `when` is not needed, since you can pattern match with
```elixir
  %Cluster{
          cluster_id: cluster_id,
          selected_checks: []
        }
```

lib/trento/domain/cluster/cluster.ex
```diff
@@ -472,6 +487,18 @@ defmodule Trento.Domain.Cluster do
...
   end
 
+  defp maybe_emit_cluster_health_changed_event(%Cluster{
+         cluster_id: cluster_id,
+         discovered_health: discovered_health,
+         selected_checks: selected_checks,
+         health: health
+       })
+       when selected_checks == [] do
+    if discovered_health != health do
```
fabriziosestito, line 497: guard is not needed here.
Also, we could maybe refactor this by creating a function called `maybe_add_checks_health(checks_health, selected_cheacks)`
to pipe it in the `new_health` match in the function below to remove repetition.

```elixir
    new_health =
      [discovered_health]
      |> maybe_add_checks_health(checks_health, selected_checks)
      |> Enum.filter(& &1)
      |> HealthService.compute_aggregated_health()
```

fabriziosestito (changes requested):

arbulu89:
@fabriziosestito Suggested changes done: https://github.com/trento-project/web/pull/555/commits/86301ab655efd80a740b1a63869c8e1ce127aa15

Thanks dude!
Elixir rules!

fabriziosestito (approved):
@arbulu89 some conflicts but god to go

PR #556, opened 2022-05-19
Speed up CI process

arbulu89:
Speed up the CI execution. 3 major changes:
- Cache as much as possible. At least the re-runs will be faster
- Disable the video recording on Cypress. This boosts a lot the speed, and we don't use almost the videos. The screenshots are still there
- Disable photofinish usage and the login process in all contexts. They are done just at the beginning of the suite

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -364,6 +364,7 @@ jobs:
       - name: Cypress run
         uses: cypress-io/github-action@v2
         env:
+          cypress_video: false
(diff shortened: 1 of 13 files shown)


nelsonkopliku (approved):
FSAT! :rocket:

nelsonkopliku:
This is more appropriate :joy: 

laws-of-time

rtorrero (approved):
You just made my day better with this! LGTM

PR #568, opened 2022-05-20
Event handlers failure context 

fabriziosestito:
This PR adds a failure context for event handlers (and projectors) that can be used to retry/skip an event that couldn't be handled. The failure context accepts two callbacks (`after_retry` and `after_max_retries_reached`),

### DIFF
--- a/lib/trento/support/event_handler_failure_context.ex
+++ b/lib/trento/support/event_handler_failure_context.ex
@@ -1,0 +1,72 @@
+defmodule Trento.Support.EventHandlerFailureContext do
+  @moduledoc """
+  Event handler failure context
+
+  max_retries: max retries before the event handler is shut down (defauklt: 3)
+  retry_after: time between retries in ms (default: 0)
+  callback: callback to be called when the max retries are reached
+  skip: if skip is true, the event will be skipped, otherwise the process will stop (default: false)
+  """
+
+  defmacro __using__(opts \\ []) do
+    quote do
+      alias Commanded.Event.FailureContext
+
+      require Logger
+
+      def error(
+            error,
+            event,
+            %FailureContext{context: context, metadata: metadata} = failure_context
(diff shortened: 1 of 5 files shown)


lib/trento/support/event_handler_failure_context.ex
```diff
@@ -0,0 +1,73 @@
...
+              )
+
+              :skip
+            else
+              Logger.error(
+                "#{__MODULE__} failed to handle event, stopping after #{max_retries} retries"
+              )
+
+              {:error, :max_retries_reached}
+            end
```
dottorblaster, line 55: Just personal taste, but I would separate this into two separate functions:

```elixir
defp skip_failure?(true) do
  Logger.error(
    "#{__MODULE__} failed to handle event, skipping after #{max_retries} retries"
  )
  :skip
end
```

either the `false` case. But really, an if is ok too.

dottorblaster (approved):
LGTM

nelsonkopliku (approved):
Great stuff! You rock! :rocket:

lib/trento/support/event_handler_failure_context.ex
dottorblaster: Fair enough :+1:

lib/trento/support/event_handler_failure_context.ex
```diff
@@ -0,0 +1,72 @@
+defmodule Trento.Support.EventHandlerFailureContext do
+  @moduledoc """
+  Event handler failure context
+
+  max_retries: max retries before the event handler is shut down (defauklt: 3)
+  retry_after: time between retries in ms (default: 0)
```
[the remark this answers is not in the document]
dottorblaster: Also: typo `defauklt` :smile: good catch @nelsonkopliku

PR #580, opened 2022-05-23
Rollup integration test

fabriziosestito:
This PR adds the rollup integration test.
Adds an event store case helper that resets the test eventstore every time a test runs against the ES.

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -14,11 +14,13 @@ config :trento, Trento.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
   pool_size: 10
 
-config :trento, Trento.Commanded,
-  event_store: [
-    adapter: Commanded.EventStore.Adapters.InMemory,
-    serializer: Commanded.Serialization.JsonSerializer
-  ]
+config :trento, Trento.EventStore,
+  username: "postgres",
+  password: "postgres",
+  database: "trento_eventstore_test#{System.get_env("MIX_TEST_PARTITION")}",
+  hostname: "localhost",
+  port: 5433,
+  pool: Ecto.Adapters.SQL.Sandbox
 
 # We don't run a server during test. If one is required,
 # you can enable the server option below.
--- a/lib/trento/application/usecases/rollup/rollup.ex
+++ b/lib/trento/application/usecases/rollup/rollup.ex
@@ -1,0 +1,29 @@
+defmodule Trento.Rollup do
+  @moduledoc """
+  Probides the rollup functionality
+  """
+
+  def rollup_aggregate(aggregate_id, event) do
+    {:ok, pid} = Postgrex.start_link(Trento.EventStore.config())
+
+    Postgrex.transaction(pid, fn conn ->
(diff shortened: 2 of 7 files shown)


config/test.exs
```diff
@@ -14,11 +14,13 @@ config :trento, Trento.Repo,
   pool: Ecto.Adapters.SQL.Sandbox,
   pool_size: 10
 
-config :trento, Trento.Commanded,
-  event_store: [
-    adapter: Commanded.EventStore.Adapters.InMemory,
-    serializer: Commanded.Serialization.JsonSerializer
-  ]
+config :trento, Trento.EventStore,
```
nelsonkopliku, line 17: This gives us a real event store for test instead of an `InMemory` one.

I don't remember if we have tests needing an eventstore other than this one:
in that case those would now hit the database instead of using `InMemory` implementation.

I don't see that necessarily as a bad thing, I would just keep in mind that given the nature of the test, we might need a different implementation of the EventStore.

test/support/event_store_case.ex
```diff
@@ -0,0 +1,25 @@
...
+  use ExUnit.CaseTemplate
+
+  setup do
+    :ok = Application.stop(:trento)
+    :ok = Application.stop(:commanded)
+    :ok = Application.stop(:eventstore)
+
+    config = Trento.EventStore.config()
+    {:ok, conn} = Postgrex.start_link(config)
+    reset_event_store(conn, config)
```
nelsonkopliku, line 15: Disclaimer: I am ok with this.

just a nitpick: is this for readability?
I mean `reset_event_store` instead of just `EventStore.Storage.Initializer.reset!(conn, config)`

nelsonkopliku (approved):
Great stuff! Thanks!

test/support/event_store_case.ex
fabriziosestito: ah yeah, refactoring stuff i missed that part, I was initializing the connection in that function and then I moved it on top to pass the connection in the context. i will fix it.

config/test.exs
fabriziosestito: This was configured, but we never used any EventStore in the tests.
We test our domain logic and handlers with unit tests/integration tests that do not pass from the EventStore.
The implementation of the rollup is specific to the Postgres event store, which is what we use, so I don't find particularly useful testing against the `InMeory` one.

lib/trento/application/usecases/rollup/rollup.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Trento.Rollup do
+  @moduledoc """
+  Probides the rollup functionality
```
arbulu89, line 3: Typo `Provides`

arbulu89 (approved):
Nice!

I don't see we test that old events are indeed removed, right? We are just testing that the rollup event is there

fabriziosestito:
> Nice!
> 
> I don't see we test that old events are indeed removed, right? We are just testing that the rollup event is there

@arbulu89 
We match against a list with just one event when we read the stream in the test after the rollup action, so we know that the other events have been removed.

PR #582, opened 2022-05-23
Ecto decode cloud discovery

dottorblaster:
Decoding cloud discovery payload using Ecto.


nelsonkopliku (approved):
LGTM

lib/trento/application/integration/discovery/policies/host_policy.ex
```diff
@@ -113,6 +84,49 @@ defmodule Trento.Integration.Discovery.HostPolicy do
...
+         provider: provider,
+         metadata: metadata
+       }) do
+    UpdateProvider.new(%{
+      host_id: agent_id,
+      provider: provider,
+      provider_data:
+        case metadata do
+          %{
+            compute: %{
```
fabriziosestito, line 97: could we have a function here instead of a case?

lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex
```diff
@@ -0,0 +1,83 @@
+defmodule Trento.Integration.Discovery.CloudDiscoveryPayload do
+  @moduledoc """
+  Cloud discovery integration event payload
+  """
+
+  @required_fields []
```
arbulu89, line 6: Wouldn't be the `provider` field required here?

lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex
```diff
@@ -0,0 +1,83 @@
...
+      on_type_not_found: :nilify,
+      on_replace: :update
+  end
+
+  def changeset(event, attrs) do
+    enriched_attrs = enrich_metadata_type(attrs)
+
+    event
+    |> cast(enriched_attrs, [:provider])
+    |> cast_polymorphic_embed(:metadata, required: false)
```
arbulu89, line 25: Ok, now I understand why `ExMachina` is not an issue here, as we use the polymorphism just as a decoder, and not in the read models. So in fact, we don't create factories from it

lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  defp enrich_metadata_type(%{"provider" => provider, "metadata" => %{} = metadata} = attrs),
+    do: %{attrs | "metadata" => Map.put(metadata, "__type__", provider)}
+
+  defp enrich_metadata_type(attrs), do: attrs
+
+  defmodule AzureMetadata do
```
arbulu89, line 33: I think having the `AzureMetadata` in other file would be nicer, as this will expand when we have other providers. But well, we can change overtime of course!

lib/trento/application/integration/discovery/policies/host_policy.ex
arbulu89: I would have something like:
```
defp build_update_provider_command(agent_id, %CloudDiscoveryPayload{
         provider: :azure,
         metadata: metadata
       }) do
...
````

To handle the different provider overtime, instesad of the `case` option

**Review by arbulu89 — APPROVED** [review]
Nice work @dottorblaster !
I agree on the comment given by @fabriziosestito on splitting the metadata building process by providers

File: lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex (hunk 3 above)
Comment (dottorblaster) [reply lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex:25]: Exactly! Fortunately here we can use polymorphic embed :smiling_face_with_tear:

File: lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex (hunk 4 above)
Comment (dottorblaster) [reply lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex:33]: I thought that too, in the future for sure we will have the need for separate files, I agree. :+1:

File: lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex (hunk 2 above)
Comment (dottorblaster) [reply lib/trento/application/integration/discovery/payloads/cloud_discovery_payload.ex:6]: Not really since we default to `:unknown` :sweat_smile: 

```
iex(4)> %{"metadata" => %{"compute" => %{"name" => "alessio"}}} |> Trento.Integration.Discovery.CloudDis
coveryPayload.new
{:ok,
 %Trento.Integration.Discovery.CloudDiscoveryPayload{
   metadata: nil,
   provider: :unknown
 }}
```

PR #587, opened 2022-05-24
Add e2e tests clusters overview

rtorrero:
This PR adds:
 - Basic mocking to simulate that the checks execution from the runner has been completed
 - Additional e2e tests coverage to the clusters overview
 

Sadly, I've realized that the mocking right now breaks some checks functionality as it causes
```(Ecto.StaleEntryError) attempted to update a stale struct:``` when handling the `Trento.Domain.Events.ChecksExecutionCompleted` event.

The result is that the cluster-details e2e test fails.

### CODE REVIEW

File: test/e2e/cypress/integration/clusters_overview.js [hunk 1]
```diff
@@ -44,5 +44,226 @@ context('Clusters Overview', () => {
...
+        result: 'critical',
+        executionId: '8d840ba1-828f-4cb8-bce0-6e34224fe4f9',
+      };
+
+      before(() => {
+        cy.task('selectChecks', {
+          clusterId: clusterIdByName(healthyClusterScenario.clusterName),
+          checkIds: healthyClusterScenario.checks,
+        });
+        cy.task('mockStartChecksExecution', {
```
nelsonkopliku, line 76: for an execution to start on the runner there's the need to have first requested for that execution to start.

I am not completely sure the issue you're referring to is an actual bug, I have the impression we're missing something in the flow.
is the command RequestChecksExecution  being dispatched to the aggregate? And the following ChecksExecutionRequested  event emitted?

rtorrero: No, I was actually avoiding it as I understood that this might create the random mocked results, which don't allow the test to fail or succeed in a predictable way.

Is there any better way to do this?

nelsonkopliku: I understand, but you _have to_ go through that command/event pair, because it changes state of the Aggregate.

You might need to look into `Trento.Integration.Checks.MockRunner` which should allow you to set some `expected_results`.

I don't know whether that's enough or some extra pieces are needed to instruct during e2e tests.

The thing I am pretty sure is that that piece of the flow is missing, and some business logic has not been applied.

rtorrero: > You might need to look into Trento.Integration.Checks.MockRunner which should allow you to set some expected_results.

Thanks! I'll take a look at `Trento.Integration.Checks.MockRunner`

>The thing I am pretty sure is that that piece of the flow is missing, and some business logic has not been applied.

Yes, agreed 100%. As I later clarified in slack, I don't think this exposes a bug per se as it's an edge case that a runner would reach the callback endpoint without a prior request. It would be more of an improvement: handling those scenarios correctly by e.g. discarding the event.

nelsonkopliku: > handling those scenarios correctly by e.g. discarding the event.
>
What does this mean? :thinking:

test/e2e/cypress/plugins/index.js
```diff
@@ -55,5 +55,91 @@ module.exports = (on, config) => {
       });
       return null;
     },
+
+    selectChecks({ clusterId, checkIds }) {
```
arbulu89, line 59: We already have a command for this:
https://github.com/trento-project/web/blob/main/test/e2e/cypress/support/commands.js#L64

I think commands are the way to go, but well, in any case, let's have it only in 1 place hehe

test/e2e/cypress/integration/clusters_overview.js
```diff
@@ -44,5 +44,226 @@ context('Clusters Overview', () => {
         cy.get(`a:contains(${clusterID})`).should('be.visible');
       });
     });
+
+    describe('Health status for each cluster is correct', () => {
+      const healthyClusterScenario = {
```
arbulu89, line 49: This data looks something that should go in the fixtures, in `clusters-overview` folder I guess

test/e2e/cypress/integration/clusters_overview.js
arbulu89: Let's try to create an additional e2e environment so we can enable the actual runner, instead of the mocked one, as it doesn't allow us doing real integration tests

rtorrero: > > handling those scenarios correctly by e.g. discarding the event.
> 
> What does this mean? thinking

This is about the original issue I was exposing:
the `Elixir.Trento.Domain.Events.ChecksExecutionCompleted` gets stuck endlessly as the projection fails with `(Ecto.StaleEntryError) attempted to update a stale struct:`. 

> Let's try to create an additional e2e environment so we can enable the actual runner, instead of the mocked one, as it doesn't allow us doing real integration tests

Yes, I do actually think this is the right path :+1:

rtorrero:
The `e2e.exs` environment idea to disable the mock runner lead to actually complicating things so instead as per @fabriziosestito suggestion, I simply added a conditional endpoint to pre-set the checks result. This endpoint is only present in the `dev` environment that the e2e tests use so it be available in production.

This endpoint allows to set check results in a predictable way so that we can check the checks/health status indicator in the clusters overview.

test/e2e/cypress/support/commands.js
```diff
@@ -97,3 +97,111 @@ Cypress.Commands.add('addTagByColumnValue', (columnValue, tagValue) => {
...
+
+  const headers = {
+    'Content-Type': 'application/json;charset=UTF-8',
+  };
+
+  const url = `http://${webAPIHost}:${webAPIPort}/api/clusters/${clusterId}/checks/request_execution`;
+  cy.request({ method: 'POST', url: url, headers: headers });
+});
+
+Cypress.Commands.add('mockStartChecksExecution', (clusterId, executionId) => {
```
nelsonkopliku, line 138: I am wondering whether we need 
`mockStartChecksExecution` and `mockCompleteChecksExecution`.

In fact I believe we might not need them.

Why you may ask.

The reason is that once we 
- had instructed the `MockRunner` on the results that it should return on next execution `requestMockRunnerResult`
- requested the execution to start `requestChecksExecution`

then the `MockRunner` itself takes care of calling the appropriate callback functions when needed.

See here https://github.com/trento-project/web/blob/main/lib/trento/application/integration/checks/adapter/mock_runner.ex#L88

rtorrero: Nice catch! As originally I was trying to go around the mock runner I didn't notice this wasn't needed anymore +1

lib/trento_web/router.ex
```diff
@@ -124,6 +124,13 @@ defmodule TrentoWeb.Router do
     get "/openapi", OpenApiSpex.Plug.RenderSpec, []
   end
 
+  if Mix.env() == :dev do
+    scope "/api", TrentoWeb do
```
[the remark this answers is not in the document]
nelsonkopliku: Hope this was a good suggestion :stuck_out_tongue_closed_eyes:

test/e2e/cypress/support/commands.js
nelsonkopliku: Great!

I don't know if you already considered the fact that the `MockRunner` takes a randomic amount of time (between 2000 and 4000 ms) to return the results.

Not sure if this is handled by the test.

rtorrero: In principle, it should be covered by the default wait time of 4000 ms that cypress implements to get elements from the DOM. I didn't find any flakiness at least on my machine but I'll keep an eye on it.

nelsonkopliku: I see, let's keep an eye if there's any flakiness. :+1:

test/e2e/cypress/support/commands.js
```diff
@@ -97,3 +97,40 @@ Cypress.Commands.add('addTagByColumnValue', (columnValue, tagValue) => {
       cy.get('span').contains('Add Tag').type(`${tagValue}{enter}`);
     });
 });
+
+Cypress.Commands.add('requestMockRunnerResult', (result) => {
```
arbulu89, line 101: I would rename this to `setMockRunnerExpectedResult`

test/e2e/cypress/integration/clusters_overview.js
```diff
@@ -44,5 +46,193 @@ context('Clusters Overview', () => {
...
+          .parent()
+          .prev()
+          .get('div > svg')
+          .should('have.class', 'fill-red-500');
+      });
+    });
+
+    describe('Clusters Tagging', () => {
+      before(() => {
+        cy.get('body').then(($body) => {
```
arbulu89, line 106: We already have a command for this (removing tags):
https://github.com/trento-project/web/blob/main/test/e2e/cypress/support/commands.js#L81

test/e2e/cypress/integration/clusters_overview.js
```diff
@@ -44,5 +46,193 @@ context('Clusters Overview', () => {
...
+        ['hana_cluster_3', 'env3'],
+      ];
+
+      taggingRules.forEach(([pattern, tag]) => {
+        describe(`Add tag '${tag}' to all clusters with '${pattern}' in the cluster name`, () => {
+          availableClusters
+            .filter(clustersByMatchingPattern(pattern))
+            .forEach((clusterName) => {
+              it(`should tag cluster '${clusterName}'`, () => {
+                cy.get('td > span > a > span')
```
arbulu89, line 129: The same here, we have a command for this:
https://github.com/trento-project/web/blob/main/test/e2e/cypress/support/commands.js#L92

arbulu89 (approved):
Nice @rtorrero ,
Few comments about renaming and reusing already existing commands

test/e2e/cypress/integration/clusters_overview.js
```diff
@@ -44,5 +46,193 @@ context('Clusters Overview', () => {
...
+    describe('Filtering the Clusters overview', () => {
+      describe('Filtering by tags', () => {
+        const tagsScenarios = [
+          ['env1', 1],
+          ['env2', 1],
+          ['env3', 1],
+        ];
+        tagsScenarios.forEach(([tag, expectedTaggedClusters]) => {
+          it(`should have ${expectedTaggedClusters} clusters tagged with tag '${tag}'`, () => {
+            cy.get(':nth-child(5) > .mt-1 > .relative').click();
```
arbulu89, line 150: Here a reference on how I did the hosts overview tests on filtering. If you want to have a look:
https://github.com/trento-project/web/blob/main/test/e2e/cypress/integration/hosts_overview.js#L197

nelsonkopliku (approved):
Thanks Ruben!
Let's just keep an eye on unwanted flakiness.

PR #601, opened 2022-05-26
Add the possibility to decode lists of items with deftype

dottorblaster:
Adding a new `from_list` function to `deftype` macro. This will allow us to decode payloads incoming as lists without too much boilerplate, returning meaningful errors in case of them, or just plain results. Of course the default behavior uses tagged tuples for flow control, if we want we can add banged functions to raise. At the moment I didn't feel the need for that.

Tests are pretty self-explaining.


nelsonkopliku (approved):
Very nice but... where are lasagne? :smile:

dottorblaster:
@nelsonkopliku you're right! There they are :smile:

rtorrero (approved):
LGTM, thanks!

PR #613, opened 2022-05-30
Expose a valid OpenApi specification

nelsonkopliku:
I noticed our openapi spec is not valid (shame on me :smile: )

image

here are the reasons https://validator.swagger.io/validator/debug?url=http%3A%2F%2Fdemo.trento-project.io%2Fapi%2Fopenapi

```json
{
  "messages": [
    "attribute paths.'/api/clusters/{cluster_id}/checks'(post).parameters.[selected_checks].in is not of type `string`",
    "attribute paths.'/api/runner/callback'(post).parameters.[callback_event].in is not of type `string`",
    "attribute components.schemas.Schema name IP address doesn't adhere to regular expression ^[a-zA-Z0-9\\.\\-_]+$"
  ],
  "schemaValidationMessages": [
    {
      "level": "error",
      "domain": "validation",
      "keyword": "oneOf",
      "message": "instance failed to match exactly one schema (matched 0 out of 2)",
      "schema": {
        "loadingURI": "#",
        "pointer": "/definitions/Operation/properties/parameters/items"
      },
      "instance": {
        "pointer": "/paths/~1api~1clusters~1{cluster_id}~1checks/post/parameters/1"
      }
    }
  ]
}
```

with these changes we it's green again
image

image

image

### DIFF
--- a/lib/trento_web/controllers/cluster_controller.ex
+++ b/lib/trento_web/controllers/cluster_controller.ex
@@ -68,13 +68,7 @@ defmodule TrentoWeb.ClusterController do
     tags: ["Checks"],
     description:
       "The Runner executing the Checks Selection on the target infrastructure, publishes updates about the progress of the Execution.",
-    parameters: [
-      callback_event: [
-        in: :body,
-        required: true,
-        type: Schema.Runner.CallbackEvent
-      ]
-    ],
+    request_body: {"Callback Event", "application/json", Schema.Runner.CallbackEvent},
     responses: [
       accepted:
         {"The Operation has been accepted, and the proper followup processes will trigger",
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):
Nice thing!
How did you generate the swagger validation output?
Maybe we could integrate something like this in our CI

nelsonkopliku:
> Nice thing! How did you generate the swagger validation output? Or is it just this online service? Maybe we could integrate something like this in our CI

Here https://validator.swagger.io/ then you provide the url of a reachable openapi spec.
Because it doesn't work with localhost I cheated and used ngrok and provided this URL https://5faa-79-18-240-47.eu.ngrok.io/api/doc which is basically my machine :tm: 

We can put to production our machines eheh :joy: 

Anyway, I believe there should be some validator to be used programmatically, even in CI, but I didn't go there yet.

PR #616, opened 2022-05-31
Use failure context rollup event handler

fabriziosestito:
This PR uses the event handler failure context in the rollup event handler to "unlock" the cluster aggregate in case something went wrong with the cluster rollup, so the cluster aggregate can self-heal from a stale state.


nelsonkopliku (approved):
Very nice! :shipit:

dottorblaster (approved):

PR #619, opened 2022-06-01
Handle empty list in from_list function

arbulu89:
Handle empty lists in the payload, in the `from_list` function. This scenario happens in the SAP system payload, when no sap system is found.

```
    ** (FunctionClauseError) no function clause matching in Trento.Integration.Discovery.SapSystemDiscoveryPayload.decoding_results/1
        (trento 1.0.0) lib/trento/application/integration/discovery/payloads/sap_system_discovery_payload.ex:14: Trento.Integration.Discovery.SapSystemDiscoveryPayload.decoding_results(%{})
        (trento 1.0.0) lib/trento/application/integration/discovery/policies/sap_system_policy.ex:32: Trento.Integration.Discovery.SapSystemPolicy.handle/1
```

### DIFF
--- a/lib/trento/support/type.ex
+++ b/lib/trento/support/type.ex
@@ -112,6 +112,7 @@ defmodule Trento.Type do
 
       defp decoding_results(%{error: decoding_errors}), do: {:error, decoding_errors}
       defp decoding_results(%{ok: decoding_results}), do: {:ok, decoding_results}
+      defp decoding_results(_), do: {:ok, []}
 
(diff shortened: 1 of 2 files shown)


fabriziosestito (approved):

arbulu89:
> LGTM, just a small comment for the future I would rename this function to `new` as well and use the `is_list` guard. should I track this as a tech debt?

Yes, let's track it. It looks an easy fix. I will create a new PR with it tomorrow

fabriziosestito:
@arbulu89 #620

PR #622, opened 2022-06-02
Persist cib last written field

arbulu89:
Persist the `cib_last_written` field in the database. 

*Before checking the code:* This change is only done to have a `CIB last written` field in the cluster details view always populated. Before this, this fields is only changed when we receive a ws message on a discovery. So if we open the page, at the beginning this field is empty as long as we don't receive other cluster discovery message.
image

In my opinion, it is not a big deal, but if we always want to have the field populated we need a change. If we are fine with having the field unpopulated for a while, we can discard this PR. (this question is more for @abravosuse as well)

Now...

The PR basically stores this field in the database through the read model.
The field is stored in 2 scenarios:
- Cluster is 1st registered, on `ClusterRegisteredEvent`
- In `RegisterClusterHost` event, in the enrich function. It only works if the cluster exists.

So, we continue **not** having any event on this field updates (which happens super frequently). The unique major difference is that we add a read/update database actions in every command. It is not even stored in the aggregate, as we don't use it to compare new changes on this field

### DIFF
--- a/lib/trento/application/integration/discovery/protocol/enrich_register_cluster_host.ex
+++ b/lib/trento/application/integration/discovery/protocol/enrich_register_cluster_host.ex
@@ -2,16 +2,27 @@ defimpl Trento.Support.Middleware.Enrichable,
   for: Trento.Domain.Commands.RegisterClusterHost do
   alias Trento.Domain.Commands.RegisterClusterHost
 
+  alias Trento.Clusters
+
   @spec enrich(RegisterClusterHost.t(), map) :: {:ok, map} | {:error, any}
   def enrich(
         %RegisterClusterHost{cluster_id: cluster_id, cib_last_written: cib_last_written} =
           command,
         _
       ) do
-    TrentoWeb.Endpoint.broadcast("monitoring:clusters", "cluster_cib_last_written_updated", %{
-      cluster_id: cluster_id,
-      cib_last_written: cib_last_written
-    })
+    case Clusters.update_cib_last_written(cluster_id, cib_last_written) do
+      {:ok, _} ->
+        TrentoWeb.Endpoint.broadcast("monitoring:clusters", "cluster_cib_last_written_updated", %{
+          cluster_id: cluster_id,
+          cib_last_written: cib_last_written
+        })
+
+      {:error, :cluster_not_found} ->
+        nil
+
+      error ->
+        error
+    end
 
     {:ok, command}
   end
--- a/lib/trento/application/integration/discovery/protocol/enrich_register_cluster_host.ex
+++ b/lib/trento/application/integration/discovery/protocol/enrich_register_cluster_host.ex
@@ -11,14 +11,12 @@ defimpl Trento.Support.Middleware.Enrichable,
         _
       ) do
     case Clusters.update_cib_last_written(cluster_id, cib_last_written) do
-      {:ok, _} ->
-        TrentoWeb.Endpoint.broadcast("monitoring:clusters", "cluster_cib_last_written_updated", %{
-          cluster_id: cluster_id,
-          cib_last_written: cib_last_written
-        })
-
-      {:error, :cluster_not_found} ->
-        nil
+      {:ok, cluster} ->
+        TrentoWeb.Endpoint.broadcast(
+          "monitoring:clusters",
+          "cluster_cib_last_written_updated",
+          cluster
+        )
 
       error ->
         error
--- a/lib/trento/application/usecases/clusters/clusters.ex
+++ b/lib/trento/application/usecases/clusters/clusters.ex
@@ -56,10 +56,22 @@ defmodule Trento.Clusters do
     |> Repo.all()
   end
 
+  @spec enrich_cluster_model_query(Ecto.Query.t()) :: Ecto.Query.t()
   def enrich_cluster_model_query(query) do
     query
     |> join(:left, [c], e in EnrichedCluster, on: c.id == e.cluster_id)
-    |> select_merge([_c, e], %{cib_last_written: e.cib_last_written})
+    |> select_merge([c, e], %{cib_last_written: e.cib_last_written})
+  end
+
+  @spec enrich_cluster_model(ClusterReadModel.t()) :: ClusterReadModel.t()
+  def enrich_cluster_model(%ClusterReadModel{id: id} = cluster) do
+    case Repo.get(EnrichedCluster, id) do
+      nil ->
+        cluster
+
+      enriched_data ->
+        %ClusterReadModel{cluster | cib_last_written: enriched_data.cib_last_written}
+    end
   end
 
   @spec request_clusters_checks_execution :: :ok | {:error, any}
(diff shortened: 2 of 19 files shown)


nelsonkopliku (changes requested):
Hey @arbulu89 thanks for addressing this, not an easy one :smile: 

The problem, as you mentioned, is that the `cib_last_written` comes with the cluster discovery, which might happen very frequently - every 10s by default, yet it can be configured - and if taken into account in the Aggregate command-event flow, it might raise a too many events and grow the eventstore too much.

The proposed solution allows us to project the discovered `cib_last_written` value the very first time we register a cluster, so that the information is immediately available to read.
And then update that value every time we get a new discovery (regardless of what happens to that discovery)
I see a bit of inconsistency, though, in the fact that we leverage the command-event-projection flow when registering (because it happens only once) and then updating that value during command enrichment, regardless of what happens to the enriched command.
I provided my thoughts and some alternative options in the comments.

fabriziosestito:
@arbulu89 the choice about having the `cib last written` not passing from the event store was mainly made to not have an event every one minute for each cluster. 
My original idea was to set up a cache and use it to persist the `cib last written` (alongside other things).
Since we still do not have the use-case for rolling out a cache I would instead store the `cib last written` in a separate crud table in the command enrichment (since writing directly to a read model would violate the CQRS pattern).
This way we would have the same result, but we would not pass from an event to update the whole thing.

arbulu89:
> @arbulu89 the choice about having the `cib last written` not passing from the event store was mainly made to not have an event every one minute for each cluster. My original idea was to set up a cache and use it to persist the `cib last written` (alongside other things). Since we still do not have the use-case for rolling out a cache I would instead store the `cib last written` in a separate crud table in the command enrichment (since writing directly to a read model would violate the CQRS pattern). This way we would have the same result, but we would not pass from an event to update the whole thing.

Ok, that looks interesting. And it doesn't require many changes. My initial implementation didn't increment the number of events stored in the event-store, but now I see that it breaks the CQRS pattern as you say.

I will explore the option of having this field in an additional table, and not use the read model.

lib/trento/application/integration/discovery/protocol/enrich_register_cluster_host.ex
```diff
@@ -2,16 +2,27 @@ defimpl Trento.Support.Middleware.Enrichable,
...
   def enrich(
         %RegisterClusterHost{cluster_id: cluster_id, cib_last_written: cib_last_written} =
           command,
         _
       ) do
-    TrentoWeb.Endpoint.broadcast("monitoring:clusters", "cluster_cib_last_written_updated", %{
-      cluster_id: cluster_id,
-      cib_last_written: cib_last_written
-    })
+    case Clusters.update_cib_last_written(cluster_id, cib_last_written) do
```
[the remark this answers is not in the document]
arbulu89: @nelsonkopliku Thank you dude!
My brain almost exploded when I saw this comment first thing in the morning after 4 days off hehe
At the end I tried to do the most minimalist change possible. Most of the thing were already in place. The unique thing I did is to include the new field in the read model, and as @fabriziosestito suggested nicely, this breaks the CQRS pattern, so having the data stored in an additional table makes a lot of sense.

If you don't mind, I will implement this option, without changing more things. Broadcasting before anything else looks like a smell as you say, but as you say again XD, it doesn't play an important role, and the frontend works perfectly fine in either case.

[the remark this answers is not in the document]
fabriziosestito: > Broadcasting before anything else looks like a smell as you say

@arbulu89  the suggested`after_dispatch` callback will be handy for this. I honestly didn't know/remember of that option, thanks @nelsonkopliku  for doing some research. 

We just need to split the `Enrichable` protocol into before/after/failure and hook the broadcast to the after one.
Let's maybe keep this for a follow-up PR.

arbulu89:
@nelsonkopliku @fabriziosestito 
Finally, I have decided to the with the `virtual` field solution, as it is the most straightforward to avoid changes in the frontend, as we already expect a data struct that directly has this field without nesting structs.

Basically, `virtual` fields allow us to set fields that are not stored in the database, but still continue using ECTO validation and changeset functionalities.

The enriched cluster data goes in the `enriched_clusters` table, which is joined to the original `clusters` table when we retrieve the clusters, so we get the complete dataset.

The next step would to create the clusters api DTO so we can define the contract between the server and frontend in a more reliable way.

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -47,10 +48,30 @@ defmodule Trento.Clusters do
...
+      order_by: [asc: c.name],
+      preload: [:tags, :hosts_executions, :checks_results]
+    )
+    |> enrich_cluster_model_query()
     |> Repo.all()
-    |> Repo.preload([:tags, :hosts_executions, :checks_results])
+  end
+
+  @spec enrich_cluster_model_query(Ecto.Query.t()) :: Ecto.Query.t()
+  def enrich_cluster_model_query(query) do
```
nelsonkopliku, line 60: this could be a `defp`

lib/trento/application/usecases/clusters/enriched_cluster.ex
```diff
@@ -0,0 +1,22 @@
...
+
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @derive {Jason.Encoder, except: [:__meta__, :__struct__]}
+  @primary_key {:cluster_id, :binary_id, autogenerate: false}
+  schema "enriched_clusters" do
```
nelsonkopliku, line 14: To me an `Enrichedcluster` is the result of a cluster plus information available for enrichment.

What do you think about `clusters_enriching_information`, `clusters_enriching_dataa` or something like this?

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -91,4 +112,14 @@ defmodule Trento.Clusters do
     results ->
       {:ok, results}
   end
+
+  @spec update_cib_last_written(String.t(), String.t()) :: {:ok, Ecto.Schema.t()} | {:error, any}
+  def update_cib_last_written(cluster_id, cib_last_written) do
+    case Repo.get(EnrichedCluster, cluster_id) do
+      nil -> %EnrichedCluster{cluster_id: cluster_id}
+      cluster -> cluster
```
nelsonkopliku, line 120: for consistency `cluster` should be the actual little data-lake we'd use for enrichment.
In this specific case could be `enriched_cluster` or something reflecting a different name if we're going to change it.

lib/trento/application/usecases/clusters/enriched_cluster.ex
arbulu89: Renamed to `clusters_enrichment_data`, and the module to `ClusterEnrichmentData`

arbulu89:
@nelsonkopliku The suggested changes applied, mostly renaming.
Let's go for another round! XD

PR #623, opened 2022-06-02
Use NavLink in the navigation elements

arbulu89:
Use `NavLink` in the navigation elements to highlight the children elements. For example:

image

When we are in the cluster details page (before this change, it sidebar entry wouldn't be highlighted as active).

Find more info about `NavLink` in: https://www.geeksforgeeks.org/link-and-navlink-components-in-react-router-dom/

> A <NavLink> is a special kind of <Link> that knows whether or not it is “active”. Now, let’s understand  this with the help of an example. 

I have included a couple of tests in the host entries.


nelsonkopliku (approved):
Alright, thanks! :rocket:

dottorblaster (approved):
LGTM

PR #663, opened 2022-06-14
Allow a failover payload to pass validations

rtorrero:
This PR fixes an issue introduced in PR #602 that caused a payload validation failure in the event of a failover. This caused that some information in the SAP systems overview and HANA database details overview was wrong/missing.

Now both views show the correct information:
sap-details
db-details


lib/trento/application/integration/discovery/payloads/sap_system_discovery_payload.ex
```diff
@@ -276,11 +276,9 @@ defmodule Trento.Integration.Discovery.SapSystemDiscoveryPayload do
     @required_fields [
       :pid,
       :name,
-      :starttime,
```
arbulu89, line 279: I don't really understand this. Why are you removing these fields? They should be always coming.
Do they have some empty field, and that's why ecto is discarding the messages?

What are the values that you are receiving in the failed message?

lib/trento/application/integration/discovery/payloads/sap_system_discovery_payload.ex
```diff
@@ -343,6 +341,10 @@ defmodule Trento.Integration.Discovery.SapSystemDiscoveryPayload do
     defp parse_local_site_id(%{"local_site_id" => local_site_id}), do: local_site_id
     defp parse_local_site_id(_), do: 1
 
+    defp validate_replication_mode(changeset, "0") do
```
arbulu89, line 344: We should rename the methods adding a `maybe_` here, as this methods doesn't validate in all the cases

arbulu89 (commented):
Thanks @rtorrero ,
I think we should add a new test (unit test) at least on this, using the payload which caused the issue. This way we continue covering the most standard use-cases with tests, so we know that we don't break them

lib/trento/application/integration/discovery/payloads/sap_system_discovery_payload.ex
rtorrero: Forn @abravosuse to trigger the failover scenario for me, he basically was stopping the primary instance; causing that the process information in the payload was something like:

```
"Processes" => [ % {
	"description" => "HDB Daemon",
	"dispstatus" => "SAPControl-GRAY",
	"name" => "hdbdaemon",
	"pid" => 26205,
	"textstatus" => "Stopped"
}],
```

As you see, that process is in status Stopped and doesn't include these fields.

arbulu89: But this is already the elixir code. I would like to see the content of the real payload, to understand why is not casting those fields. Maybe there are not coming. And that's why having a test with the real payload would be ideal

dottorblaster (changes requested):
As Xabi said, we have to add this case to the unit tests to be sure we don't end up in regressions :+1: 

Good job! Once we are done with this we can merge

dottorblaster (approved):
Now that we've got the test for the failing case, LGTM

arbulu89 (approved):
Thanks for adding the test @rtorrero 

In my opinion, we shouldn't call this tests and fixture "failover". The failover is a cluster thing, not a sap system thing. In this case, we are just watching a stopped SAP instance, which launches a failover in the cluster (making the other secondary instance primary).
I would try to be meticulous with the wording, and call this scenarios as `stopped` rather than `failover`.

lib/trento/application/integration/discovery/payloads/sap_system_discovery_payload.ex
arbulu89: Thanks @rtorrero

rtorrero:
Added the `maybe_` prefix for the validation functions and changed a bit the wording for the test case.

PR #674, opened 2022-06-20
Add a first version of the DSL specification

dottorblaster:
Just stating the progress about a potential DSL definition with a design document. Nothing fancy, just notes!

### DIFF
--- a/docs/development/dsl.md
+++ b/docs/development/dsl.md
@@ -1,0 +1,49 @@
+# Trento Checks DSL Specification
+
+## Foreword
+The need this DSL aims to fulfill is to provide users a simple way to declare what we (the Trento Core Team) often refer to as "checks". Checks are, in Trento's domain, the crystallization of SUSE's best practices when it comes to SAP clusters configuration in a form that both a man and a machine can read.
+
+We get several benefits from this approach:
+
+- Humans can formalize best practices with no space for ambiguity;
+- Machines can assert systems' state, automatically, with no space for ambiguity;
+- The development of new best practices gets streamlined through a common definition that allows to firestart shared efforts.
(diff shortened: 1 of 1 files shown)


docs/development/dsl.md
```diff
@@ -0,0 +1,49 @@
...
+- Machines can assert systems' state, automatically, with no space for ambiguity;
+- The development of new best practices gets streamlined through a common definition that allows to firestart shared efforts.
+
+## Facts
+Fact declaration and gathering revolves all around the concept of _gatherer_. Gatherers are functions directly built inside the interpretation level of the language, and reflect the various types of information that the platform can discover on the hosts, and on a given cluster. Examples include (but are not limited to) installed packages, open ports, and configuration files' content. Gathered facts can be matched against values inside expectations, using the name that we give them inside the fact declaration.
+
+A fact gathering declaration contains:
+
+- The fact name;
+- The function used to gather the fact;
```
rtorrero, line 18: I'd write this as:

`- The pre-defined gatherer function used for this fact`



(Not that it's wrong as it is, just maybe to use the `gatherer` term again instead of function)


Nitpicking, I know :smile:

nelsonkopliku:
Great stuff! :rocket: :heart:

dottorblaster:
We are at a crossroad now. When matching two values in the expectation value matching phase, the values have to be of the same type to avoid comparison of apples with oranges.

Now:

- Should the values be compared casting all of them to strings?
- Should the values be compared trying to apply automatic typecasting? (So the value _has_ a type, and we try to cast the right-hand value to the type of the left-hand value)
- Should we annotate types for expectations' values?

PR #679, opened 2022-06-22
Add hosts_execution to cluster_registered broadcast

rtorrero:
This PR is a fix for an issue currently present in `main` that causes the checks result not being displayed after a check execution is triggered:

broken

We now preload the :hosts_executions before the `cluster_registered` broadcast to avoid `undefined` errors.


rtorrero:
The CI doesn't really like this, I'll leave it open for discussion and meanwhile try to figure out why the cause for the original issue is

dottorblaster (approved):
LGTM

PR #680, opened 2022-06-23
Add a modal to display remediation info on check results

rtorrero:
This PR adds a modals where users can check the remediation details from the check results screen by clicking on the check result row. In action:

remediation-info

### DIFF
--- a/assets/js/components/ClusterDetails/ChecksResults.jsx
+++ b/assets/js/components/ClusterDetails/ChecksResults.jsx
@@ -2,6 +2,8 @@ import React, { useEffect, useState } from 'react';
 import { useNavigate, useParams } from 'react-router-dom';
 import { useSelector, useDispatch } from 'react-redux';
 
+import Modal from '@components/Modal';
+
 import {
   EOS_ERROR,
   EOS_ARROW_BACK,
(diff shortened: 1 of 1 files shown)


abravosuse (approved):
Leaving code aside, this looks awesome @rtorrero !

assets/js/components/ClusterDetails/ChecksResults.jsx
```diff
@@ -69,6 +73,14 @@ export const ChecksResults = () => {
     ]
   );
 
+  const checksRemediationById = Object.assign(
```
arbulu89, line 76: This approach works, but I find more "normal" to run a `find` function:
```
export getRemediationById = (id) => (catalogData) =>
  catalogData.flat().find((check) => check.id === id);
```
(didn't test, but I guess you get the idea)

assets/js/components/ClusterDetails/ChecksResults.jsx
```diff
@@ -54,6 +56,8 @@ const getResultIcon = (executionState, health) => {
 };
 
 export const ChecksResults = () => {
+  const [modalOpen, setModalOpen] = useState(false);
+  const [selectedCheck, setSelectedCheck] = useState(false);
```
arbulu89, line 60: Maybe better to default to an empty string or null? This value is a string after all, and not a boolean
`useState("")`

arbulu89 (commented):
The code looks good overall.
Not a big fan of the modals though, more when we are using collapsible rows everywhere else.

But if the PO like, I'm fine with it.
You have green light from me, just have a look on my comments, and feel free to change (if you want of course hehe)

PD: I guess that at some point we could use this information box to show some more information about why the check failed too

abravosuse:
> PD: I guess that at some point we could use this information box to show some more information about why the check failed too

Me linking this idea VERY MUCH :-) I will add a user story for it...

PR #694, opened 2022-07-06
Add support script

rtorrero:
This PR is a (draft) proposal to have a support script that customers or our L3 team can call to collect data about the installation.

The base idea is to collect everything that might help us determine what could be going wrong on the trento-server installation.

Some random thoughts:
 - Do we need something on the agent side too for agent related issues? Not sure if it would make sense to have something that needs to be executed on each agent or if instead we can rely on trento-server to collect such information from here
 - It would be nice if we can integrate this with `supportconfig` used in SLES. This script should still remain here (for non-SLES users), but could write a very simple wrapper plugin for supportconfig that call this script to collect all that it needs. The only change we would need is that we'd have to output to STDOUT instead of a file (as IIRC, supportconfig plugins are expect to output to STDOUT)

Suggestions welcome!

### DIFF
--- a/hack/trento-support.sh
+++ b/hack/trento-support.sh
@@ -1,0 +1,122 @@
+#!/bin/bash
+
+set -e
+
+readonly ARGS=("$@")
+readonly NOW="$(date +%x_%H%M%S)"
+
+indent() { sed 's/^/  /'; }
+
+trento_configuration() {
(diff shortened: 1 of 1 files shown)


rtorrero:
Closing this one in favour of implementing this in the `helm-charts` repo:

https://github.com/trento-project/helm-charts/pull/32

PR #695, opened 2022-07-07
Allow custom email sender

nelsonkopliku:
Allow the user to provide a custom email sender for alerting.
Relates to https://github.com/trento-project/helm-charts/pull/31

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -102,7 +102,8 @@ if config_env() == :prod do
 
   config :trento, :alerting,
     enabled: System.get_env("ENABLE_ALERTING", "false") == "true",
-    recipient: System.get_env("ALERT_RECIPIENT") || "admin@trento-project.io"
+    sender: System.get_env("ALERT_SENDER", "alerts@trento-project.io"),
+    recipient: System.get_env("ALERT_RECIPIENT", "admin@trento-project.io")
 
   config :trento, Trento.Mailer,
     adapter: Swoosh.Adapters.SMTP,
--- a/config/config.exs
+++ b/config/config.exs
@@ -32,6 +32,7 @@ config :swoosh, :api_client, false
 # configure the recipient for alert notifications
 config :trento, :alerting,
   enabled: true,
+  sender: "alerts@trento-project.io",
   recipient: "admin@trento.io"
 
 # Configure esbuild (the version is required)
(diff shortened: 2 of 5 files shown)


CDimonaco (approved):
LGTM

rtorrero (approved):
LGTM

config/runtime.exs
```diff
@@ -102,6 +102,7 @@ if config_env() == :prod do
 
   config :trento, :alerting,
     enabled: System.get_env("ENABLE_ALERTING", "false") == "true",
+    sender: System.get_env("ALERT_SENDER") || "alerts@trento-project.io",
```
fabriziosestito, line 105: ````elixir System.get_env("ALERT_SENDER", "alerts@trento-project.io")```

nelsonkopliku: right :smile: 
Changed also the recipient to 
```elixir
System.get_env("ALERT_RECIPIENT", "admin@trento-project.io")
```

stefanotorresi:
Since we're touching this code, do you think it could be feasible to make both recipient and sender follow the customary mailbox format `name <address>` according to https://datatracker.ietf.org/doc/html/rfc5322#section-3.4, so that we users can customize the entire header value in one shot?

nelsonkopliku:
> Since we're touching this code, do you think it could be feasible to make both recipient and sender follow the customary mailbox format `name <address>` according to https://datatracker.ietf.org/doc/html/rfc5322#section-3.4, so that we users can customize the entire header value in one shot?

Not an expert here, but I am wondering whether that could open to some sort of security issue where bad things could be injected :thinking: 
If so, that would then add the complexity of validation.

It might be interesting allowing that customization, yet I believe the added value is not enough compared to the security risks, if any, and the validation added complexity.

Do you see other options?

stefanotorresi:
> Not an expert here, but I am wondering whether that could open to some sort of security issue where bad things could be injected thinking If so, that would then add the complexity of validation.
> 
> It might be interesting allowing that customization, yet I believe the added value is not enough compared to the security risks, if any, and the validation added complexity.

I was expecting the validation to be done by Swoosh, but after I looked it up it seems it doesn't support mailbox headers, so I guess the answer to my question is "no, it's not easily feasible". I would never want you to re-implement RFC5322 email header validation, that's what framework should do for us. 😉

PR #704, opened 2022-07-13
Reset pagination when a filter is selected

arbulu89:
Fix filtering when a non 1st page is selected.

Before the fix, this was the behaviour:
Peek 2022-07-12 09-39

Now:
Peek 2022-07-13 09-46

PD: I know, the number of hosts in the feature file and the test don't match... but it comes from before, and I don't want to start changing that in this PR


stefanotorresi (approved):
LGTM

PR #723, opened 2022-07-22
Add automated deployments to the demo environment

rtorrero:
This PR makes use of self hosted GitHub actions runner to enable the automatic deployment of our demo environment. As of now, it's just the trento-server being deployed (no agents) and no mocked data. That will follow on a separate PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -84,7 +84,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
-          node-version: ${{ env.NODE_VERSION }}        
+          node-version: ${{ env.NODE_VERSION }}
 
       - name: Install NPM dependencies
         if: steps.npm-cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -421,6 +421,43 @@ jobs:
...
+          k3s-channel: latest
+
+      - name: Add bitnami helm deps
+        run: |
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm repo update
+
+      - name: Download and unzip helm chart
+        run: |
+          rm rolling.zip | true
```
rtorrero, line 445: I'm having some doubts about this: should each deployment fetch the rolling helm charts or should we make exceptions for the releases so that they also attempt to install the helm charts that correspond to that same release?

IMO going always for the `rolling` tag is safer but I'd like to have more opinions on it

rtorrero, line 445: I'm having some doubts about this: should each deployment fetch the rolling helm charts or should we make exceptions for the releases so that they also attempt to install the helm charts that correspond to that same release?

IMO going always for the `rolling` tag is safer but I'd like to have more opinions on it

arbulu89: By don't use simply use `$IMAGE_TAG`. It would actually do the same, but it looks more elegant at least. I guess the archives with the tag names are created. Otherwise, you can always use `main.zip` instead of rolling as well,

.github/workflows/ci.yaml
```diff
@@ -421,6 +421,43 @@ jobs:
...
+  deploy-demo-env:
+    name: Deploy updated images to the demo environment
+    runs-on: self-hosted
+    if: github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
+    env:
+      IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}/trento-web
+      IMAGE_TAG: "${{ (github.event_name == 'release' && github.event.release.tag_name) || (github.event_name == 'push' && github.ref_name == 'main' && 'rolling') || github.sha }}"
+    needs: build-and-push-container-images
+    steps:
+      - name: Start a local k8s cluster
```
arbulu89, line 433: Are these initial 2 steps needed? I would expect that these utilities are already in our server. Anyway, if they are not harmful go ahead

.github/workflows/ci.yaml
```diff
@@ -421,6 +421,43 @@ jobs:
...
+          rm rolling.zip | true
+          rm -rf helm-charts-rolling | true
+          wget https://github.com/trento-project/helm-charts/archive/refs/tags/rolling.zip
+          unzip rolling.zip
+
+      - name: Install trento-server helm chart
+        run: |
+          cd helm-charts-rolling/charts/trento-server
+          helm dependency update
+          helm uninstall --wait trento | true
```
arbulu89, line 454: I would try to use the `update` option rather than `uninstall -> install`. It looks like something more closer to what an actual user would do

arbulu89 (commented):
It looks good as an initial way for deploying the web server as it is now.
I have written down some comments before approving. Waiting for your answer.

.github/workflows/ci.yaml
rtorrero: Yep, I noticed that `upgrade --install` is a much better option, thanks!

.github/workflows/ci.yaml
rtorrero: The trouble then would be that it would enforce a requirement that the helm chart version x.y.z is already released when version x.y.z gets released here. The rolling is a more generic one

.github/workflows/ci.yaml
rtorrero: They are not needed, but I'd keep then as they are useful in the case when we need to deploy a new demo VM where to deploy to. It removes the requirement of preinstalling anything on the server while at the same time we are being more declarative. wdyt?

lib/trento/application/usecases/hosts/heartbeats_fake.ex
```diff
@@ -0,0 +1,17 @@
...
+  Heartbeat faker for demo environment
+  """
+  require Logger
+
+  alias Trento.Hosts
+
+  def send_heartbeats do
+    hosts = Hosts.get_all_hosts()
+
+    Enum.map(hosts, fn host ->
```
fabriziosestito, line 12: `Enum.each`

fabriziosestito (approved):

nelsonkopliku (approved):
As long as you find agreement also with Xabi, green light :green_circle:

.github/workflows/ci.yaml
```diff
@@ -421,6 +421,99 @@ jobs:
...
+    if: github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
+    needs: [static-code-analysis, test, test-e2e]
+    permissions:
+      contents: read
+      packages: write
+    env:
+      MIX_ENV: demo
+      REGISTRY: ghcr.io
+      IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}/trento-web
+      IMAGE_TAG: "${{ (github.event_name == 'release' && github.event.release.tag_name) || (github.event_name == 'push' && github.ref_name == 'main' && 'rolling') || github.sha }}"
```
arbulu89, line 436: This `IMAGE_TAG` variable is not used anywhere, we could remove it

Dockerfile
```diff
@@ -22,18 +23,18 @@ RUN npm run build
 FROM elixir-build AS release
 COPY --from=assets-build /build /build
 WORKDIR /build
-ENV MIX_ENV=prod
```
arbulu89, line 25: Is this removal safe? Does the `mix release` still use the expected value?

config/demo.exs
```diff
@@ -0,0 +1,31 @@
...
+# config :logger, level: :info
+
+config :trento, Trento.Scheduler,
+  jobs: [
+    clusters_checks_execution: [
+      schedule: {:extended, "@hourly"}
+    ],
+    heartbeat_fake: [
+      schedule: {:extended, "*/5"},
+      task: {Trento.Heartbeats.Faker, :send_heartbeats, []},
```
arbulu89, line 20: Just as a naming convention, it looks strange that we use `Faker` here and the `Mock` suffix in the others hehe

arbulu89 (commented):
Thanks @rtorrero ,
Added some more small comments, but besides that, it looks good

Dockerfile
rtorrero: I got this as a suggestion while pairing with @stefanotorresi but after checking it a bit more I think you are right. I'll re-add for now

config/demo.exs
rtorrero: This was intentional, there are some differences in the terms and @fabriziosestito and I agreed that "faker" was probably a better term here

fabriziosestito: oh i missed this, i think `Fake` is the right naming for this kind of stuff.
The Mock runner we use in dev should be a `Fake` as well. 

See: https://stackoverflow.com/questions/346372/whats-the-difference-between-faking-mocking-and-stubbing

PR #725, opened 2022-07-27
Highlight check result row on hover and id in green

arbulu89:
The colors are coming from https://tailwindcss.com/docs/background-color
@jagabomb Feel free to recommend other color set.

check-result-highlight

Edit: @dottorblaster @jagabomb I was actually thinking if all "clickable" rows in our app (check results, catalog, sap system/hana database expandable rows, etc) should follow the same styling


dottorblaster (approved):
If the background color looks good to @jagabomb then LGTM :smile:

jagabomb:
This looks good to me and I agree with @arbulu89 point of standardising all clickable rows in the ui to this styling. But before you do I just want to check it out in a few design mockups first before you action this as a global change.

PR #726, opened 2022-07-27
Enable telemetry community

arbulu89:
Enable telemetry publishing in the community flavor. Basically, the code removes the need of having the premium flavor to publish telemetry. This means that the EULA is shown in the 2 versions.

Besides that, a new field, `installation_flavor` is sent as other metric, so the option to filter by the options is available in the telemetry server side.

As a todo thing, the eula message is the same for the community flavor, and it relates to the premium version, something to change on the road.

Once this is accepted and merged, the next step would be to parse the new coming field in the telemetry server, and use it filter the grafana dashboards.

### DIFF
--- a/assets/js/state/sagas/index.js
+++ b/assets/js/state/sagas/index.js
@@ -113,10 +113,10 @@ function* initialDataFetch() {
   yield loadSapSystemsHealthSummary();
 
   const {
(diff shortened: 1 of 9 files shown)


fabriziosestito (approved):
lgtm

dottorblaster (approved):
LGTM

PR #728, opened 2022-07-29
Add installation_source field to the host telemetry

arbulu89:
Add the installation_source field to the host telemetry model.

This field is not required in the agent side, we so it is still backwards compatible with older agent version. In this cases, it is populated with `Unknown` string.

Depends on: https://github.com/trento-project/agent/pull/64

### DIFF
--- a/lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
+++ b/lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
@@ -3,16 +3,7 @@ defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
   Host discovery integration event payload
   """
 
-  @required_fields [
-    :hostname,
-    :ip_addresses,
-    :ssh_address,
-    :agent_version,
-    :cpu_count,
-    :total_memory_mb,
-    :socket_count,
-    :os_version
-  ]
+  @required_fields :all
 
   use Trento.Type
 
@@ -32,9 +23,7 @@ defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
   end
 
   def changeset(host, attrs) do
-    modified_attrs =
-      attrs
-      |> installation_source_to_downcase
+    modified_attrs = installation_source_to_downcase(attrs)
 
     host
     |> cast(modified_attrs, fields())
--- a/lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
+++ b/lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
@@ -25,6 +25,24 @@ defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
     field :total_memory_mb, :integer
     field :socket_count, :integer
     field :os_version, :string
-    field :installation_source, :string, default: "Unknown"
+
+    field :installation_source, Ecto.Enum,
+      values: [:community, :suse, :unknown],
+      default: :unknown
   end
+
+  def changeset(host, attrs) do
+    modified_attrs =
+      attrs
+      |> installation_source_to_downcase
+
+    host
+    |> cast(modified_attrs, fields())
+    |> validate_required_fields(@required_fields)
+  end
+
+  defp installation_source_to_downcase(%{"installation_source" => installation_source} = attrs),
+    do: %{attrs | "installation_source" => String.downcase(installation_source)}
+
+  defp installation_source_to_downcase(attrs), do: attrs
 end
--- a/lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
+++ b/lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
@@ -3,7 +3,16 @@ defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
   Host discovery integration event payload
   """
 
-  @required_fields :all
+  @required_fields [
+    :hostname,
+    :ip_addresses,
+    :ssh_address,
+    :agent_version,
+    :cpu_count,
+    :total_memory_mb,
+    :socket_count,
+    :os_version
+  ]
 
   use Trento.Type
 
@@ -16,5 +25,6 @@ defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
     field :total_memory_mb, :integer
     field :socket_count, :integer
     field :os_version, :string
+    field :installation_source, :string, default: "Unknown"
   end
 end
--- a/lib/trento/application/read_models/host_telemetry_read_model.ex
+++ b/lib/trento/application/read_models/host_telemetry_read_model.ex
@@ -18,6 +18,7 @@ defmodule Trento.HostTelemetryReadModel do
     field :socket_count, :integer
     field :total_memory_mb, :integer
     field :sles_version, :string
+    field :installation_source, :string
     field :provider, Ecto.Enum, values: [:azure, :aws, :gcp, :unknown]
 
     timestamps()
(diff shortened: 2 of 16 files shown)


rtorrero (approved):
I had imaged here he would use something like the repo url to identify the installation source, but I guess that what you propose here is more flexible. Looks good to me @arbulu89, thanks!

fabriziosestito (changes requested):

lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
```diff
@@ -16,5 +25,6 @@ defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
     field :total_memory_mb, :integer
     field :socket_count, :integer
     field :os_version, :string
+    field :installation_source, :string, default: "Unknown"
```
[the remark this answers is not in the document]
arbulu89: Done in: https://github.com/trento-project/web/pull/728/commits/6f6c1c750e44b143c890ce858a0ca94cdb2f9606

lib/trento/application/read_models/host_telemetry_read_model.ex
```diff
@@ -18,6 +18,7 @@ defmodule Trento.HostTelemetryReadModel do
     field :socket_count, :integer
     field :total_memory_mb, :integer
     field :sles_version, :string
+    field :installation_source, :string
```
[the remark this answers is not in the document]
arbulu89: Done: https://github.com/trento-project/web/pull/728/commits/6f6c1c750e44b143c890ce858a0ca94cdb2f9606

arbulu89:
Moving to `Draft` as this change creates a risky tech debt as the event store is not properly upcasted

fabriziosestito:
Unblocked by #751

arbulu89:
Upcasting applied. Waiting for https://github.com/trento-project/web/pull/751 to be rebased

arbulu89:
Hey @fabriziosestito ,
The upcasting implemented https://github.com/trento-project/web/pull/728/commits/ea9ddb932819a6985cf55e5653229ce19ca6b73b. Finally I have added a couple of basic tests, as they are quick to write.

lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
```diff
@@ -16,5 +25,24 @@ defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
...
+
+    field :installation_source, Ecto.Enum,
+      values: [:community, :suse, :unknown],
+      default: :unknown
+  end
+
+  def changeset(host, attrs) do
+    modified_attrs =
+      attrs
+      |> installation_source_to_downcase
```
fabriziosestito, line 37: `installation_source_to_downcase(attrs)`

lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
```diff
@@ -3,7 +3,16 @@ defmodule Trento.Integration.Discovery.HostDiscoveryPayload do
...
+  @required_fields [
+    :hostname,
+    :ip_addresses,
+    :ssh_address,
+    :agent_version,
+    :cpu_count,
+    :total_memory_mb,
+    :socket_count,
+    :os_version
+  ]
```
fabriziosestito, line 15: just curious, is this needed also we the default? isn't `:all` enough?

lib/trento/domain/host/commands/register_host.ex
```diff
@@ -17,5 +17,9 @@ defmodule Trento.Domain.Commands.RegisterHost do
     field :total_memory_mb, :integer
     field :socket_count, :integer
     field :os_version, :string, default: "Unknown"
+
+    field :installation_source, Ecto.Enum,
+      values: [:community, :suse, :unknown],
+      default: :unknown
```
fabriziosestito, line 23: is the default needed here in the command? i find it a bit opaque that we can build a command without that field and having a default value in place. this is already managed by the payload parsing, so I would not use any default when building the command.

lib/trento/domain/host/events/host_details_updated.ex
```diff
@@ -15,5 +15,11 @@ defmodule Trento.Domain.Events.HostDetailsUpdated do
     field :total_memory_mb, :integer
     field :socket_count, :integer
     field :os_version, :string
+
+    field :installation_source, Ecto.Enum,
+      values: [:community, :suse, :unknown],
+      default: :unknown
```
fabriziosestito, line 21: the default is not needed in the event

lib/trento/domain/host/events/host_registered.ex
```diff
@@ -15,6 +15,13 @@ defmodule Trento.Domain.Events.HostRegistered do
     field :total_memory_mb, :integer
     field :socket_count, :integer
     field :os_version, :string
+
+    field :installation_source, Ecto.Enum,
+      values: [:community, :suse, :unknown],
+      default: :unknown
```
fabriziosestito, line 21: same, default is not needed

test/support/factory.ex
```diff
@@ -64,6 +64,7 @@ defmodule Trento.Factory do
       total_memory_mb: Enum.random(1..128),
       socket_count: Enum.random(1..16),
       os_version: Faker.App.semver(),
+      installation_source: Enum.random([:community, :suse, :unknown]),
```
fabriziosestito, line 67: great!

test/trento/domain/host/host_test.exs
```diff
@@ -522,4 +531,26 @@ defmodule Trento.HostTest do
...
+
+    test "should upcast HostDetailsUpdated event properly" do
+      assert %HostDetailsUpdated{
+               version: 2,
+               installation_source: :unknown
+             } ==
+               %{}
+               |> HostDetailsUpdated.upcast(%{})
+               |> HostDetailsUpdated.new!()
+    end
```
fabriziosestito, line 554: I would split these tests into two different files with the event module name to reflect the module structure (e.g. domain/hosts/events/host_details_updated.exs`

Also, the test is a bit unclear, we need to upcast a map that is at version 1, not an empty map because this is what we get from the event store when dealing with old events.

Something like 
```elixir
%{
                 "version" => 1,
# insert all the host details updated events fields that are present in version 1 possibly with a faker
....
               }
```

fabriziosestito (changes requested):
Some minor stuff then good to go

lib/trento/application/integration/discovery/payloads/host_discovery_payload.ex
arbulu89: The `installation_source` field is not mandatory, as old versions of the agent won't send it, and we need to cover this use case

arbulu89: You are right, having a `default` value already fills the value, so `:all` works perfectly fine. I will revert it

test/trento/domain/host/host_test.exs
fabriziosestito: or event without the version field, since it should default it at version 1

arbulu89:
> Some minor stuff then good to go

Thanks @fabriziosestito ,
Changes done!

fabriziosestito (approved):

fabriziosestito:
> > Some minor stuff then good to go
> 
> Thanks @fabriziosestito , Changes done!

@arbulu89 go!

PR #735, opened 2022-08-01
Set version with git

arbulu89:
In order to get the version of the rolling container, this PR includes a new way of getting the application version. It uses the next order:
1. `VERSION` env variable during the release creation
2. `hack/get_version_from_git.sh` script output
3. Default version

image

PD: I have already checked solutions like [edeliver](https://github.com/edeliver/edeliver), but they look more complicated than the thing we really need. Anyway, if there is any better way of doing so, I'm all open to apply it!

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -1,11 +1,26 @@
 defmodule Trento.MixProject do
   use Mix.Project
 
+  @version "1.1.0"
+  defp get_version do
+    case System.get_env("VERSION") do
+      nil -> get_version_from_git()
+      version -> version
+    end
+  end
+
+  defp get_version_from_git do
+    case File.cwd!() |> Path.join("hack/get_version_from_git.sh") |> System.cmd([]) do
+      {version, 0} -> version |> String.trim("\n")
+      _ -> @version
+    end
+  end
+
   def project do
     [
       app: :trento,
       description: "Easing your life in administering SAP applications",
-      version: "1.1.0",
+      version: get_version(),
       elixir: "~> 1.13",
       elixirc_paths: elixirc_paths(Mix.env()),
       compilers: [:gettext] ++ Mix.compilers(),
--- a/mix.exs
+++ b/mix.exs
@@ -2,19 +2,6 @@ defmodule Trento.MixProject do
   use Mix.Project
 
   @version "1.1.0"
-  defp get_version do
-    case System.get_env("VERSION") do
-      nil -> get_version_from_git()
-      version -> version
-    end
-  end
-
-  defp get_version_from_git do
-    case File.cwd!() |> Path.join("hack/get_version_from_git.sh") |> System.cmd([]) do
-      {version, 0} -> version |> String.trim("\n")
-      _ -> @version
-    end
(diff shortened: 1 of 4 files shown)


mix.exs
```diff
@@ -1,11 +1,26 @@
...
+  @version "1.1.0"
+  defp get_version do
+    case System.get_env("VERSION") do
+      nil -> get_version_from_git()
+      version -> version
+    end
+  end
+
+  defp get_version_from_git do
+    case File.cwd!() |> Path.join("hack/get_version_from_git.sh") |> System.cmd([]) do
```
nelsonkopliku, line 13: I am wondering whether this is necessary here.
What I mean is that we're setting the `VERSION` env variable at build time by using `get_version_from_git.sh` so, do we really need to do that here? :thinking: 

The alternative I see would be: either at build time the `VERSION` has been set, or we fallback to the default version.

Am I missing something? would that work?

nelsonkopliku (commented):
Looks reasonable to me. Just a tiny thought to better understand.
I am also open to learn for better ways of doing it :smile:

mix.exs
arbulu89: Yes, I guess you are right. I just implemented this way to have the way to debug it in my development environment. 
We can delete this case, no issues with it.

Ok, I will remove it for the sake of simplicity

nelsonkopliku (approved):
LGTM

mix.exs
arbulu89: Edit: Ok, now I know why I have added it...
It was for the community version. The `VERSION` replacement only affect the suse container, but the community doesn't have such a replacement or Dockerfile edition, and the `docker/build-push-action@v3` job doesn't make it that straightforward

nelsonkopliku: Oh I see, is that the reason why e2e tests are failing?

Anyway, no strong opinion from my side. It is just fine like you did. :muscle:

mix.exs
```diff
@@ -1,11 +1,26 @@
 defmodule Trento.MixProject do
   use Mix.Project
 
+  @version "1.1.0"
+  defp get_version do
+    case System.get_env("VERSION") do
```
fabriziosestito, line 6: you dont need a case here, just `System.get_env("VERSION", get_version_from_git())` should work

mix.exs
fabriziosestito: > and the docker/build-push-action@v3 job doesn't make it that straightforward.
ah I see, have to say I am not a big fan of executing shell scripts from elixir but if there is not another way i am ok with it

mix.exs
```diff
@@ -1,11 +1,26 @@
 defmodule Trento.MixProject do
   use Mix.Project
 
+  @version "1.1.0"
+  defp get_version do
```
fabriziosestito, line 5: `defp` should be at the end of the module by convention

mix.exs
arbulu89: Thanks! I actually did it this way at some point... I guess too many back and forths hehe

PR #737, opened 2022-08-01
Remove gettext compiler from project configuration

dottorblaster:
Just addressing this warning I saw in my console:

```
warning: the :gettext compiler is no longer required in your mix.exs.

Please find the following line in your mix.exs and remove the :gettext entry:

    compilers: [..., :gettext, ...] ++ Mix.compilers(),
```


fabriziosestito:
that's new

PR #749, opened 2022-08-04
Set expectations for the about.js from cypress.env

mpagot:
It has only been tested against a Trento deployed in Azure. Not tested in the local simulated environment.
Tested with cypress.json like
```
{
  "baseUrl": "http://<IP_OF_MY_DEPLOYMENT>",
  "viewportWidth": 1366,
  "viewportHeight": 768,
  "defaultCommandTimeout": 10000,
  "env": {
    "web_api_host": "127.0.0.1",
    "web_api_port": 4000,
    "heartbeat_interval": 5000,
    "db_host": "127.0.0.1",
    "db_port": 5432,
    "project_root": "../..",
    "login_user": "admin",
    "login_password": "******",
    "flavour": "Premium",
    "version": "1.1.0",
    "subscriptions": 0
  }
}
``` 

Note that the env about photofinish has been removed, to disable its execution.

### DIFF
--- a/test/e2e/cypress/integration/about.js
+++ b/test/e2e/cypress/integration/about.js
@@ -17,17 +17,22 @@ describe('User account page', () => {
   });
 
   it('should show the correct server version', () => {
-    let version;
-    if (Cypress.env('version')) {
-      version = Cypress.env('version');
-    } else {
-      cy.exec(`cd ${Cypress.env('project_root')} && mix version`).then(
-        ({ stdout: out_version }) => {
-          version = out_version;
+    cy.get('div')
+      .contains('Server version')
+      .next()
+      .then(($sv) => {
+        if (Cypress.env('version')) {
+          expect($sv).to.contain(Cypress.env('version'));
+        } else {
+          cy.exec(`cd ${Cypress.env('project_root')} && mix version`)
+            .then(({ stdout: out_version }) => {
+              return out_version;
+            })
+            .then((version) => {
+              expect($sv.text()).to.contain(version);
+            });
         }
-      );
-    }
-    cy.get('div').contains('Server version').next().should('contain', version);
+      });
   });
 
   it('should show the github project link', () => {
--- a/test/e2e/cypress/integration/about.js
+++ b/test/e2e/cypress/integration/about.js
@@ -9,21 +9,25 @@ describe('User account page', () => {
   });
 
   it('should show the correct flavor', () => {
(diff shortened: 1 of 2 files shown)


test/e2e/cypress/support/commands.js
```diff
@@ -55,9 +55,11 @@ Cypress.Commands.add('loadScenario', (scenario) => {
     Cypress.env('web_api_port'),
   ];
   cy.log(`Loading scenario "${scenario}"...`);
-  cy.exec(
-    `cd ${projectRoot} && ${photofinishBinary} run --url "http://${webAPIHost}:${webAPIPort}/api/collect" ${scenario}`
-  );
+  if(photofinishBinary) {
```
dottorblaster: Could you move the `cy.log` statement inside this `if` and add an `else` clause that tells the user that `photofinish` isn't being used?

test/e2e/cypress/integration/about.js
```diff
@@ -1,42 +1,50 @@
...
+    let version;
+    if (Cypress.env("version")) {
+      version = Cypress.env("version");
+    } else {
+      cy.exec(`cd ${Cypress.env("project_root")} && mix version`).then(
+        ({ stdout: out_version }) => {
+          version = out_version;
+        }
+      );
+    }
```
dottorblaster, line 29: This expectation will run against an empty `version` variable due to the asyncronous nature of JS and Node on which Cypress is based. What you could do is turning this function into an `async` one and `await` for the `cy.exec` output. You can ping me anytime.

dottorblaster (changes requested):
Some enhancements to do, but we're definitely on the right track!

Also, congrats on your first PR :smile:

test/e2e/cypress/integration/about.js
rtorrero: What about just catching the missing `mix` binary error on the `cy.exec` and attempting to get the environment variable in that case? you'd have to run the `cy.get` in the if/else which is a bit of a smell... but I'm not a huge fan of how we are using `mix version` here anyways so I wouldn't invest too much time in it

mpagot: > catching the missing mix binary error 
Is it not more safe/clear to let the user decide, using the env variable, what to do? Is it not interesting (let the test to fails) to know if `mix version` is not able to properly run?

dottorblaster (approved):
:partying_face:

PR #751, opened 2022-08-05
Upcasting

fabriziosestito:
Missing tests


fabriziosestito:
@dottorblaster the Commanded protocol passes metadata to the upcaster but it is called _after_ the serialization that in our case is done by the `JsonbSerializer`. 
This means that the  upcaster will receive a struct.

We could remove the `new!/1` call in the `JsonbSerializer` and use the https://hexdocs.pm/commanded/Commanded.Serialization.JsonDecoder.html#content protocol in the `defevent` macro, so that we can implement Commanded upcaster protocol and make use of the metadata.

lib/trento/support/event.ex
```diff
@@ -5,9 +5,31 @@ defmodule Trento.Event do
...
+        |> upcast_params(metadata)
+      end
+
+      defp upcast_params(%{"version" => @version} = params, _) do
+        params
+      end
+
+      defp upcast_params(%{"version" => version} = params, metadata) do
+        params
+        |> upcast(metadata, version + 1)
```
arbulu89, line 29: Where is this `upcast` with 3 params defined?

arbulu89 (approved):
Excellent!
It looks great in my low knowledge of the concept hehe

2 questions:
- We assume that the version is `1` if it is not defined in the event, right?
- Where should we run the `upcast` method? In the `changeset` method for our types?

lib/trento/support/event.ex
fabriziosestito: you need to define it for every event if the upcast is needed

fabriziosestito:
> Excellent! It looks great in my low knowledge of the concept hehe
> 
> 2 questions:
> 
>     * We assume that the version is `1` if it is not defined in the event, right?

exactly

>     * Where should we run the `upcast` method? In the `changeset` method for our types?

not sure about this question, could you elaborate?

arbulu89:
> 
> not sure about this question, could you elaborate?

Sorry. I just wanted to know how and where we should use the `upcast` method. I'm not entirely sure of it. I guess we will have an example short enough in: https://github.com/trento-project/web/pull/728

fabriziosestito:
> > not sure about this question, could you elaborate?
> 
> Sorry. I just wanted to know how and where we should use the `upcast` method. I'm not entirely sure of it. I guess we will have an example short enough in: #728

@arbulu89 check https://github.com/trento-project/web/blob/d61de432e83fc5de1346a5bf8eb076e9ccdbb102/test/trento/support/event_test.exs in particular the `TestUpcastedEvent` event

dottorblaster (approved):
LGTM

PR #757, opened 2022-08-09
add nodejs to .tools-versions

stefanotorresi:
it's been too long 😛


mpagot (commented):
LGTM (locally tested with 16.15.1 and I'm able to start the server and run the cypress e2e tests)

PR #759, opened 2022-08-10
Add value files support inside the Cypress testing suite

dottorblaster:
Since we aim to reuse the same e2e tests against the CI environment and a "real world" one, this PR introduces support for different value files depending on which environment do we run the test against.

To switch the environment the cypress env variable `destination_environment` can be used.

### DIFF
--- a/test/e2e/cypress/support/common.js
+++ b/test/e2e/cypress/support/common.js
@@ -1,5 +1,5 @@
 import devValues from './values.dev.json';
-import prodValues from './values.dev.json';
+import prodValues from './values.prod.json';
 
 export const getValue = (key, defaultValue) => {
   const environment = Cypress.env('destination_environment');
--- a/test/e2e/cypress/support/common.js
+++ b/test/e2e/cypress/support/common.js
@@ -1,0 +1,15 @@
+import devValues from './values.dev.json';
+import prodValues from './values.dev.json';
+
+export const getValue = (key, defaultValue) => {
+  const environment = Cypress.env('destination_environment');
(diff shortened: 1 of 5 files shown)


test/e2e/cypress/support/common.js
```diff
@@ -0,0 +1,15 @@
+import devValues from './values.dev.json';
+import prodValues from './values.dev.json';
```
mpagot, line 2: values.dev.json --> values.prod.json

mpagot (changes requested):

test/e2e/cypress/support/common.js
dottorblaster: Oh man you're right, thanks :smile:

mpagot (approved):
LGTM

rtorrero:
This looks good :-)

But it makes me wonder: we have also a bunch of expectations defined in https://github.com/trento-project/web/tree/main/test/e2e/cypress/fixtures ; should we also move them away from there and populate the `values.dev.json` 

Also: I'm guessing `values.prod.json` will be generated dynamically?

dottorblaster:
@rtorrero my plan for now was to only get values dinamically based on the environment, I guess we can to the same for fixtures or move them to values file when needed. I opted out from fixtures because they work asynchronously and for simple values we just needed a simple sync call, maybe we can try to find a balance in the future.

About `values.prod.json`: it won't be generated dinamically, I'm faithful that @mpagot will fill it with the values he needs :smile:

PR #797, opened 2022-09-06
Refactor docker compose to add named volume for postgres persistence

CDimonaco:
We remove the local binding of docker compose in the project root.

We add a named docker-compose volume in order to maintain the persistence of postgres but avoid binding in the local directory with polluted data.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,5 +1,6 @@
 version: "3"
-
+volumes:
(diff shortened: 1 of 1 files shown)


fabriziosestito (approved):

arbulu89 (approved):
Good!

dottorblaster (approved):

PR #827, opened 2022-09-14
Downgrade to previus k3s due to PodSecurityPolicy deprecation

rtorrero:
Due to the latest k3s release deprecating the beta PodSecurityPolicy, we need to freeze and stay with 1.24.x

See [k3s 1.25.0 release notes](https://github.com/k3s-io/k3s/releases/tag/v1.25.0%2Bk3s1) for more information

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -471,7 +471,7 @@ jobs:
       - name: Start a local k8s cluster
         uses: jupyterhub/action-k3s-helm@v3
         with:
-          k3s-channel: latest
+          k3s-channel: v1.24.4+k3s1
 
       - name: Add bitnami helm deps
         run: |


fabriziosestito:
Couldn't we fix it instead of downgrading it?

rtorrero:
I suspect it's the grafana helm chart that we depend on which is causing the issue: 
```
Error: unable to build kubernetes objects from release manifest: [resource mapping not found for name: "trento-grafana" namespace: "" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
ensure CRDs are installed first, resource mapping not found for name: "trento-grafana-test" namespace: "" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
ensure CRDs are installed first]
```

dottorblaster:
I hope we just have to wait until the grafana's chart maintainers release a fix or something?

rtorrero:
I think we can close this as we managed to upgrade to a new grafana chart that doesn't expose this problem.
@stefanotorresi you mentioned that in the long term it could be interesting to pin to a stable version and add steps in the pipeline to test with other versions, can we close and add that as a task?

stefanotorresi:
Alright, so we keep the latest version for now.

You already collected a followup item in JIRA, what task do you want to add?

rtorrero:
> Alright, so we keep the latest version for now.
> 
> You already collected a followup item in JIRA, what task do you want to add?

Oh yes, I see that. Thanks!

PR #836, opened 2022-09-22
Add missing :kvm and :nutanix atoms to the provider field in events/commands/read models

fabriziosestito:
Add missing `:kvm` and `:nutanix` atoms to the provider field in events/commands/read models, except for the catalog where we just have `:aws | :gcp | :azure | :default`.

Improve testing by adding random providers to the factory.

### DIFF
--- a/lib/trento/application/integration/discovery/payloads/cluster/cluster_discovery_payload.ex
+++ b/lib/trento/application/integration/discovery/payloads/cluster/cluster_discovery_payload.ex
@@ -16,7 +16,11 @@ defmodule Trento.Integration.Discovery.ClusterDiscoveryPayload do
 
   deftype do
     field :dc, :boolean
-    field :provider, Ecto.Enum, values: [:aws, :gcp, :azure, :unknown], default: :unknown
+
+    field :provider, Ecto.Enum,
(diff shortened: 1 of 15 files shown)


arbulu89 (approved):
LGTM.
@fabriziosestito Isn't there some way to store the list as a constant somewhere where we can import it? To avoid the duplication of the same options all over the code.

fabriziosestito:
> LGTM. @fabriziosestito Isn't there some way to store the list as a constant somewhere where we can import it? To avoid the duplication of the same options all over the code.

@arbulu89 we need some macro magic to achieve that, as the list is used inside an ecto schema aka at compile time. Let's keep track of this since I need to investigate the proper way to do it.

arbulu89:
@fabriziosestito Not a big deal hehe.
It is for convenience mostly (and to avoid some error if we forget to add the values in some piece of code).
We can go like the current code by now

fabriziosestito:
> @fabriziosestito Not a big deal hehe. It is for convenience mostly (and to avoid some error if we forget to add the values in some piece of code). We can go like the current code by now

@arbulu89 I think it makes sense, I've used a library before that helped define "enums", but as we are using ecto it wouldn't work. I am tracking this as a generic debt about enum types.

dottorblaster (approved):
LGTM

PR #837, opened 2022-09-22
Create BackToCluster button component

arbulu89:
Unify the `BackToCluster` button usage.

image

### DIFF
--- a/assets/js/components/ClusterDetails/BackToCluster.jsx
+++ b/assets/js/components/ClusterDetails/BackToCluster.jsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import Button from '@components/Button';
+import { useNavigate } from 'react-router-dom';
+import { EOS_ARROW_BACK } from 'eos-icons-react';
+
+export const BackToCluster = ({ clusterId }) => {
+  const navigate = useNavigate();
+  return (
+    <div className="flex mb-8">
+      <div className="flex w-2/5">
+        <Button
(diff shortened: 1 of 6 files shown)


assets/js/components/ClusterDetails/BackToCluster.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import Button from '@components/Button';
+import { useNavigate } from 'react-router-dom';
+import { EOS_ARROW_BACK } from 'eos-icons-react';
+
+export const BackToCluster = ({ clusterId }) => {
```
fabriziosestito, line 6: maybe we could generalize that, by creating a BackButton component and passing a prop/child

arbulu89: Maybe just asking the `url` and `text`, without making it more complex.

dottorblaster (changes requested):
The BackToCluster component is a little bit too specific, you can try generalizing it more converting that to a BackLink or something named like that, also I'd like to have a render test for such a component.

assets/js/components/ClusterDetails/BackToCluster.jsx
dottorblaster: Yeah, @arbulu89 bear in mind that you can also use `children` for that and just pass the text as `<BackLink>Back to cluster details</BackLink>`

CDimonaco (approved):
LGTM

dottorblaster (approved):
LGTM

PR #840, opened 2022-09-23
Show warning unknown check selection

arbulu89:
Include warning message in the check selection and results if the provider is unknown (out of azure,aws,gcp,nutanix,kvm):

image

image

### DIFF
--- a/assets/js/components/ClusterDetails/ChecksResults.jsx
+++ b/assets/js/components/ClusterDetails/ChecksResults.jsx
@@ -11,6 +11,7 @@ import LoadingBox from '../LoadingBox';
 
 import Button from '@components/Button';
 import BackButton from '@components/BackButton';
+import WarningBanner from '@components/Banners/WarningBanner';
 
 import { getCluster } from '@state/selectors';
 import TrentoLogo from '../../../static/trento-icon.png';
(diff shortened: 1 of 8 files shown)


assets/js/components/ClusterDetails/ChecksResults.jsx
```diff
@@ -253,6 +237,14 @@ export const ChecksResults = () => {
           }
         />
       </div>
+      {cluster.provider == 'unknown' && (
```
dottorblaster, line 240: You can do

```js
const UNKNOWN = 'unknown';
```

on top of the file and use it here as

```js
cluster.provider == UNKNOWN &&
```

it would be a little more idiomatic.

dottorblaster (changes requested):
One tiny thing, the we can merge :+1:

assets/js/components/ClusterDetails/ChecksResults.jsx
arbulu89: Done: https://github.com/trento-project/web/pull/840/commits/a9156c6d81d1a1365cfe173f8bb0e6cc8c133912

dottorblaster (approved):
perfection

fabriziosestito (approved):

jagabomb:
@arbulu89 Looks great! Minor point for the future is that I should have made the border radius 8px for the warning box. I realised that the two different pages have inconsistencies with the container border radii. An 8px border is more commonly used throughout the UI.

PR #849, opened 2022-09-28
Add provider enum

rtorrero:
# Description

This PR adds:
  - A new `Trento.Support.Enum` module with:
    - A `.values()` macro: Get the possible enum values
    - Dynamically created macros based on the enum values: Can be used to validate a enum value; instead of using the `:something` directly, you can use `Trento.Domain.SomeEnum.something()` that will fail in `:something` is not a valid value for the enum.
    
  - A new `Trento.Domain.Provider` module that uses the above.

## How was this tested?

No new tests added, existing tests are passing.

### DIFF
--- a/lib/trento/application/integration/checks/dto/flat_check_dto.ex
+++ b/lib/trento/application/integration/checks/dto/flat_check_dto.ex
@@ -18,7 +18,7 @@ defmodule Trento.Integration.Checks.FlatCheckDto do
 
   deftype do
     field :id, :string
-    field :provider, Ecto.Enum, values: [:azure, :aws, :gcp, :default, :unknown]
+    field :provider, Ecto.Enum, values: [:azure, :aws, :gcp, :default]
     field :group, :string
     field :name, :string
     field :description, :string
--- a/lib/trento/application/integration/checks/dto/provider_dto.ex
+++ b/lib/trento/application/integration/checks/dto/provider_dto.ex
@@ -13,7 +13,7 @@ defmodule Trento.Integration.Checks.ProviderDto do
   alias Trento.Integration.Checks.GroupDto
 
   deftype do
-    field :provider, Ecto.Enum, values: [:azure, :aws, :gcp, :default, :unknown]
+    field :provider, Ecto.Enum, values: [:azure, :aws, :gcp, :default]
 
     embeds_many :groups, GroupDto
   end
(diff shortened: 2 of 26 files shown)


arbulu89 (changes requested):
Hey @rtorrero ,
Now I remember the difference on arrays between the ones that have the `default` entry and not.
There is not such a `default` provider. No agent will send this. We use the `unknown` value there. So i cases where we really are using the agent provider value, we should have the `[:aws, :gcp, :azure, :kvm, :nutanix, :unknown]` list.

The other case is the catalog options, which include the `default` value. So my recommendation would be to create other enum such as `CatalogOptions` (try to think a good name please XD) where we have the `[:azure, :aws, :gcp, :default]`. This must be used when we deal with the catalog options (nutanix and kvm use the default catalog)

PD:
Checking the code I miss the `health.ex` file that was in `enums` folder, but happened to it. It should follow the same convention than with provider I guess

rtorrero:
> Hey @rtorrero , Now I remember the difference on arrays between the ones that have the `default` entry and not. There is not such a `default` provider. No agent will send this. We use the `unknown` value there. So i cases where we really are using the agent provider value, we should have the `[:aws, :gcp, :azure, :kvm, :nutanix, :unknown]` list.
> 
> The other case is the catalog options, which include the `default` value. So my recommendation would be to create other enum such as `CatalogOptions` (try to think a good name please XD) where we have the `[:azure, :aws, :gcp, :default]`. This must be used when we deal with the catalog options (nutanix and kvm use the default catalog)

Ok, I was thinking initially of joining these two but it makes sense to consider the "default" catalog as a sum of all the (yet) non-supported providers such as nutanix & kvm.

> PD: Checking the code I miss the `health.ex` file that was in `enums` folder, but happened to it. It should follow the same convention than with provider I guess
More PRs are coming regarding this; this one specifically focuses the providers enums.

fabriziosestito:
@rtorrero @arbulu89 I would leave the catalog as it is (I mean with explicit atom list and without the enum macro), for two reasons. The first is that it is not really a "domain" enum, and the second one is because we are about to drop the catalog support in this project so I think it will be a little waste.

arbulu89 (approved):

fabriziosestito (changes requested):

PR #850, opened 2022-09-28
Added margin to the bottom of each result/host container.

jagabomb:
# Description

Added margin to the bottom of the each container/host in the results on the Check Results screen. This margin appears for every container except for the last one in the list.

Screenshot 2022-09-28 at 16 35 02

## How was this tested?

Visually

### DIFF
--- a/assets/js/components/ClusterDetails/ChecksResults.jsx
+++ b/assets/js/components/ClusterDetails/ChecksResults.jsx
@@ -25,6 +25,7 @@ import { getClusterName } from '@components/ClusterLink';
 import ChecksResultFilters, {
   useFilteredChecks,
 } from '@components/ClusterDetails/ChecksResultFilters';
+import classNames from 'classnames';
 
 import ReactMarkdown from 'react-markdown';
 import remarkGfm from 'remark-gfm';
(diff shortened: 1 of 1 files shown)


assets/js/components/ClusterDetails/ChecksResults.jsx
```diff
@@ -126,7 +126,7 @@ export const ChecksResults = () => {
 
     pageContent = lastExecution.map(
       ({ _cluster_id, host_id, reachable, msg }, idx) => (
-        <div key={idx} className="flex flex-col">
+        <div key={idx} className={`flex flex-col ${idx != (cluster.hosts_executions.length - 1) ? "mb-8" : ""}`}>
```
dottorblaster, line 129: You could use `classNames` (`import classNames from 'classnames'`) and just do `classNames('flex', 'flex-col', {'mb-8': idx != (cluster.hosts_executions.length - 1)})`

dottorblaster (changes requested):

assets/js/components/ClusterDetails/ChecksResults.jsx
jagabomb: Cool, I can make that change.

dottorblaster (approved):
LGTM

PR #854, opened 2022-09-29
Community eula

arbulu89:
# Description

Add `Premium` vs `Community` eula distinction.
Selection_453

Yes

## How was this tested?

Jest tests added.
PD: `intersection-observer` package is needed to run the tests correctly. Ref: https://stackoverflow.com/questions/57008341/jest-testing-react-component-with-react-intersection-observer

### DIFF
--- a/assets/js/components/Eula/Eula.jsx
+++ b/assets/js/components/Eula/Eula.jsx
@@ -1,8 +1,8 @@
 import React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 
-import Premium from '@components/Eula/Premium';
-import Community from '@components/Eula/Community';
+import Premium from './Premium';
+import Community from './Community';
 
 const Eula = () => {
   const eulaVisible = useSelector((state) => state.settings.eulaVisible);
--- a/assets/js/components/Eula/Eula.jsx
+++ b/assets/js/components/Eula/Eula.jsx
@@ -1,91 +1,14 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 
-import Button from '@components/Button';
-import Modal from '@components/Modal';
+import Premium from '@components/Eula/Premium';
 
 const Eula = () => {
   const eulaVisible = useSelector((state) => state.settings.eulaVisible);
   const dispatch = useDispatch();
 
   return (
-    <Fragment>
-      <Modal title="License agreement" open={eulaVisible} onClose={() => {}}>
-        <div className="mb-4">
-          <div className="row">
-            <div>
-              <p className="my-2">
-                Trento, an open cloud-native web console improving the life of
-                SAP Applications administrators
-              </p>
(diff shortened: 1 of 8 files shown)


assets/js/components/Eula/Eula.jsx
```diff
@@ -1,92 +1,19 @@
-import React, { Fragment } from 'react';
+import React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 
-import Button from '@components/Button';
-import Modal from '@components/Modal';
+import Premium from '@components/Eula/Premium';
+import Community from '@components/Eula/Community';
```
dottorblaster, line 5: These imports can be local, there's no need to make them absolute :+1:

assets/js/components/Eula/Eula.test.jsx
```diff
@@ -9,12 +9,24 @@ import { setEulaVisible } from '@state/settings';
...
+  // it('should render premium eula correctly', () => {
+  //   const [statefulEula, store] = withState(<Eula />);
+
+  //   store.dispatch(setEulaVisible());
+  //   store.dispatch(setIsPremium());
+
+  //   render(statefulEula);
+  //   expect(screen.getByRole('button')).toHaveTextContent('Accept');
+  //   expect(screen.getByText(/Trento Premium/)).toBeTruthy();
+  // });
```
dottorblaster: Why is this commented? I think it's fine :smile:

assets/js/state/sagas/index.js
```diff
@@ -113,11 +113,14 @@ function* initialDataFetch() {
...
   const {
-    data: { eula_accepted },
+    data: { eula_accepted, premium_subscription },
   } = yield call(get, '/api/settings');
 
   if (!eula_accepted) {
     yield put(setEulaVisible());
+    if (premium_subscription) {
+      yield put(setIsPremium());
+    }
```
dottorblaster, line 123: I think we can extract this if statement to the upper level :+1:

dottorblaster (changes requested):
Just a couple comments then we can merge

assets/js/components/Eula/Eula.test.jsx
dottorblaster: LOL you outran me

arbulu89: It was failing, because I didn't include the `intersection-observer`, now it works

assets/js/state/sagas/index.js
arbulu89: Well, we can, but as it is only used to show eula. I will move it anyway, if we need to use it somewhere else

CDimonaco (approved):

dottorblaster (approved):
LGTM

assets/js/state/sagas/index.js
dottorblaster: Yeah that was exactly my point :+1:

PR #855, opened 2022-09-29
Add tag validation

### LINKED ISSUE #464: Disallow certain chars in tags
Right now, adding e.g. a `\` char in a tag breaks the functionality as it's not able to delete the created tag.

EMaksy:
# Description

This pr adds: 
- frontend tag validation
- backend tag validation
- new test for wrong tags

When a user wants to create a tag in the frontend, every character of the input is validated.
In the case a forbidden character was used, the frontend will reject the input and don't add the character.
Additionally, a backend validation was added to ensure that the backend is protected when a user tries directly to interact with the backend.

Allowed characters are letters, numbers representable in UTF-8, and the following characters: + - = . , _ : @
This is leaned to the aws tag guidelines, for **instance tags** 
See here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html

Fixes  #464

## How was this tested?
Existing tests were adjusted to the new requirements. 
A new test was added to check if the wrong tags are handled.

### DIFF
--- a/test/trento_web/controllers/tags_controller_test.exs
+++ b/test/trento_web/controllers/tags_controller_test.exs
@@ -6,7 +6,7 @@ defmodule TrentoWeb.TagsControllerTest do
   alias Trento.Tag
 
   describe "Tag Validation" do
-    test "should validate incoming tags", %{conn: conn} do
+    test "should decline tag with whitespace", %{conn: conn} do
       conn =
         post(conn, Routes.hosts_tagging_path(conn, :add_tag, Faker.UUID.v4()), %{
           "value" => "     "
@@ -15,6 +15,19 @@ defmodule TrentoWeb.TagsControllerTest do
       assert %{
                "errors" => %{
                  "value" => ["can't be blank"]
+               }
+             } = json_response(conn, 400)
+    end
+
+    test "should decline tag with forbidden characters", %{conn: conn} do
+      conn =
+        post(conn, Routes.hosts_tagging_path(conn, :add_tag, Faker.UUID.v4()), %{
+          "value" => "This / is a \ wrong #tag"
+        })
+
+      assert %{
+               "errors" => %{
+                 "value" => ["has invalid format"]
                }
              } = json_response(conn, 400)
     end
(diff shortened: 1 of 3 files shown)


CDimonaco:
The frontend has some sort of error message on the failed validation?

EMaksy:
Currently, the frontend does not give user feedback on wrong input.
An UI pop feedback was already discussed and needs to be refined for the future implementation.

rtorrero:
> The frontend has some sort of error message on the failed validation?

I wouldn't add any message, the fact that the invalid character simply doesn't appear on the input box seems to me that it should give enough feedback. A message could be a bit annoying in this context, wdyt?

test/trento_web/controllers/tags_controller_test.exs
```diff
@@ -18,13 +18,26 @@ defmodule TrentoWeb.TagsControllerTest do
...
+             } = json_response(conn, 400)
+    end
   end
 
   describe "tagging sap systems and databases" do
     test "should add a tag to a sap system", %{conn: conn} do
       conn =
         post(conn, Routes.sap_systems_tagging_path(conn, :add_tag, Faker.UUID.v4()), %{
-          "value" => Faker.Beer.style()
+          "value" => String.replace(Faker.Beer.style(), " ", "")
```
rtorrero, line 40: I'd propose to switch `Faker.Beer` with something that maybe doesn't have spaces, wdyt?

There is e.g. https://github.com/elixirs/faker/blob/master/lib/faker/color/en.ex
(much more boring ofc... ;-p)

EMaksy: Haha that's true... well I will miss the beer, but you are right thanks :D

EMaksy:
So for me as a user, when I try to create a tag and my input was not added, I get the feeling that something is not working properly. A popup could inform me and show me the allowed character. So I directly can do it right on the second try.

For now this works fine, but I would suggest adding some user feedback in the future.

What do you think, guys?

dottorblaster (approved):
My thoughts:

- The eslint ignore is fine for now, tinkering with this doesn't stop us merging
- The visual feedback is a lot of work that we can track in a further user story to refine

LGTM

PR #856, opened 2022-09-29
Use new enum type for Health

rtorrero:
# Description
Following up on the work started in PR #849, this also uses the new Enum type macros and substitutes all the literal references to the possible health values with the macros.

## How was this tested?
No new tests added, existing tests are passing.

### DIFF
--- a/lib/trento/application/read_models/application_instance_read_model.ex
+++ b/lib/trento/application/read_models/application_instance_read_model.ex
@@ -7,6 +7,8 @@ defmodule Trento.ApplicationInstanceReadModel do
 
   import Ecto.Changeset
 
+  require Trento.Domain.Enum.Health, as: Health
+
   @type t :: %__MODULE__{}
 
   alias Trento.HostReadModel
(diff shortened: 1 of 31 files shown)


arbulu89 (approved):

nelsonkopliku (approved):
Nice!

lib/trento/application/read_models/application_instance_read_model.ex
```diff
@@ -7,6 +7,8 @@ defmodule Trento.ApplicationInstanceReadModel do
 
   import Ecto.Changeset
 
+  require Trento.Domain.Enum.Health, as: Health
```
fabriziosestito, line 10: this should be called `Trento.Domain.Enums.Health` since the folder is called enums

lib/trento/application/read_models/cluster_read_model.ex
```diff
@@ -8,6 +8,7 @@ defmodule Trento.ClusterReadModel do
   import Ecto.Changeset
 
   require Trento.Domain.Enum.Provider, as: Provider
+  require Trento.Domain.Enum.Health, as: Health
```
fabriziosestito, line 11: `Trento.Domain.Enums.Provider`
`Trento.Domain.Enums.Health`

fabriziosestito (changes requested):

lib/trento/application/read_models/cluster_read_model.ex
fabriziosestito: @rtorrero let' s fix this in another PR

rtorrero: aight!

fabriziosestito (approved):

PR #857, opened 2022-09-29
Aggregates snapshot

fabriziosestito:
Enables aggregate snapshots, see: https://hexdocs.pm/commanded/aggregates.html#aggregate-state-snapshots

To do this, I needed to refactor the aggregate state to use the `deftype` macro.

### DIFF
--- a/lib/trento/domain/sap_system/instance.ex
+++ b/lib/trento/domain/sap_system/instance.ex
@@ -3,25 +3,19 @@ defmodule Trento.Domain.SapSystem.Instance do
   This module represents a SAP System instance.
   """
 
-  alias Trento.Domain.Enums.Health
+  require Trento.Domain.Enums.Health, as: Health
 
-  defstruct [
-    :sid,
-    :instance_number,
-    :features,
-    :host_id,
-    :health,
-    :system_replication,
-    :system_replication_status
-  ]
+  @required_fields []
 
-  @type t :: %__MODULE__{
-          sid: String.t(),
-          instance_number: String.t(),
-          features: String.t(),
-          host_id: String.t(),
-          health: Health.t(),
-          system_replication: String.t(),
-          system_replication_status: String.t()
-        }
+  use Trento.Type
+
+  deftype do
+    field :sid, :string
+    field :instance_number, :integer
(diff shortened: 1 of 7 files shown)


arbulu89 (commented):
Why did you need to change to use `deftype`? In the example of the refence as see `defstruct` (which was our old implementation).
Is it for the polymorphich fields?

fabriziosestito:
> Why did you need to change to use `deftype`? In the example of the refence as see `defstruct` (which was our old implementation). Is it for the polymorphich fields?

@arbulu89 it is used to cast the snapshot saved in the db to the aggregate state as we do with events. the example assumes you have already some sort of casting in place.

CDimonaco (approved):

lib/trento/domain/sap_system/instance.ex
```diff
@@ -3,25 +3,19 @@ defmodule Trento.Domain.SapSystem.Instance do
...
-          host_id: String.t(),
-          health: Health.t(),
-          system_replication: String.t(),
-          system_replication_status: String.t()
-        }
+  use Trento.Type
+
+  deftype do
+    field :sid, :string
+    field :instance_number, :integer
```
arbulu89, line 14: I'm not sure if we should change the instance number to integer. Yes it is an integer, but we always use it as a 2 digit string. I would be conservative with this an keep string

lib/trento/domain/sap_system/instance.ex
```diff
@@ -3,25 +3,19 @@ defmodule Trento.Domain.SapSystem.Instance do
...
+    field :features, {:array, :string}
+    field :host_id, :string
```
arbulu89, line 16: `host_id` maybe as UUID?

arbulu89 (approved):
Good!
Commented some minor things that you could change without other review

lib/trento/domain/sap_system/instance.ex
fabriziosestito: mmh i think i mistyped it thank you

dottorblaster (approved):
LGTM

nelsonkopliku (approved):
Nice! Thanks!

PR #869, opened 2022-10-03
Enable single pipe check on credo

arbulu89:
# Description
Enable `SinglePipe` check on credo: https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/single_pipe.ex

On @fabriziosestito request XD (as he always forgets hehe)


fabriziosestito (approved):
Happy when credo is

PR #872, opened 2022-10-05
Add mix deps.unlock --check-unused to CI

dottorblaster:
# Description
As the team agreed upon during the planning, we are adopting `mix deps.unlock --check-unused` in our CI.

Fixes # (issue)

Aye

## How was this tested?
If the CI is happy, we are I guess


arbulu89 (commented):
Is it possible to run a non cached execution? I think this execution should fail (it does in my local machine)

dottorblaster:
@arbulu89 yeah you're right, because we had a cache hit. Let's see if removing unused dependencies triggers the step

.github/workflows/ci.yaml
```diff
@@ -56,6 +56,7 @@ jobs:
           mix local.rebar --force
           mix local.hex --force
           mix deps.get
+          mix deps.unlock --check-unused
```
fabriziosestito, line 59: I am late to the party, but this should be a listing step of "Static code analysis"

PR #878, opened 2022-10-07
Health summary with clusters and databases

CDimonaco:
# Description

The health overview api endpoint, has two new fields, the `database_id`  and `hana_cluster_id`, we need them in order to achieve new functionalities in the frontend that consumes this api endpoint.

## How was this tested?

Integration testing for both service and endpoint.


lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -48,10 +48,26 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
       database_health: compute_database_health(database_instances),
       clusters_health: compute_clusters_health(all_instances),
-      hosts_health: compute_hosts_health(all_instances)
+      hosts_health: compute_hosts_health(all_instances),
+      hana_cluster_id: extract_hana_cluster_id(database_instances),
+      database_id: extract_database_id(database_instances)
     })
   end
 
+  @spec extract_database_id([DatabaseInstanceReadModel.t()]) :: binary()
```
dottorblaster, line 57: Did you try String.t() for the return type?

dottorblaster (approved):
LGTM

lib/trento/application/usecases/sap_systems/dto/health_summary_dto.ex
```diff
@@ -15,6 +15,8 @@ defmodule Trento.Application.UseCases.SapSystems.HealthSummaryDto do
     field :sapsystem_health, Ecto.Enum, values: Health.values()
     field :database_health, Ecto.Enum, values: Health.values()
     field :clusters_health, Ecto.Enum, values: Health.values()
+    field :hana_cluster_id, :string
```
arbulu89, line 18: They should be uuid

arbulu89: And I think we should maybe name it `cluster_id`. I wouldn't start yet putting hana specific details here. Maybe in the future we will have HANA and NW clusters independently, but yet, our code is using the `cluster` name, and not `hana_cluster`

lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -48,10 +48,26 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
       database_health: compute_database_health(database_instances),
       clusters_health: compute_clusters_health(all_instances),
-      hosts_health: compute_hosts_health(all_instances)
+      hosts_health: compute_hosts_health(all_instances),
+      hana_cluster_id: extract_hana_cluster_id(database_instances),
+      database_id: extract_database_id(database_instances)
     })
   end
 
+  @spec extract_database_id([DatabaseInstanceReadModel.t()]) :: String.t()
```
arbulu89, line 57: I would put the `extract_*` functions after the `compute_*` ones. To have them sorted as the appear in the code

lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -48,10 +48,26 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
+  defp extract_database_id(database_instances) do
+    database_instances
+    |> Enum.map(fn %{sap_system_id: database_id} -> database_id end)
+    |> List.first()
+  end
+
+  @spec extract_hana_cluster_id([DatabaseInstanceReadModel.t()]) :: String.t()
+  defp extract_hana_cluster_id(database_instances) do
```
arbulu89, line 65: Just in case we change the `hana_cluster_id` the function should be renamed

arbulu89: Keep in mind, that we could have a non clustered database, which in this case, we wouldn't get anything.
I don't know if this could break some pattern matching rule. We would be get a nil value most probably.

lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -48,10 +48,26 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
+      hosts_health: compute_hosts_health(all_instances),
+      hana_cluster_id: extract_hana_cluster_id(database_instances),
+      database_id: extract_database_id(database_instances)
     })
   end
 
+  @spec extract_database_id([DatabaseInstanceReadModel.t()]) :: String.t()
+  defp extract_database_id(database_instances) do
+    database_instances
+    |> Enum.map(fn %{sap_system_id: database_id} -> database_id end)
```
arbulu89, line 60: We don' need to iterate all over the databases. Maybe simply:
```
database_instances
|> List.first()
|> Map.get(sap_system_id)
```
```

**Review by arbulu89 — CHANGES_REQUESTED** [review]
Good looks besides the naming. I'm a bit concerned about using the `hana_` prefix here. Maybe, in the future we will need to do so, but right now, I would stick to `cluster_id`.

File: lib/trento/application/usecases/sap_systems/health_summary_service.ex (hunk 5 above)
Comment (fabriziosestito) [reply lib/trento/application/usecases/sap_systems/health_summary_service.ex:60]: pay attention here since if the map is empty `List.first()` returns nil

Comment (fabriziosestito) [reply lib/trento/application/usecases/sap_systems/health_summary_service.ex:60]: Or better

```elixir
  def extract_database_id([]), do: nil
  def extract_database_id([%DatabaseInstance{sap_system_id: sap_system_id} | _]), do: sap_system_id
```

lib/trento/application/usecases/sap_systems/health_summary_service.ex
fabriziosestito: + 1

lib/trento/application/usecases/sap_systems/health_summary_service.ex
CDimonaco: This choice was made because the readmodel is directly consumed by the frontend and we have a different naming on the frontend for these things. Anyway no strong opinions, I could change it :D

lib/trento/application/usecases/sap_systems/health_summary_service.ex
CDimonaco: Done with the match clause as @fabriziosestito suggested, thanks guys

lib/trento/application/usecases/sap_systems/health_summary_service.ex
CDimonaco: Removed the hana_ naming stuff

lib/trento/application/usecases/sap_systems/health_summary_service.ex
CDimonaco: done

lib/trento/application/usecases/sap_systems/dto/health_summary_dto.ex
CDimonaco: done!

CDimonaco:
Ready to rereview @arbulu89

test/trento/application/usecases/health_summary_service_test.exs
```diff
@@ -47,19 +49,43 @@ defmodule Trento.HealthSummaryServiceTest do
     end
 
     test "should determine health summary for a SAP System" do
-      %{
-        sap_system_id: sap_system_id,
+      %Trento.ClusterReadModel{id: cluster_id} =
+        insert(:cluster, type: ClusterType.hana_scale_up(), health: Health.passing())
+
+      %Trento.HostReadModel{id: host_1_id} =
```
arbulu89, line 55: We could rename `host_1_id` to `host_id`

test/trento/application/usecases/health_summary_service_test.exs
```diff
@@ -47,19 +49,43 @@ defmodule Trento.HealthSummaryServiceTest do
...
                  sid: ^sid,
-                 sapsystem_health: :passing,
-                 database_health: :critical,
-                 clusters_health: :warning,
-                 hosts_health: :unknown
+                 sapsystem_health: :critical,
+                 database_health: :warning,
+                 clusters_health: :passing,
+                 hosts_health: :unknown,
+                 database_id: ^sap_system_id,
```
arbulu89, line 87: now I bit confused it this is the database id or the sap system id hehe
Because if it is the same, we are sending it twice, in the `id` and in the `database_id`.

Now i'm thinking both of them might have the same id, the difference in the frontend is that we get one with `/sap_systems/id` and the other with `/databases/id`.
Could you double check this?

arbulu89 (commented):
@CDimonaco the changes look good.
Now I have the doubt if we need to send the `database_id`, because as far as I can see, they sap_system_id and database_id are the same.
Maybe we can remove this `database_id`

test/trento/application/usecases/health_summary_service_test.exs
CDimonaco: As far as I understood, they are not the same 👀

CDimonaco: They can but can also be different

PR #883, opened 2022-10-10
Navigate filtered checkresults

rtorrero:
# Description

This PR allows to navigate from the cluster details overview directly to the check results filtered based on which check result state is clicked. See the attached gif:

filtered-links

## How was this tested?

E2E tests should be passing


assets/js/components/ClusterDetails/ClusterDetails.jsx
```diff
@@ -198,8 +198,8 @@ const ClusterDetails = () => {
         <div className="tn-cluster-checks-overview mt-4 bg-white shadow rounded-lg py-4 xl:w-1/4 w-full">
           <ChecksResultOverview
             {...checkResults}
-            onCheckClick={() =>
-              navigate(`/clusters/${clusterID}/checks/results`)
+            onCheckClick={(health) =>
+              navigate(`/clusters/${clusterID}/checks/results?filter=${health}`)
```
arbulu89, line 202: If we are filtering by health, it shouldn't be this better: `/check/results?health=${health}`

assets/js/components/ClusterDetails/ChecksResultOverview.jsx
```diff
@@ -55,10 +55,18 @@ const ChecksResultOverview = ({
       </h6>
 
       <div className="flex flex-col self-start w-full px-4 mt-2">
-        <CheckResult onClick={onCheckClick} value={passing} result="passing" />
-        <CheckResult onClick={onCheckClick} value={warning} result="warning" />
         <CheckResult
-          onClick={onCheckClick}
+          onClick={() => onCheckClick('passing')}
```
arbulu89, line 59: I see single and double quotes mixed here hehe
Can we use the same? XD

assets/js/components/ClusterDetails/ChecksResultFilters.jsx
```diff
@@ -38,10 +39,23 @@ export const useFilteredChecks = (cluster) => {
 
 const ChecksResultFilters = ({ onChange }) => {
   const [filtersForField, setFiltersForField] = useState({});
-
+  let [searchParams, setSearchParams] = useSearchParams();
```
arbulu89, line 42: Why not `const`?

arbulu89 (changes requested):
@rtorrero Great! Just some small comments.
I think we should add a new test here, whether it is a e2e or unit test with jest. Having the current tests passing doesn't cover this new feature

assets/js/components/ClusterDetails/ChecksResultOverview.jsx
rtorrero: I actually used double quotes but the linter on autosave changed them for single quotes

assets/js/components/ClusterDetails/ClusterDetails.jsx
rtorrero: Fair enough

assets/js/components/ClusterDetails/ChecksResultFilters.jsx
rtorrero: Yes, I knew I was forgetting something!

assets/js/components/ClusterDetails/ChecksResultOverview.jsx
dottorblaster: Yeah single quotes are fine here, don't worry, Prettier is always right :smile:

dottorblaster (approved):
I was about to suggest to turn let `let`s to `const`s as well, super good job

arbulu89 (approved):
Thanks @rtorrero

PR #885, opened 2022-10-10
Add providers KVM & Nutanix to Host Details view

jamie-suse:
# Description

Currently in the Hosts Detail view of a host with KVM and Nutanix as the provider, the user would be presented the message

> Provider not recognized

image

This change adds the providers KVM and Nutanix to the Hosts Detail view as per example below:

image

Note that currently we do not get any provider details for these two providers, but the message "_Provider not recognized_" is not correct.

## How was this tested?

Added React component tests for the `ProviderDetails` component.


dottorblaster (approved):
image

arbulu89 (commented):
@jamie-suse At the end you didn't add any e2e (cypress) test, for what you created the fixtures

PR #886, opened 2022-10-10
Make Pacemaker Clusters status icon interactive

EMaksy:
# Description
Makes the Pacemaker Status icon clickable in the Dashboard and takes the user to the details view of the corresponding database.

## How was this tested?
This was tested manually
A test will be added in a future implementation when the database and the sap instance icons are also clickable.

### DIFF
--- a/assets/js/components/Health/HealthIcon.jsx
+++ b/assets/js/components/Health/HealthIcon.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { computedIconCssClass } from '@lib/icon';
+import { Link } from 'react-router-dom';
 
 import {
   EOS_CHECK_CIRCLE_OUTLINED,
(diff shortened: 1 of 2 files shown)


arbulu89 (commented):
Nice @EMaksy ,
Could you add a gif to the implementation. Always good to have hehe

PD: I don't understand why don't you add the test directly in the PR. I don't find any reason to postpone it. Once you have the test, to extend them to the other links would be simple.

dottorblaster:
> I don't understand why don't you add the test directly in the PR.

@arbulu89 I told him to do so, because I wanted to focus on the implementation first

EMaksy:
> Nice @EMaksy , Could you add a gif to the implementation. Always good to have hehe
Hey, thanks for the suggestion, do you mean to make the icon visually more interactive for the user.
In this case, I already asked @jagabomb his vision for the UI.
Other icons are planed, so we can add some kind of hoover effect to make it more user-friendly, but this may need some refinement.

EMaksy:
About the tests, I will implement interactive status icons for the rest of the dashboard rows and add a test.

dottorblaster (approved):
LGTM

PR #891, opened 2022-10-12
Checks results refactor

dottorblaster:
# Description
The time for a massive rework of the checks results' view layer is coming! In order to ease the pain of the transition from something barrique to something even new-ish, we want the checks result component to be isolated. This is the first step towards a more maintainable `ChecksResults` component.

## How was this tested?

E2E testing should be ok for now, then we're going to write some Jest tests in the coming days


EMaksy (approved):
LGTM!
Looking forward for the upcoming tests

PR #904, opened 2022-10-17
Refactor frontend events mapping

jamie-suse:
# Description

This change refactors the frontend event mapping to be more manageable. To be more specific, it refactors the process of mapping socket events to Redux actions.

## How was this tested?

Existing unit and end-to-end tests should be able to check this change for a regression in behavior.


assets/js/state/index.js
```diff
@@ -45,104 +45,50 @@ const joinChannel = (channel) => {
...
-  );
+const registerEvents = (socket, channelName, events) => {
+  const channel = socket.channel(channelName, {});
 
-  clustersChannel.on('cluster_health_changed', (payload) =>
-    store.dispatch({ type: 'CLUSTER_HEALTH_CHANGED', payload })
-  );
+  for (const event of events) {
+    channel.on(event, (payload) =>
+      store.dispatch({ type: event.toUpperCase(), payload })
```
dottorblaster, line 53: We don't have local access to the `store` variable here. You could pass it as a parameter to the `processChannelEvents` function and to the `registerEvents` function

dottorblaster (changes requested):
Could you try moving these functions to a new file too? Awesome job BTW!

dottorblaster (approved):
LGTM

PR #906, opened 2022-10-17
Publish execution requested

arbulu89:
# Description
Wanda checks adapter. Used with `MIX_ENV=wanda`. The dummy catalog is needed to fullful the behaviour, it will be removed in the future (as it will be queried to wanda).
- I have reused the current checks behaviour, because I forsee a `MockWanda` adapter on the road
- It still uses the current commands and events. That pipeline will be removed in a next PR to reduce the amount of changes in the same PR

## How was this tested?

Tests added


lib/trento/infrastructure/messaging/adapters/amqp.ex
```diff
@@ -9,8 +9,13 @@ defmodule Trento.Messaging.Adapters.AMQP do
 
   require Logger
 
+  # FIXME: fix Trento.Contracts.to_event/2
```
nelsonkopliku, line 12: What needs to be fixed in the contracts?

lib/trento/application/integration/checks/adapter/wanda/wanda.ex
```diff
@@ -0,0 +1,49 @@
...
+  alias Trento.Messaging
+  alias Trento.Messaging.Mapper
+
+  alias Trento.Integration.Checks.{
+    FlatCatalogDto,
+    FlatCheckDto
+  }
+
+  @impl true
+  def request_execution(execution_id, cluster_id, provider, hosts_settings, selected_checks) do
```
nelsonkopliku, line 17: I am wondering whether we should talk about `targets` rather than `hosts_settings` :thinking: 

Yet, I understand this is comes from what we currently have and it might be subject to change as we progress, so not a big deal.

test/trento/application/integration/checks/wanda/wanda_test.exs
```diff
@@ -0,0 +1,35 @@
...
+      selected_checks = ["check_1", "check_2"]
+
+      hosts = [
+        %{host_id: "agent_1"},
+        %{host_id: "agent_2"}
+      ]
+
+      expect(Trento.Messaging.Adapters.Mock, :publish, fn topic, message ->
+        assert "executions" = topic
+        assert %ExecutionRequested{execution_id: ^execution_id} = message
```
nelsonkopliku, line 26: We could add a couple of assertions to check that the correct targets and checks selections are involved.
And while writing I realize that those have been considered in the `MapperTest`.
Ignore this :sweat_smile:

nelsonkopliku (approved):
Nice! I left a couple of minor comments that anyway do not block going forward.

lib/trento/application/integration/checks/adapter/wanda/wanda.ex
arbulu89: Exactly, the signature comes from the behaviour. I will rename it to the new nomenclature

lib/trento/infrastructure/messaging/adapters/amqp.ex
arbulu89: There is an issue with a signature there. It is due abacus, I don't know how important it is, as we will most probably remove it. Anyway, we have tech debt ticket in our backlog to track this issue

fabriziosestito (approved):

PR #908, opened 2022-10-18
Add filtering through search params

rtorrero:
# Description

Peek 18-10-2022 16-02

This PR was initially supposed to just add linking from the dashboard to the hosts view with filters set in the search params, but after starting, it was clear that by allowing the filter component to handle the search params we could enable it in every view.

@CDimonaco actually has another approach on the works, if that works/looks better we can just close this, I just didn't want so many hours of bashing my head against the keyboard to end in nothing :smile: 

## How was this tested?

A new e2e test was added to check that the link from the dashboard includes the filter preset.

### DIFF
--- a/assets/js/components/HealthSummary/HomeHealthSummary.jsx
+++ b/assets/js/components/HealthSummary/HomeHealthSummary.jsx
@@ -56,8 +56,13 @@ const healthSummaryTableConfig = {
       title: 'Hosts',
       key: 'hostsHealth',
       className: 'text-center',
-      render: (content) => {
-        return <HealthIcon health={content} centered={true} />;
+      render: (content, item) => {
+        const linkToHosts = `/hosts?sid=${item.sid}`;
+        return (
+          <Link to={linkToHosts}>
+            <HealthIcon health={content} centered={true} />
+          </Link>
+        );
       },
     },
   ],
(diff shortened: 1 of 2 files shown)


assets/js/components/Table/filters.jsx
```diff
@@ -30,41 +31,62 @@ const setFilter = (filters, filterKey, filterValue, filterFunction) => {
...
+      useEffect(() => {
+        onChange(
+          setFilter(
+            filters,
+            column.key,
+            searchParams.getAll(column.key),
+            filterFunction(filterValue, column)
+          )
+        );
+      }, [JSON.stringify(filterValue)]);
```
rtorrero, line 67: Just wanted to shine a bit of light on this -> https://github.com/facebook/react/issues/14476#issuecomment-471199055

dottorblaster, line 67: Using an `onChange` inside a `useEffect` is an antipattern, are you sure there's no other way to do this? :frowning:

rtorrero: I think this is actually one of the parts where we smell that maybe we should do this in a different place (refactor the flow in a top down fashion, as @CDimonaco suggests). Can't think of any other way of doing this without a major refactor. I'd suggest to wait for the refactor and use this PR to be able to move forward and test the links with filters. Once it's ready, we can switch the code for the better approach.

rtorrero:
Since PR #925 has already been merged and included improvements over the work that I did here, I've removed one of my commits and rebased. Should be good now.

dottorblaster (approved):
LGTM

PR #914, opened 2022-10-19
General browser fixes for collapsed sidebar

jagabomb:
# Description

Minor improvements to remove issues in Safari browser with collapsed sidebar elements. Also changed implementation of Double Arrow Left icon to library standard due to fix with icon library.

Fixes:
Issue with icon not changing direction when the sidebar is collapsed in Safari.
Alignment of Trento logo not being centred when the sidebar is collapsed in Safari.

## How was this tested?

Visually in Safari and Chrome

Screenshot 2022-10-19 at 10 11 58


assets/js/components/Layout/Layout.jsx
```diff
@@ -10,10 +10,10 @@ import {
   EOS_STORAGE,
   EOS_LIST,
   EOS_SETTINGS,
+  EOS_KEYBOARD_DOUBLE_ARROW_LEFT,
```
arbulu89, line 13: Dude, good catcth!

assets/js/components/Layout/Layout.jsx
```diff
@@ -89,23 +89,24 @@ const Layout = () => {
...
                 title={isCollapsed ? 'Expand Sidebar' : 'Collapse Sidebar'}
               >
-                <EosKeyboardDoubleArrowLeft
-                  className="text-gray-400 hover:text-gray-300"
-                  flip={isCollapsed}
+                <EOS_KEYBOARD_DOUBLE_ARROW_LEFT
+                  size="24"
+                  color="currentColor"
+                  className={classNames('text-gray-400 hover:text-gray-300', {
+                    '-scale-100': isCollapsed,
```
arbulu89, line 96: Is this `scale` thing to change arrows direction?
Other option I guess would be to use `EOS_KEYBOARD_DOUBLE_ARROW_RIGHT` depending on the `isCollapsed` value, instead of this direction change. Both work

assets/js/components/Layout/Layout.jsx
```diff
@@ -89,23 +89,24 @@ const Layout = () => {
                 onClick={handleSidebar}
                 title={isCollapsed ? 'Expand Sidebar' : 'Collapse Sidebar'}
               >
-                <EosKeyboardDoubleArrowLeft
-                  className="text-gray-400 hover:text-gray-300"
-                  flip={isCollapsed}
+                <EOS_KEYBOARD_DOUBLE_ARROW_LEFT
+                  size="24"
+                  color="currentColor"
```
arbulu89, line 94: I'm not sure if this `currentColor` is needed, maybe it is

arbulu89 (approved):
Looks good to me!
Maybe you could even add a test to check the arrow direction XD

assets/js/components/Layout/Layout.jsx
dottorblaster: Yeah he's been waiting for days for this :smile:

assets/js/components/Layout/Layout.jsx
jagabomb: I will look at what happens if I remove it. If its is unnecessary then I will remove it. It was only there initially as I saw it in an SVG implementation.

jagabomb: @arbulu89 If I remove `currentColor` then it reverts to black and the classes don't have any effect. I think I will need to retain setting the `color` attribute to enable the classes to work too.

assets/js/components/Layout/Layout.jsx
jagabomb: Both EOS_KEYBOARD_DOUBLE_ARROW_RIGHT and EOS_KEYBOARD_DOUBLE_ARROW_LEFT are now being imported.

assets/js/components/Layout/Layout.jsx
```diff
@@ -89,13 +90,19 @@ const Layout = () => {
...
                 title={isCollapsed ? 'Expand Sidebar' : 'Collapse Sidebar'}
               >
-                <EOS_KEYBOARD_DOUBLE_ARROW_LEFT
-                  size="24"
-                  color="currentColor"
-                  className={classNames('text-gray-400 hover:text-gray-300', {
-                    '-scale-100': isCollapsed,
-                  })}
-                />
+                {isCollapsed ? (
```
jagabomb, line 93: @dottorblaster So I am now using two implementations of the `EOS Keyboard Double Arrow` icons but it @arbulu89 had a suggestion to create a React component for both icons so that we can reduce the repeated attributes for size, color and classname. What do you think?

dottorblaster: @jagabomb a component would be an overkill to me, maybe we can try just defining these things as `const`?

jagabomb: @dottorblaster Cool, I have defined the three const values in a new commit.

PR #918, opened 2022-10-20
Remove events from checks execution

arbulu89:
# Description

Not use commanded commands/events to start checks execution. I have created an adapter to only use the new feature on `wanda` env.

As we don't use events anymore, using commands or using aggregate data/metadata is not possible. If we want to store information such as `issuer` and other things, we will need to send this info to wanda to store it there.

## How was this tested?

Tests added

### DIFF
--- a/lib/trento/application/usecases/clusters/adapters/gen.ex
+++ b/lib/trento/application/usecases/clusters/adapters/gen.ex
@@ -1,0 +1,7 @@
+defmodule Trento.Clusters.Gen do
+  @moduledoc """
+  Request checks execution adapter
+  """
+
+  @callback request_checks_execution(cluster_id :: String.t()) :: :ok | {:error, any}
+end
--- a/lib/trento/application/usecases/clusters/adapters/wanda.ex
+++ b/lib/trento/application/usecases/clusters/adapters/wanda.ex
@@ -1,0 +1,51 @@
+defmodule Trento.Clusters.Wanda do
+  @moduledoc """
+  Wanda integration adapter
+  """
+
+  @behaviour Trento.Clusters.Gen
+
+  import Ecto.Query
+
+  require Logger
+
+  alias Trento.{
+    ClusterReadModel,
+    HostReadModel
+  }
+
+  alias Trento.Integration.Checks
+
+  alias Trento.Repo
+
+  @impl true
+  def request_checks_execution(cluster_id) do
+    hosts_query =
+      from h in HostReadModel,
+        select: %{host_id: h.id},
+        where: h.cluster_id == ^cluster_id
+
+    with %{provider: provider, selected_checks: selected_checks} <-
+           Repo.get(ClusterReadModel, cluster_id),
+         hosts_data <- Repo.all(hosts_query) do
+      Checks.request_execution(
+        UUID.uuid4(),
+        cluster_id,
+        provider,
+        hosts_data,
+        selected_checks
+      )
+    else
+      nil ->
+        Logger.error("Cluster with ID #{cluster_id} not found")
+        {:error, :cluster_not_found}
+
+      {:error, reason} ->
+        Logger.error("Failed to request checks execution for cluster #{cluster_id}: #{reason}",
+          error: reason
+        )
+
+        {:error, reason}
+    end
+  end
+end
--- a/lib/trento/application/usecases/clusters/adapters/wanda.ex
+++ b/lib/trento/application/usecases/clusters/adapters/wanda.ex
@@ -20,32 +20,34 @@ defmodule Trento.Clusters.Wanda do
 
   @impl true
   def request_checks_execution(cluster_id) do
-    hosts_query =
-      from h in HostReadModel,
-        select: %{host_id: h.id},
-        where: h.cluster_id == ^cluster_id
-
-    with %{provider: provider, selected_checks: selected_checks} <-
-           Repo.get(ClusterReadModel, cluster_id),
-         hosts_data <- Repo.all(hosts_query) do
-      Checks.request_execution(
-        UUID.uuid4(),
-        cluster_id,
-        provider,
-        hosts_data,
-        selected_checks
-      )
-    else
-      nil ->
-        Logger.error("Cluster with ID #{cluster_id} not found")
-        {:error, :cluster_not_found}
-
-      {:error, reason} ->
-        Logger.error("Failed to request checks execution for cluster #{cluster_id}: #{reason}",
-          error: reason
-        )
-
-        {:error, reason}
-    end
+    ClusterReadModel
+    |> Repo.get(cluster_id)
+    |> maybe_request_checks_execution()
   end
+
+  defp maybe_request_checks_execution(nil), do: {:error, :cluster_not_found}
+  defp maybe_request_checks_execution(%{selected_checks: []}), do: :ok
+
+  defp maybe_request_checks_execution(%{
+         id: cluster_id,
+         provider: provider,
+         selected_checks: selected_checks
+       }) do
+    hosts_data =
+      Repo.all(
+        from h in HostReadModel,
+          select: %{host_id: h.id},
+          where: h.cluster_id == ^cluster_id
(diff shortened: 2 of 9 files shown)


test/trento/application/usecases/clusters_test.exs
```diff
@@ -30,6 +33,47 @@ defmodule Trento.ClustersTest do
     end
   end
 
+  describe "checks execution with wanda adapter" do
```
nelsonkopliku, line 36: What about moving this tests in its separate module, Wanda related?

At the end what we are testing here is the behavior in wanda scenario
```elixir
assert :ok = Clusters.Wanda.request_checks_execution(cluster_id)
```

test/trento/application/usecases/clusters_test.exs
```diff
@@ -30,6 +33,47 @@ defmodule Trento.ClustersTest do
...
+        assert [%{host_id: host_id_1}, %{host_id: host_id_2}] == expected_hosts
+        assert ^selected_checks = expected_checks
+        :ok
+      end)
+
+      assert :ok = Clusters.Wanda.request_checks_execution(cluster_id)
+    end
+
+    test "should not start checks execution if the cluster is not registered" do
+      assert {:error, :cluster_not_found} = Clusters.Wanda.request_checks_execution(UUID.uuid4())
```
nelsonkopliku, line 57: I believe we could explicitly assert also that the `Trento.Integration.Checks.request_execution` is not called at all, which together with `assert {:error, :cluster_not_found} = Clusters.Wanda...` would reflect what you described in the test `should not start checks execution`

test/trento/application/usecases/clusters_test.exs
```diff
@@ -30,6 +33,47 @@ defmodule Trento.ClustersTest do
...
+    end
+
+    test "should not start checks execution if no checks are selected" do
+      %{id: cluster_id} = insert(:cluster, selected_checks: [])
+
+      assert :ok = Clusters.Wanda.request_checks_execution(cluster_id)
```
nelsonkopliku, line 63: Same as before. To me what would tell that `should not start checks execution` is the fact that we do not call `Trento.Integration.Checks.request_execution`, next to the `:ok = Clusters.Wanda...`

lib/trento/application/usecases/clusters/adapters/wanda.ex
```diff
@@ -0,0 +1,53 @@
...
+
+  defp maybe_request_checks_execution(nil), do: {:error, :cluster_not_found}
+  defp maybe_request_checks_execution(%{selected_checks: []}), do: :ok
+
+  defp maybe_request_checks_execution(%{
+         id: cluster_id,
+         provider: provider,
+         selected_checks: selected_checks
+       }) do
+    hosts_data =
```
nelsonkopliku, line 36: What about moving this second query next to the first one? Or even doing only one query?
So we can separate querying for the needed data for the execution and maybe requesting an execution based on the queried data as we're doing in `maybe_request_checks_execution`

nelsonkopliku: And: what happens/should happen if `hosts_data` ends up in being an empty list? Should it skip requesting an execution to start, similarly to when we get an empty selection?

lib/trento/application/usecases/clusters/adapters/gen.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Trento.Clusters.Gen do
```
nelsonkopliku, line 1: I was thinking that this abstraction name `Trento.Clusters` might be narrowed down to what could be thought as the `ChecksEngine`, or similar names, as its responsibility is to provide an entry point interface to functionalities about executing checks, not really providing abstraction to the whole cluster thing.

It might become something like

```elixir
# config/config.exs
config :trento, Trento.Clusters, checks_engine: Trento.Checks.Engine.Runner

# config/wanda.exs
config :trento, Trento.Clusters, checks_engine: Trento.Checks.Engine.Wanda

# lib/trento/application/usecases/clusters/clusters.ex
defmodule Trento.Clusters do
  ...
  @spec request_checks_execution(String.t()) :: :ok | {:error, any}
  def request_checks_execution(cluster_id) do
    engine().request_checks_execution(cluster_id)
  end
  ...
  defp engine do
    Application.fetch_env!(:trento, __MODULE__)[:checks_engine]
  end
end
```

nelsonkopliku (commented):
Great @arbulu89, thanks!
It looks generally reasonable, main comments revolve around alternative abstraction and encapsulation.
Happy to share ideas :smile:

lib/trento/application/usecases/clusters/adapters/gen.ex
```diff
@@ -0,0 +1,7 @@
...
+  @moduledoc """
+  Request checks execution adapter
```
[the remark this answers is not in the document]
arbulu89: Honestly, I find hard to understand the difference.
I find it as an adapter, because it adapts certain functionalities. In this case for Wanda and old runner.
I would say adapter and behaviour are used for the same thing

test/trento/application/usecases/clusters_test.exs
arbulu89: I just added the wanda word in the text to make differentiation easy. At some point the old runner code will be removed, and we won't have such a differentiation (neither this adapter/behaviour code), so most probably wanda terminology will go away from here, because it will simply be how checks execution is requested. And I would be simple to know that these tests must not be deleted (and we can rename the describe text as well).
What do you think?

test/trento/application/usecases/clusters_test.exs
arbulu89: ok, fair enough!

lib/trento/application/usecases/clusters/adapters/wanda.ex
arbulu89: It shouldn't be possible to have a cluster without any host (unless we have some bug in our code I guess hehe).

About the queries order. I did it this way because it looked the easiest. If we start joining the cluster and host tables, we would already have multiple rows, and this would make all pattern matching more complex and not so obvious. With the current code, I feel that everyone coming to the code can easily understand the logic in a simple glance.

Anyway, if you have some code suggestion that looks good enough we can give a try. I couldn't get antyhing much better than this XD

lib/trento/application/usecases/clusters/adapters/gen.ex
arbulu89: Ok!
This adapter is a temporary thing, but well, let's name it as it is

lib/trento/application/usecases/clusters/adapters/gen.ex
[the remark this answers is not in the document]
nelsonkopliku: > adapter and behaviour are used for the same thing

probably here's where the misunderstanding comes from :sweat_smile:
I thought an `adapter` being a concretization of an abstraction, the `behaviour`, not also the abstraction itself.

Looking at [Ecto](https://github.com/elixir-ecto/ecto/tree/master/lib/ecto/adapter) [repo](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/adapter.ex) clarifies that both abstraction and implementations might be referred to both as adapter.

:rocket:

test/trento/application/usecases/clusters_test.exs
nelsonkopliku: Alright, makes sense, thanks!

lib/trento/application/usecases/clusters/adapters/wanda.ex
nelsonkopliku: I see, I tried myself and yes, it can become really fancy :sweat_smile: 

Fine with this :wink:

nelsonkopliku (approved):
Thanks! Makes sense to me!

fabriziosestito (approved):

rtorrero (approved):
Green light from me too

PR #924, opened 2022-10-21
Checks results refactor part two

dottorblaster:
# Description
To better orchestrate the switch between legacy logic and the new checks engine, we need to refactor the ChecksResults component making the view layer better isolated.

## How was this tested?

Uni tests added, more unit tests will come

### DIFF
--- a/assets/js/components/ChecksResults/ChecksResults.jsx
+++ b/assets/js/components/ChecksResults/ChecksResults.jsx
@@ -19,7 +19,7 @@ import { UNKNOWN_PROVIDER } from '@components/ClusterDetails/ClusterSettings';
 import { getClusterName } from '@components/ClusterLink';
 
 import ChecksResultFilters, { useFilteredChecks } from './ChecksResultFilters';
-import HintingProvider from './HintingProvider';
+import ResultsContainer from './ResultsContainer';
 import CheckResult from './CheckResult';
 import HostResultsWrapper from './HostResultsWrapper';
 import {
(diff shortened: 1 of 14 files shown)


assets/js/components/ChecksResults/HintingProvider.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
arbulu89, line 1: Why is the file named `HintingProvider`?
As far as I see, it doesn't have any relationship with the provider. It looks more some sort of placeholder

assets/js/components/ChecksResults/ChecksResults.jsx
```diff
@@ -251,7 +112,43 @@ const ChecksResults = () => {
...
       )}
-      {pageContent}
+      <HintingProvider
+        catalogError={catalogError}
+        clusterID={clusterID}
+        hasAlreadyChecksResults={hasAlreadyChecksResults}
+        selectedChecks={cluster?.selected_checks}
+        onCatalogRefresh={dispatchUpdateCatalog}
+      >
+        {sortHosts(hosts).map(({ host_id: hostId, reachable, msg }, idx) => (
```
arbulu89, line 122: Maybe `hostID` better

arbulu89 (approved):
It looks great @dottorblaster 
I hope it helps to work in the integration with wanda (which will require much more changes as well).
I wouldn't obsess on adding tests to this current version, as I expect "major" changes during the new integration

assets/js/components/ChecksResults/HintingProvider.jsx
arbulu89: `ResultsContainer`, `ResultsPlaceholder`, `ResultsTable`...
My prefence from left to right.
I find the component as the `content` container of this view.
But I understand it is difficult to find an obvious name

dottorblaster: I'll go with `ResultsContainer`, keep in mind that we have a `provider` notion in our domain and with React you can implement a pattern called Provider, where a top-layer component passes data to underlying ones. Here we adopt that logic not to pass data but to encapsulate some of the rendering logic. Anyway it has nothing to do with the cloud provider thing :joy:

nelsonkopliku (approved):
Looks good!
Was that `.slice()` doing all the magic? :sweat_smile:

assets/js/components/ChecksResults/HintingProvider.jsx
arbulu89: Ok, I totally missunderstood the term hehe. It could be `ResultsProvider` them as well

EMaksy (approved):
Looks great, the code is now much more readable and not stuffed!

rtorrero (approved):
I usually don't like sequels but I might have to make a exception with this... good job :smile:

assets/js/components/ChecksResults/HintingProvider.test.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+  it('should render the suggestion box', () => {
+    renderWithRouter(
+      <HintingProvider catalogError={false} hasAlreadyChecksResults={false} />
+    );
+
+    expect(screen.getByRole('button')).toHaveTextContent('Select Checks now!');
+  });
+
+  it('should render an hello', () => {
```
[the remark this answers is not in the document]
dottorblaster: Damn thank you mate

PR #925, opened 2022-10-21
Frontend table filters from query string

CDimonaco:
# Description

This pr derives from the pr #908, following @rtorrero proposal on query string filters feature.

This pr enables all the tables in trento webapp application to be filtered with query string parameters.
The configuration is declarative and enables these capabilities with  option fields on columns.

### Demo
trento_table_filters

Configuration examples:

```js
{
        title: 'SID',
        key: 'sid',
        filterFromParams: true,
        filter: true
}
```

`filter` field and `filterFromParams` field should be present.

You can enable the filters an all columns and filters type.

For demo/feature purpose the sid and tags filters are enabled with query string on HostOverview screen.

## How was this tested?

E2E testing with cypress, the component is compliant to unit testing with jest, but for pure testing purposes the filtering capabilities are tested with an e2e tests on the HostOverview screen with SID and Tags filters.

### DIFF
--- a/assets/js/components/HostsList.jsx
+++ b/assets/js/components/HostsList.jsx
@@ -1,6 +1,7 @@
 import React, { Fragment } from 'react';
 import Table from './Table';
 import HealthIcon from '@components/Health/HealthIcon';
+import { useSearchParams } from 'react-router-dom';
 import Tags from './Tags';
 import { addTagToHost, removeTagFromHost } from '@state/hosts';
 import HostLink from '@components/HostLink';
@@ -43,6 +44,8 @@ const HostsList = () => {
   const { applicationInstances, databaseInstances } = useSelector(
     (state) => state.sapSystemsList
   );
+
+  const [searchParams, setSearchParams] = useSearchParams();
 
   const dispatch = useDispatch();
 
@@ -171,7 +174,12 @@ const HostsList = () => {
   return (
     <Fragment>
       <ComponentHealthSummary data={data} />
-      <Table config={config} data={data} />
+      <Table
+        config={config}
+        data={data}
+        searchParams={searchParams}
+        setSearchParams={setSearchParams}
+      />
     </Fragment>
   );
 };
--- a/assets/js/components/HostsList.jsx
+++ b/assets/js/components/HostsList.jsx
@@ -90,6 +90,7 @@ const HostsList = () => {
       {
         title: 'SID',
         key: 'sid',
+        filterFromParams: true,
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
         render: (sids, { sap_systems }) => {
@@ -120,6 +121,7 @@ const HostsList = () => {
         title: 'Tags',
         key: 'tags',
         className: 'w-80',
+        filterFromParams: true,
         filter: (filter, key) => (element) =>
           element[key].some((tag) => filter.includes(tag)),
         render: (content, item) => (
--- a/assets/js/components/Table/Table.jsx
+++ b/assets/js/components/Table/Table.jsx
@@ -1,11 +1,13 @@
 import React, { Fragment, useState } from 'react';
 import classNames from 'classnames';
-
+import { useSearchParams } from 'react-router-dom';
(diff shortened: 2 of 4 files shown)


assets/js/components/HostsList.jsx
```diff
@@ -120,6 +124,7 @@ const HostsList = () => {
         title: 'Tags',
         key: 'tags',
         className: 'w-80',
+        filterFromParams: true,
```
rtorrero, line 127: Are we enabling tags to use the searchparams filtering for something in particular? I'd either enable them all or enable only what we immediately need.

rtorrero (changes requested):
Good stuff! As I said, I like the ability of enabling/disabling the search params per column. I made some minor comments, let me know if you agree

assets/js/components/Table/Table.jsx
```diff
@@ -43,6 +44,59 @@ const Table = ({ config, data = [] }) => {
...
+      return [...acc, { key: curr.key, value: curr.value }];
+    }, []);
+
+    filtersBindedToQs.forEach((f) => {
+      currentParams.delete(f.key);
+
+      f.value.forEach((v) => {
+        currentParams.append(f.key, v);
+      });
+    });
```
dottorblaster, line 72: Can we have a function for this? Something like `updateParameters` would be ok ç_ç

assets/js/components/Table/filters.jsx
```diff
@@ -4,11 +4,11 @@ import { uniq } from '@lib/lists';
 
 import Filter from './Filter';
 
-const getDefaultFilterFunction = (filter, key) => (element) => {
+export const getDefaultFilterFunction = (filter, key) => (element) => {
   return filter.includes(element[key]);
 };
 
-const setFilter = (filters, filterKey, filterValue, filterFunction) => {
+export const setFilter = (filters, filterKey, filterValue, filterFunction) => {
```
dottorblaster, line 11: Why didn't you rename this as `createFilter` directly? I would be okay with that.

assets/js/components/Table/Table.jsx
```diff
@@ -43,6 +44,59 @@ const Table = ({ config, data = [] }) => {
...
+
+  const columnFiltersBindToParams = columns.filter(
+    (c) => c.filter && c.filterFromParams
+  );
+
+  useEffect(() => {
+    if (!searchParamsEnabled) return;
+    const currentParams = searchParams;
+    const filtersBindedToQs = filters.reduce((acc, curr) => {
+      const filterBoundToQs = columnFiltersBindToParams.find(
```
dottorblaster, line 57: Lots of stuff named Bound and Bind and James Bound here, plenty of room for renaming

dottorblaster (changes requested):
A couple things to improve but generally speaking, great job!

assets/js/components/HostsList.jsx
CDimonaco: It's enabled for feature purpose, I enabled it because I want to have tests for multiple filters. I think we set to all the table filters in an upcoming pr

assets/js/components/Table/filters.jsx
CDimonaco: done

assets/js/components/Table/Table.jsx
CDimonaco: Agent zero zero bound, mission complete 🤣

assets/js/components/Table/Table.jsx
CDimonaco: done

assets/js/components/Table/Table.jsx
```diff
@@ -43,6 +44,59 @@ const Table = ({ config, data = [] }) => {
   const [filters, setFilters] = useState([]);
   const [currentPage, setCurrentPage] = useState(1);
 
+  const searchParamsEnabled = searchParams && setSearchParams;
```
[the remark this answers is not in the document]
CDimonaco: done! thanks for spotting that

rtorrero (approved):
Looks good now!

dottorblaster (approved):
LGTM

PR #941, opened 2022-11-02
Remove container name from rabbitmq container in docker compose

CDimonaco:
# Description
As the title suggest 

## How was this tested?

Manually


dottorblaster (approved):
LGTM

PR #943, opened 2022-11-02
Enable query string filtering in existing views

CDimonaco:
# Description

This PR enables the query string filter capability on every filtered table column in the application.

This enable the users to share their links with filters in order to have a more consistent UX across the tables.

Feature enabled in

- Hosts List
- Clusters List
- Sap Systems List
- Databases List

The gif will show the filtering feature in every screen.

trento_filters_all_tables

Tag @jagabomb for UX review.

## How was this tested?

The table filtering feature was already test with e2e test in another PR.

### DIFF
--- a/assets/js/lib/test-utils/index.jsx
+++ b/assets/js/lib/test-utils/index.jsx
@@ -10,7 +10,7 @@ import { setSapSystems } from '@state/sapSystems';
 
 import hosts from './data/hosts';
 import clusters from './data/clusters';
-import sapSystems from './data/hosts';
+import sapSystems from './data/sapSystems';
 
 export const withState = (component) => {
   store.dispatch(setHosts(hosts));
--- a/assets/jest.config.js
+++ b/assets/jest.config.js
@@ -86,6 +86,7 @@ module.exports = {
   moduleNameMapper: {
     '^@lib(.*)$': '<rootDir>/js/lib$1',
     '^@components(.*)$': '<rootDir>/js/components$1',
+    '^@hooks(.*)$': '<rootDir>/js/hooks$1',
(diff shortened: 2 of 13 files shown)


abravosuse:

dottorblaster (changes requested):
LGTM but could we also cover some of this with Jest? Consider using `renderWithRouter` and `withState` helpers :+1:

CDimonaco:
> LGTM but could we also cover some of this with Jest? Consider using `renderWithRouter` and `withState` helpers +1

I think we could test the table component with jest because we don't have any dependency to the router, but I will investigate on that stuff

jagabomb (approved):
Tested this manually and it seems to be updating the url correctly across the relevant pages. Links also seem to work when loaded in a new browser window with a pre-authenticated session.

rtorrero (approved):
Green light from my side but if we can get those tests it'd be great :-)

assets/js/lib/test-utils/index.jsx
```diff
@@ -10,7 +10,7 @@ import { setSapSystems } from '@state/sapSystems';
 
 import hosts from './data/hosts';
 import clusters from './data/clusters';
-import sapSystems from './data/hosts';
+import sapSystems from './data/sapSystems';
```
dottorblaster, line 13: Ouch ._. Sorry about that

dottorblaster (approved):
Aw yeeeeeah

CDimonaco:
Add jest tests for every view except the database view, but we already tested the functionality in the tables. I skipped the database view testing for some problems in the fixtures (I need to connect some dots 👁️ ) so maybe we could track in another pr @dottorblaster what do you think?

PR #946, opened 2022-11-03
Replace mock usage for dispatch

arbulu89:
# Description
Remove `mock` usage on commanded dispatched usages. I will send the removal of the DateTime mock in a next PR

## How was this tested?
Using Mox instead of Mock

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -53,6 +53,8 @@ config :logger, :console,
 # Use Jason for JSON parsing in Phoenix
 config :phoenix, :json_library, Jason
 
+config :trento, Trento.Commanded, adapter: Trento.Commanded
+
 config :trento, Trento.Commanded,
   event_store: [
     adapter: Commanded.EventStore.Adapters.EventStore,
--- a/config/test.exs
+++ b/config/test.exs
@@ -70,6 +70,13 @@ config :trento, Trento.Integration.Checks.Wanda.Messaging.AMQP,
     ]
   ]
 
+config :trento, Trento.Scheduler,
+  jobs: [
+    heartbeat_check: [
+      state: :inactive
+    ]
+  ]
+
 config :trento,
   extra_children: [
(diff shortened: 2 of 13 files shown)


config/test.exs
```diff
@@ -70,6 +70,13 @@ config :trento, Trento.Integration.Checks.Wanda.Messaging.AMQP,
     ]
   ]
 
+config :trento, Trento.Scheduler,
```
arbulu89, line 73: This is needed to avoid unexpected heartbeat failure messages, as the scheduler runs the checks randomnly and as the dispatch is mocked now, it raises a Mox error.
This was the easiest solution. Other would be to not start the scheduler at all on tests, but this requires changes on how app children are passed, which at the end makes things more confusing.

config/config.exs
```diff
@@ -53,6 +53,8 @@ config :logger, :console,
 # Use Jason for JSON parsing in Phoenix
 config :phoenix, :json_library, Jason
 
+config :trento, Trento.Commanded, adapter: Trento.Commanded
```
arbulu89, line 56: As always, the hardest is to find an appropriate name XD

config/test.exs
fabriziosestito: that's super ok

fabriziosestito (approved):
We can remove the `mock` dependency and clean the mix.lock

otherwise, LGTM!

arbulu89:
@fabriziosestito 

> We can remove the mock dependency and clean the mix.lock

Not yet, I need to remove some other mocks used with `DateTime` XD

config/config.exs
fabriziosestito: We could bake our own "command dispatcher" in the future, which will follow an extended behaviour (dispatch_many, dispatch_retry) and so on, so as for now the naming is fine I think.

PR #947, opened 2022-11-03
Mock DateTime using a custom date service and mox

arbulu89:
# Description
Remove completely `Mock` usage using a Date service module

## How was this tested?
Mox used instead of Mock


lib/trento/application/usecases/hosts/heartbeats.ex
```diff
@@ -37,17 +39,19 @@ defmodule Trento.Heartbeats do
...
-    Enum.each(get_all_expired_heartbeats(), fn %{agent_id: agent_id} ->
+  @spec dispatch_heartbeat_failed_commands(DateService) :: :ok
+  def dispatch_heartbeat_failed_commands(data_service \\ DateService) do
+    Enum.each(get_all_expired_heartbeats(data_service), fn %{agent_id: agent_id} ->
       dispatch_command(agent_id, :critical)
     end)
   end
 
-  defp get_all_expired_heartbeats do
+  defp get_all_expired_heartbeats(data_service) do
```
dottorblaster, line 49: `s/data_service/date_service/g` :smile:

dottorblaster: Also, is it fine to have `DateService` in the spec? Shouldn't it be `DateService.t()`?

dottorblaster (approved):
LGTM but I'm not a big fan of passing whole modules into function signatures, @fabriziosestito is this ok to you?

lib/trento/application/usecases/hosts/heartbeats.ex
arbulu89: yes, thanks. updating

dottorblaster: Thanks for the explanation!

arbulu89: Amended

test/trento/application/usecases/heartbeats_test.exs
```diff
@@ -24,32 +23,35 @@ defmodule Trento.HeartbeatsTest do
...
-                   heartbeat: :passing
-                 } = command
-
-          :ok
-        end
-      )
+    expect(
+      Trento.Support.DateService.Mock,
+      :utc_now,
+      fn -> now end
```
nelsonkopliku, line 29: I found [this article](https://codeincomplete.com/articles/testing-dates-in-elixir/) that deals with dates and tests.
What do you think?

lib/trento/application/usecases/hosts/heartbeats.ex
```diff
@@ -37,17 +39,19 @@ defmodule Trento.Heartbeats do
...
   end
 
-  defp get_all_expired_heartbeats do
+  defp get_all_expired_heartbeats(date_service) do
     query =
       from h in Heartbeat,
-        where: h.timestamp < ^DateTime.add(DateTime.utc_now(), -@heartbeat_interval, :millisecond)
+        where:
+          h.timestamp <
+            ^DateTime.add(date_service.utc_now(), -@heartbeat_interval, :millisecond)
```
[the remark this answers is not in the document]
arbulu89: You cannot mock that using Mox output. It would exactly be what we had before

test/trento/application/usecases/heartbeats_test.exs
arbulu89: I find it more complex and difficult to follow. You still need to have a mocked functionality, that is hidden within that service with a bunch of helper methods for testing. It looks like it tries to create its own mocking system. And you cannot use tools like Mox

The best thing about the current solution is that it is straightforward and doesn't require and configuration. And the tests are super simple to understand.

Some refs: https://blog.carbonfive.com/lightweight-dependency-injection-in-elixir-without-the-tears/

CDimonaco (approved):
LGTM

Missing my RoR times using timecop 🥲

lib/trento/application/usecases/hosts/heartbeats.ex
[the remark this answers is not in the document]
nelsonkopliku: Yep I see that mocking it is the driver to inject it. Completely fine with it.

test/trento/application/usecases/heartbeats_test.exs
nelsonkopliku: Fair, I am fine with this approach.

PR #949, opened 2022-11-03
Use uuid format for id fields

arbulu89:
# Description
Use uuid format for cluster, host, sap system ids in the domain events/commands/aggregates

## How was this tested?

Tests were alright


CDimonaco (approved):
LGTM

fabriziosestito (approved):

PR #958, opened 2022-11-04
Add the possibility to add environment configuration to the bundle

dottorblaster:
# Description

We need to be able to configure variables depending upon `NODE_ENV`, or to put it clearly, to switch between development and production values. In `esbuild` this can be done through the `define` field. Inside the app the value can be accessed through `process.env.WANDA_URL`.

## How was this tested?
Manually!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,5 +33,6 @@ trento-*.tar
 # In case you use Node.js/npm, you want to ignore these.
 npm-debug.log
 /assets/node_modules/
+/assets/.env
 
 .pgdata/
--- a/assets/build.js
+++ b/assets/build.js
@@ -5,8 +5,18 @@ const alias = require('esbuild-plugin-path-alias');
 
 const resolvePath = (p) => path.resolve(__dirname, p);
 
+const WANDA_URL =
+  process.env.NODE_ENV === 'production'
+    ? ''
+    : JSON.stringify('http://localhost:4001');
+
+const define = {
+  'process.env.WANDA_URL': WANDA_URL,
+};
+
 require('esbuild')
   .build({
+    define,
     entryPoints: ['js/app.js', 'js/trento.jsx'],
     outdir: resolvePath('../priv/static/assets'),
     bundle: true,
--- a/assets/build.js
+++ b/assets/build.js
@@ -2,13 +2,16 @@
 /* eslint-disable no-console */
 const path = require('path');
 const alias = require('esbuild-plugin-path-alias');
+const { config } = require('dotenv');
+
+config();
 
 const resolvePath = (p) => path.resolve(__dirname, p);
 
 const WANDA_URL =
   process.env.NODE_ENV === 'production'
     ? ''
-    : JSON.stringify('http://localhost:4001');
+    : JSON.stringify(process.env.WANDA_URL);
 
 const define = {
   'process.env.WANDA_URL': WANDA_URL,
--- a/assets/.env.example
+++ b/assets/.env.example
@@ -1,0 +1,1 @@
+WANDA_URL=http://localhost:4001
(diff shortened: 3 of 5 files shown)


arbulu89 (commented):
Hey @dottorblaster ,
In order to make it a bit more flexible, what if we include `dotenv`. Something like:
```
require('dotenv').config();

const WANDA_URL =
  process.env.NODE_ENV === 'production'
    ? ''
    : JSON.stringify(process.env.WANDA_URL);

const define = {
  'process.env.WANDA_URL': WANDA_URL,
};
```

This way, we can have an optional `.env` file with somethin like:
```
WANDA_URL=http://localhost:4001
```

We could even load the whole process.env, but we would lack some control about specific values that we need to have yes or yes, even with empty values.

CDimonaco:
> Hey @dottorblaster , In order to make it a bit more flexible, what if we include `dotenv`. Something like:
> 
> ```
> require('dotenv').config();
> 
> const WANDA_URL =
>   process.env.NODE_ENV === 'production'
>     ? ''
>     : JSON.stringify(process.env.WANDA_URL);
> 
> const define = {
>   'process.env.WANDA_URL': WANDA_URL,
> };
> ```
> 
> This way, we can have an optional `.env` file with somethin like:
> 
> ```
> WANDA_URL=http://localhost:4001
> ```
> 
> We could even load the whole process.env, but we would lack some control about specific values that we need to have yes or yes, even with empty values.

+1, if we could add dotenv easily I think it could be an improvement. Maybe we could add a specific env file and instruct dotenv to load only that specific env file

dottorblaster:
@arbulu89 @CDimonaco you are both right, I'll work on that during the day :+1:

assets/build.js
```diff
@@ -2,13 +2,16 @@
...
+
+config();
 
 const resolvePath = (p) => path.resolve(__dirname, p);
 
 const WANDA_URL =
   process.env.NODE_ENV === 'production'
     ? ''
-    : JSON.stringify('http://localhost:4001');
+    : JSON.stringify(process.env.WANDA_URL);
```
CDimonaco, line 14: the stringify is a nice sanitation touch ;D

dottorblaster: It didn't work otherwise LOL

assets/.env
```diff
@@ -0,0 +1 @@
+WANDA_URL=http://localhost:4001
```
arbulu89, line 1: was this file uploaded in purpose? 
I would actually put the `.env` file in the gitignore. Having the example file is good enough, so everyone can configure their own environment

.gitignore
```diff
@@ -33,5 +33,6 @@ trento-*.tar
 # In case you use Node.js/npm, you want to ignore these.
 npm-debug.log
 /assets/node_modules/
+/assets/.gitignore
```
arbulu89, line 36: I don't get this. Do we need other `.gitignore` file? We already have one in the root folder.
Maybe you meant there `/assets/.env` ?

arbulu89 (approved):
Hey @dottorblaster ,
The code looks good. I'm a bit confused with the `gitignore` thing and the usage of the `.env` file already having an example file/

PD: once you try to use this, eslint complains that the `process` variable does not exist hehe

.gitignore
dottorblaster: What I was thinking while committing this?

assets/build.js
```diff
@@ -1,12 +1,26 @@
 /* eslint-disable no-undef */
 /* eslint-disable no-console */
+/* eslint-disable no-undef */
```
arbulu89, line 3: `/* eslint-disable no-undef */`  line now is duplicated XD

dottorblaster: You won't believe this. I had vim open and first line was just truncated LOL

arbulu89: You could always use emacs.... XD

dottorblaster: DUDE

PR #960, opened 2022-11-07
Catalog from wanda

arbulu89:
# Description
Load catalog data from wanda. The same functionality as before, except there is no provider selection. The new view is available through `http://localhost:4000/catalog_new`

image

PD: I have included a pair of dummy checks to show the groups in the result

## How was this tested?
Tests on jest added


assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -0,0 +1,40 @@
...
+      .get(`${wandaURL}/api/checks/catalog`)
+      .then((catalog) => {
+        setLoaded(false);
+        setCatalog(catalog.data);
+      })
+      .catch(function (error) {
+        setLoaded(false);
+        setError(error.message);
+      });
+  };
```
dottorblaster, line 30: You could just use an arrow function here `(error) => {setError(error.message)}`

assets/js/components/ChecksCatalog/CatalogContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+                          )}
+                        </div>
+                        <div className="mt-2 sm:flex sm:justify-between">
+                          <div className="sm:flex">
+                            <ReactMarkdown
+                              className="markdown text-sm"
+                              remarkPlugins={[remarkGfm]}
+                            >
+                              {check.description}
+                            </ReactMarkdown>
```
dottorblaster, line 75: Do we wanna extract all this markup in some more granular components?

dottorblaster (changes requested):
Nice! Some tweaks and I think we can merge this :+1:

assets/js/components/ChecksCatalog/CatalogContainer.jsx
arbulu89: Do you mean to create our own markdown component with those defaults? 
In this case `text-sm` would be an additional style.

arbulu89: Can we do the next? Let me merge this as it is about this topic, and I will create other PR with a new markdown component and use it all around the code. What do you think?

dottorblaster: Not the markdown, I mean the whole HTML body of this component :sweat_smile: 

In a similar way to https://github.com/trento-project/web/blob/main/assets/js/components/ChecksResults/ChecksResults.jsx, we can make this code more readable just having a wrapper component for the `div`s and the `ul` and another component for the `li`

dottorblaster: Anyway if you wanna do that in a subsequent PR, feel free.

arbulu89: @dottorblaster CheckItem extracted to an individual component. Going further looks like an overkill to me

assets/js/components/ChecksCatalog/CatalogContainer.jsx
```diff
@@ -0,0 +1,75 @@
...
+          >
+            <div className="bg-white px-4 py-5 border-b border-gray-200 sm:px-6">
+              <h3 className="text-lg leading-6 font-medium text-gray-900">
+                {group}
+              </h3>
+            </div>
+            <ul role="list" className="divide-y divide-gray-200">
+              {checks.map((check) => (
+                <CheckItem
+                  key={check.id}
```
arbulu89, line 60: This `key` variable is needed to avoid `key prop not defined issue`. Besides that, just as key, it doesn't render properly the text, so I need to send it in `checkID` as well

dottorblaster (changes requested):
Sorry, really. Last thing then we can merge :+1:

CDimonaco:
Looks good to me, I'm with @dottorblaster on all the other comments, I think that we could add some data-testid attributes/props on component to better test them even in the future, like we already done other jest tests.

For reference: https://testing-library.com/docs/queries/bytestid/ 

It's a nice to have btw, @arbulu89 good job!

arbulu89:
> Looks good to me, I'm with @dottorblaster on all the other comments, I think that we could add some data-testid attributes/props on component to better test them even in the future, like we already done other jest tests.
> 
> For reference: https://testing-library.com/docs/queries/bytestid/
> 
> It's a nice to have btw, @arbulu89 good job!

Hey @CDimonaco ,
I see the point of using the `test-ids`. In this case, finding the elements was fairly simple and the tests look pretty easy to understand, that's why I didn't include them. But I agree that once you have difficult dom findings etc they are super useful

CDimonaco:
> > Looks good to me, I'm with @dottorblaster on all the other comments, I think that we could add some data-testid attributes/props on component to better test them even in the future, like we already done other jest tests.
> > For reference: https://testing-library.com/docs/queries/bytestid/
> > It's a nice to have btw, @arbulu89 good job!
> 
> Hey @CDimonaco , I see the point of using the `test-ids`. In this case, finding the elements was fairly simple and the tests look pretty easy to understand, that's why I didn't include them. But I agree that once you have difficult dom findings etc they are super useful

Perfect! LGTM

CDimonaco (approved):
LGTM

dottorblaster (approved):
LGTM

fabriziosestito:
> > > Looks good to me, I'm with @dottorblaster on all the other comments, I think that we could add some data-testid attributes/props on component to better test them even in the future, like we already done other jest tests.
> > > For reference: https://testing-library.com/docs/queries/bytestid/
> > > It's a nice to have btw, @arbulu89 good job!
> > 
> > 
> > Hey @CDimonaco , I see the point of using the `test-ids`. In this case, finding the elements was fairly simple and the tests look pretty easy to understand, that's why I didn't include them. But I agree that once you have difficult dom findings etc they are super useful
> 
> Perfect! LGTM

FYI
In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible.

PR #963, opened 2022-11-08
Phoenix views refactoring

CDimonaco:
# Description

Refactor all the controllers to use the Phoenix view mechanism instead of Dtos.

Some phoenix broadcast messages are refactored to use phoenix views, in particular only the ones that uses StructHelper.to_map. The `to_map` function is now deprecated, and used only in some tests.

Phoenix views cover the needs of the `to_map` function providing a better and more lean api and execution.

## How was this tested?

Automated test, old tests refactored, the controller tests now uses OpenApi spec, some specs fixed.

The broadcast mechanism was tested manually using photofinish, no tests were present and no test is added by this pr. We need to cover that in another moment with a more structural approach to the problem.


test/trento_web/controllers/installation_controller_test.exs
```diff
@@ -0,0 +1,15 @@
...
+  use TrentoWeb.ConnCase, async: true
+
+  import OpenApiSpex.TestAssertions
+
+  alias TrentoWeb.OpenApi.ApiSpec
+
+  test "should return the api key", %{conn: conn} do
+    api_spec = ApiSpec.spec()
+
+    get(conn, Routes.installation_path(conn, :get_api_key))
```
arbulu89, line 11: We agreed here that we are going to use "absolute" paths for testing. So this would be `api/installation` or whatever, without using the `Routes` option

lib/trento_web/views/cluster_view.ex
```diff
@@ -0,0 +1,45 @@
...
+
+  def render("cluster.json", %{cluster: cluster}), do: cluster
+
+  def render("cluster_registered.json", %{cluster: cluster}) do
+    cluster
+    |> Map.from_struct()
+    |> Map.delete(:__meta__)
+    |> Map.delete(:tags)
+  end
+
```
arbulu89, line 16: Shouldn't we have a `cluster_details_updated` view? We have it for the hosts
But I understand that it is not that simple in this case

lib/trento_web/views/cluster_view.ex
```diff
@@ -0,0 +1,45 @@
...
+  def render("clusters.json", %{clusters: clusters}) do
+    render_many(clusters, __MODULE__, "cluster.json")
+  end
+
+  def render("cluster.json", %{cluster: cluster}), do: cluster
+
+  def render("cluster_registered.json", %{cluster: cluster}) do
+    cluster
+    |> Map.from_struct()
```
arbulu89, line 12: As in hosts, maybe having the `Map.from_struct` in the `cluster.json` view would look more logical. With the removal of the common things (meta, tags).
This way, we wouldn't be sending the `__meta__` tag in the `clusters.json` multi rendered element

test/trento_web/controllers/cluster_controller_test.exs
```diff
@@ -1,35 +1,24 @@
...
     test "should list all clusters", %{conn: conn} do
-      [
-        %{id: cluster_id_1, name: cluster_name_1},
-        %{id: cluster_id_2, name: cluster_name_2},
-        %{id: cluster_id_3, name: cluster_name_3}
-      ] =
-        0..2
-        |> Enum.map(fn _ -> insert(:cluster) end)
-        |> Enum.sort_by(& &1.name)
+      0..2
```
arbulu89, line 12: This looks really strange. Why not `insert_list`?

lib/trento_web/openapi/schema/cluster.ex
```diff
@@ -138,14 +139,19 @@ defmodule TrentoWeb.OpenApi.Schema.Cluster do
           items: %Schema{type: :string}
         },
         health: ResourceHealth,
-        resources_number: %Schema{type: :integer, description: "Resource number"},
-        hosts_number: %Schema{type: :integer, description: "Hosts number"},
-        cib_last_written: %Schema{type: :string, description: "CIB last written date"},
+        resources_number: %Schema{type: :integer, description: "Resource number", nullable: true},
```
arbulu89, line 142: why are all of these things nullable now?

lib/trento_web/views/sap_system_view.ex
```diff
@@ -0,0 +1,157 @@
...
+    %{
+      sap_system_id: sap_system_id,
+      host_id: host_id,
+      instance_number: instance_number,
+      system_replication: system_replication,
+      system_replication_status: system_replication_status
+    }
+  end
+
+  defp add_system_replication_status_to_secondary_instance(
```
arbulu89, line 126: I'm not sure we should move this here. This looks an important domain logic component, not simply a view thing. Maybe we want to have this information somewhere else.

lib/trento_web/views/sap_system_view.ex
```diff
@@ -0,0 +1,157 @@
+defmodule TrentoWeb.SapSystemView do
+  use TrentoWeb, :view
+
+  def render("databases.json", %{databases: databases}) do
```
arbulu89, line 4: I find the ordering hard to follow here. Maybe we could have `database` first and sap system after order. What do you think? Or does it break some other ordering rule?

arbulu89 (commented):
Hey @CDimonaco ,
I finished checking the whole PR hehe
It looks totally correct in most of the places. I have commented a ton of things, but most of them are minor things (some are more relevant). Have a look on the comments.
After that, we can maybe talk about them as they are many things

Some general comments/questions here.
- The openapi spec `nullable` changes. Did you do them after finding them on the tests somehow? I didn't check all if they can be nullable or not
- Let's avoid using the `Routes` module in the controller testing. We agreed in to use real paths there, to have a more thorough testing
- In the controller tests, I only see that we check that the messages follow the openapi schema, but I don't see tests for the views for example for the hosts/about/prometheus/settings, etc. Actually, I see very few tests on the views. Should we have some view tesitng there?

CDimonaco:
@arbulu89 I'm reviewing all the comments, thanks for the review

> The openapi spec nullable changes. Did you do them after finding them on the tests somehow? I didn't check all if they can be nullable or not

After reviewing some controller tests, I discovered that we had partial assertions in tests, and we were not using the openapi schema in controller testing, after moving to openapi schema testing I discovered this issues in the openapi schema, after a conversation with @fabriziosestito we agreed that the fields should be nullable, for the eventual consistent nature of the application, in which at some point is allowed to have this fields nullable before the computation

> In the controller tests, I only see that we check that the messages follow the openapi schema, but I don't see tests for the views for example for the hosts/about/prometheus/settings, etc. Actually, I see very few tests on the views. Should we have some view tesitng there?

We have some view testing, when the view are not trivial, the majority of the views are tested with the output of controller tests, they are very trivial and results in pure manipulation of the schemas, deleting the association not loaded and other not needed metadatatas, so checking the output of the view through the controller seems a good tradeoff between test redundancy and actual coverage.
The views used by broadcast channels are different, and you are right I need to test them in a more unit fashion way, because we don't have any broadcasting channel testing right know.

I'm reviewing the other comments as well, I will update the pr with more comments and in dedicated review threads.

lib/trento_web/views/host_view.ex
```diff
@@ -0,0 +1,28 @@
+defmodule TrentoWeb.HostView do
+  use TrentoWeb, :view
+
+  def render("hosts.json", %{hosts: hosts}) do
+    render_many(hosts, __MODULE__, "host.json")
+  end
+
+  def render("host.json", %{host: host}), do: host
```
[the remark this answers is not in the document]
CDimonaco: Absolutely, you are right, this was made before the other views, before I started converting to a map directly, that view works because there are all loaded association where it's used, so the jason encoder of the schema works without hassles, but I will make that explicit converting to a map and inherit that. This works for a side effect. Thanks for the catch!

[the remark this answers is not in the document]
CDimonaco: done

test/trento_web/controllers/cluster_controller_test.exs
CDimonaco: Yes it is, some tests are like that, I already created a tech debt task to tackle it as the next thing in another prs

lib/trento_web/openapi/schema/cluster.ex
CDimonaco: asnwered in the main comment

lib/trento_web/views/host_view.ex
```diff
@@ -0,0 +1,28 @@
...
+
+  def render("host_details_updated.json", %{host: host}) do
+    host
+    |> Map.from_struct()
+    |> Map.delete(:__meta__)
+    |> Map.delete(:sles_subscriptions)
```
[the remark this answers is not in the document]
CDimonaco: This is used by a project in a message broadcast, I investigated about the actual payload we send and it's like that, we don't have right know any testing on that behavior so I had to investigate manually, Before this exact things was done by the to_map function of StructHelper, but we are getting rid of it so I had to remove that things manually

lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -17,41 +17,39 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
       id: id,
       sid: sid,
       sapsystem_health: health,
       database_health: compute_database_health(database_instances),
       clusters_health: compute_clusters_health(all_instances),
       hosts_health: compute_hosts_health(all_instances),
-      cluster_id: extract_cluster_id(database_instances),
-      database_id: extract_database_id(database_instances)
-    })
+      database_instances: database_instances
```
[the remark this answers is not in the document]
CDimonaco: It's strange right know, but the view should not have any reference to any repository or something external, so the trick was that, it was used also in dto approach, speaking freely, some days ago I discussed that with @fabriziosestito and we agreed that this service should have some refactor, so we can deal with it when we refactor the whole service

test/trento_web/controllers/installation_controller_test.exs
CDimonaco: answered in the main comment

lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -17,41 +17,39 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
 
-  @spec sap_system_to_summary(SapSystemReadModel.t()) :: HealthSummaryDto.t()
-  defp sap_system_to_summary(%SapSystemReadModel{
-         id: id,
-         sid: sid,
-         health: health,
-         application_instances: application_instances,
-         database_instances: database_instances
-       }) do
+  @spec summary_from_sap_system(SapSystemReadModel.t()) :: map()
```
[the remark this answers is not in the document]
CDimonaco: done

test/trento/application/usecases/hosts_test.exs
```diff
@@ -73,13 +73,11 @@ defmodule Trento.HostsTest do
                %{
                  host_id: ^a_host_id,
                  hostname: ^a_hostname,
-                 default_user: "root",
```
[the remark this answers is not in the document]
CDimonaco: It's tested in the view, it's computed from other values, the `cluster_view` contains that logic

lib/trento_web/views/sap_system_view.ex
CDimonaco: I don't know, in the old code this was used only when the function in which was called was used by the controller, so I assumed it was only related to the view section, we can certainly move that if we think it's valuable

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -82,9 +82,7 @@ defmodule Trento.Hosts do
         order_by: [asc: h.hostname]
       )
 
-    query
-    |> Repo.all()
-    |> Enum.map(&enrich_with_default_user/1)
```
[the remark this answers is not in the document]
CDimonaco: It was a view concern for the settings, that behavior is now embeeded into settings render, in cluster view

lib/trento_web/views/cluster_view.ex
CDimonaco: done

arbulu89:
@CDimonaco Being the PR that big, I'm fine on creating tech-debt tasks on working on them later on, so we don't overload anymore the current code.

lib/trento_web/views/sap_system_view.ex
arbulu89: Ok, I have checked the code and now understand what's going on.
The thing is that we do something strange in the read models projection (and in our domain logic in general). Having an `ACTIVE` or not sync status usually is attached to the secondary node, which is the node that is doing the sync (so this is the one that is active).
In the other hand, due how we handle the data (and how is sent by some HANA native outputs), we attach this value to the primary, and it is stored this way in the read model.
That's why we do this transition in the view to put the state in the secondary.

So long story short, how it is done, yes, this code belongs to the view, because we simply use it to display correctly things.
But I find that we could do this better, and store the `status` correctly in the projections, to avoid having this strange behaviour.
It is in any case a low hanging fruit. I will create tech-debt ticket myself.
So, you can let this code as you changed XD

arbulu89 (approved):
@CDimonaco 
Great job! Thank you for looking at to all my comments hehe
Feel free to work on the remaining tech-debts now or create tech-debt items in our backlog

lib/trento_web/views/sap_system_view.ex
CDimonaco: done!

PR #970, opened 2022-11-09
Use mocked redux store

arbulu89:
# Description
Use the [https://github.com/reduxjs/redux-mock-store](https://github.com/reduxjs/redux-mock-store) package to mock the tests redux store. Until now, we were using the real state, which doesn't help much on the testing phase.

## How was this tested?

New test utility added


CDimonaco (approved):
LGTM for me, good starting point we can adjust the usability in the future!

EMaksy (commented):
LGTM :D

dottorblaster (approved):
LGTM, I don't like particularly the idea of injecting a state knowing its shape but maybe I'm just biased

PR #975, opened 2022-11-10
Fix tagging test in hosts overview e2e suite

arbulu89:
# Description
Fix hosts overview tagging test in the e2e  hosts overview suite.
As far as I could debug, the test was not giving enough time to type the last tag and store the data in the backend.

`cy.intercept('POST', '/api/hosts/*/tags');` fixes this (hopefully)

PD: tagging tests continue being inconsistent on the different views where filters are used, but "fixing" those would take much more time


dottorblaster (approved):
LGTM, maybe we could review this in mob for 5 minutes to see if somebody spots something that I missed, but it really looks good

CDimonaco (approved):
lgtm

PR #976, opened 2022-11-10
Checks selection from wanda

arbulu89:
# Description
Replace current checks selection view to use data coming from wanda.
Almost the complete work consists on having the checks selection view components in smaller pieces, beforehand everything was a unique view.
Besides that, I have introduced in the last commit the concept of  `actions` in a new folder. There, we can have the actions and their constants defined and reuse the dispatched actions. I think we should do this for all the logical components at some point, at the moment where we do the folder/file restructure

## How was this tested?
Tests added

### DIFF
--- a/assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
+++ b/assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
@@ -4,6 +4,7 @@ import { useSelector, useDispatch } from 'react-redux';
 import { groupBy } from '@lib/lists';
 
 import { getCatalog } from '@state/selectors/catalog';
+import { dispatchUpdateCatalog } from '@state/actions/catalog';
 import CatalogContainer from './CatalogContainer';
 import CheckItem from './CheckItem';
 
@@ -17,19 +18,12 @@ export const ChecksCatalogNew = () => {
   } = useSelector(getCatalog());
 
   useEffect(() => {
-    dispatchUpdateCatalog();
+    dispatchUpdateCatalog()(dispatch);
   }, [dispatch]);
-
-  const dispatchUpdateCatalog = () => {
-    dispatch({
-      type: 'UPDATE_CATALOG_NEW',
-      payload: {},
-    });
-  };
 
   return (
     <CatalogContainer
-      onRefresh={() => dispatchUpdateCatalog()}
+      onRefresh={() => dispatchUpdateCatalog()(dispatch)}
       isCatalogEmpty={catalogData.length === 0}
       catalogError={catalogError}
       loading={loading}
--- a/assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
+++ b/assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
@@ -4,7 +4,7 @@ import { useSelector, useDispatch } from 'react-redux';
 import { groupBy } from '@lib/lists';
 
 import { getCatalog } from '@state/selectors/catalog';
-import { dispatchUpdateCatalog } from '@state/actions/catalog';
+import { updateCatalog } from '@state/actions/catalog';
 import CatalogContainer from './CatalogContainer';
 import CheckItem from './CheckItem';
 
@@ -18,12 +18,12 @@ export const ChecksCatalogNew = () => {
   } = useSelector(getCatalog());
 
   useEffect(() => {
-    dispatchUpdateCatalog()(dispatch);
+    dispatch(updateCatalog());
   }, [dispatch]);
 
   return (
     <CatalogContainer
-      onRefresh={() => dispatchUpdateCatalog()(dispatch)}
+      onRefresh={() => dispatch(updateCatalog())}
       isCatalogEmpty={catalogData.length === 0}
       catalogError={catalogError}
       loading={loading}
--- a/assets/js/components/ClusterDetails/ChecksSelectionGroup.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelectionGroup.jsx
@@ -1,0 +1,91 @@
+import React from 'react';
+
+import classNames from 'classnames';
+
+import { Disclosure, Switch, Transition } from '@headlessui/react';
+
+import { EOS_KEYBOARD_ARROW_RIGHT } from 'eos-icons-react';
+
+const ChecksSelectionGroup = ({
+  children,
+  group,
+  allSelected,
+  someSelected,
+  noneSelected,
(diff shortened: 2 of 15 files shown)


assets/js/components/ClusterDetails/ClusterSettings.jsx
```diff
@@ -17,20 +18,26 @@ import { ConnectionSettings, ClusterInfoBox } from '@components/ClusterDetails';
...
+      <ChecksSelectionNew clusterId={clusterID} cluster={cluster} />
+    ) : (
       <ChecksSelection clusterId={clusterID} cluster={cluster} />
     ),
     'Connection Settings': (
       <ConnectionSettings clusterId={clusterID} cluster={cluster} />
     ),
   };
 
+  if (!cluster) {
```
nelsonkopliku, line 37: Not a biggie: would it make sense to move this guard right after the selection from the state?

assets/js/components/ClusterDetails/ClusterSettings.jsx
```diff
@@ -17,20 +18,26 @@ import { ConnectionSettings, ClusterInfoBox } from '@components/ClusterDetails';
...
-export const ClusterSettings = () => {
+export const ClusterSettings = ({ newChecksSelectionView = false }) => {
   const { clusterID } = useParams();
 
   const cluster = useSelector(getCluster(clusterID));
 
   const tabsSettings = {
-    'Checks Selection': (
+    'Checks Selection': newChecksSelectionView ? (
+      <ChecksSelectionNew clusterId={clusterID} cluster={cluster} />
```
nelsonkopliku, line 28: I guess in wanda context the `ChecksSelectionNew` component would not need to be tabbed with the `ConnectionSettings`. I believe this is going to be iterated on.

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -17,19 +18,12 @@ export const ChecksCatalogNew = () => {
   } = useSelector(getCatalog());
 
   useEffect(() => {
-    dispatchUpdateCatalog();
+    dispatchUpdateCatalog()(dispatch);
```
nelsonkopliku, line 21: As mentioned elsewhere could this simply be `dispatchUpdateCatalog(dispatch);`?

And also other similar usages.

nelsonkopliku (commented):
Good job!
I could make it work locally and looks going in the right direction. 

I left a couple of comments about some doubts and in relation to the connection settings, which we'd need to get rid of when using wanda.

assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
```diff
@@ -0,0 +1,169 @@
...
+
+  useEffect(() => {
+    if (loading === true) {
+      setLocalSavingError(null);
+      setLocalSavingSuccess(null);
+    }
+  }, [loading]);
+
+  const isSelected = (checkId) =>
+    selectedChecks ? selectedChecks.includes(checkId) : false;
```
dottorblaster, line 84: You can move this outside the component and make it a function

dottorblaster (changes requested):
Some things to adjust, then we are good

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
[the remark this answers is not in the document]
arbulu89: Man, totally right! 
That way looks obvious now XD

assets/js/state/actions/cluster.js
```diff
@@ -0,0 +1,9 @@
+export const checksSelectedAction = 'CHECKS_SELECTED';
+
+export const dispatchChecksSelected =
+  (selectedChecks, clusterID) => (dispatch) => {
```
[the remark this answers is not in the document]
arbulu89: Solved as explained here: https://github.com/trento-project/web/pull/976#discussion_r1034444440

assets/js/components/ClusterDetails/ClusterSettings.jsx
arbulu89: Yes, we will see what we do about this

assets/js/components/ClusterDetails/ClusterSettings.jsx
arbulu89: I prefer to leave it as it is to have consistency on order. `const` first and returned code afterwards.

assets/js/components/ClusterDetails/ChecksSelectionGroup.jsx
```diff
@@ -0,0 +1,91 @@
...
+import { Disclosure, Switch, Transition } from '@headlessui/react';
+
+import { EOS_KEYBOARD_ARROW_RIGHT } from 'eos-icons-react';
+
+const ChecksSelectionGroup = ({
+  children,
+  group,
+  allSelected,
+  someSelected,
+  noneSelected,
```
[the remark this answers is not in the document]
arbulu89: I have ended adding the states into an "enum" (js doesn't have proper enums, so a dictionary does the work, the values of the map don't have any meaning)

assets/js/components/ClusterDetails/ChecksSelectionGroup.jsx
```diff
@@ -0,0 +1,101 @@
...
+
+import { Disclosure, Switch, Transition } from '@headlessui/react';
+
+import { EOS_KEYBOARD_ARROW_RIGHT } from 'eos-icons-react';
+
+export const groupState = {
+  None: 0,
+  Some: 1,
+  All: 2,
+};
```
dottorblaster, line 13: You could also do

```js
export const SOME = 'SOME';

// and then
import { ALL, SOME, NONE } from './ChecksSelectionGroup';
```

Then you could also use the constants this way:

```js
const switchClasses = {
  [NONE]: 'bg-gray-200',
  [SOME]: 'bg-green-300',
  [ALL]: 'bg-jungle-green-500',
};
```

dottorblaster (approved):
LGTM, a tiny comment but really your call

nelsonkopliku (approved):
Good to go.

assets/js/components/ClusterDetails/ChecksSelectionGroup.jsx
arbulu89: Yep, gone this way.
Basically, because it is much better to find compile time errors, as the "enum" dict can get anything after the dot, creating some potential issues.

PR #982, opened 2022-11-14
Setup structured docs

arbulu89:
Apply a more advanced documentation settings that are published in the github pages.
It includes the already existing docs in a `GUIDES` tag.
Besides that, I have added the host aggregate documentation to see how it looks like, as it is by far the simpler.
Ref: https://hexdocs.pm/ex_doc/Mix.Tasks.Docs.html

image

image

PD: As the version is so long, the icon is rendered quite ugly...
We would have this with a short version string:
image


mix.exs
```diff
@@ -129,6 +132,57 @@ defmodule Trento.MixProject do
...
 
+  defp docs do
+    [
+      main: "readme",
+      logo: "priv/static/images/trento.svg",
+      extra_section: "GUIDES",
+      source_url: @source_url,
+      skip_undefined_reference_warnings_on: [
+        "CHANGELOG.md",
+        "docs/development/environment_variables.md",
```
arbulu89, line 143: This docs point to actual configuration files, and we can only attach md to the docs, so I just put this to remove the warning. But the links don't work.
I didn't make them work at least...

fabriziosestito: this is easily fixed by using absolute links in the md files

dottorblaster:
THE LOGO :heart:

mix.exs
arbulu89: done!

lib/trento/domain/host/host.ex
```diff
@@ -1,5 +1,20 @@
...
+
+  Each host running a Trento agent is registered as a new aggregate entry. The host
+  aggregate stores information such as:
+
+  - Host basic information as the hostname and ip addresses
+  - Hardware specifications
+  - Platform where the host is running (the cloud provider for instance)
+  - Registered SLES4SAP subscriptions
+
+  Besides these mostly static values, the aggregate takes care of detecting the host availability
```
nelsonkopliku, line 14: I would make also a reference to the fact that the aggregate detects changed discovered information and updates the system.

Not a biggie tho.

nelsonkopliku (approved):
Nice job!

fabriziosestito (approved):

PR #985, opened 2022-11-15
Document the cluster aggregate module

arbulu89:
Basic documentation of the cluster aggregate.
I have tried to simply specify the core concepts of the aggregate, without going too technical.
At the end, the idea is that anyone coming to this code has an initial knowledge about this aggregate and boundaries.

### DIFF
--- a/lib/trento/domain/cluster/cluster.ex
+++ b/lib/trento/domain/cluster/cluster.ex
@@ -1,5 +1,51 @@
 defmodule Trento.Domain.Cluster do
-  @moduledoc false
+  @moduledoc """
+  The cluster aggregate manages all the domain logic related to
+  deployed HA Clusters (Pacemaker, Corosync, etc).
+  The HA cluster is used to handle the high availability scenarios on the installed
+  SAP infrastructure. That's why this domain is tailored to work on clusters managing
+  SAP workloads. Any other type of resource is not handled properly.
+
+  Each deployed cluster is registered as a new aggregate entry, meaning that all the hosts belonging
+  to the same cluster are part of the same stream. Among all the hosts, only the cluster discovery messages
+  coming from the **designated controller** update the aggregate. In any case, all the hosts are listed in
+  the `hosts` field.
+
+  The cluster aggregate stores and updates information coming in the cluster discovery messages such as:
+
+  - Cluster name
+  - Number of hosts and cluster resources
+  - Platform where the host is running (the cloud provider for instance)
+  - Managed SAP workload SID
+
+  ## Cluster health
+
+  The cluster health is one of the most relevant concepts of this domain.
+  It shows if the cluster is working as expected or not, and in the second case,
+  what is the roout cause of the issue and if there is some possible remediation.
+  It is composed by sub-health elements:
+
+  - Discovered health
+  - Checks health
+
+  The main cluster health is computed using the values from these two. This means that the cluster health is the
+  worst of the two.
+
+  ### Discovered health
+
+  The discovered health comes from the cluster discovery messages and it depends on the cluster type.
+  Each cluster type has a different way of evaluating the health.
+
+  ### Checks health
+
+  The checks health is obtained from the [Checks Engine executions](https://github.com/trento-project/wanda/).
+  Every time a checks execution is started, the selected checks for this cluster are executed, and based on the result
+  the health value is updated. The checks are started from a user request or periodically following the
+  project scheduler configuration.
+
+  This domain only knows about the health, the details about the execution are stored in the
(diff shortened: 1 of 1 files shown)


lib/trento/domain/cluster/cluster.ex
```diff
@@ -1,5 +1,51 @@
...
+  @moduledoc """
+  The cluster aggregate manages all the domain logic related to
+  deployed HA Clusters (Pacemaker, Corosync, etc).
+  The HA cluster is used to handle the high availability scenarios on the installed
+  SAP infrastructure. That's why this domain is tailored to work on clusters managing
+  SAP workloads. Any other type of resource is not handled properly.
+
+  Each deployed cluster is registered as a new aggregate entry, meaning that all the hosts belonging
+  to the same cluster are part of the same stream. Among all the hosts, only the cluster discovery messages
+  coming from the **designated controller** update the aggregate. In any case, all the hosts are listed in
```
nelsonkopliku, line 11: > Among all the hosts, only the cluster discovery messages coming from the **designated controller** update the aggregate

Not really...

We take into account discoveries from non designated controllers to add them to the cluster, if it wasn't already registerd.
We do take into account those only after the designated controller was registered.

```elixir
# If no DC node was received yet, no cluster was registered.
def execute(%Cluster{cluster_id: nil}, %RegisterClusterHost{designated_controller: false}),
  do: {:error, :cluster_not_found}

def execute(
      %Cluster{} = cluster,
      %RegisterClusterHost{
        host_id: host_id,
        designated_controller: false
      }
    ) do
  maybe_emit_host_added_to_cluster_event(cluster, host_id)
end
```

nelsonkopliku (commented):
Nice! I noticed a little inconsistency on what happens with the designated controller.

lib/trento/domain/cluster/cluster.ex
```diff
@@ -1,5 +1,51 @@
 defmodule Trento.Domain.Cluster do
-  @moduledoc false
+  @moduledoc """
+  The cluster aggregate manages all the domain logic related to
+  deployed HA Clusters (Pacemaker, Corosync, etc).
+  The HA cluster is used to handle the high availability scenarios on the installed
+  SAP infrastructure. That's why this domain is tailored to work on clusters managing
+  SAP workloads. Any other type of resource is not handled properly.
```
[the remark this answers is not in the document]
arbulu89: Well, in reality they are handled. We have `drbd` clusters for instance. I simply meant that many of the functionalities won't be applied.
I will remove this sentence entirely.

lib/trento/domain/cluster/cluster.ex
arbulu89: Ok, you are right. I meant that all the values "except" the host list are only updated in those conditions. That's why I added the next sentence. But I will make it more clear

lib/trento/domain/cluster/cluster.ex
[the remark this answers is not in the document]
arbulu89: I have removed the sentence at the end. Even though other cluster types could be handled, giving this information is not valuable, as they we won't have this scenario

lib/trento/domain/cluster/cluster.ex
arbulu89: Rephrased

nelsonkopliku (approved):
DOCH! :scroll:

lib/trento/domain/cluster/cluster.ex
```diff
@@ -1,5 +1,52 @@
...
+  Each cluster type has a different way of evaluating the health.
+
+  ### Checks health
+
+  The checks health is obtained from the [Checks Engine executions](https://github.com/trento-project/wanda/).
+  Every time a checks execution is started, the selected checks for this cluster are executed, and based on the result
+  the health value is updated. The checks are started from a user request or periodically following the
+  project scheduler configuration.
+
+  This domain only knows about the health, the details about the execution are stored in the
```
fabriziosestito, line 47: just a little change here: "This bounded context" is more accurate.

fabriziosestito (changes requested):
Great work and thanks for this!

Just a mini nitpick on DDD terminology and LGTM

lib/trento/domain/cluster/cluster.ex
nelsonkopliku: Question: who is **This bounded context**?

If we refer to **web** or **wanda** I'd be fine, but if we refer to the Cluster Aggregate, I wouldn't agree.

Then what about other references to `domain` in this Aggregate's documentation?

fabriziosestito (approved):

PR #991, opened 2022-11-17
Phoenix channels broadcasting testing

CDimonaco:
# Description

This PR adds the phoenix channels broadcasting tests.
We refactor the broadcast mechanism to use the phoenix views for rendering the messages

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/application/projectors/check_result_projector.ex
+++ b/lib/trento/application/projectors/check_result_projector.ex
@@ -20,6 +20,8 @@ defmodule Trento.CheckResultProjector do
     HostChecksExecutionsReadModel
   }
 
+  alias TrentoWeb.ChecksResultView
+
   project(
     %ChecksExecutionRequested{
       cluster_id: cluster_id,
@@ -133,15 +135,25 @@ defmodule Trento.CheckResultProjector do
         _
       ) do
     Enum.each(hosts, fn host_id ->
-      TrentoWeb.Endpoint.broadcast("monitoring:clusters", "checks_results_updated", %{
-        cluster_id: cluster_id,
+      view_data = %{
         host_id: host_id,
-        hosts_executions: [%{cluster_id: cluster_id, host_id: host_id, reachable: true, msg: ""}],
+        cluster_id: cluster_id,
+        reachable: true,
+        msg: "",
         checks_results:
           Enum.map(checks, fn check_id ->
             %{host_id: host_id, check_id: check_id, result: :unknown}
           end)
-      })
+      }
+
+      TrentoWeb.Endpoint.broadcast(
+        "monitoring:clusters",
+        "checks_results_updated",
+        ChecksResultView.render(
+          "checks_results_updated.json",
+          data: view_data
+        )
+      )
     end)
   end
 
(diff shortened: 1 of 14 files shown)


lib/trento/application/projectors/check_result_projector.ex
```diff
@@ -133,15 +135,25 @@ defmodule Trento.CheckResultProjector do
         _
       ) do
     Enum.each(hosts, fn host_id ->
-      TrentoWeb.Endpoint.broadcast("monitoring:clusters", "checks_results_updated", %{
-        cluster_id: cluster_id,
+      view_data = %{
         host_id: host_id,
-        hosts_executions: [%{cluster_id: cluster_id, host_id: host_id, reachable: true, msg: ""}],
+        cluster_id: cluster_id,
+        reachable: true,
```
arbulu89, line 141: Nice! Now the magic happens in the view!

By the way, why have you decided to use a variable here, when in the others the data goes direclty in the `render` function

lib/trento/application/projectors/check_result_projector.ex
```diff
@@ -157,21 +169,32 @@ defmodule Trento.CheckResultProjector do
...
+    TrentoWeb.Endpoint.broadcast(
+      "monitoring:clusters",
+      "checks_results_updated",
+      ChecksResultView.render("checks_results_updated.json",
+        data: %{
+          cluster_id: cluster_id,
+          host_id: host_id,
+          reachable: reachable,
+          msg: msg,
+          checks_results:
```
arbulu89, line 181: We could even move this `checks_results` rendering to the view. Maybe sending a new field like `on_request: true` to set the result to `unknown` or use the actual result otherwise.
But maybe it is too much XD

test/trento/application/projectors/check_result_projector_test.exs
```diff
@@ -2,6 +2,9 @@ defmodule Trento.CheckResultProjectorTest do
   use ExUnit.Case
   use Trento.DataCase
 
+  import Phoenix.ChannelTest
+  import TrentoWeb.ChannelCase
```
arbulu89, line 6: So at the end the trick was to import this instead of `use`ing.
That's why you need to import the `Phoenix.ChannelTest`. 
Most probably you cannot `use` both because there is a conflict starting the sandbox

arbulu89 (commented):
Hey @CDimonaco
Promising work! 
I have my doubts on testing the broadcast together with the project, as we already start adding high timeouts (1 second looks high for unit testing)

test/trento/application/projectors/check_result_projector_test.exs
```diff
@@ -65,6 +79,19 @@ defmodule Trento.CheckResultProjectorTest do
                                           } ->
              reachable == nil
            end)
+
+    assert_broadcast "checks_results_updated",
```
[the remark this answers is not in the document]
CDimonaco: I considered, that, we already have tests in place for projector, and they are a sort of guideline in assertion making, the tests are always fast, I think we can go this way

test/trento/application/projectors/check_result_projector_test.exs
CDimonaco: exactly, only one type of test case should be "used" because then the sandbox init will conflict on both

fabriziosestito (approved):

lib/trento/application/projectors/check_result_projector.ex
fabriziosestito: this projector is going to go away so I would not spend a lot of time trying to refactor it

test/trento/application/projectors/check_result_projector_test.exs
```diff
@@ -17,6 +20,17 @@ defmodule Trento.CheckResultProjectorTest do
     HostChecksExecutionCompleted
   }
 
+  @endpoint TrentoWeb.Endpoint
+
+  setup do
```
arbulu89, line 25: @CDimonaco 
Maybe we could've have put this `setup` in a macro and use it.
Duplicating over and over again looks tedious hehe

Maybe it is a good exercise to learn about elixir macros

fabriziosestito: @CDimonaco @arbulu89 ExUnit offers a way to do this through [ExUnit.CaseTemplate](https://hexdocs.pm/ex_unit/1.12.3/ExUnit.CaseTemplate.html).
We could define a `Trento.ProjectorCase` and pass the channel option by using the [using](https://hexdocs.pm/ex_unit/1.12.3/ExUnit.CaseTemplate.html#functions) macro to add the shared imports, and then we could access the `tags` in the setup function see `TrentoWeb.ConnCase` for instance.

`use Trento.ProjectorCase, channel: "somechannel"`

PR #995, opened 2022-11-18
Add HA scenario & provider to `Cluster Checks Results` & `Cluster Settings` views

jamie-suse:
# Description

Adds the HA scenario and provider to the _Cluster Checks Results_ & _Cluster Settings_ views.

image

image

## How was this tested?

Tests have been added that validate the HA scenario and provider for a cluster in the _Cluster Checks Results_ view.

### DIFF
--- a/assets/js/components/ChecksResults/ChecksResults.jsx
+++ b/assets/js/components/ChecksResults/ChecksResults.jsx
@@ -29,6 +29,7 @@ import {
   sortHosts,
   sortChecks,
 } from './checksUtils';
+import { ClusterInfoBox } from '@components/ClusterDetails';
 
 const truncatedClusterNameClasses =
   'font-bold truncate w-60 inline-block align-top';
@@ -112,6 +113,7 @@ const ChecksResults = () => {
           use results with caution
         </WarningBanner>
       )}
+      <ClusterInfoBox haScenario={cluster.type} provider={cluster.provider} />
       <ResultsContainer
         catalogError={catalogError}
         clusterID={clusterID}
--- a/assets/js/components/ClusterDetails/ClusterInfoBox.jsx
+++ b/assets/js/components/ClusterDetails/ClusterInfoBox.jsx
@@ -1,0 +1,30 @@
+import React from 'react';
+
+import ListView from '@components/ListView';
+import ProviderLabel from './ProviderLabel';
+
+const haScenarioToString = {
+  hana_scale_up: 'HANA scale-up',
+  hana_scale_out: 'HANA scale-out',
+};
+
+export const ClusterInfoBox = ({ haScenario, provider }) => (
+  <div className="tn-cluster-details w-full my-4 mr-4 bg-white shadow rounded-lg px-8 py-8">
+    <ListView
+      className="grid-flow-row"
+      titleClassName="text-lg"
+      orientation="vertical"
+      data={[
+        {
+          title: 'HA Scenario',
+          content: haScenarioToString[haScenario] ?? 'Unknown',
+        },
+        {
+          title: 'Provider',
+          content: provider,
+          render: (content) => <ProviderLabel provider={content} />,
+        },
+      ]}
+    />
+  </div>
+);
--- a/assets/js/components/ClusterDetails/ClusterInfoBox.test.jsx
+++ b/assets/js/components/ClusterDetails/ClusterInfoBox.test.jsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { ClusterInfoBox } from './ClusterInfoBox';
+
+describe('Cluster Info Box', () => {
+  [
+    {
+      haScenario: 'hana_scale_up',
+      haScenarioText: 'HANA scale-up',
+      provider: 'aws',
+      providerText: 'AWS',
+    },
+    {
+      haScenario: 'hana_scale_out',
+      haScenarioText: 'HANA scale-out',
+      provider: 'aws',
+      providerText: 'AWS',
+    },
+    {
+      haScenario: 'hana_scale_up',
+      haScenarioText: 'HANA scale-up',
+      provider: 'azure',
+      providerText: 'Azure',
+    },
+    {
+      haScenario: '',
+      haScenarioText: 'Unknown',
+      provider: 'gcp',
+      providerText: 'GCP',
+    },
+    {
+      haScenario: 'unknown',
+      haScenarioText: 'Unknown',
+      provider: 'kvm',
+      providerText: 'KVM',
+    },
+    {
+      haScenario: 'hana_scale_up',
+      haScenarioText: 'HANA scale-up',
+      provider: 'nutanix',
+      providerText: 'Nutanix',
+    },
+  ].forEach(({ haScenario, haScenarioText, provider, providerText }) => {
+    it(`should display ${providerText} as the provider and HA Scenario: ${haScenarioText}`, () => {
+      const { getByText } = render(
+        <ClusterInfoBox haScenario={haScenario} provider={provider} />
+      );
+      expect(getByText(providerText)).toBeTruthy();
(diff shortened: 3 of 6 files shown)


jamie-suse:
I think this should be abstracted into a component as it needs to be reused in the _Cluster Settings_ view also.

jamie-suse:
Also open to a better name/location for `@components/ClusterDetails/ClusterInfoBox.jsx`

assets/js/components/ChecksResults/ChecksResults.jsx
```diff
@@ -29,6 +29,7 @@ import {
   sortHosts,
   sortChecks,
 } from './checksUtils';
+import ClusterInfoBox from '../ClusterDetails/ClusterInfoBox';
```
arbulu89, line 32: I would go with `@components/Clust...` here.
As it belongs to other component, let's use the alias. Otherwise, we can have issues if we change folders or anything

assets/js/components/ClusterDetails/ClusterInfoBox.jsx
```diff
@@ -0,0 +1,27 @@
...
+      titleClassName="text-lg"
+      orientation="vertical"
+      data={[
+        {
+          title: 'HA Scenario',
+          content: haScenario === 'hana_scale_up' ? 'HANA scale-up' : 'Unknown',
+        },
+        {
+          title: 'Provider',
+          content: provider || 'Not defined',
```
arbulu89, line 19: Still not convinced with this `Not defined` string that is not used anywhere.
What happens if we remove it?
Or we put an empty string maybe otherwise...

assets/js/components/ClusterDetails/ClusterInfoBox.jsx
```diff
@@ -0,0 +1,27 @@
...
+const ClusterInfoBox = ({ haScenario, provider }) => (
+  <div className="tn-cluster-details w-full my-4 mr-4 bg-white shadow rounded-lg px-8 py-8">
+    <ListView
+      className="grid-flow-row"
+      titleClassName="text-lg"
+      orientation="vertical"
+      data={[
+        {
+          title: 'HA Scenario',
+          content: haScenario === 'hana_scale_up' ? 'HANA scale-up' : 'Unknown',
```
arbulu89, line 15: We could go with a map here:
`hana_scale_up` -> `HANA scale-up`
`hana_scale_out` -> `HANA scale-out`
unknown otherwise

assets/js/components/ClusterDetails/ClusterInfoBox.test.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import ClusterInfoBox from './ClusterInfoBox';
+
+describe('Provider Label', () => {
+  it('should display AWS as the provider and HA Scenario: "HANA scale-up"', () => {
+    render(<ClusterInfoBox haScenario="hana_scale_up" provider="aws" />);
+    expect(screen.getAllByText(/AWS/)).toBeTruthy();
```
arbulu89, line 9: We could even test tha the scenario is follwing the `HA scenario` header and the provider text the `Provider` header. Just for completness

assets/js/components/ClusterDetails/ClusterInfoBox.test.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import ClusterInfoBox from './ClusterInfoBox';
+
+describe('Provider Label', () => {
+  it('should display AWS as the provider and HA Scenario: "HANA scale-up"', () => {
```
arbulu89, line 7: Just now thinking about this.
Is it possible to create tests with data coming from a loop?
Something like:
```
[{haScenario: "hana_scale_up", haScenarioText: "HANA Scale-up", provider: "aws", providerText: "AWS"},{{haScenario: "hana_scale_out", haScenarioText: "HANA Scale-out",...}].forEach(({haScenario: haScenario, haScenarioText: haScenarioText) => {
        it(`should display ${providerText} as the provider and HA Scenario: ${haScenarioText}`, () => {
              render(<ClusterInfoBox haScenario="hana_scale_up" provider="aws" />);
              expect(screen.getAllByText(/AWS/)).toBeTruthy();
...
```

This way we can easily add all the scenario (if we want) without adding more code

arbulu89 (changes requested):
Hey @jamie-suse 
some little adjustments

assets/js/components/ChecksResults/ChecksResults.jsx
dottorblaster: Xabi's right, also, I would turn this to `import { ClusterInfoBox } from '@components/ClusterDetails` adjusting `ClusterDetails`' `index.js` file accordingly

assets/js/components/ClusterDetails/ClusterInfoBox.test.jsx
dottorblaster: > Is it possible to create tests with data coming from a loop?

I don't know actually? We need to try :D

assets/js/components/ClusterDetails/ClusterInfoBox.test.jsx
```diff
@@ -0,0 +1,51 @@
...
+      providerText: 'Nutanix',
+    },
+  ].forEach(({ haScenario, haScenarioText, provider, providerText }) => {
+    it(`should display ${providerText} as the provider and HA Scenario: ${haScenarioText}`, () => {
+      render(<ClusterInfoBox haScenario={haScenario} provider={provider} />);
+      expect(screen.getAllByText(providerText)).toBeTruthy();
+      expect(screen.getAllByText(haScenarioText)).toBeTruthy();
+    });
+  });
+});
```
dottorblaster, line 51: HELL YEAH MAN

assets/js/components/ClusterDetails/ClusterSettings.jsx
```diff
@@ -14,6 +14,7 @@ import TriggerChecksExecutionRequest from '@components/TriggerChecksExecutionReq
 import { truncatedClusterNameClasses } from './ClusterDetails';
 import { getClusterName } from '@components/ClusterLink';
 import WarningBanner from '@components/Banners/WarningBanner';
+import { ClusterInfoBox } from '@components/ClusterDetails/ClusterInfoBox';
```
dottorblaster, line 17:
```suggestion
import { ClusterInfoBox } from '@components/ClusterDetails';
```

assets/js/components/ClusterDetails/ClusterInfoBox.test.jsx
arbulu89: Nice!

assets/js/components/ClusterDetails/ClusterInfoBox.test.jsx
```diff
@@ -0,0 +1,51 @@
...
+    },
+    {
+      haScenario: 'hana_scale_up',
+      haScenarioText: 'HANA scale-up',
+      provider: 'nutanix',
+      providerText: 'Nutanix',
+    },
+  ].forEach(({ haScenario, haScenarioText, provider, providerText }) => {
+    it(`should display ${providerText} as the provider and HA Scenario: ${haScenarioText}`, () => {
+      render(<ClusterInfoBox haScenario={haScenario} provider={provider} />);
```
arbulu89, line 46: This is still a bit risky. You can have the scenario where all the components are rendered, and you are checking in other screen part:
```
{ component } = render(<ClusterInfo...
expect(component.getAll...
```

arbulu89 (approved):
Thanks @jamie-suse ,
A couple of minor things that don't need a new review.
Ready to merge afterwards from my side

dottorblaster (approved):
LGTM

PR #996, opened 2022-11-21
Add SBD_OPTS to cluster discovery events

nelsonkopliku:
# Description

This change in the agent https://github.com/trento-project/agent/pull/145 exposes to discoveries also empty SBD configuration entries like `SBD_OPTS=` (becoming in the json `"SBD_OPTS": ""`)

With this PR we test that the change is supported by trento web as well.

## How was this tested?

Changed fixtures and let e2e tests run.

### DIFF
--- a/test/fixtures/discovery/ha_cluster_discovery_hana_scale_up.json
+++ b/test/fixtures/discovery/ha_cluster_discovery_hana_scale_up.json
@@ -1018,6 +1018,7 @@
         }
       ],
       "Config": {
+        "SBD_OPTS": "",
         "SBD_DELAY_START": "no",
         "SBD_DEVICE": "/dev/vdc;/dev/vdb",
         "SBD_MOVE_TO_ROOT_CGROUP": "auto",
--- a/test/fixtures/discovery/ha_cluster_discovery_unnamed.json
+++ b/test/fixtures/discovery/ha_cluster_discovery_unnamed.json
@@ -1018,6 +1018,7 @@
         }
       ],
       "Config": {
+        "SBD_OPTS": "",
         "SBD_DELAY_START": "no",
         "SBD_DEVICE": "/dev/vdc;/dev/vdb",
         "SBD_MOVE_TO_ROOT_CGROUP": "auto",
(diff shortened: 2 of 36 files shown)


dottorblaster (dismissed):
Photofinish scenarios are not supposed to fiddle with, mainly because the majority of the fixtures we have is extracted from real landscapes and should reflect what the customers are running.

https://github.com/trento-project/agent/pull/145/commits/c51ca4e27b053546ac59c7c099815664b676916c#r1027690491

> in case we could remove the extra entries so the output is the same to what we had with the regex

I think this is spot on, maybe we could pursue that path?

arbulu89:
> Photofinish scenarios are not supposed to fiddle with, mainly because the majority of the fixtures we have is extracted from real landscapes and should reflect what the customers are running.
> 
> [trento-project/agent@c51ca4e#r1027690491](https://github.com/trento-project/agent/commit/c51ca4e27b053546ac59c7c099815664b676916c#r1027690491)
> 
> > in case we could remove the extra entries so the output is the same to what we had with the regex
> 
> I think this is spot on, maybe we could pursue that path?

@dottorblaster In the other hand, I think this is a pretty logical movement.
I think that the change @nelsonkopliku did caught an "weakness" our discovery process had, not sending empty values.
I think it is valuable to have them.
Here the thing is that we don't even use the sbd config data in the web side...
Most probably we did in the old golang based backend, but we don't do it anymore.
That's why this change is harmless.
I don't know if the users would have those values changes (99% not) but modifying this values is something that we should consider.
As the agent evolves, the payloads must evolve together

arbulu89:
> @arbulu89 @dottorblaster with this [trento-project/agent#145](https://github.com/trento-project/agent/pull/145) being merged it looks safe enough updating the fixtures accordingly.
> 
> In addition, the updated fixtures is what would be extracted from real landscapes and would reflect what the customers are running (ofc I am referring to the latest version of the agent)

I'm fine with it. Mostly, because it is pretty irrelevant as we don't use this data anywhere (so it is not harmful for sure). But it is true that this is what the agent would be sending so I'm in favor of the change.

arbulu89 (approved):

fabriziosestito (approved):
Missing verified commit

dottorblaster (commented):
I don't like the idea of fiddling with fixtures but as long as we verify that this is exactly the payload that we can expect... I think we don't have really a choice :smile:

PR #997, opened 2022-11-21
Bump contracts

fabriziosestito:
...and remove dialyzer exceptions


arbulu89 (approved):
The same applies to wanda I guess

PR #1005, opened 2022-11-23
Help tailwind's JIT figure which classes we need

rtorrero:
# Description

Since tailwind 3, [JIT mode](https://v2.tailwindcss.com/docs/just-in-time-mode) is enabled by default. This causes that classes are only added if needed. When we use variables to construct the name of a class, we prevent tailwind from being able to understand what class it needs to add.

This caused that we missed the `fill-yellow-600` class needed for the warning summary and rendering as black text instead.

By specifying the classname completely without string formatting or concatenation, it works again. Result:

Captura desde 2022-11-23 13-32-37

## How was this tested?

It was tested manually. Unfortunately, I'm not sure we are able to add a test that checks the presence or absence of a CSS class.

### DIFF
--- a/assets/js/components/ClusterDetails/ChecksResultOverview.jsx
+++ b/assets/js/components/ClusterDetails/ChecksResultOverview.jsx
@@ -8,20 +8,32 @@ import React from 'react';
 
 const uiForResult = {
   passing: {
-    color: 'green',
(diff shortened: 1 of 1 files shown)


rtorrero:
The failure seems to be unrelated (flaky test?)

nelsonkopliku (approved):
LGTM

PR #1006, opened 2022-11-23
Frontend coveralls

arbulu89:
Add combined coveralls coverage for backend and frontend.
We cannot have 2 different badges, so the coverage is combined


janvhs:
@arbulu89 just wondering if this PR is still relevant, because the README now shows only one badge and it is open for two years at this point.

arbulu89:
> @arbulu89 just wondering if this PR is still relevant, because the README now shows only one badge and it is open for two years at this point.

It was a nice to have thing that got forgotten.
We can close.

PR #1007, opened 2022-11-23
Disable link to unavailable pacemaker cluster detail in home/health summary

nelsonkopliku:
# Description

This PR disables links to pacemaker cluster in the dashboard when a pacemaker cluster is not available.

no-cluster-link

## How was this tested?

An additional jest test was added and it was tested manually against the scenario provided by Alberto.

### DIFF
--- a/assets/js/components/Health/HealthIcon.jsx
+++ b/assets/js/components/Health/HealthIcon.jsx
@@ -11,13 +11,22 @@ import {
 import Spinner from '@components/Spinner';
 import classNames from 'classnames';
 
-const HealthIcon = ({ health = undefined, centered = false }) => {
+const HealthIcon = ({
+  health = undefined,
+  centered = false,
+  hoverOpacity = true,
+}) => {
+  const hoverOpacityClass = {
+    'hover:opacity-75': hoverOpacity,
+    'hover:opacity-100': !hoverOpacity,
+  };
+
   switch (health) {
     case 'passing':
       return (
         <EOS_CHECK_CIRCLE_OUTLINED
           className={classNames(
-            'hover:opacity-75',
+            hoverOpacityClass,
             computedIconCssClass('fill-jungle-green-500', centered)
           )}
         />
@@ -26,7 +35,7 @@ const HealthIcon = ({ health = undefined, centered = false }) => {
       return (
         <EOS_WARNING_OUTLINED
           className={classNames(
-            'hover:opacity-75',
+            hoverOpacityClass,
             computedIconCssClass('fill-yellow-500', centered)
           )}
         />
@@ -35,7 +44,7 @@ const HealthIcon = ({ health = undefined, centered = false }) => {
       return (
         <EOS_ERROR_OUTLINED
           className={classNames(
-            'hover:opacity-75',
+            hoverOpacityClass,
             computedIconCssClass('fill-red-500', centered)
           )}
         />
@@ -46,7 +55,7 @@ const HealthIcon = ({ health = undefined, centered = false }) => {
       return (
         <EOS_LENS_FILLED
           className={classNames(
-            'hover:opacity-75',
+            hoverOpacityClass,
             computedIconCssClass('fill-gray-500', centered)
           )}
         />
--- a/assets/js/components/HealthSummary/HomeHealthSummary.jsx
+++ b/assets/js/components/HealthSummary/HomeHealthSummary.jsx
@@ -51,10 +51,13 @@ const healthSummaryTableConfig = {
       className: 'text-center',
       render: (content, item) => {
         const linkToCluster = `/clusters/${item.clusterId}`;
-        return (
+
+        return item?.clusterId ? (
           <Link to={linkToCluster}>
             <HealthIcon health={content} centered={true} />
           </Link>
+        ) : (
+          <HealthIcon health={content} centered={true} hoverOpacity={false} />
         );
       },
     },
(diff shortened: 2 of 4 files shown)


CDimonaco:
I think we have the "core" functionality right here, but the hovering animation is still present, I think we should disable the hover transition and the pointer events, so that the item can completely act as a "non button", what do you think? Tag @jagabomb for further UI insights 👁️

assets/js/components/HealthSummary/HomeHealthSummary.jsx
```diff
@@ -52,7 +52,10 @@ const healthSummaryTableConfig = {
       render: (content, item) => {
         const linkToCluster = `/clusters/${item.clusterId}`;
```
EMaksy, line 53: Hey can we use an if statement to check if there is an clusterId? 
If there is none, we don't render the Link. 
Otherwise, we just hide the rendered Link.

What do you think ?

EMaksy (changes requested):

nelsonkopliku:
> I think we have the "core" functionality right here, but the hovering animation is still present, I think we should disable the hover transition and the pointer events, so that the item can completely act as a "non button"

hey @CDimonaco I agree with you, the fact is that to me it seems to be already behaving that way :smile: 

The hovering animation you are referring to is bound to `HealthIcon` component, and by disabling `pointerEvents` on the outer `Link`, those are not propagated to the contained icon and there is no hovering.

This is at least what I can observe when running it locally and in the attached gif

What am I missing?

assets/js/components/HealthSummary/HomeHealthSummary.jsx
nelsonkopliku: As discussed, that would mean having something like
```javascript
if (!item?.clusterId) {
  # here the HealthIcon needs to be instructed about the hover opacity
  return <HealthIcon health={content} centered={true} hoverOpacity={'0'}/>;
}

const linkToCluster = `/clusters/${item.clusterId}`;
return (
  <Link to={linkToCluster}>
    <HealthIcon health={content} centered={true} />
  </Link>
);
```
which I'd be fine with, yet it would require making an extra change to the `HealthIcon` component to accept the opacity, because it defaults to `hover:opacity-75` while in a specific scenario has to be 0.

I find enough disabling `pointerEvents` when we do not have a clusterId, keeping the scope narrow.
Still open to other's stance on this.

dottorblaster: > I find enough disabling pointerEvents when we do not have a clusterId

We don't change how event propagation works for a corner case. We change the rendered DOM instead. That's the point about React.

@EMaksy is right, we'll have to modify the `HealthIcon` using a prop accordingly, maybe a boolean one. Also, you might find useful what React docs say about conditional rendering :smile: 

https://reactjs.org/docs/conditional-rendering.html

nelsonkopliku: right, already two in that direction. Changing it!

assets/js/components/HealthSummary/HomeHealthSummary.test.jsx
```diff
@@ -53,6 +60,44 @@ describe('HomeHealthSummary component', () => {
...
+      container
+        .querySelector(unavailablePacemakerClusterSelector)
+        .getAttribute('href')
+    ).toContain(`/clusters/null`);
+
+    expect(
+      container.querySelector(unavailablePacemakerClusterSelector).style
+        .pointerEvents
+    ).toBe('none');
+  });
```
dottorblaster, line 100: Actually I would expect no href to be present if the cluster id is null :sweat_smile:

assets/js/components/HealthSummary/HomeHealthSummary.jsx
```diff
@@ -52,7 +52,10 @@ const healthSummaryTableConfig = {
...
         return (
-          <Link to={linkToCluster}>
+          <Link
+            to={linkToCluster}
+            style={{ pointerEvents: item?.clusterId ? 'auto' : 'none' }}
+          >
             <HealthIcon health={content} centered={true} />
           </Link>
```
dottorblaster, line 60: You can use a ternary and render the link conditionally

```suggestion
          item.clusterId ? (
            <Link
              to={linkToCluster}
              style={{ pointerEvents: item?.clusterId ? 'auto' : 'none' }}
            >
              <HealthIcon health={content} centered={true} />
            </Link>
          ) : (
            <HealthIcon health={content} centered={true} />
          )
```

dottorblaster (changes requested):

assets/js/components/HealthSummary/HomeHealthSummary.jsx
nelsonkopliku: Actually let me rephrase: changing because you are right. 

It semantically does not make sense having a Link which is actually not going to be a Link.
I overlooked that semantics, my bad :sweat_smile: 

Thanks guys! :heart:

assets/js/components/Health/HealthIcon.jsx
```diff
@@ -11,13 +11,22 @@ import {
...
-const HealthIcon = ({ health = undefined, centered = false }) => {
+const HealthIcon = ({
+  health = undefined,
+  centered = false,
+  hoverOpacity = true,
+}) => {
+  const hoverOpacityClass = {
+    'hover:opacity-75': hoverOpacity,
+    'hover:opacity-100': !hoverOpacity,
+  };
```
dottorblaster, line 22: Your call but keep in mind that you can also just pass `className` down to the rendered component

dottorblaster (approved):
LGTM

EMaksy (approved):
LGTM, thanks for applying the changes :+1:

PR #1008, opened 2022-11-24
Prevent flaky tests by using UUIDs on the tags values

rtorrero:
# Description
This PR changes the tag_factory to use UUIDs for the tag value.

Why?

 - We have a constraint that prevents the same tag/resource combination:
 ```|> unique_constraint([:resource_id, :value])```
 
 - The tests attempt to insert two tags generated in the factory `using Faker.Beer.hop()` which, given the sample size, causes collisions.
 
With this PR we shouldn't have any collision in the near future :sweat_smile: 

## How was this tested?

I tested the tests by manually running the tests :smile:

### DIFF
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -382,7 +382,7 @@ defmodule Trento.Factory do
 
   def tag_factory do
     %Tag{
-      value: Faker.Beer.hop(),
+      value: Faker.UUID.v4(),
       resource_id: Faker.UUID.v4(),
       resource_type: :host
     }
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -382,7 +382,7 @@ defmodule Trento.Factory do
 
   def tag_factory do
     %Tag{
-      value: Faker.UUID.v4(),
+      value: sequence(:value, &"#{Faker.Beer.hop()}0#{&1}"),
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
LOL wow

test/support/factory.ex
```diff
@@ -382,7 +382,7 @@ defmodule Trento.Factory do
 
   def tag_factory do
     %Tag{
-      value: Faker.Beer.hop(),
+      value: Faker.UUID.v4(),
```
arbulu89, line 385: We should use the `sequence` option of exmachina here
https://github.com/thoughtbot/ex_machina#overview

arbulu89 (changes requested):
Hey @rtorrero ,
Let's use the `sequence` option

test/support/factory.ex
rtorrero: It wouldn't make much of a diference but... ok, it looks prettier and we get to keep the hops :smile:

arbulu89 (approved):

test/support/factory.ex
arbulu89: It is to keep things meaningful. This value is regular string, not a uuid. Otherwise, we could put UUID everywhere hehe

PR #1010, opened 2022-11-24
Unified Api Response for errors

CDimonaco:
# Description

As the title suggest, this pr adds a unified way to deal with api errors, using ErrorView and phoenix fallback controller

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento_web/controllers/fallback_controller.ex
+++ b/lib/trento_web/controllers/fallback_controller.ex
@@ -1,0 +1,17 @@
+defmodule TrentoWeb.FallbackController do
+  use TrentoWeb, :controller
+
+  alias TrentoWeb.ErrorView
+
+  def call(conn, {:error, {:bad_request, reason}}) do
+    conn
+    |> put_status(:bad_request)
+    |> render_error(reason)
+  end
+
+  defp render_error(conn, reason) do
+    conn
+    |> put_view(ErrorView)
+    |> render("error.json", reason: reason)
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/trento_web/controllers/fallback_controller.ex
```diff
@@ -0,0 +1,35 @@
...
+    |> put_status(:bad_request)
+    |> render_error(reason)
+  end
+
+  def call(conn, {:error, {:internal_error, reason}}) do
+    conn
+    |> put_status(:internal_server_error)
+    |> render_error(reason)
+  end
+
```
arbulu89, line 23: I don't know if we should have a last pattern match with something more generic:
```
  def call(conn, {:error, reason}) do
```
But I guess that we are on control of sending these values, so we should be fine.
Just we need to be careful as the fallback action is not that obvious and we could en up sending something that is not matched here

test/trento_web/controllers/cluster_controller_test.exs
```diff
@@ -20,6 +24,66 @@ defmodule TrentoWeb.ClusterControllerTest do
     end
   end
 
+  describe "runner_callback" do
+    test "should return 400 when the request is invalid", %{conn: conn} do
```
arbulu89, line 28: It is strange to have tests on failed cases and not in the happy path hehe
What it is what it is

arbulu89 (approved):
Nice work!
Now we need to be careful on the controllers, so we always use known cases for the fallback controller.
@CDimonaco do you know if the callback controller doesn't match the returned value if the controller sends something different?

lib/trento_web/controllers/fallback_controller.ex
CDimonaco: Yes, it could happen, I tried to match all the errors we produce directly, the rest of the errors should be catched by other exception handlers, like ecto or somehting generic about 500

PR #1012, opened 2022-11-25
Add photofinish to the cache

rtorrero:
# Description

This PR enables the use of GH actions cache to store the photofinish binary used in the e2e testing.

In terms of speed there isn't a huge gain (though it seems to download faster from the cache than from the regular http endpoint) but it should prevent errors caused sometimes by download limits. See:
```
+ wget -P /tmp https://github.com/trento-project/photofinish/releases/download/v1.1.0/photofinish
--2022-11-23 13:54:35--  https://github.com/trento-project/photofinish/releases/download/v1.1.0/photofinish
...
Warning: Resolving objects.githubusercontent.com (objects.githubusercontent.com)... 185.199.111.133, 185.199.108.133, 185.199.109.133, ...
Connecting to objects.githubusercontent.com (objects.githubusercontent.com)|185.199.111.133|:443... connected.
HTTP request sent, awaiting response... 503 Egress is over the account limit.
2022-11-23 13:54:36 ERROR 503: Egress is over the account limit..
```

## How was this tested?

I runned it manually on my fork a few times to check if the cache asset was generated properly and used properly when available

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -364,19 +364,21 @@ jobs:
         run: mix setup
       - name: Run trento detached
         run: mix phx.server &
-      - name: Download Photofinish for e2e testing
-        if: steps.photofinish-cache.outputs.cache-hit != 'true'
-        run: |
-          set -x
-          wget -P /tmp https://github.com/trento-project/photofinish/releases/download/v1.1.0/photofinish
-          chmod +x /tmp/photofinish
+      - name: Install photofinish
+        uses: jaxxstorm/action-install-gh-release@v1.5.0
+        with:
+          repo: trento-project/photofinish
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Give executable permissions to photofinish
+        run: chmod +x $(whereis photofinish | cut -d" " -f2)
       - name: Cypress run
         uses: cypress-io/github-action@v4
         env:
           cypress_video: false
           cypress_db_host: postgres
           cypress_db_port: 5432
-          cypress_photofinish_binary: /tmp/photofinish
+          cypress_photofinish_binary: $(whereis photofinish | cut -d" " -f2)
         with:
           working-directory: test/e2e
           headless: true
@@ -511,13 +513,16 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Get Photofinish for e2e testing
-        run: |
-          set -x
-          wget -P /tmp https://github.com/trento-project/photofinish/releases/download/v1.1.2/photofinish
-          chmod +x /tmp/photofinish
+      - name: Install photofinish
+        uses: jaxxstorm/action-install-gh-release@v1.5.0
+        with:
+          repo: trento-project/photofinish
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Give executable permissions to photofinish
+        run: chmod +x $(whereis photofinish | cut -d" " -f2)
       - name: Push data
-        run: /tmp/photofinish run healthy-27-node-SAP-cluster -u "http://$TRENTO_DEMO_IP/api/collect" "$TRENTO_API_KEY"
+        run: photofinish run healthy-27-node-SAP-cluster -u "http://$TRENTO_DEMO_IP/api/collect" "$TRENTO_API_KEY"
 
   obs-commit:
     name: Commit the project on OBS
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -352,12 +352,6 @@ jobs:
           path: |
             test/e2e/node_modules
           key: ${{ runner.os }}-${{ env.NODE_VERSION }}-${{ hashFiles('test/e2e/package-lock.json') }}
-      - name: Retrieve cached photofinish binary
-        uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


arbulu89:
@rtorrero I think you wanted to put us as reviewers, not assignee hehe

arbulu89 (commented):
@rtorrero I miss how we could not use the cache if we update the photofinish version. With this code, once it is downloaded it won't be downloaded never again.
Or am I missing something?

arbulu89:
> Yes, the idea is that when we update photofinish we update this too. In the end, I think it's always a good thing that CI tools are pinned to specific versions rather than always downloading the latest which may reduce reproducibility of failed tasks.

Update what? You would need to update the photofinish version in the download stage, but remove this cache step together as well. Otherwise, as the cache still finds the binary there, it wouldn't never allow to run the download stage.
You would need to put some `key` that changes based on the photofinish.
I think that we need to do some improvements, we cannot be adding/removing the cache step everytime we do a photofinish upgrae

arbulu89:
@rtorrero besides that, we even have other stage in the CI that uses photofinish:
https://github.com/trento-project/web/blob/552cf4e2cc1aa44c7822b79a65a239bb57ef250f/.github/workflows/ci.yaml#L517

Funnily enough, with a newer version. Maybe we should even extract the photofinish download to an individual job. To manage the download and cache, and use it in the other 2 jobs

rtorrero:
> > Yes, the idea is that when we update photofinish we update this too. In the end, I think it's always a good thing that CI tools are pinned to specific versions rather than always downloading the latest which may reduce reproducibility of failed tasks.
> 
> Update what? You would need to update the photofinish version in the download stage, but remove this cache step together as well. Otherwise, as the cache still finds the binary there, it wouldn't never allow to run the download stage. You would need to put some `key` that changes based on the photofinish. I think that we need to do some improvements, we cannot be adding/removing the cache step everytime we do a photofinish upgrae

Yes, the idea would be to add a version identifier to the cache key when it needs to be updated. We can do it now for clarity if that helps.

I don't think that photofinish is going to be updated so often that this will become a burden, if you ask me. In any case, nothing changes here, right now if we want to use a newer photofinish version we still need to update the URL.

> @rtorrero besides that, we even have other stage in the CI that uses photofinish:
Will update, thanks for the catch!

fabriziosestito (changes requested):
This problem is already solved by this action so we should consider using it:
https://github.com/marketplace/actions/install-a-binary-from-github-releases

It downloads the latest release if available and uses a cache if configured.

This also addresses @arbulu89 concerns.

.github/workflows/ci.yaml
```diff
@@ -352,24 +352,33 @@ jobs:
...
         run: mix phx.server &
-      - name: Get Photofinish for e2e testing
-        run: |
-          set -x
-          wget -P /tmp https://github.com/trento-project/photofinish/releases/download/v1.1.0/photofinish
-          chmod +x /tmp/photofinish
+      - name: Install photofinish
+        uses: jaxxstorm/action-install-gh-release@v1.5.0
+        with:
+          repo: trento-project/photofinish
```
arbulu89, line 370: We should still pin the binary to a version here (`1.2.0`) and the action itself already have a caching system, using the `cache: enable` entry

.github/workflows/ci.yaml
```diff
@@ -352,24 +352,33 @@ jobs:
           path: |
             test/e2e/node_modules
           key: ${{ runner.os }}-${{ env.NODE_VERSION }}-${{ hashFiles('test/e2e/package-lock.json') }}
+      - name: Retrieve cached photofinish binary
```
arbulu89, line 355: I think we can remove this and use the cache option of the new action

arbulu89 (commented):
Thanks @rtorrero ,
I think we can use the cache option of the `action-install-hh-release`

.github/workflows/ci.yaml
```diff
@@ -358,18 +358,23 @@ jobs:
...
+      - name: Install photofinish
+        uses: jaxxstorm/action-install-gh-release@v1.5.0
+        with:
+          repo: trento-project/photofinish
+          tag: v1.1.2
+          cache: enable
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Give executable permissions to photofinish
+        run: chmod +x $(whereis photofinish | cut -d" " -f2)
```
arbulu89, line 370: Maybe we could replace `whereis` by `which`. This 2nd returns the binary path

arbulu89 (approved):
Thanks @rtorrero 
All good from my side.
PD: Maybe `which` would work better than `whereis`

fabriziosestito (approved):

.github/workflows/ci.yaml
rtorrero: @arbulu89 I had already tried but from what I could see, `which` returns only results that already have the correct permissions:

```
[rtorrero@galileo ~]$ which photofinish
which: no photofinish in (/home/rtorrero/.mix/escripts:/home/rtorrero/go/bin:/home/rtorrero/.local/bin...)

[rtorrero@galileo ~]$ whereis photofinish
photofinish: /usr/bin/photofinish

[rtorrero@galileo ~]$ sudo chmod +x /usr/bin/photofinish 
[rtorrero@galileo ~]$ which photofinish
/usr/bin/photofinish

```

arbulu89: So you could use `which` in the 2nd call XD

PR #1013, opened 2022-11-25
Add test and refactor the  AboutPage component

EMaksy:
# Description

Refactor AboutPage component and added multiple components.
Add an AboutPageText component which includes text description.
Add an AboutPageLogo component which includes the Logo.

This will make the AboutPage component better accessible, easier to read and test. 

## How was this tested?

Added test for AboutPage component.
Missing tests for AboutPageText and AboutPageLogo components.
This will be added in a follow-up pr.

### DIFF
--- a/assets/js/components/AboutPage/AboutPage.test.jsx
+++ b/assets/js/components/AboutPage/AboutPage.test.jsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { aboutPageFactory } from '@lib/test-utils/factories';
+import AboutPage from './index';
+import { act, screen } from '@testing-library/react';
+import { renderWithRouter } from '@lib/test-utils';
+
+describe('AboutPage component', () => {
+  const apiRequestData = aboutPageFactory.build();
+  it('should render the about page with content from the api', async () => {
+    await act(async () => {
+      renderWithRouter(
+        <AboutPage onFetch={() => Promise.resolve({ data: apiRequestData })} />
+      );
+    });
+
+    expect(screen.getByText(apiRequestData.flavor)).toBeTruthy();
+    expect(screen.getByText(apiRequestData.version)).toBeTruthy();
+    expect(
+      screen.getByText(`${apiRequestData.sles_subscriptions} found`)
+    ).toBeTruthy();
+  });
+
+  it('should render the about page with state values if api get request fails', async () => {
+    const stateValues = { flavor: 'N/A', subscriptions: 0, version: 'v0.0.0' };
+    const errorMessage = { messages: "Get request '/api/about' failed" };
+    await act(async () => {
+      renderWithRouter(
+        <AboutPage onFetch={() => Promise.reject(errorMessage)} />
+      );
+    });
+
+    expect(screen.getByText(stateValues.flavor)).toBeTruthy();
+    expect(screen.getByText(stateValues.version)).toBeTruthy();
+    expect(screen.getByText(`${stateValues.subscriptions} found`)).toBeTruthy();
+  });
+});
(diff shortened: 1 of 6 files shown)


assets/js/lib/test-utils/factories.js
```diff
@@ -87,3 +87,9 @@ export const hostnameFactory = Factory.define(() => ({
   id: faker.datatype.uuid(),
   hostname: faker.hacker.noun(),
 }));
+
+export const aboutPageFactory = Factory.define(() => ({
+  flavor: faker.animal.cat(),
+  sles_subscriptions: faker.datatype.number(),
+  version: faker.git.branch(),
```
arbulu89, line 94: Maybe this better: https://fakerjs.dev/api/system.html#semver

assets/js/lib/test-utils/factories.js
```diff
@@ -87,3 +87,9 @@ export const hostnameFactory = Factory.define(() => ({
   id: faker.datatype.uuid(),
   hostname: faker.hacker.noun(),
 }));
+
+export const aboutPageFactory = Factory.define(() => ({
```
arbulu89, line 91: I would name this `aboutPage`, simply `aboutFactory` as it returns the about data, even though it is only used in the about page by now hehe

assets/js/lib/api/about.js
```diff
@@ -0,0 +1,3 @@
+import { get } from '@lib/network';
+
+export const getAboutPageData = () => get('/api/about');
```
arbulu89, line 3: Maybe better `getAboutData` ?

assets/js/components/AboutPage/AboutPage.test.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { aboutPageFactory } from '@lib/test-utils/factories';
+import AboutPage from './index';
```
arbulu89, line 4: You could even to do `import AboutPage from './';` or `import AboutPage from `./AboutPage';`

assets/js/components/AboutPage/AboutPage.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+import { act, screen } from '@testing-library/react';
+import { renderWithRouter } from '@lib/test-utils';
+
+describe('AboutPage component', () => {
+  const apiRequestData = aboutPageFactory.build();
+  it('should render the about page with content from the api', async () => {
+    await act(async () => {
+      renderWithRouter(
```
arbulu89, line 12: Usually is a good practice to do something like:
```
{ getByText } = renderWithRouter(..
expect(getByText(api..
```
This way you are 100% sure that the used `getByText` is applied in the rendered component

assets/js/components/AboutPage/AboutPage.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+    });
+
+    expect(screen.getByText(apiRequestData.flavor)).toBeTruthy();
+    expect(screen.getByText(apiRequestData.version)).toBeTruthy();
+    expect(
+      screen.getByText(`${apiRequestData.sles_subscriptions} found`)
+    ).toBeTruthy();
+  });
+
+  it('should render the about page with state values if api get request fails', async () => {
```
arbulu89, line 24: Maybe with `default values` ?

arbulu89 (approved):
Really nice @EMaksy ,
I wrote some cosmetic recommendations, but it is really nice as it is as well

assets/js/lib/test-utils/factories.js
EMaksy: Yeah thanks!
Naming is hard haha

assets/js/components/AboutPage/AboutPage.test.jsx
EMaksy: Yeah good point

PR #1016, opened 2022-11-28
Eslint fix

CDimonaco:
# Description

Reconfigure the eslint linter based on `airbnb` styleguide with the appropriate linting rules.

This is a first process, hundreds of error will be fixed, but this is a good starting point for further iterations when needed.

We will disable some rules to match our coding standards and our usecases.

## How was this tested?

Automated tests still running


dottorblaster (approved):
LGTM

PR #1023, opened 2022-12-01
Broadcast execution completed

fabriziosestito:
Adds the updated execution completed action dispatch when the BE receives an execution completed event from wanda.

NOTE:
I had to add the channel and the callback manually and couldn't rely on the `registerEvents` utility function since the name of the action and the message sent to the phoenix channel differs.
Also, I'd like to use action constants instead of generating the name of the action so I will track a debt to track an improvement on this side.

### DIFF
--- a/assets/js/state/index.js
+++ b/assets/js/state/index.js
@@ -14,7 +14,9 @@ import catalogNewReducer from './catalogNew';
 import lastExecutionsReducer from './lastExecutions';
 import liveFeedReducer from './liveFeed';
 import settingsReducer from './settings';
-import registerEvents from './registerSocketEvents';
+import registerEvents, { joinChannel } from './registerSocketEvents';
+
+import { updateLastExecution } from '@state/actions/lastExecutions';
 
 import rootSaga from './sagas';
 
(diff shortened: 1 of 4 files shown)


assets/js/state/index.js
```diff
@@ -72,6 +74,19 @@ const processChannelEvents = (store) => {
...
+  // FIXME: This is to overcome the fact that we are generating names with registerEvents
+  // in the future we want to remove this and use the constants directly,
+  // since events and actions may have different names and parameters.
+  const channel = socket.channel('monitoring:executions', {});
+  channel.on('execution_completed', ({ group_id }) => {
+    store.dispatch({
+      type: UPDATE_LAST_EXECUTION,
+      payload: { groupID: group_id },
+    });
+  });
```
dottorblaster, line 87:
```suggestion
  channel.on('execution_completed', ({ group_id: groupID }) => {
    store.dispatch({
      type: UPDATE_LAST_EXECUTION,
      payload: { groupID },
    });
  });
```

dottorblaster (changes requested):
Just a tiny thing then we can merge

assets/js/state/index.js
```diff
@@ -72,6 +74,19 @@ const processChannelEvents = (store) => {
     'database_instance_health_changed',
     'database_instance_system_replication_changed',
   ]);
+
+  // FIXME: This is to overcome the fact that we are generating names with registerEvents
+  // in the future we want to remove this and use the constants directly,
+  // since events and actions may have different names and parameters.
+  const channel = socket.channel('monitoring:executions', {});
+  channel.on('execution_completed', ({ group_id: groupID }) => {
+    store.dispatch({
```
arbulu89, line 83: We should import and use the appropriate action here: 
https://github.com/trento-project/web/blob/main/assets/js/state/actions/lastExecutions.js#L3
```
store.dispatch(updateLastExecution(groupID))
```

arbulu89 (approved):
Looks good @fabriziosestito ,
Let's use the action function instead of importing the action key and creating the payload manually

arbulu89 (approved):

dottorblaster (approved):
LGTM

PR #1032, opened 2022-12-02
Fix get health

arbulu89:
Improve the old `getHealth` (now `getCheckHealthByAgent`) to consider all the possible scenarios: agentCheckError, expect_same, fact gathering error, etc

I have added more factory improvements in order to make the check results mocks creation easy.
PD: I have fixed some other functions like `getHosts` by this will be removed once we get the other PR using the `targets` to create the results table


assets/js/components/ChecksResults/checksUtils.js
```diff
@@ -32,31 +34,71 @@ export const getCheckResults = (executionData) => {
...
+        type === 'expect_same' &&
+        !checkResult.expectation_results.find(
+          ({ name: resultName }) => resultName === name
+        )?.result
+    ).length;
+
+  const failedExpectations =
+    evaluationErrors + failedExpectEvaluations + failedExpectSameEvaluations;
+
+  const health = failedExpectations > 0 ? checkResult.result : 'passing';
```
nelsonkopliku, line 96: What about relying on wanda's response for the health :thinking: 

And since this has been introduced https://github.com/trento-project/wanda/pull/104 (counts for passing/warning/critical checks), could also those be used in the frontend?

arbulu89: We cannot. The way the Frontend is now, we need to know the result for each host, which wanda doesn't store, as it only stores the complete check result.
Once we change this view, using a more checks result oriented rather than host result, we will need to revisit this whole thing

nelsonkopliku: Alright, got it. Thanks!

dottorblaster: Yeah, @arbulu89 is right, the counters are for checks results, we need to count single expectations too :smile:

assets/js/components/ChecksResults/checksUtils.js
```diff
@@ -32,31 +34,71 @@ export const getCheckResults = (executionData) => {
...
+  // agentCheckError
+  if (agentCheckResult?.type) {
+    return {
+      health: 'critical',
+      error: agentCheckResult.message,
+    };
+  }
+
+  // expectation evaluation error , malformed expression most probably
+  const evaluationErrors = agentCheckResult?.expectation_evaluations.filter(
```
rtorrero, line 73: I personally like adding a `Count` suffix to variables that are just counting something, in this case e.g.: `evaluationErrorCount` as it makes operations with the variable more obvious (e.g. the `+` below is immediately understood as a sum and not a concatenation). Wdyt?

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -83,3 +113,70 @@ export const withCompletedResults = (execution, result, checkResults) => ({
...
+
+  return {
+    ...checkResult,
+    agents_check_results: agents,
+    expectation_results: [],
+  };
+};
+
+const addExpectation = (checkResult, name, expec, result) => {
+  const { type } = expec;
```
rtorrero, line 131: expect maybe?

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -83,3 +113,70 @@ export const withCompletedResults = (execution, result, checkResults) => ({
...
+  const expec = executionExpectationEvaluationFactory.build({
+    name,
+    type,
+    return_value: false,
+  });
+
+  return addExpectation(checkResult, name, expec, false);
+};
+
+export const addExpectationWithError = (checkResult) => {
```
rtorrero, line 175: maybe `addErroredExpectation` sound better and matches the same wording as the other functions?

rtorrero (approved):
LGTM @arbulu89 , just a bit of (opinionated) nitpicking on naming

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -1,60 +1,82 @@
 import React from 'react';
+import { screen } from '@testing-library/react';
+
 import { renderWithRouter } from '@lib/test-utils';
 import {
+  addCriticalExpectation,
```
arbulu89, line 6: Is there any reason not to put the `executions` under the factories import. All of them are exported in the index file.
`@lib/test-utils/factories`

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -87,7 +108,8 @@ describe('ExecutionResults', () => {
     const {
       clusterID,
       hostnames,
-      checkID,
+      checkID1,
```
arbulu89, line 111: Should we return maybe `checks`? As we do with the hostnames

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -1,60 +1,82 @@
...
-  expectationResultFactory,
+  withEmptyExpectations,
 } from '@lib/test-utils/factories/executions';
 import { catalogFactory, hostnameFactory } from '@lib/test-utils/factories';
 
 import ExecutionResults from './ExecutionResults';
 
 const prepareStateData = (checkExecutionStatus) => {
-  const hostnames = hostnameFactory.buildList(2);
+  const hostnames = hostnameFactory.buildList(4);
```
arbulu89, line 18: Why a `4` here?

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -1,60 +1,82 @@
...
+  let checkResult2 = checkResultFactory.build({
+    agents_check_results: [agentCheckResult3, agentCheckResult4],
+    result: 'critical',
+  });
+
+  checkResult2 = withEmptyExpectations(checkResult2);
+  checkResult2 = addPassingExpectation(checkResult2, 'expect');
+  checkResult2 = addCriticalExpectation(checkResult2, 'expect');
+
+  const executionResult = checksExecutionCompletedFactory.build({
```
arbulu89, line 50: Here the order of things looks strange.
I think we should create `targets` before this, and assign it there, as the targets belongs to the execution result.

```
 const targets = [
    { agent_id: agentID1, checks: [checkID1, checkID2] },
    { agent_id: agentID2, checks: [checkID1, checkID2] },
  ];

const executionResult = checksExecutionCompletedFactory.build({
    check_results: [checkResult1, checkResult2],
    targets: targets,
    result: 'critical',
  });
```

After that, instead of creating a new execution data for the `lastExecution` element, we should directly pass this value.

```
  const lastExecution = {
    executionLoading: false,
    executionData: executionResult,
    error: '',
  };
```

arbulu89 (commented):
@dottorblaster It looks good.
Have a look only in one of the factory functions that I think you updated accidentaly.
Besides that, a small change in the test to make it more logical.
Green light, otherwise. 
I cannot approve though, as I opened the PR XD

assets/js/lib/test-utils/factories/executions.js
dottorblaster: This should be `expectation` to me, you're right

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
dottorblaster: Leftover from an experiment LOL

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
dottorblaster: Fair point

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -1,60 +1,85 @@
...
+    error: null,
   });
 
   const lastExecution = {
     executionLoading: false,
     executionData: {
       status: checkExecutionStatus,
       targets,
-      check_results: [checkResult],
+      check_results: [checkResult1, checkResult2],
```
arbulu89, line 82: @dottorblaster You fortot to change this `executionData: executionResult`

PR #1041, opened 2022-12-07
Fix frontend formatting tasks

CDimonaco:
# Description

As the title suggest, fix the integration of prettier with eslint rules.

## How was this tested?

Manually


dottorblaster (commented):
LOL LGTM

PR #1042, opened 2022-12-07
Jwt authentication

CDimonaco:
# Description

This pr change the authentication method of the web project from stateful cookie auth to stateless jwt auth.

Changes the web auth logic and implement on the frontend a jwt auth flow with the all the client side authentication flow handled within the single page application. Route guard on the frontend will be added.

The application will use a access/refresh token flow, with the SPA responsible for all the api interaction, including the login and the token refresh part. Here some diagrams

## Login flow

flowchart_trento_login

## Refresh token success flow

flow_chart_trento_refresh_ok

## Refresh token failure flow

flowchart_trento_refresh_bad

The frontend routes are protected with a react router `route guard`. The guard will check that the user is logged in, and the token is valid. The route uses the `/api/me` endpoint.

The backend jwts, both access and refresh token are signed with a HS256 (HMAC with SHA-256)  key, on runtime the key should be provided as config variable.

**The AUTHENTICATION IS ENABLED IN THE DEV ENVIRONMENT**

## How was this tested?

Automatic tests.

The e2e tests are refactored to use cypress version 12 with proper session support across tests, we need that in order to ensure that the authentication mechanism is present in e2e tests and well tested.


CDimonaco:
DO NOT MERGE, BEFORE THE HELM CHART IS UPDATED

nelsonkopliku (dismissed):
Amazing.

lib/trento_web/auth/jwt_auth_plug.ex
```diff
@@ -0,0 +1,93 @@
...
+  use Pow.Plug.Base
+
+  require Logger
+
+  alias Plug.Conn
+  alias TrentoWeb.Auth.AccessToken
+  alias TrentoWeb.Auth.RefreshToken
+
+  @impl true
+  @spec fetch(Plug.Conn.t(), any) :: {Plug.Conn.t(), nil | %{optional(<<_::40>>) => binary}}
```
fabriziosestito, line 19: the spec is funky and not needed since we have already an @impl true
I would anyway add a docstring here

lib/trento_web/auth/jwt_auth_plug.ex
```diff
@@ -0,0 +1,93 @@
...
+        |> Conn.put_private(:user_id, claims["sub"])
+
+      {conn, %{"access_token" => jwt_token, "user_id" => claims["sub"]}}
+    else
+      _ -> {conn, nil}
+    end
+  end
+
+  @impl true
+  def create(conn, user, _config) do
```
fabriziosestito, line 35: I would  add a docstring here

lib/trento_web/auth/jwt_auth_plug.ex
```diff
@@ -0,0 +1,93 @@
...
+
+        {:ok, conn}
+
+      {:error, reason} ->
+        {:error, reason}
+    end
+  end
+
+  @impl true
+  def delete(conn, _config) do
```
fabriziosestito, line 74: I would  add a docstring here

lib/trento_web/auth/jwt_auth_plug.ex
```diff
@@ -0,0 +1,93 @@
+defmodule TrentoWeb.Auth.JWTAuthPlug do
+  @moduledoc """
+    The JWTAuthPlug is a Pow compatible authorization flow.
+    Handles the login and the credentials recovery at each request
+
+    Uses Joken for jwt management
+
+    See the pow documentation for further details.
```
fabriziosestito, line 8: It is worth to add a link to Pow.Plug.Base

lib/trento_web/templates/session/new.html.eex
```diff
@@ -24,7 +24,7 @@
             Password
           </label>
           <div class="mt-1">
-            <input id="user_password" name="user[password]" type="password" autocomplete="current-password" required class="appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-jungle-green-500 focus:border-jungle-green-500 sm:text-sm">
+            <input id="user_password" name="password" type="password" autocomplete="current-password" required class="appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-jungle-green-500 focus:border-jungle-green-500 sm:text-sm">
```
fabriziosestito, line 27: shouldn't we port this view completely to React SPA?

test/trento_web/auth/access_token_test.exs
```diff
@@ -0,0 +1,56 @@
...
+
+  setup [:set_mox_from_context, :verify_on_exit!]
+
+  setup do
+    expect(
+      Joken.CurrentTime.Mock,
+      :current_time,
+      3,
+      fn ->
+        1_671_641_814
```
fabriziosestito, line 18: this number can go to a module attribute for clarity since we repeat it a lot in the tests below

test/trento_web/auth/jwt_auth_plug_test.exs
```diff
@@ -0,0 +1,218 @@
...
+      expect(
+        Joken.CurrentTime.Mock,
+        :current_time,
+        0,
+        fn ->
+          1_671_715_992
+        end
+      )
+
+      bad_refresh =
```
fabriziosestito, line 77: To make the test even more robust, can't these tokens be generated programmatically instead of using a magic string in the tests?

test/trento_web/auth/refresh_token_test.exs
```diff
@@ -0,0 +1,58 @@
...
+
+  setup [:set_mox_from_context, :verify_on_exit!]
+
+  setup do
+    expect(
+      Joken.CurrentTime.Mock,
+      :current_time,
+      3,
+      fn ->
+        1_671_641_814
```
fabriziosestito, line 18: constant in module attribute

test/trento_web/controllers/session_controller_test.exs
```diff
@@ -0,0 +1,198 @@
...
+          "refresh_token" => "malformed"
+        })
+
+      resp = json_response(conn, 401)
+
+      assert %{"error" => "You don't have access to the resource."} = resp
+    end
+
+    test "should return unauthorized if the refresh token is valid but expired", %{conn: conn} do
+      expired_jwt =
```
fabriziosestito, line 99: same comment about generating those instead of using magic strings

fabriziosestito (changes requested):
Great work overall! 

I did a first pass of the review looking only at the elixir code, I will take the time to look at the FE part soonish.

assets/js/state/user.js
```diff
@@ -0,0 +1,36 @@
...
+    setAuthError(state, { payload: { error } }) {
+      state.authInProgress = false;
+      state.authError = error;
+    },
+    setUserAsLogged(state, _payload) {
+      state.loggedIn = true;
+      state.authInProgress = false;
+      state.authError = null;
+    },
+    setUser(state, { payload: { username } }) {
```
arbulu89, line 27: Why do we need an independent `setUser` function? Couldn't we put it in `setUserAsLogged` and pass the username there?

test/e2e/cypress/e2e/about.cy.js
```diff
@@ -2,7 +2,7 @@ import { getValue } from '../support/common';
 
 describe('User account page', () => {
   before(() => {
-    cy.navigateToItem('About');
+    cy.visit('/about');
     cy.url().should('include', '/about');
```
nelsonkopliku, line 6: Is this assertion `cy.url().should('include', '/about');` needed when we do `cy.visit('/about');`

test/e2e/cypress/e2e/checks_catalog.cy.js
```diff
@@ -5,7 +5,7 @@ import {
 
 context('Checks catalog', () => {
   before(() => {
-    cy.navigateToItem('Checks catalog');
+    cy.visit('/catalog');
     cy.url().should('include', '/catalog');
```
nelsonkopliku, line 9: same question about the assertion. And for other places as well.

assets/js/state/sagas/user.test.js
```diff
@@ -0,0 +1,98 @@
...
+        password: 'good',
+      },
+    });
+
+    expect(dispatched).toContainEqual(setAuthInProgress());
+    expect(dispatched).toContainEqual(setUser({ username: 'good' }));
+    expect(dispatched).toContainEqual(setUserAsLogged());
+  });
+
+  it('should set the access token and refresh token in the storage when the login is successful', async () => {
```
arbulu89, line 74: Couldn't we combine this test with the previous? They run the same, so it would be a matter of adding the extra expects in the prev test. What do you think?

arbulu89 (changes requested):
Hey @CDimonaco ,
It looks great. I couldn't do a really deep review as it is huge.
Besides some cosmetic comments, 2 things:
- The routes in `trento.jsx` look duplicated
- The login page component html template now is duplicated in the frontend and backend, is this needed?

PD: I have tested the happy path of the branch, and it works great. And really appreciated for the massive amount of tests you added, this give a lot of confidence to the change

test/e2e/cypress/e2e/sap_systems_overview.cy.js
```diff
@@ -198,10 +212,16 @@ context('SAP Systems Overview', () => {
 
   describe('SAP Systems Tagging', () => {
     before(() => {
-      cy.removeTagsFromView();
+      cy.apiLogin().then(() => {
+        cy.removeTagsFromView();
```
nelsonkopliku, line 216: Inside `cy.removeTagsFromView()` there is already an `apiLogin()`. Is this intended?

lib/trento_web/templates/session/new.html.eex
CDimonaco: Yes, this is a leftover, I will remove the file, good catch :D

test/trento_web/auth/jwt_auth_plug_test.exs
CDimonaco: The tokens need to be crafted accordingly, in order to set a proper time and date we will need to remock the current time and regenerate, or tweak the library to override the claims, imho it's better like that

test/trento_web/controllers/session_controller_test.exs
CDimonaco: Same as the last comment

assets/js/lib/auth/index.js
```diff
@@ -0,0 +1,51 @@
+import axios from 'axios';
+
+const STORAGE_ACCESS_TOKEN_IDENTIFIER = 'access_token';
+const STORAGE_REFRESH_TOKEN_IDENTIFIER = 'refresh_token';
+
+export const authClient = axios.create();
+
+export const login = (credentials) =>
+  authClient.post('/api/session', credentials).then((response) => {
```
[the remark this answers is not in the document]
CDimonaco: This functions are part of the "auth library", the error handling is handled by the callers, and it's actually handled

assets/js/state/user.js
CDimonaco: Yes, we can but I prefer to distinguish the two things and have two separate action, in order to attach sagas to both of the authentication actions if needed, but we can discuss about that

assets/js/state/sagas/user.test.js
CDimonaco: you're right, squashed into one test

test/e2e/cypress/e2e/about.cy.js
CDimonaco: I was thinking about that, but since we have some route guards that can change the route url, I just added as a defensive movement to debug the e2e tests in case of failing, but we can remove

test/e2e/cypress/support/commands.js
```diff
@@ -31,19 +31,59 @@ Cypress.Commands.add('login', () => {
     Cypress.env('login_user'),
     Cypress.env('login_password'),
   ];
-  cy.visit('/login');
-  cy.get('#user_username').type(username);
-  cy.get('#user_password').type(password);
-  cy.get(':nth-child(5) > .w-full').click();
+  cy.visit('/session/new');
```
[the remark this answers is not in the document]
CDimonaco: done!

test/e2e/cypress/e2e/sap_systems_overview.cy.js
CDimonaco: Nope, we can remove it good catch!

fabriziosestito (approved):

assets/js/components/Login.test.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+    await act(async () => {
+      const usernameField = screen.getByTestId('login-username');
+      const passwordField = screen.getByTestId('login-password');
+
+      await user.type(usernameField, 'admin');
+      await user.type(passwordField, 'admin');
+
+      const submitButton = screen.getByTestId('login-submit');
+      await user.click(submitButton);
```
dottorblaster, line 107: are you sure you can't get those fields by _any_ other query?

dottorblaster (changes requested):
Some changes but overall really a good job

assets/js/components/Login.test.jsx
CDimonaco: I think I can get them with Id certainly, do you think it's better?

assets/js/components/Layout/Layout.jsx
```diff
@@ -65,14 +67,17 @@ function Layout() {
...
   const sidebarIconColor = 'currentColor';
   const sidebarIconClassName = 'text-gray-400 hover:text-gray-300';
   const sidebarIconSize = '24';
 
+  const logout = (e) => {
+    e.preventDefault();
+
+    clearCredentialsFromStore();
+    window.location.href = '/session/new';
+  };
```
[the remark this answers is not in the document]
CDimonaco: done

assets/js/components/Login.test.jsx
```diff
@@ -0,0 +1,133 @@
...
+        matches: false,
+        media: query,
+        onchange: null,
+        addEventListener: jest.fn(),
+        removeEventListener: jest.fn(),
+        dispatchEvent: jest.fn(),
+      })),
+    });
+  });
+  it('should redirect to the / path, when the user is already logged in', async () => {
```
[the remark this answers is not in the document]
CDimonaco: idk formatter goes this way 🤣

arbulu89 (approved):
Here we go! Thanks

nelsonkopliku (approved):

dottorblaster (approved):

stefanotorresi:
excellent work, @CDimonaco !

PR #1056, opened 2022-12-15
Transition start execution running

arbulu89:
Add transition since the execution is started until it is received from wanda. It uses a "fake it until make it" pattern which fakes the redux state setting an execution in a running state.
Now on, all wanda related views are under `clusters_new/${clusterID}` endpoint. This is done with the `usingWanda` boolean variable

demo

PD: I have fixed many small (and not so small) details on the way
PD2: The gif is a bit faked, as in order to work we need to merge https://github.com/trento-project/web/pull/1032 first

### DIFF
--- a/assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
@@ -158,7 +158,7 @@ function ChecksSelectionNew({ clusterId, cluster }) {
           {localSavingSuccess && selectedChecks.length > 0 && (
             <SuggestTriggeringChecksExecutionAfterSettingsUpdated
               clusterId={clusterId}
-              usingWanda={true}
+              usingWanda
               onClose={() => setLocalSavingSuccess(null)}
             />
           )}
--- a/assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
@@ -158,6 +158,7 @@ function ChecksSelectionNew({ clusterId, cluster }) {
           {localSavingSuccess && selectedChecks.length > 0 && (
             <SuggestTriggeringChecksExecutionAfterSettingsUpdated
               clusterId={clusterId}
+              usingWanda={true}
               onClose={() => setLocalSavingSuccess(null)}
             />
           )}
--- a/assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
@@ -158,7 +158,7 @@ function ChecksSelectionNew({ clusterId, cluster }) {
           {localSavingSuccess && selectedChecks.length > 0 && (
             <SuggestTriggeringChecksExecutionAfterSettingsUpdated
               clusterId={clusterId}
-              usingWanda
+              usingNewChecksEngine
               onClose={() => setLocalSavingSuccess(null)}
             />
           )}
--- a/assets/js/components/TriggerChecksExecutionRequest/TriggerChecksExecutionRequest.test.jsx
+++ b/assets/js/components/TriggerChecksExecutionRequest/TriggerChecksExecutionRequest.test.jsx
@@ -2,47 +2,43 @@ import React from 'react';
 
 import { screen, act } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
+import { faker } from '@faker-js/faker';
 import { withState, renderWithRouter } from '@lib/test-utils';
+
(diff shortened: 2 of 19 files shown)


assets/js/components/TriggerChecksExecutionRequest/TriggerChecksExecutionRequest.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+describe('TriggerChecksExecutionRequest', () => {
+  it('should dispatch execution requested on click', async () => {
+    const user = userEvent.setup();
+
+    const initialState = {
+      clustersList: {
+        clusters: [
+          {
+            id: 'cluster1',
+            selected_checks: ['check1', 'check3'],
```
dottorblaster, line 18: Can we use faker for these?

dottorblaster (changes requested):
Just a tiny comment on a test then we can merge

nelsonkopliku (approved):

jamie-suse (approved):

rtorrero (approved):

assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
```diff
@@ -158,6 +158,7 @@ function ChecksSelectionNew({ clusterId, cluster }) {
           {localSavingSuccess && selectedChecks.length > 0 && (
             <SuggestTriggeringChecksExecutionAfterSettingsUpdated
               clusterId={clusterId}
+              usingWanda
```
dottorblaster, line 161: Can we avoid referencing the raw name of the service? something like `usingNewChecksEngine` would be fine.

Also, we use the same pattern in the router. Having a shared name for all these props would allow us for a simpler `grep` in the future when we have to remove all of these switches.

assets/js/state/actions/lastExecutions.js
```diff
@@ -1,6 +1,12 @@
...
 
 export const updateLastExecution = (groupID) => ({
   type: UPDATE_LAST_EXECUTION,
   payload: { groupID },
 });
+
+export const executionRequested = (clusterID, hosts, checks) => ({
+  type: EXECUTION_REQUESTED,
+  payload: { clusterID, hosts, checks },
+});
```
dottorblaster, line 12: Maybe just `requestExecution`?

dottorblaster (changes requested):
Sorry, a couple things more :sweat_smile:

assets/js/state/selectors/cluster.test.js
```diff
@@ -0,0 +1,52 @@
+import { getClusterHostIDs, getClusterSelectedChecks } from './cluster';
+
+describe('Cluster selector', () => {
+  it('should return the cluster hosts IDs', () => {
+    const state = {
+      hostsList: {
+        hosts: [
+          {
+            id: 'id1',
```
EMaksy, line 9: Can faker and factories be used here ?

assets/js/components/TriggerChecksExecutionRequest/TriggerChecksExecutionRequest.test.jsx
arbulu89: Done: https://github.com/trento-project/web/pull/1056/commits/57a90f03cab3f41f78f0cdaa467cc8a5b3db4b3c

EMaksy (approved):
Hey nice!
Looks good to me :D 
Just a small note about  the cluster test in a comment.

assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
arbulu89: I replaced all the places where we have this pattern by this `usingWanda`, so there is not any other wording.
This variable will be removed soonish, and I preferred the wanda wording, as it makes it cristal clear what it does.
But well, I will check if you prefer the other way around

assets/js/state/actions/lastExecutions.js
arbulu89: Let's avoid this discussion now. We already have a tech debt ticket to talk about present/past term in the actions/events, etc

assets/js/components/ClusterDetails/ChecksSelectionNew.jsx
dottorblaster: It's ok :+1:

arbulu89: https://github.com/trento-project/web/pull/1056/commits/e1b59b597e937fc9034c38de6a484fed3e21e684

dottorblaster (approved):
LGTM

PR #1064, opened 2022-12-19
Add filter to Execution Results page

jamie-suse:
# Description

This change adds a filter to the new Execution Results page, the same as the current Checks Results page.

## How was this tested?

Adjusted `ExecutionResults.test.jsx` to accommodate for the addition of the filter.


dottorblaster (changes requested):
We can't use the old hook, we have to have a new one.

dottorblaster (approved):
LGTM

PR #1071, opened 2022-12-21
Add ChecksSelectionHints in the ExecutionResults

EMaksy:
# Description

A selected cluster should show a check selection hints, in the "Show Results" view,  when no tests were executed and selected.

## How was this tested?

Added test for ChecksSelectionHints.
Additional test will be added later.


assets/js/components/ChecksResults/ChecksSelectionHints.jsx
```diff
@@ -36,7 +39,11 @@ function ChecksSelectionHints({ clusterId, selectedChecks }) {
             <Button
               className="bg-waterhole-blue mx-auto px-2 py-2 w-1/4 xs:w-full"
               onClick={() => {
-                navigate(`/clusters/${clusterId}/settings`);
+                navigate(
+                  usingNewChecksEngine
+                    ? `/clusters_new/${clusterId}/settings`
+                    : `/clusters/${clusterId}/settings`
+                );
```
arbulu89, line 46: Here in this file, below, we need to add the `usingNewChecksEngine` to the `TriggerChecksExecutionRequest` component

assets/js/components/ChecksResults/ChecksSelectionHints.test.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
```
arbulu89, line 1: I wouldn't create this test. As it is now, this logic will be removed soon and it simply tests the difference in the url.
If we want to check this component (which would be out of scope for this PR), we should test something else

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -121,9 +122,10 @@ function ExecutionResults({
       <ResultsContainer
         catalogError={false}
         clusterID={clusterID}
-        hasAlreadyChecksResults
+        hasAlreadyChecksResults={!!(executionData || executionLoading)}
         selectedChecks={checkResults}
         onCatalogRefresh={onCatalogRefresh}
+        usingNewChecksEngine="true"
```
arbulu89, line 128: Change this by simply `usingNewChecksEngine`. That `"true"` is not even a real boolean hehe

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -121,9 +122,10 @@ function ExecutionResults({
       <ResultsContainer
         catalogError={false}
         clusterID={clusterID}
-        hasAlreadyChecksResults
+        hasAlreadyChecksResults={!!(executionData || executionLoading)}
```
arbulu89, line 125: The next line `selectedChecks`, we would need to change it.
This way, you are using current check results, but that field is not for that, it is to get the currently selected checks.
To make it proper, we would need to get the selected checks in the `ExecutionResultsPage`, like `cluster?.selected_checks`, pass them to this view, and use them in that param

arbulu89 (changes requested):
Hey @EMaksy 
Added some comments.
Besides that, I think we need other kind of testing. We would need to add a new test to the `ExecutionResults` view, where we need to test the result when we give `executionData=null` and `executionLoading=false`. In this scenario, we need to test that we render the hint view, checking some basic text like `It looks like there is no recent execution for current cluster` or similar

assets/js/components/ExecutionResults/ExecutionResults.jsx
arbulu89: Yes, that's why you only need to put `usingNewChecksEngine` without anything as value:
```
<ResultsContainer
        catalogError={false}
        clusterID={clusterID}
        hasAlreadyChecksResults={!!(executionData || executionLoading)}
        usingNewChecksEngine
>
```

EMaksy: Ah it's always true then... 
Thanks!

assets/js/components/ChecksResults/ChecksSelectionHints.test.jsx
EMaksy: Oki i will remove this test

assets/js/components/ExecutionResults/ExecutionResults.jsx
EMaksy: Done that,  ExecutionResultPage passes me the selected checks to the ExecutionResults

assets/js/components/ChecksResults/ChecksSelectionHints.jsx
EMaksy: You mean to pass usingNewChecksEngine as prop to TriggerChecksExecutionRequest? 
Or something else ?

arbulu89: Yes, exactly that

assets/js/components/ExecutionResults/ExecutionResultsPage.jsx
```diff
@@ -38,17 +37,13 @@ function ExecutionResultsPage() {
   if (!cluster) {
     return <div>Loading...</div>;
   }
+  const clusterSelectedChecks = cluster?.selected_checks;
```
arbulu89, line 40: I don't think we need a new constant here. Simply pass `cluster?.selected_checks` to `clusterSelectedChecks` as we do with other props

EMaksy: Oki yeah!

EMaksy: Yeah sure :D

arbulu89 (approved):
Thanks @EMaksy 
Fantastic!

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { screen } from '@testing-library/react';
-
+import '@testing-library/jest-dom/extend-expect';
```
dottorblaster, line 3: Could you move this to the bottom of these imports?

dottorblaster (changes requested):
Just a tiny thing then LGTM

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
EMaksy: Sure :zap:

dottorblaster (approved):
LGTM

PR #1073, opened 2022-12-22
Fix Checks results filters

jamie-suse:
# Description

Fixes changes to the new checks results filters from bd2f538.

Fixes [#1064](https://github.com/trento-project/web/pull/1064)

filters

## How was this tested?

Added new tests for filtering in `ExecutionResults.test.jsx`.

### DIFF
--- a/assets/js/components/ExecutionResults/ExecutionResults.jsx
+++ b/assets/js/components/ExecutionResults/ExecutionResults.jsx
@@ -133,43 +133,39 @@ function ExecutionResults({
         onCatalogRefresh={onCatalogRefresh}
         usingNewChecksEngine
       >
-        {executionData?.targets.map(({ agent_id: hostID, checks }) => {
-          const filteredChecks = filterChecks(checks, predicates);
+        {executionData?.targets.map(({ agent_id: hostID, checks }) => (
+          <HostResultsWrapper
+            key={hostID}
+            hostname={hostnames.find(({ id }) => hostID === id)?.hostname}
+          >
+            {filterChecks(checks, predicates).map((checkID) => {
+              const { health, error, expectations, failedExpectations } =
+                getCheckHealthByAgent(checkResults, checkID, hostID);
 
-          return (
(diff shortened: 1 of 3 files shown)


assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -138,32 +136,57 @@ function ExecutionResults({
...
-                  }}
-                />
-              );
-            })}
+                return {
+                  checkID,
+                  error,
+                  expectations,
+                  failedExpectations,
+                  result: health,
```
arbulu89, line 149: Why this change from `health` to `result`?
We use `health` all around the frontend code.
Is it because the `predicate` function expects that?

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -206,4 +206,88 @@ describe('ExecutionResults', () => {
...
+        catalog={catalog}
+        catalogError={error}
+        executionLoading={executionLoading}
+        executionData={executionData}
+        executionError={executionError}
+      />,
+      { route: `/clusters_new/${clusterID}/executions/last?health=passing` }
+    );
+
+    expect(screen.getByText('test-cluster')).toBeTruthy();
```
arbulu89, line 240: I don't think we need all of these expects, they are already tested somewhere else.
I would only use the last 3 ones, which are the good ones for this scenario.
The same for the next test

arbulu89 (approved):
Hey @jamie-suse ,
The code looks great! Thank you for adding tests ;)
I would appreciate a gif to see it working

assets/js/components/ExecutionResults/ExecutionResults.jsx
jamie-suse: Yep that's correct, `predicate()` relies on `result` being present.

dottorblaster (approved):
LGTM

PR #1094, opened 2023-01-09
Remove connection settings when using wanda

nelsonkopliku:
# Description

This keeps only the checks selection when in Wanda context.
As a proposal, page title has also been changed from `Cluster Settings for <cluster-name>` to `Checks Selection for <cluster-name>`.

image

## How was this tested?

Added a test case in `ChecksSelectionNew.test.jsx`

### DIFF
--- a/assets/js/components/ClusterDetails/ChecksSelectionNew.test.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelectionNew.test.jsx
@@ -29,9 +29,9 @@ describe('ClusterDetails ChecksSelectionNew component', () => {
 
     await act(async () => renderWithRouter(statefulChecksSelection));
 
-    const groupItem = await waitFor(() => screen.getByRole('heading'));
+    const groupItem = await waitFor(() => screen.getByText(group));
 
-    await user.click(groupItem.parentNode);
+    await user.click(groupItem);
     let switches = screen.getAllByRole('switch');
 
     expect(switches[0]).not.toBeChecked();
@@ -73,9 +73,9 @@ describe('ClusterDetails ChecksSelectionNew component', () => {
 
     await act(async () => renderWithRouter(statefulChecksSelection));
 
-    const groupItem = await waitFor(() => screen.getByRole('heading'));
+    const groupItem = await waitFor(() => screen.getByText(group));
 
-    await user.click(groupItem.parentNode);
+    await user.click(groupItem);
     let switches = screen.getAllByRole('switch');
 
     expect(switches[0]).toBeChecked();
@@ -117,7 +117,9 @@ describe('ClusterDetails ChecksSelectionNew component', () => {
 
     await waitFor(() => screen.getAllByRole('heading'));
 
-    const saveButton = screen.getByRole('button');
+    const saveButton = screen.getByRole('button', {
+      name: 'Select Checks for Execution',
+    });
     await user.click(saveButton);
 
     const actions = store.getActions();
--- a/assets/js/components/ClusterDetails/ClusterSettings.jsx
+++ b/assets/js/components/ClusterDetails/ClusterSettings.jsx
@@ -8,7 +8,6 @@ import classNames from 'classnames';
 import BackButton from '@components/BackButton';
 import { Tab } from '@headlessui/react';
 import { ChecksSelection } from '@components/ClusterDetails/ChecksSelection';
-import ChecksSelectionNew from '@components/ClusterDetails/ChecksSelectionNew';
 import { getCluster } from '@state/selectors';
 import TriggerChecksExecutionRequest from '@components/TriggerChecksExecutionRequest';
 import { getClusterName } from '@components/ClusterLink';
@@ -18,15 +17,13 @@ import { truncatedClusterNameClasses } from './ClusterDetails';
 
 export const UNKNOWN_PROVIDER = 'unknown';
 
-export function ClusterSettings({ usingNewChecksEngine = false }) {
+export function ClusterSettings() {
   const { clusterID } = useParams();
 
   const cluster = useSelector(getCluster(clusterID));
 
   const tabsSettings = {
-    'Checks Selection': usingNewChecksEngine ? (
-      <ChecksSelectionNew clusterId={clusterID} cluster={cluster} />
-    ) : (
+    'Checks Selection': (
       <ChecksSelection clusterId={clusterID} cluster={cluster} />
     ),
     'Connection Settings': (
@@ -40,13 +37,7 @@ export function ClusterSettings({ usingNewChecksEngine = false }) {
 
   return (
     <div className="w-full px-2 sm:px-0">
-      <BackButton
-        url={
-          usingNewChecksEngine
-            ? `/clusters_new/${clusterID}`
-            : `/clusters/${clusterID}`
-        }
-      >
+      <BackButton url={`/clusters/${clusterID}`}>
         Back to Cluster Details
       </BackButton>
       <div className="flex mb-2">
--- a/assets/js/components/ClusterDetails/ClusterSettingsNew.jsx
+++ b/assets/js/components/ClusterDetails/ClusterSettingsNew.jsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import { useParams } from 'react-router-dom';
+
+import { getCluster } from '@state/selectors';
+
+import BackButton from '@components/BackButton';
+import { getClusterName } from '@components/ClusterLink';
+import { ClusterInfoBox } from '@components/ClusterDetails';
+import WarningBanner from '@components/Banners/WarningBanner';
+import ChecksSelectionNew from './ChecksSelectionNew';
+import { UNKNOWN_PROVIDER } from './ClusterSettings';
+
+export function ClusterSettingsNew() {
+  const { clusterID } = useParams();
+
+  const cluster = useSelector(getCluster(clusterID));
+
+  if (!cluster) {
+    return <div>Loading...</div>;
+  }
+
+  return (
+    <div className="w-full px-2 sm:px-0">
+      <BackButton url={`/clusters_new/${clusterID}`}>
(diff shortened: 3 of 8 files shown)


assets/js/components/ClusterDetails/ClusterSettings.jsx
```diff
@@ -51,54 +40,83 @@ export function ClusterSettings({ usingNewChecksEngine = false }) {
...
+          >
+            {tabContent}
+          </Tab.Panel>
+        ))}
+      </Tab.Panels>
+    </Tab.Group>
+  );
+}
+
+export function InfoBox({ haScenario, provider }) {
```
arbulu89, line 108: I wouldn't create a new exported `InfoBox` function in a component that will be deleted soonish.
I would replicate this if needed in the `ChecksSelectionNew` file, so later on the removal is seamless

arbulu89 (commented):
I'm a bit lost with the changes.
If the intention is to use the `ChecksSelectionNew` directly when we go to the settings i wouldn't touch the `ClusterSettings` code, I would directly render the selection page in the `trento.jsx` file:
```
<Route
  path="clusters_new/:clusterID/settings"
  element={<CheckSelectionNew />}
 />
```

We can even think of changing the path from `settings` to something else.
The setting component looks like it will be removed soon, so I wouldn't do changes there if possible

dottorblaster (changes requested):
I totally agree with the approach @arbulu89 is suggesting, do see any potential issue in that?

assets/js/components/ClusterDetails/ChecksSelectionNew.test.jsx
```diff
@@ -117,7 +144,7 @@ describe('ClusterDetails ChecksSelectionNew component', () => {
 
     await waitFor(() => screen.getAllByRole('heading'));
 
-    const saveButton = screen.getByRole('button');
+    const saveButton = screen.getByTestId('save-selection-button');
```
dottorblaster, line 147: Did the previous query work the same or you replaced `getByRole` with `getByTestId` because it actually didn't work?

Keep in mind that:

> In the spirit of [the guiding principles](https://testing-library.com/docs/guiding-principles), it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible.

https://testing-library.com/docs/queries/bytestid/

nelsonkopliku:
Alright, I followed the suggestion with the only difference that instead of binding directly `CheckSelectionNew` to the route, I added an intermediate container Component `ClusterSettingsNew` for consistency with what we have now (yes we can rethink the structure there) and to keep testing simple.

How does this work?

assets/js/components/ClusterDetails/ClusterInfoBox.jsx
```diff
@@ -11,7 +11,10 @@ const haScenarioToString = {
 // eslint-disable-next-line import/prefer-default-export
 export function ClusterInfoBox({ haScenario, provider }) {
   return (
-    <div className="tn-cluster-details w-full my-4 mr-4 bg-white shadow rounded-lg px-8 py-4">
+    <div
+      data-testid="cluster-info-box"
```
dottorblaster, line 15: I think you can strip this and use `.getByText`

assets/js/components/ClusterDetails/ChecksSelectionNew.test.jsx
```diff
@@ -11,6 +11,29 @@ import { catalogCheckFactory, clusterFactory } from '@lib/test-utils/factories';
...
+    const [statefulChecksSelection] = withState(
+      <ChecksSelectionNew clusterId={cluster.id} cluster={cluster} />,
+      initialState
+    );
+
+    renderWithRouter(statefulChecksSelection);
+
+    const clusterInfoBox = screen.getByTestId('cluster-info-box');
+
+    expect(clusterInfoBox).toBeVisible();
```
dottorblaster, line 32: `expect(screen.getByText('Provider')).toBeVisible()`

assets/js/components/ClusterDetails/ChecksSelectionNew.test.jsx
```diff
@@ -11,6 +11,29 @@ import { catalogCheckFactory, clusterFactory } from '@lib/test-utils/factories';
...
+    expect(clusterInfoBox).toHaveTextContent('Azure');
```
dottorblaster, line 33:
```suggestion
    expect(screen.getByText('Azure')).toBeVisible();
```

assets/js/components/ClusterDetails/ChecksSelectionNew.test.jsx
```diff
@@ -11,6 +11,29 @@ import { catalogCheckFactory, clusterFactory } from '@lib/test-utils/factories';
...
+    expect(screen.getByTestId('catalog-container')).toBeVisible();
```
dottorblaster, line 34:
```suggestion

```

assets/js/components/ClusterDetails/ChecksSelectionNew.test.jsx
```diff
@@ -11,6 +11,29 @@ import { catalogCheckFactory, clusterFactory } from '@lib/test-utils/factories';
...
+      clusterChecksSelection: {},
+    };
+    const [statefulChecksSelection] = withState(
+      <ChecksSelectionNew clusterId={cluster.id} cluster={cluster} />,
+      initialState
+    );
+
+    renderWithRouter(statefulChecksSelection);
+
+    const clusterInfoBox = screen.getByTestId('cluster-info-box');
```
dottorblaster, line 30:
```suggestion

```

assets/js/components/ClusterDetails/ClusterSettingsNew.jsx
```diff
@@ -0,0 +1,14 @@
...
+import { getCluster } from '@state/selectors';
+import ChecksSelectionNew from './ChecksSelectionNew';
+
+export function ClusterSettingsNew() {
+  const { clusterID } = useParams();
+
+  const cluster = useSelector(getCluster(clusterID));
+
+  return <ChecksSelectionNew clusterId={clusterID} cluster={cluster} />;
+}
```
dottorblaster, line 14: Good job, we should move all the dispatches and the selectors here but I think it's something to be done in a second moment

dottorblaster (changes requested):
Much better now! I'm sure the strangling of the old part will be much easier :sweat_smile: 

Some more stuff to take care off then we are ok I think :+1:

assets/js/components/ClusterDetails/ClusterSettingsNew.jsx
arbulu89: To be honest, I find this component not really useful, as it doesn't do anything. If we at least ported here all the dispatches, etc to make the testing of the check selection page more pure.
Or bring here the back button, the info box and all the "container" things.

What is the "functional" logic of this component, is we don't even have "settings" anymore?

assets/js/components/ClusterDetails/index.js
```diff
@@ -7,3 +7,5 @@ export { ClusterInfoBox } from './ClusterInfoBox';
 
 export default ClusterDetails;
 export { ClusterDetailsNew } from './ClusterDetailsNew';
+export { ClusterSettingsNew } from './ClusterSettingsNew';
+export { truncatedClusterNameClasses } from './ClusterDetails';
```
arbulu89, line 11: Why do we need to export the `truncatedClusterNameClasses` object?
We were not exporting before, and it is not used anywhere else outside of this component

assets/js/components/ClusterDetails/ChecksSelectionNew.test.jsx
nelsonkopliku: Thanks for the suggestions! Adjusted to the following
```js
expect(screen.getByText('Provider')).toBeVisible();
expect(screen.getByText('Azure')).toBeVisible();
```
and in order to assert that the checks catalog is there, I also traded `expect(screen.getByTestId('catalog-container')).toBeVisible();` for
```js
expect(
  screen.getByRole('button', { name: 'Select Checks for Execution' })
).toBeVisible();
```
Does it work?

assets/js/components/ClusterDetails/ClusterSettingsNew.test.jsx
```diff
@@ -0,0 +1,45 @@
...
+    const group = faker.animal.cat();
+    const catalog = catalogCheckFactory.buildList(2, { group });
+
+    const [StatefulChecksSettingsNew, state] = withExtendedDefaultState(
+      <ClusterSettingsNew />,
+      { catalogNew: { loading: false, data: catalog, error: null } }
+    );
+
+    const {
+      clusters: [, , , { id: clusterId }],
```
arbulu89, line 24: `clusterId` -> `clusterID`

arbulu89 (approved):
Great job! Thanks @nelsonkopliku

assets/js/lib/test-utils/index.jsx
```diff
@@ -49,6 +50,12 @@ export const renderWithRouter = (ui, { route = '/' } = {}) => {
   };
 };
 
+export const withExtendedDefaultState = (component, additionalState = {}) =>
+  withState(component, {
+    ...defaultInitialState,
+    ...additionalState,
+  });
+
```
dottorblaster, line 58: I'd rather `export defaultInitialState` and then compose it using `withState` inside the test

dottorblaster (changes requested):
One more change then we are ok to merge :+1:

assets/js/lib/test-utils/index.jsx
nelsonkopliku: Initially I started like that and then it looked natural following the same `with...` pattern, to keep the same level of encapsulation for `defaultInitialState`, composability and to reduce a bit the boiler of merging the default state with a piece of extra state when needed.
However since you prefer that, I changed as per the suggestion so we can proceed.

dottorblaster (approved):

PR #1098, opened 2023-01-11
Allow getCatalog wanda call to filter by provider

rtorrero:
# Description

This PR adds a `provider` search param to wanda's getCatalog call. This is particularly useful in the checks selection view, as we might want to limit the display of checks that can't be applied to certain provider.

## How was this tested?

Manually by running wanda and checking that the checks selection view causes the search param to be added to the request while the catalog view doesn't.


assets/js/lib/api/wanda.js
```diff
@@ -12,4 +12,10 @@ export const getExecutionResult = (executionID) =>
...
   wandaClient.get(`/api/checks/groups/${groupID}/executions/last`);
 
-export const getCatalog = () => wandaClient.get('/api/checks/catalog');
+export const getCatalog = (provider = "") => {
+  if (provider === "") {
+    return wandaClient.get('/api/checks/catalog');
+  }
+
+  return wandaClient.get(`/api/checks/catalog?provider=${provider}`);
+};
```
dottorblaster, line 21:
```suggestion
export const getCatalog = (env) => wandaClient.get(`/api/checks/catalog?provider=${provider}`, { params: env });
```

assets/js/state/sagas/catalog.js
```diff
@@ -8,10 +8,12 @@ import {
...
 } from '@state/catalogNew';
 
-export function* updateCatalog() {
+export function* updateCatalog({ payload }) {
+  const { provider } = payload;
+  
   yield put(setCatalogLoading());
   try {
-    const { data } = yield call(getCatalog);
+    const { data } = yield call(getCatalog, provider);
```
dottorblaster, line 16:
```suggestion
    const { data } = yield call(getCatalog, payload);
```

assets/js/state/sagas/catalog.js
```diff
@@ -8,10 +8,12 @@ import {
   setCatalogError,
 } from '@state/catalogNew';
 
-export function* updateCatalog() {
+export function* updateCatalog({ payload }) {
+  const { provider } = payload;
```
dottorblaster, line 12:
```suggestion

```

assets/js/state/actions/catalog.js
```diff
@@ -1,6 +1,8 @@
 export const UPDATE_CATALOG = 'UPDATE_CATALOG_NEW';
 
-export const updateCatalog = () => ({
+export const updateCatalog = ({ provider} ) => ({
```
dottorblaster, line 3:
```suggestion
export const updateCatalog = (env ) => ({
```

assets/js/state/actions/catalog.js
```diff
@@ -1,6 +1,8 @@
...
   type: UPDATE_CATALOG,
-  payload: {},
+  payload: {
+    provider,
+  },
```
dottorblaster, line 7:
```suggestion
  payload: env,
```

dottorblaster (changes requested):
Overall a good job, I made some suggestions so that we end basically collecting an `env` object that we end up passing to the request. I'm uncertain about the suggestion I made on `assets/js/state/actions/catalog.js` because actually having a RORO there would ensure some extra parsing.

Ref about RORO: https://www.freecodecamp.org/news/elegant-patterns-in-modern-javascript-roro-be01e7669cbd/

dottorblaster (approved):
LGTM

PR #1105, opened 2023-01-12
Fix e2e

dottorblaster:
Turns out opening a dropdown, selecting an item and then reopening it right after makes an end to end test extremely prone to flakiness.

In this PR we try to use the clear function of a filter, that should avoid the second dropdown interaction, therefore making the test more reliable.

### DIFF
--- a/assets/js/components/Table/Filter.jsx
+++ b/assets/js/components/Table/Filter.jsx
@@ -21,12 +21,14 @@ function Filter({ options, title, value, onChange }) {
     .filter((option) => option.toLowerCase().includes(query.toLowerCase()));
 
   useOnClickOutside(ref, () => setOpen(false));
+
   return (
     <div className="w-64 w-72 top-16 mr-4" ref={ref}>
       <div className="mt-1 relative">
         {value !== '' && (
           <button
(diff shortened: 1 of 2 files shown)


fabriziosestito:
best pr 2023

assets/js/components/Table/Filter.jsx
```diff
@@ -21,12 +21,14 @@ function Filter({ options, title, value, onChange }) {
...
 
   useOnClickOutside(ref, () => setOpen(false));
+
   return (
     <div className="w-64 w-72 top-16 mr-4" ref={ref}>
       <div className="mt-1 relative">
         {value !== '' && (
           <button
             type="button"
+            data-testid={`filter-${title}-clear`}
```
fabriziosestito, line 31: will we remember to remove testid related stuff?

fabriziosestito (approved):

nelsonkopliku (approved):

assets/js/components/Table/Filter.jsx
dottorblaster: Right, I'll create a card for that

PR #1110, opened 2023-01-13
Rollup fixes

fabriziosestito:
This PR fixes the rollup flow since it could happen that a snapshot is built concurrently with the rollup execution, causing the snapshot to be taken with a lock.
It also fixes the aggregate lifespan that wasn't stopping the aggregate correctly if the aggregate is locked due to a bad pattern match.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -67,10 +67,6 @@ config :trento, Trento.Commanded,
       snapshot_every: 200,
       snapshot_version: 1
     ],
-    Trento.Domain.Cluster => [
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):
:eyes:

dottorblaster (approved):
LGTM

CDimonaco (approved):
LGTM

arbulu89 (approved):
I hope we remember this once we add other aggregate rollup code

PR #1115, opened 2023-01-16
Add premium pill to the check selection view

arbulu89:
Add premium flag/pill to the catalog and to the check selection view (temporary `premium` flag added to a random check to do some more manual testing)

image

image

### DIFF
--- a/assets/js/components/PremiumPill.jsx
+++ b/assets/js/components/PremiumPill.jsx
@@ -1,10 +1,16 @@
 import React from 'react';
+import classNames from 'classnames';
 
-function PremiumPill() {
+import Pill from '@components/Pill';
+
+function PremiumPill({ className }) {
   return (
-    <p className="ml-1 px-2 inline-flex text-xs leading-5 font-semibold rounded-full bg-green-100 text-green-800">
+    <Pill
+      className={classNames(className, 'bg-green-100 text-green-800')}
+      size="xs"
+    >
       Premium
-    </p>
+    </Pill>
   );
 }
 
--- a/assets/js/components/PremiumPill.jsx
+++ b/assets/js/components/PremiumPill.jsx
@@ -1,0 +1,11 @@
+import React from 'react';
+
+function PremiumPill() {
+  return (
+    <p className="ml-1 px-2 inline-flex text-xs leading-5 font-semibold rounded-full bg-green-100 text-green-800">
+      Premium
+    </p>
+  );
+}
+
+export default PremiumPill;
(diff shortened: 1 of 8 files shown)


CDimonaco (approved):
LGTM

assets/js/lib/test-utils/factories/index.js
```diff
@@ -39,6 +39,7 @@ export const catalogCheckFactory = Factory.define(() => ({
   group: faker.animal.cat(),
   description: faker.lorem.paragraph(),
   remediation: faker.lorem.paragraph(),
+  premium: false,
```
dottorblaster, line 42: Maybe having this as a `true | false` could help randomizing?

assets/js/components/PremiumPill.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+function PremiumPill() {
+  return (
+    <p className="ml-1 px-2 inline-flex text-xs leading-5 font-semibold rounded-full bg-green-100 text-green-800">
+      Premium
+    </p>
```
dottorblaster, line 7: I'm noticing that this markup is pretty much identical to the one used to declare the check ID, maybe we could have a component for both purposes?

assets/js/components/PremiumPill.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+function PremiumPill() {
```
dottorblaster, line 3: We are missing `className` as a prop here. Keep in mind that having `className` as a prop always allows for better composability into the DOM

(for example having more of these inline could allow us to specify individual margins or similar stuff)

dottorblaster (changes requested):

assets/js/components/PremiumPill.jsx
CDimonaco: Nice catch, binding classNames with a prop is always a good Idea +1

arbulu89: Done

assets/js/components/PremiumPill.jsx
arbulu89: Hey @dottorblaster ,
I have modified the `PremiumPill` to use the default `Pill` component.
I modified the `Pill` component to accept a `size` prop, which is not nothing more than more options to add classnames.
I didn't want to have a hardcoded map, as sizing is always really cusotmizable.
What do you think?

EMaksy (approved):
LGTM!

nelsonkopliku (approved):

assets/js/components/PremiumPill.jsx
dottorblaster: I would prefer the hardcoded map because paddings and sizes should be a little bit opinionated, but your call :+1:

dottorblaster (approved):

assets/js/components/PremiumPill.jsx
arbulu89: Ok. I will go with the map. If we need further customization we will have a look at that point

PR #1121, opened 2023-01-17
Fix prettier checker

dottorblaster:
What the heck? We forgot `--check` somewhere but we had this problem today. When in doubt, just fix it


nelsonkopliku (approved):
:shipit:

PR #1125, opened 2023-01-18
Fix start execution interaction

nelsonkopliku:
# Description

Currently if a request to `POST /api/clusters/:cluster_id/checks/request_execution` fails the error is not caught on the client causing the webapp to crash.

<details>
  <summary>See gif</summary>
  
  failing-request-execution-bug
</details>

The agreed behaviour is that a user gets redirected to the checks execution detail only if requesting the execution to start succeeds, otherwise an error notification is shown and the user stays in the page from where the action was triggered.

<details>
  <summary>On failed execution start</summary>
  
  failed-execution-start
</details>
<details>
  <summary>On successful execution start</summary>
  
  successful-execution-start
</details>

Outline of the changes:
- used [react context](https://beta.reactjs.org/learn/passing-data-deeply-with-context) to simplify passing props down the components tree
- moved relevant selectors/actions/sagas in appropriate selector/action/saga location
- minor cleanupz

## How was this tested?

Manually + existing automated tests + added needed tests

### DIFF
--- a/assets/js/components/ChecksExecutionContext.jsx
+++ b/assets/js/components/ChecksExecutionContext.jsx
@@ -1,0 +1,29 @@
+import React, { createContext, useCallback } from 'react';
+import { useDispatch } from 'react-redux';
+import { Outlet } from 'react-router-dom';
+
+import { executionRequested } from '@state/actions/lastExecutions';
+
+export const ChecksExecutionContext = createContext(() => {});
+
+export default function ChecksExecutionContextProvider({
+  handleExecutionStart,
+}) {
+  const dispatch = useDispatch();
+
+  const startExecution =
+    typeof handleExecutionStart !== 'undefined'
+      ? handleExecutionStart
+      : useCallback(
+          (clusterId, hostList, selectedChecks) => {
+            dispatch(executionRequested(clusterId, hostList, selectedChecks));
+          },
+          [dispatch]
+        );
+
+  return (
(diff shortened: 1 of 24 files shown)


dottorblaster (changes requested):
A couple things then we can merge :+1: 

- Besides the bug fix, I see a pretty big refactoring, could you split the refactoring and the bugfix in two separate PRs?
- Could you switch the approach from this one that leverages contexts to just using the `navigate` call inside the saga? It should be fine to have the `navigate` passed as a payload, or you can use an `history.push`-based approach.

nelsonkopliku:
Hey @dottorblaster, thanks for the review, I made some adjustments to the PR to only keep the necessary changes for the bugfix. I cannot go thinner than this :+1: 

Regarding the suggestion to not use a context: would you mind sharing more about the rationale please?

dottorblaster:
> Regarding the suggestion to not use a context: would you mind sharing more about the rationale please?

@nelsonkopliku absolutely! Two points overall:

- We tried to achieve having stateless components with logic handled through props instantiated inside a "controller" component, and we are using that pattern successfully;
- In this specific case having a context doesn't add much readability, looks less idiomatic, increases complexity and does exactly the same job of the controller component we already have in place, while bringing in dependencies that could be abstracted simply by refactoring the saga and bringing the `navigate` call to the upper layer.
- I deliberately chose to avoid contexts and custom hooks when kickstarting the project to leverage all the potential of sagas and redux shared/deterministic state.

PR #1126, opened 2023-01-18
Remove ssh address references

arbulu89:
# Description

First step on the legacy runner removal. Remove `ssh_address` field references.

**Important: once this is merged the checks with the legacy runner won't work anymore (the "demo" env will continue working as it fakes this communication)**

## How was this tested?

Nothing to be done


lib/trento/application/integration/checks/adapter/runner.ex
```diff
@@ -57,7 +57,7 @@ defmodule Trento.Integration.Checks.Runner do
         Enum.map(host_settings, fn host ->
           %{
             host_id: host.host_id,
-            address: host.ssh_address,
+            address: "",
```
arbulu89, line 60: Setting this empty by now. This complete function will be removed once we remove the connection settings code. I didn't want to do it here, to not overload the PR more

lib/trento_web/views/prometheus_view.ex
```diff
@@ -10,7 +10,7 @@ defmodule TrentoWeb.PrometheusView do
 
   def render("target.json", %{target: target}) do
     %{
-      targets: ["#{target.ssh_address}:#{@node_exporter_port}"],
+      targets: ["#{List.first(target.ip_addresses, target.hostname)}:#{@node_exporter_port}"],
```
arbulu89, line 13: Use the 1st obtained IP address (or the hostname as fallbak) as node exporter address. I think it should be enough, but if in the future we need more customization, we will need to add back some settings to be added by the user maybe

lib/trento/application/integration/checks/adapter/runner.ex
fabriziosestito: could you add a FIXME reminder?

lib/trento/domain/host/events/host_registered.ex
```diff
@@ -9,7 +9,6 @@ defmodule Trento.Domain.Events.HostRegistered do
     field :host_id, Ecto.UUID
     field :hostname, :string
     field :ip_addresses, {:array, :string}
```
fabriziosestito, line 11: We need to agree on when we will upcast or, in general, what we consider a breaking change. This is valid for API versioning but also event versioning (internal events and integration events).
My proposal is to go like this but track it somewhere.

lib/trento/application/integration/checks/adapter/runner.ex
arbulu89: The complete file will be gone once we finish removing the legacy runner usage. 
I'm actually already working on the removal in other branches. I don't think we will forget anyhow about this

lib/trento/domain/host/events/host_registered.ex
arbulu89: @fabriziosestito I will create a ticket in our backlog to discuss this.
I think that we might agree that in the particular scenario (removing a field), we don't have any breaking change, so upgrading the version would only be "informative"

lib/trento_web/views/prometheus_view.ex
fabriziosestito: Could we track this in a new card as well?

arbulu89: Done

fabriziosestito (approved):

PR #1129, opened 2023-01-19
Freeze demo deployment stages in ci

arbulu89:
Freeze demo deployment now as the transition from old runner to wanda has started.

In order to disable or enable the demo deployment, the repository variable `DEPLOY_DEMO` must be changed.
Set to `'true'` to enable. Everything else should disable it.
I already set it to `false`.

I don't know if we should document this somewhere, we don't really have any documentation about the CI or the demo environment by now

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -467,7 +467,7 @@ jobs:
   deploy-demo-env:
     name: Deploy updated images to the demo environment
     runs-on: self-hosted
-    if: github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
+    if: vars.DEPLOY_DEMO == 'true' && (github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch')
     env:
       IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}/trento-web
     needs: build-demo-img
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -467,7 +467,8 @@ jobs:
   deploy-demo-env:
     name: Deploy updated images to the demo environment
     runs-on: self-hosted
-    if: github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch'
+    if: ${{ false }}
```
fabriziosestito, line 470: we could use an actions variable, https://docs.github.com/en/actions/learn-github-actions/variables so we don't need to commit next time we want to activate/deactivate

arbulu89: Damn! Unfortunately the `env` value is not available in the `if` at `job` level: https://github.com/actions/runner/issues/1189
It can only be used in the `step` level.
Does it make sense to put it in the last step of the involved jobs?
It would work, but many the initial steps would be executed, and I don't want to put it everywhere...

arbulu89: Variables are accepted though...

fabriziosestito (approved):
Nice and clean!

nelsonkopliku (approved):

rtorrero:
Looks good to me if you want to merge it like this, but I'd suggest to change it so that it deploys by default and skips the demo deployment if a variable is present, e.g.:

`SKIP_DEMO_DEPLOYMENT`

PR #1132, opened 2023-01-23
Use JWT to authenticate /api/checks calls

fabriziosestito:
Adds missing JWT authentication to checks related API calls in the FE, by simply replacing the axios client.

### DIFF
--- a/assets/js/lib/api/wanda.js
+++ b/assets/js/lib/api/wanda.js
@@ -1,16 +1,13 @@
 import axios from 'axios';
+import { networkClient } from '@lib/network'
 
 const baseURL = process.env.WANDA_URL;
 
-export const wandaClient = axios.create({
-  baseURL,
-});
-
 export const getExecutionResult = (executionID) =>
-  wandaClient.get(`/api/checks/executions/${executionID}`);
+  networkClient.get(`/api/checks/executions/${executionID}`, { baseURL: baseURL });
 
 export const getLastExecutionByGroupID = (groupID) =>
-  wandaClient.get(`/api/checks/groups/${groupID}/executions/last`);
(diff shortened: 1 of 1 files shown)


arbulu89 (approved):
Nice trick!
So now it used the client which already has the token.
Let's test before merging hehe

assets/js/lib/api/wanda.js
```diff
@@ -1,16 +1,13 @@
...
 
 const baseURL = process.env.WANDA_URL;
 
-export const wandaClient = axios.create({
-  baseURL,
-});
-
 export const getExecutionResult = (executionID) =>
-  wandaClient.get(`/api/checks/executions/${executionID}`);
+  networkClient.get(`/api/checks/executions/${executionID}`, { baseURL: baseURL });
```
dottorblaster, line 7: You could declare a default config for all the calls and pass it over to functions or spread when needed

```js
const defaultConfig = { baseURL };

export const getCatalog = (env) =>
  networkClient.get(`/api/checks/catalog`, { ...defaultConfig, params: env });
```

dottorblaster (changes requested):
Just a tiny change then we can merge

CDimonaco:
@dottorblaster Nice catch, but I would choose another name, default config maybe is misleading, the default config for the client is in the network library, but idk

dottorblaster:
@CDimonaco I suck at naming things, I agree :smile:

fabriziosestito:
Closing as duplicate

assets/js/lib/api/wanda.js
fabriziosestito: @CDimonaco i am going with config since the doc states `axios.get(url[, config])`

PR #1133, opened 2023-01-23
Frontend wanda client with authentication interceptors

fabriziosestito:
Adds auth to wanda related API calls in the FE.

### DIFF
--- a/assets/js/lib/api/wanda.js
+++ b/assets/js/lib/api/wanda.js
@@ -1,14 +1,16 @@
 import { networkClient } from '@lib/network';
 
 const baseURL = process.env.WANDA_URL;
+const defaultConfig = { baseURL };
 
 export const getExecutionResult = (executionID) =>
-  networkClient.get(`/api/checks/executions/${executionID}`, { baseURL });
+  networkClient.get(`/api/checks/executions/${executionID}`, defaultConfig);
 
 export const getLastExecutionByGroupID = (groupID) =>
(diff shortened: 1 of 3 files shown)


assets/js/lib/api/wanda.js
```diff
@@ -1,16 +1,14 @@
...
 
 const baseURL = process.env.WANDA_URL;
 
-export const wandaClient = axios.create({
-  baseURL,
-});
-
 export const getExecutionResult = (executionID) =>
-  wandaClient.get(`/api/checks/executions/${executionID}`);
+  networkClient.get(`/api/checks/executions/${executionID}`, { baseURL });
```
nelsonkopliku, line 6: Instead of having to pass the `baseURL` on each call, we could export a function from `@lib/network` that returns an authenticated client on a given `baseURL` (aka service?)

So we'd need to only change how we build the `wandaClient`
```js
export const wandaClient = createAuthenticatedClient({ baseURL });
```
wdyt?

CDimonaco: The authenticated client should be a singleton for all the application, like now, we could achieve that with a little of closure magic ofc, but no hard feelings

Having a shared object in the Wanda client with the base URL option and spreading it through the calls like @dottorblaster suggested in the previous pr should be done

fabriziosestito: Addressed #1132 comments

nelsonkopliku (approved):
LGTM

dottorblaster (approved):
Go :+1:

PR #1134, opened 2023-01-23
Remove legacy catalog frontend

arbulu89:
# Description

Remove old catalog code usage. Even though the `git diff` looks confusing, I simply removed the old code and renamed the files with the `new` suffix removing it.

The unique real changes are in the e2e tests, which are now cleaner as I used factories (imported from the main assets folder). This way, we don't need more and more fixture files to test basic things even in the e2e tests.

PD: The changes `+/-` looks huge, but all of it comes from the `package-lock.json` of the e2e tests

## How was this tested?

As commented above, I have updated the e2e tests to use factories coming from the main frontend code. This helps us to have a high degree of customization, without the real need of adding more and more fixture files for small changes. This includes a dependency between the frontend tests and e2e tests, but I think it is worthy, as the factories should be correct for both cases

# Edit
The `hana_cluster_details` e2e tests has been skipped by now, as the transition creates incompatibilities with the code between new and legacy code. Besides that, many of those tests won't apply to the new execution flow and views
https://github.com/trento-project/web/pull/1142


test/e2e/cypress/e2e/checks_catalog.cy.js
```diff
@@ -1,45 +1,63 @@
-import {
-  availableChecks,
-  checkDataByProvider,
-} from '../fixtures/checks-catalog/available_checks';
+import { catalogCheckFactory } from '@lib/test-utils/factories';
+import { groupBy } from '@lib/lists';
 
 context('Checks catalog', () => {
+  const wandaCatalogURL = `**/api/checks/catalog`;
```
fabriziosestito, line 5: Let's avoid to use wanda naming

test/e2e/cypress/e2e/checks_catalog.cy.js
```diff
@@ -1,45 +1,63 @@
...
+
+  const group1 = catalogCheckFactory.buildList(2, { group: 'Group 1' });
+  const group2 = catalogCheckFactory.buildList(2, { group: 'Group 2' });
+  const group3 = catalogCheckFactory.buildList(2, { group: 'Group 3' });
+  const catalog = group1.concat(group2, group3);
+
   before(() => {
     cy.visit('/catalog');
     cy.url().should('include', '/catalog');
+    cy.intercept(wandaCatalogURL, {
```
fabriziosestito, line 15: wanda naming

test/e2e/cypress/e2e/checks_catalog.cy.js
```diff
@@ -1,45 +1,63 @@
...
   describe('Provider selection', () => {
-    checkDataByProvider.forEach((checkData, provider) => {
-      it(`should include the correct check data '${checkData} for provider '${provider}'`, () => {
+    [
+      ['aws', 'AWS', 2],
+      ['azure', 'Azure', 3],
+      ['gcp', 'GCP', 4],
+    ].forEach(([provider, label, checkCount]) => {
+      it(`should query the correct checks data filtered by provider ${label}`, () => {
+        cy.intercept(`${wandaCatalogURL}?provider=${provider}`, {
```
fabriziosestito, line 49: wanda naming

test/e2e/cypress/e2e/checks_catalog.cy.js
```diff
@@ -58,4 +76,13 @@ context('Checks catalog', () => {
         });
     });
   });
+
+  describe('Catalog error', () => {
+    it('should show an error notification if the catalog cannot be obtained', () => {
+      cy.intercept(wandaCatalogURL, { forceNetworkError: true });
```
fabriziosestito, line 82: wanda naming

fabriziosestito (changes requested):

fabriziosestito (approved):

PR #1136, opened 2023-01-23
Add a checks provider filter for catalog_new view

EMaksy:
# Description
Add a filter option for the catalog_new view so incoming checks from wanda can be filtered like priviously in the catalog view

[Screencast from 2023-01-24 12-22-19.webm](https://user-images.githubusercontent.com/54111255/214279234-384192aa-35a6-4f6a-a29c-be79ae6e3939.webm)

## How was this tested?

Added test for the ProviderLabels component functions.
Reworking the existing ChecksCatalogNew test


assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -1,66 +1,81 @@
...
   const dispatch = useDispatch();
-
+  const [selected, setSelected] = useState(providers[0]);
+  const provider = selected;
   const {
     data: catalogData,
     error: catalogError,
     loading,
-  } = useSelector(getCatalog());
+  } = useSelector(getCatalog(provider));
```
arbulu89, line 24: The `getCatalog` selector doesn't accept any incoming param.
In any case, you don't need it, as this only gets the current redux state.
Simply revert the change

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -1,66 +1,81 @@
...
-            <div
-              key={idx}
-              className="check-group bg-white shadow overflow-hidden sm:rounded-md mb-8"
-            >
-              <div className="bg-white px-4 py-5 border-b border-gray-200 sm:px-6">
-                <h3 className="text-lg leading-6 font-medium text-gray-900">
-                  {group}
-                </h3>
+    <div>
+      <ProviderSelection
```
arbulu89, line 40: mmm, in not sure if this `ProviderSelection` shouldn't go inside the `CatalogContainer`, to don't show the filter if the API to wanda fails

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -1,66 +1,81 @@
...
 import CatalogContainer from './CatalogContainer';
 import CheckItem from './CheckItem';
+import ProviderSelection from './ProviderSelection';
 
+const providers = ['default'].concat(providersList.slice());
 // eslint-disable-next-line import/prefer-default-export
 export function ChecksCatalogNew() {
   const dispatch = useDispatch();
-
+  const [selected, setSelected] = useState(providers[0]);
```
arbulu89, line 18: Call this directly `provider` and rename maybe `setSelected` by `setSelectedProvider`

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -1,66 +1,81 @@
...
 import { groupBy } from '@lib/lists';
 
 import { getCatalog } from '@state/selectors/catalog';
 import { updateCatalog } from '@state/actions/catalog';
+import { providersList } from '../ProviderLabel/ProviderLabel';
 import CatalogContainer from './CatalogContainer';
 import CheckItem from './CheckItem';
+import ProviderSelection from './ProviderSelection';
 
+const providers = ['default'].concat(providersList.slice());
```
arbulu89, line 14: I think the labels approach is wrong, as you would be sending the labels in the query, and we might not recognize them. We would need to send the keys:
We don't want to send `provider=AWS` we want to send `provider=aws`

assets/js/components/ProviderLabel/ProviderLabel.jsx
```diff
@@ -50,4 +50,14 @@ function ProviderLabel({ provider }) {
   );
 }
 
+const getLabels = (labelValue) => {
```
arbulu89, line 53: You could have this as:
```
const getLabels = () =>
    Object.values(providerData).map((value) => value.label);
```

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -1,66 +1,81 @@
...
-    dispatch(updateCatalog());
-  }, [dispatch]);
+    setSelected(providers[0]);
+  }, [providers[0]]);
 
   useEffect(() => {
-    getCatalog();
-  }, []);
+    if (provider === 'default') {
+      dispatch(updateCatalog());
```
arbulu89, line 32: I don't think we need this distinction.
If you set provider to `null`, it makes the same work

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
arbulu89, line 14: I wouldn't call this `default`. I will name `All`, because we would get all the checks

arbulu89 (changes requested):
Hey @EMaksy ,
I think we need to revisit some things, as the code uses the labels to create the queries, and we need to use the provider key.

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -1,17 +1,30 @@
...
 import CatalogContainer from './CatalogContainer';
 import CheckItem from './CheckItem';
+import ProviderSelection from './ProviderSelection';
+
+const ALL_FILTER = 'All';
 
 // eslint-disable-next-line import/prefer-default-export
 export function ChecksCatalogNew() {
   const dispatch = useDispatch();
+  const updatedProvider = providerData;
```
arbulu89, line 23: Hey @EMaksy ,
I have looked at this, and we cannot do this direct assignment, as this simply copies the reference of the variable, so changing `updatedProvider` changes `providerData` afterwards. In order to avoid this we need to do a shallow copy, like this: `const updatedProviderData = {...providerData, '': ALL_FILTER}`.
This shallow copies the variable and adds the new empty field to the new object.

After that, we need to change where `providerData` is used by `updatedProviderData` all around the code in this file

EMaksy: Hey good point, i applied the  suggested changes. Thanks

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -1,17 +1,31 @@
...
 
 import { groupBy } from '@lib/lists';
 
 import { getCatalog } from '@state/selectors/catalog';
 import { updateCatalog } from '@state/actions/catalog';
+import {
+  providerData,
+  getLabels,
+  getProviderByLabel,
+} from '../ProviderLabel/ProviderLabel';
```
dottorblaster, line 13:
```suggestion
} from '@components/ProviderLabel/ProviderLabel';
```

dottorblaster (changes requested):
Last touch!

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
```diff
@@ -20,47 +34,61 @@ export function ChecksCatalogNew() {
...
-                    key={check.id}
-                    checkID={check.id}
-                    premium={check.premium}
-                    description={check.description}
-                    remediation={check.remediation}
-                  />
-                ))}
-              </ul>
-            </div>
+    <div>
```
arbulu89, line 45: I think we can safely remove this `div`

arbulu89 (approved):
Perfect! Thanks @EMaksy

assets/js/components/ChecksCatalog/ChecksCatalogNew.jsx
EMaksy: yes :D

dottorblaster (approved):
LGTM

PR #1137, opened 2023-01-23
Remove legacy runner backend

arbulu89:
# Description
Remove legacy runner code from backend. The `wanda` config env has gone now, so I have moved the config there to `config` file.
So, we don't need to do `MIX_ENV=wanda` anymore.

Besides this, the PR removes everything related to the connection settings apis/router/coder.

I have not removed the part of the domain (events and aggregate). It will be done after this, so the PRs are not that huge.

## How was this tested?

No new tests needed


config/config.exs
```diff
@@ -119,8 +119,27 @@ config :trento, Trento.Scheduler,
   debug_logging: false
 
 config :trento, Trento.Integration.Telemetry, adapter: Trento.Integration.Telemetry.Suse
-config :trento, Trento.Integration.Checks, adapter: Trento.Integration.Checks.Runner
-config :trento, Trento.Clusters, checks_adapter: Trento.Clusters.Runner
+
+config :trento, :messaging, adapter: Trento.Messaging.Adapters.AMQP
+config :trento, Trento.Integration.Checks, adapter: Trento.Integration.Checks.Wanda
```
arbulu89, line 124: This adapter remains here as we might really likely create the new wanda mock "faker"

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -96,9 +93,26 @@ defmodule Trento.Clusters do
     end)
   end
 
+  @spec update_cib_last_written(String.t(), String.t()) :: {:ok, Ecto.Schema.t()} | {:error, any}
```
arbulu89, line 96: Simple code movement to respect the `def` and `defp` order

config/runtime.exs
```diff
@@ -75,6 +75,21 @@ if config_env() in [:prod, :demo] do
...
       """
 
+  amqp_url =
+    System.get_env("AMQP_URL") ||
+      raise """
+      environment variable AMQP_URL is missing.
+      For example: amqp://USER:PASSWORD@HOST
+      """
+
+  config :wanda, Wanda.Messaging.Adapters.AMQP,
```
fabriziosestito, line 78: the first argument is the OTP app, here we are configuring `:trento` not `:wanda`

fabriziosestito (changes requested):
Apart from the config, LGTM as a removal PR.
We still need to refactor the adapters/messaging and remove any reference to wanda service tho.
We should just know how to publish on the `trento.checks` exchange and how to consume from the queue, we are not interested on who deals with the execution, so we might want to revisit naming, adapters  and folder structure. I am tracking this in a new task.

config/config.exs
```diff
@@ -119,8 +119,28 @@ config :trento, Trento.Scheduler,
...
+config :trento, Trento.Messaging.Publisher, adapter: Trento.Messaging.Adapters.AMQP
+config :trento, Trento.Clusters, checks_adapter: Trento.Clusters.Wanda
+
+config :trento, Trento.Messaging.Adapters.AMQP,
+  publisher: [
+    exchange: "trento.checks",
+    connection: "amqp://wanda:wanda@localhost:5674"
+  ]
+
+config :trento, Trento.Integration.Checks.Wanda.Messaging.AMQP,
```
fabriziosestito, line 133: I Think we still need the wanda configuration. 
I see it this way:
- dev config (uses the local docker compose forwarded port)
- wanda config (it's a dev config that connects to wanda docker compose forwarded port)

By doing this we can develop web alongside wanda or standalone.

config/config.exs
```diff
@@ -119,8 +119,27 @@ config :trento, Trento.Scheduler,
...
+  processor: Trento.Integration.Checks.Wanda.Messaging.AMQP.Processor,
+  consumer: [
+    queue: "trento.checks.results",
+    exchange: "trento.checks",
+    routing_key: "results",
+    prefetch_count: "10",
+    connection: "amqp://wanda:wanda@localhost:5672",
+    retry_delay_function: fn attempt -> :timer.sleep(2000 * attempt) end
```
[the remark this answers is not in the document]
fabriziosestito: Here the error for reference: 

image

rtorrero (commented):
Just some minor comments while the discussion on the wanda config settles

config/runtime.exs
```diff
@@ -133,7 +139,7 @@ if config_env() in [:prod, :demo] do
     jobs: [
       clusters_checks_execution: [
         # Runs every five minutes by default
-        schedule: "*/#{System.get_env("RUNNER_INTERVAL", "5")} * * * *"
+        schedule: "*/#{System.get_env("CHECKS_INTERVAL", "5")} * * * *"
```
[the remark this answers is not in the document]
arbulu89: No, nor wasn't previously, in any case, I would say this is an internal usage env variable, which shouldn't be used by users.

config/config.exs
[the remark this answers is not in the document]
arbulu89: Thanks man! What I would do without you XD

config/runtime.exs
arbulu89: damn copy/paste XD

config/config.exs
arbulu89: I don't really know a current use case where I would like to have the web running without wanda, but there might be some...
Anyway, it's true that with the current change, the current docker-compose wouldn't have much sense.
I will add back the wanda config

arbulu89: Done!

fabriziosestito: the idea is to start web "standalone" when developing. not every time you need wanda to write code on trento web.

fabriziosestito (approved):
LGTM!

PR #1138, opened 2023-01-23
Fix layout rendering when not logged-in

fabriziosestito:
When accessing a protected page, the layout was rendered for a split second due to the wrong route ordering in the React router.
This PR fixes it.


dottorblaster (approved):
LGTM

arbulu89 (approved):

PR #1139, opened 2023-01-23
Fix login error

fabriziosestito:
This PR fixes the login page. 

Before:

On a 401 we were showing a toast with a generic error message.
There was a bug and the toast was shown only once after the first attempt.

After:
We show an invalid credential message and we highlight with a red border the form fields if we get a 401.
We show a toast when any other error occurs.

It also disables the login button when submitting the form.
login

### DIFF
--- a/assets/js/components/Login.jsx
+++ b/assets/js/components/Login.jsx
@@ -37,6 +37,8 @@ export default function Login() {
     dispatch(performLogin({ username, password }));
   };
 
+  const isUnauthorized = authError && authError.code === 401;
+
   return (
     <div className="min-h-screen bg-gray-50 flex flex-col justify-center py-12 sm:px-6 lg:px-8">
       <div className="sm:mx-auto sm:w-full sm:max-w-md">
@@ -69,7 +71,7 @@ export default function Login() {
                   required
                   className={classNames(
                     'appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-jungle-green-500 focus:border-jungle-green-500 sm:text-sm',
-                    { 'border-red-300': authError && authError.code === 401 }
+                    { 'border-red-300': isUnauthorized }
                   )}
                 />
               </div>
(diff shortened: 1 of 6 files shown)


CDimonaco (approved):

assets/js/state/sagas/user.js
```diff
@@ -24,7 +24,9 @@ export function* performLogin({ payload: { username, password } }) {
     yield call(storeRefreshToken, refreshToken);
     yield put(setUserAsLogged());
   } catch (error) {
-    yield put(setAuthError({ error: error.message }));
+    yield put(
+      setAuthError({ message: error.message, code: error.response.status })
```
arbulu89, line 28: Could this somehow fails on some other error?
Like network error where we don't have a status code.
Would it go to undefined?

assets/js/components/Login.jsx
```diff
@@ -64,7 +67,10 @@ export default function Login() {
                   name="username"
                   autoComplete="username"
                   required
-                  className="appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-jungle-green-500 focus:border-jungle-green-500 sm:text-sm"
+                  className={classNames(
+                    'appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-jungle-green-500 focus:border-jungle-green-500 sm:text-sm',
+                    { 'border-red-300': authError && authError.code === 401 }
```
arbulu89, line 72: Could we wrap this in a function?
```
const unauthorized = (authError) => authError && authError.code === 401;
```

arbulu89 (approved):
Great!
I couple of comments, that don't need a new review

assets/js/components/Login.jsx
dottorblaster: Maybe just having a boolean `const` defined in the function body is enough, I was thinking about the function as well but I don't know if it's the best

EMaksy (commented):
LGTM :D

nelsonkopliku (approved):

rtorrero (approved):
Other than what has already been said, LGTM

assets/js/state/sagas/user.js
fabriziosestito: nice catch!

assets/js/components/Login.jsx
fabriziosestito: done

PR #1140, opened 2023-01-24
Execution started event

CDimonaco:
# Description

Adds the ExecutionStarted event consuming to the web dashboard.
When the event is processed a broadcast message is dispatched to the frontend, to set the execution as "started", with the proper checks executed.

In the frontend

image

## How was this tested?

Automated tests


assets/js/state/lastExecutions.js
```diff
@@ -5,6 +5,7 @@ const initialState = {};
 const initialExecutionState = {
   data: null,
   loading: false,
+  executionStarted: false,
```
arbulu89, line 8: Couldn't we use the current `status` field instead of creating a new one?

assets/js/state/lastExecutions.js
```diff
@@ -42,11 +44,36 @@ export const lastExecutionsSlice = createSlice({
 
       const lastExecutionState = {
         ...initialExecutionState,
+        executionStarted: true,
         error,
       };
 
       state[groupID] = lastExecutionState;
     },
+    setExecutionStarted: (state, { payload }) => {
```
arbulu89, line 53: I think that the next would be a better option:
- Change the current `setExecutionRequested` function to simply put ` requested` value in the status field
- Change this `setExecutionStarted` to do what the current `setExecutionRequested` does, which was working fine, except the data was not coming from wanda

assets/js/state/lastExecutions.js
```diff
@@ -42,11 +44,36 @@ export const lastExecutionsSlice = createSlice({
...
+      // eslint-disable-next-line prefer-destructuring
+      const targets = state[clusterID].data.targets;
+      const targetsWithCheckUsedInExecution = targets.map((target) => ({
+        ...target,
+        checks: target.checks.filter((checkID) =>
+          startedChecks.includes(checkID)
+        ),
+      }));
+
+      state[clusterID].data.targets = targetsWithCheckUsedInExecution;
```
arbulu89, line 73: This would better be:
```
const lastExecutionState = {
        ...initialExecutionState,
        executionStarted: true,
        data: {
          status: 'running',
          targets,
        },
      }
```
Or similar, instead of dot access assignment

arbulu89 (changes requested):
Hey @CDimonaco ,
Some suggestions from my side.
We can discuss over them

assets/js/state/lastExecutions.js
CDimonaco: I think dot assignment is fully justified here, it's one of the reason we use redux toolkit because we don't have to deal with immutability and reference

assets/js/state/lastExecutions.js
```diff
@@ -42,11 +44,36 @@ export const lastExecutionsSlice = createSlice({
...
+      // The execution started event contains the information
+      // about the checks "really" started in the execution
+      // this checks could differ from the ones requested
+      // because for some reason the checks could be skipped by
+      // some conditions
+      // we want to make sure that only the real checks executed
+      // are present into the state when an execution occur
+
+      // eslint-disable-next-line prefer-destructuring
+      const targets = state[clusterID].data.targets;
```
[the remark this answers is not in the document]
CDimonaco: I think this is totally doable by the frontend, I discussed with @fabriziosestito and this + the web transformation seems like a good solution

lib/trento/application/integration/checks/adapter/wanda/messaging/amqp/processor.ex
```diff
@@ -25,6 +29,25 @@ defmodule Trento.Integration.Checks.Wanda.Messaging.AMQP.Processor do
     end
   end
 
+  defp handle(%ExecutionStarted{
+         execution_id: execution_id,
+         group_id: group_id,
+         targets: targets
+       }) do
+    checks_for_execution =
```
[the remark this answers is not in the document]
CDimonaco: Same as above, we can discuss further, the change is minimal @fabriziosestito

assets/js/state/lastExecutions.js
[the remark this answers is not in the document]
fabriziosestito: if the targets are == to what we have in the event, the transformation is superfluous. I suggested just having the checks in the broadcasted event since I wrongly believed that we needed just the list of the checks selection. I now realize that the FE is working with a targets-based hierarchy so I agree with @arbulu89 on this one.

lib/trento/application/integration/checks/adapter/wanda/messaging/amqp/processor.ex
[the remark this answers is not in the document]
fabriziosestito: + 1 see comment above

assets/js/state/lastExecutions.js
fabriziosestito: +1 to keep the dot notation as Redux toolkit uses [Immer](https://github.com/immerjs/immer) internally

assets/js/state/lastExecutions.js
fabriziosestito: +1 using the status and transforming it to a tristate,

assets/js/state/lastExecutions.js
arbulu89: It was more about aesthetics, I find it simply more beatiful/readable. But not hard opinion

assets/js/state/lastExecutions.js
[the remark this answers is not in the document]
CDimonaco: done

lib/trento/application/integration/checks/adapter/wanda/messaging/amqp/processor.ex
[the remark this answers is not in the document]
CDimonaco: done

assets/js/state/lastExecutions.js
CDimonaco: done

assets/js/state/lastExecutions.js
CDimonaco: done

dottorblaster (approved):
LGTM

fabriziosestito (approved):

PR #1142, opened 2023-01-24
Skip temporarily the hana cluster e2e tests

arbulu89:
As commented offline, skipping the E2E tests that are attached to the legacy runner code.
Once the whole transition is done, the tests must be updated, most probably after the wanda faker is implemented


dottorblaster (approved):
LGTM

PR #1147, opened 2023-01-25
Remove legacy cluster details

arbulu89:
# Description
More removals. In this case, the legacy check results view. It doesn't have any real change, even though the git diff looks like that, as the PR has some file renaming.

I have removed the old `ChecksResults` component folder and move the needed files to the `ExecutionResults` as now all of our logic about executions is in that component.

Besides that, everything related to the connection settings topic in the redux code is deleted.

## How was this tested?
No changes needed


assets/js/components/ClusterDetails/ChecksResultOverview.jsx
```diff
@@ -1,36 +1,60 @@
...
+  }
+
+  if (!data) {
+    return (
+      <div className="flex flex-col items-center mt-2 px-4">
+        <div className="text-center text-xs">No check results available.</div>
+      </div>
+    );
+  }
+
```
dottorblaster, line 36: Just a nitpick: I would pass individual values as props, just to reduce the dot notation access inside the component and avoid assuming a specific shape for the `data` object.

Also, you can avoid `if (loading || data?.status === 'running')` by doing

```js
<CheckResultOverview
  loading={loading || lastExecution.status === 'running'}
/>
```

when using the component

assets/js/components/ClusterDetails/ClusterDetails.jsx
```diff
@@ -70,18 +71,20 @@ const getStatusPill = (status) =>
     <Pill className="bg-red-200 text-red-800 mr-2">Unhealthy</Pill>
   );
 
-function ClusterDetails() {
+export function ClusterDetails() {
```
dottorblaster, line 74: Can we `export default` this?

dottorblaster (changes requested):
LGTM, just a couple comments

rtorrero (approved):
@arbulu89 lgtm, we are getting closer :-)

dottorblaster (approved):
Curse git diffs, LGTM

assets/js/components/ClusterDetails/ChecksResultOverview.jsx
arbulu89: Tech-debt ticket added to our backlog.
The purpose of this PR is simply to replace the old legacy code by the new

PR #1148, opened 2023-01-25
Chore: auth clean-up

fabriziosestito:
Changes the wrong field name in auth error (we use a username not an email to authenticate).
Removal of repetition in JWT auth plug test.

### DIFF
--- a/test/trento_web/auth/jwt_auth_plug_test.exs
+++ b/test/trento_web/auth/jwt_auth_plug_test.exs
@@ -37,7 +37,7 @@ defmodule TrentoWeb.JWTAuthPlugTest do
   end
 
   describe "renew/2" do
-    test "should put in conn private a renewd access token if the refresh token is valid", %{
+    test "should renew a token and put it in the conn private if the refresh token is valid", %{
       conn: conn
     } do
       valid_refresh = RefreshToken.generate_refresh_token!(%{"sub" => 1})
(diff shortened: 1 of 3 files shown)


test/trento_web/auth/jwt_auth_plug_test.exs
```diff
@@ -83,15 +68,6 @@ defmodule TrentoWeb.JWTAuthPlugTest do
     test "should return an error is the refresh token signature is valid but it's expired", %{
       conn: conn
     } do
-      expect(
```
arbulu89, line 86: Is not the expiration test based on this value?
Simply commenting because I see two different mocked values for the time through the tests

arbulu89 (approved):

test/trento_web/auth/jwt_auth_plug_test.exs
fabriziosestito: as long as `expired  > current` and `non-expired < current`  the test is ok

nelsonkopliku (approved):

test/trento_web/auth/jwt_auth_plug_test.exs
```diff
@@ -28,15 +40,6 @@ defmodule TrentoWeb.JWTAuthPlugTest do
     test "should put in conn private a renewd access token if the refresh token is valid", %{
       conn: conn
     } do
-      expect(
```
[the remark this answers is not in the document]
fabriziosestito: fixed it

PR #1170, opened 2023-02-06
Remove legacy runner events

arbulu89:
# Description

Remove legacy events usage and code around it like unused read models and their database configuration. The code around host connection settings is remove as well.

The legacy events are moved to a `legacy` folder, because they must exist in the code, as they are part of the stream, but they don't apply any change to the aggregate anymore.

## How was this tested?

Mostly test were removed. A new single test to check the legacy events don't change the aggregate was added


lib/trento/domain/cluster/cluster.ex
```diff
@@ -510,6 +336,8 @@ defmodule Trento.Domain.Cluster do
     snapshot
   end
 
+  def apply(cluster, %legacy_event{}) when legacy_event in @legacy_events, do: cluster
```
arbulu89, line 339: This is the unique new piece of code

nelsonkopliku (approved):
image

fabriziosestito (changes requested):
We need also to clean up `TrentoWeb.OpenApi.Schema.Cluster`, mega LGTM for the rest.

arbulu89:
> We need also to clean up `TrentoWeb.OpenApi.Schema.Cluster`, mega LGTM for the rest.

Thanks @fabriziosestito good catch.
Updated

dottorblaster (approved):
LGTM

PR #1180, opened 2023-02-08
Pin Api version v1

CDimonaco:
# Description

Pin current api endpoints to version v1.
Add a redirect from /api/ to /api/<latest api version>

Session and Page routes not involved by api versioning.
Frontend receives the v1 api prefix.

The OpenApi swagger is prepared to multiple schema definition, the library needs some attention for multiple definition support, but for the moment we have only v1, so one definition.

The ApiRedirector plug will be copied into wanda project, then extracted into a separate configured library.

## How was this tested?

Automated tests


lib/trento_web/plugs/api_redirector.ex
```diff
@@ -0,0 +1,72 @@
+defmodule TrentoWeb.Plugs.ApiRedirector do
+  @moduledoc """
+    This Plug is responsible for redirect api requests without a specific version
```
arbulu89, line 3: `for redirecting api..`

lib/trento_web/plugs/api_redirector.ex
```diff
@@ -0,0 +1,72 @@
...
+    to the latest version, when the requested path exists
+
+    For example:
+      Requesting /api/test, will redirect to /api/<latest version/test, only if the /api/<latest version/test exists.
+
+    router and latest_version options should be provided.
+
+    latest_version option should be a string, will be interpolated with the path.
```
arbulu89, line 11: maybe `a string, will` -> `a string, which will be`

lib/trento_web/router.ex
```diff
@@ -49,75 +53,89 @@ defmodule TrentoWeb.Router do
...
+      post "/hosts/:id/heartbeat", HostController, :heartbeat
+    end
   end
 
-  scope "/api", TrentoWeb do
+  scope "/api" do
     pipe_through :api
 
-    get "/prometheus/targets", PrometheusController, :targets
+    scope "/v1", TrentoWeb.V1 do
```
arbulu89, line 121: Same again, should be change the prometheus configuration?

test/trento_web/plugs/api_redirector_test.exs
```diff
@@ -0,0 +1,54 @@
...
+
+  alias TrentoWeb.Plugs.ApiRedirector
+
+  defmodule StubRouter do
+    def __routes__ do
+      [%{verb: :get, path: "/api/v1/test"}]
+    end
+  end
+
+  describe "call/2" do
```
arbulu89, line 13: Simply fantastic testing!

lib/trento_web/plugs/api_redirector.ex
```diff
@@ -0,0 +1,72 @@
...
+
+  alias TrentoWeb.ErrorView
+
+  import Plug.Conn
+
+  @impl true
+  def init(opts), do: opts
+
+  @impl true
+  def call(%Plug.Conn{path_info: [_ | path_parts], method: method} = conn, opts) do
```
arbulu89, line 25: I don't know if this looks really "elixiric", but I find it nice hehe

arbulu89 (approved):
Dude! This work looks so fantastic!
I just opened some questions whether the frontend api and agent connection api should have the same version, and if we should change the agent explicitily to adapt to this.
In any case, even though we might want to discuss about those things, a big green light

lib/trento_web/router.ex
```diff
@@ -49,75 +53,89 @@ defmodule TrentoWeb.Router do
...
+    end
   end
 
-  scope "/api", TrentoWeb do
+  scope "/api" do
     pipe_through [:api, :apikey_authenticated]
 
-    post "/collect", DiscoveryController, :collect
-    post "/hosts/:id/heartbeat", HostController, :heartbeat
+    scope "/v1", TrentoWeb.V1 do
```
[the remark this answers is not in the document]
nelsonkopliku: Behavior of an `/api/an-endpoint` operation can not be considered stable, thus I guess it'd be safer pointing agent/prometheus/whatever to the actual latest stable version `/api/v1`.
This clarity would at least save us from some headache in the need to debug :sweat_smile: 

---

The thought about the version number opens to a different segregation of APIs based on component/product/service prior to version segregation, for example `/api/<agent|discovery>/v1/collect` `/api/<landscape>/v1/hosts`, `/api/<platform>/v1/accept-eula`

It's something I've encountered in the past and did work pretty well, however we had 15 sets of API products on 8 teams :sweat_smile:. I mean I don't dislike that, maybe we're just not there yet :thinking:

test/trento_web/plugs/api_redirector_test.exs
```diff
@@ -0,0 +1,54 @@
+defmodule TrentoWeb.Plugs.ApiRedirectorTest do
+  use TrentoWeb.ConnCase, async: true
+  use Plug.Test
+
+  alias TrentoWeb.Plugs.ApiRedirector
+
+  defmodule StubRouter do
+    def __routes__ do
+      [%{verb: :get, path: "/api/v1/test"}]
```
nelsonkopliku, line 9: I guess this handles also dynamic routes like `/api/v1/some-resource/:resource_id` by properly intercepting `/api/some-resource/12345` and redirecting to `/api/v1/some-resource/12345`.

Maybe worth adding a scenario in the tests?

assets/js/lib/network/index.js
```diff
@@ -12,7 +12,9 @@ export const unrecoverableAuthError = Error(
   'could not authenticate the user, session destroyed'
 );
 
-export const networkClient = axios.create({});
+export const networkClient = axios.create({
+  baseURL: '/api/v1/',
```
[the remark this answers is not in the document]
nelsonkopliku: I'm also fine with calling `get('/something')` rather than `get('something')`. But again, both work.

nelsonkopliku (commented):
Nice, just a couple of considerations.

assets/js/lib/network/index.js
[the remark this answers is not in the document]
fabriziosestito: +1

test/trento_web/plugs/api_redirector_test.exs
CDimonaco: nice catch! done

CDimonaco:
Addressed all the changes.

test/trento_web/plugs/api_redirector_test.exs
```diff
@@ -0,0 +1,69 @@
...
+  use TrentoWeb.ConnCase, async: true
+  use Plug.Test
+
+  alias TrentoWeb.Plugs.ApiRedirector
+
+  defmodule StubRouter do
+    def __routes__ do
+      [
+        %{verb: :get, path: "/api/v1/test"},
+        %{verb: :get, path: "/api/v1/some-resource/12345"}
```
nelsonkopliku, line 11: I thought we would have the parametrized route here `/api/v1/some-resource/:resourceId`, otherwise the two tests are basically the same.

However, I could be missing something, so up to you deciding what's best.

CDimonaco: We don't have all the routing chain in the test, but you spotted a bug, thanks man, reverting to draft

CDimonaco:
Refactored logic, please rereview.

The status code changed to 307, this is more correct and preserves the method and body when redirecting. This allows to be recognized by all the modern clients and it's semantically more correct.

test/trento_web/plugs/api_redirector_test.exs
CDimonaco: solved that bug for subrouting, thx

arbulu89 (approved):

nelsonkopliku (approved):
Besides the pipeline LGTM

PR #1196, opened 2023-02-16
Setup ci demo wanda

arbulu89:
Remove some legacy runner leftovers and add wanda image values to deploy on demo
PD: on draft as we should have the wanda side finished first, with the `demo` image published in our registry. Once that is done, we could merge this


rtorrero (approved):
LGTM

fabriziosestito (approved):

PR #1198, opened 2023-02-16
Check base url injected into the client via global config object

CDimonaco:
# Description

The base url of the check service, was previously injected into the frontend via an env variable (WANDA_URL), populated at build time. 

This behavior leads to a bug where the user cannot add the base url of the check service because that can be only set at assets build time, when we build the image basically. We need to give the user the ability to specify on which path/url is present the check service in their infrastructure. This can be a path "/check_service" or a full url like we can see on the development configuration.

This configuration is then used by axios to compose the base url of the checks frontend library.

This is similar to grafana public url handling.

Jest configuration updated to perform tests with the global config.

The helm chart needs to be updated.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento_web/templates/page/index.html.heex
+++ b/lib/trento_web/templates/page/index.html.heex
@@ -3,6 +3,7 @@
 <script>
 const config = {
     grafanaPublicUrl: '<%= @grafana_public_url %>',
+    checksServiceBaseUrl: '<%= @check_service_base_url %>'
 };
 </script>
 
--- a/assets/build.js
+++ b/assets/build.js
@@ -10,24 +10,9 @@ config();
 
 const resolvePath = (p) => path.resolve(__dirname, p);
 
-const stringify = (variable) => {
-  if (!variable) {
-    return JSON.stringify('');
-  }
-  return JSON.stringify(variable);
-};
-
-const WANDA_URL =
-  process.env.NODE_ENV === 'production' ? '' : stringify(process.env.WANDA_URL);
-
-const define = {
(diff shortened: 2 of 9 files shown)


lib/trento_web/templates/page/index.html.heex
```diff
@@ -3,7 +3,8 @@
 <script>
 const config = {
     grafanaPublicUrl: '<%= @grafana_public_url %>',
+    checksServiceBaseUrl: '<%= @check_service_base_url %>'
 };
 </script>
 
-<script defer phx-track-static type="text/javascript" src={Routes.static_path(@conn, "/assets/trento.js")}></script>
\ No newline at end of file
+<script defer phx-track-static type="text/javascript" src={Routes.static_path(@conn, "/assets/trento.js")}></script>
```
dottorblaster: God, did I forget about this.

dottorblaster (approved):

nelsonkopliku (approved):

arbulu89 (commented):
With a superficial review, we would need to update the `hack_to_trento` guides, as we wouldn't depend anymore on the `.env` file and this is what the guides describes.
And we could even drop some `dotenv` dependencies as well

lib/trento_web/templates/page/index.html.heex
CDimonaco: done

CDimonaco:
> With a superficial review, we would need to update the `hack_to_trento` guides, as we wouldn't depend anymore on the `.env` file and this is what the guides describes. And we could even drop some `dotenv` dependencies as well

you're right, done!

PR #1201, opened 2023-02-17
Add new demo-deploy dispatch action

rtorrero:
# Description
This PR allows the demo deployment to be triggered using the GH API. This is useful for wanda's CI so that it can also trigger a deployment to the demo env when it finishes updating the demo image.

## How was this tested?
Needs to be tested using a fork with curl

Setting to draft as we need first to create a organization PAT (personal access token) so that we can test it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
     types: [published]
   workflow_dispatch:
   repository_dispatch:
-    types: [start-ci]
+    types: [start-ci, demo-deploy]
 
 env:
   ELIXIR_VERSION: 1.14.3
@@ -300,7 +300,7 @@ jobs:
         run: cd test/e2e && npm install
 
   test-e2e:
-    name: End to end tests 
+    name: End to end tests
     needs: [elixir-deps, npm-deps, npm-e2e-deps]
     runs-on: ubuntu-20.04
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -481,7 +481,7 @@ jobs:
   deploy-demo-env:
     name: Deploy updated images to the demo environment
     runs-on: self-hosted
-    if: vars.DEPLOY_DEMO == 'true' && (github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch')
+    if: (vars.DEPLOY_DEMO == 'true' || github.event.action == 'demo-deploy') && (github.event_name == 'release' || (github.event_name == 'push' && github.ref_name == 'main') || github.event_name == 'workflow_dispatch')
```
arbulu89, line 484: Shouldn't this be an `and`? I mean, otherwise, we won't be able to skip the deployment.
Other option would be to add the `DEPLOY_DEMO` variable to the wanda project as well, so we can disable it from there.
What do you think?

.github/workflows/ci.yaml
```diff
@@ -8,7 +8,7 @@ on:
     types: [published]
   workflow_dispatch:
   repository_dispatch:
-    types: [start-ci]
+    types: [start-ci, demo-deploy]
```
arbulu89, line 11: ocd: as the variable is named `DEPLOY_DEMO` we could call this `deploy-demo` instead of `demo-deploy`

arbulu89 (approved):

.github/workflows/ci.yaml
rtorrero: I prefer the approach of also adding the variable in wanda's CI. I already added it :+1:

PR #1202, opened 2023-02-17
Fix Check selection hint style

nelsonkopliku:
# Description

In the execution page if there's no last execution and there's no check selected the user gets a suggestion to select some checks first - aka a button to go to the checks selection page - which is inconsistent with the rest of the styling.

image

Changed so the style matches the hint to start a checks execution

Scenario 1: No checks selected and no last execution available -> hint to select checks

image

Scenario 1: Some checks selected and no last execution available -> hint to start execution

image

Extra point: adjusted icons color which was lost.

## How was this tested?
Visually :smile: 

@abravosuse @jagabomb makes sense?

### DIFF
--- a/assets/js/components/ExecutionResults/ChecksSelectionHints.jsx
+++ b/assets/js/components/ExecutionResults/ChecksSelectionHints.jsx
@@ -3,7 +3,6 @@ import { useNavigate } from 'react-router-dom';
 
 import { EOS_SETTINGS, EOS_PLAY_CIRCLE } from 'eos-icons-react';
 
-import Button from '@components/Button';
 import TriggerChecksExecutionRequest from '@components/TriggerChecksExecutionRequest';
 
 import TrentoLogo from '@static/trento-icon.png';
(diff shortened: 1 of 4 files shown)


abravosuse (approved):
Code apart, LGTM @nelsonkopliku

jagabomb:
@nelsonkopliku From a simple solution, I would say that this is good for now. Only suggestion I can think of is for **Scenario 2** where you have "Start Execution Now", I would recommend that you use a primary button style (green background with white text) and remove the word now because is is repetitive (too many **now** words) based on the bloody copy above the button... I have included a suggestion below
Screenshot 2023-02-20 at 09 47 07

dottorblaster (approved):
No problems so far code-wise, I don't know if you wanna follow @jagabomb's indications for now

nelsonkopliku:
@jagabomb here's how the two scenarios look like:

image

and

image
would that be enough for now?

@dottorblaster I had to juggle a bit with styles to not break other buttons :sweat_smile: 
We could keep track an improvement item for those buttons and their reusability.

jagabomb:
> @jagabomb here's how the two scenarios look like:
> 
> image
> 
> and
> 
> image would that be enough for now?
> 
> @dottorblaster I had to juggle a bit with styles to not break other buttons 😅 We could keep track an improvement item for those buttons and their reusability.

Awesome, thanks for this @nelsonkopliku. This looks like a better solution to me.

jagabomb (approved):
LGTM

PR #1205, opened 2023-02-20
Fix home page filters background color on selected

arbulu89:
Basic fix to change the background color of the home page filter foxes. Previously the white background was prevailing always. The unique change is to move the `bg-white` to the non selected condition.
fix


dottorblaster (approved):
LGTM

PR #1206, opened 2023-02-20
Fix start execution interaction

nelsonkopliku:
# Description

Alternative approach to https://github.com/trento-project/web/pull/1125 (no [react context](https://beta.reactjs.org/learn/passing-data-deeply-with-context) involved)

Currently if a request to `POST /api/clusters/:cluster_id/checks/request_execution` fails the error is not caught on the client causing the webapp to crash.

<details>
  <summary>See gif</summary>
  
  failing-request-execution-bug
</details>

The agreed behaviour is that a user gets redirected to the checks execution detail only if requesting the execution to start succeeds, otherwise an error notification is shown and the user stays in the page from where the action was triggered.

<details>
  <summary>On failed execution start</summary>
  
  failed-execution-start
</details>
<details>
  <summary>On successful execution start</summary>
  
  successful-execution-start
</details>

Outline of the changes:
- passes `navigate` as an argument to `onStartExecution`
- fixed lastExecution saga
- minor cleanupz

## How was this tested?

Manually + existing automated tests + added needed tests

### DIFF
--- a/assets/js/state/lastExecutions.js
+++ b/assets/js/state/lastExecutions.js
@@ -54,17 +54,14 @@ export const lastExecutionsSlice = createSlice({
     setExecutionStarted: (state, { payload }) => {
       const { groupID: clusterID, targets } = payload;
 
-      if (typeof state[clusterID] === 'undefined') {
-        state[clusterID] = initialExecutionState;
-      }
-
-      if (state[clusterID].data === null) {
-        state[clusterID].data = {};
-      }
-
-      state[clusterID].data.targets = targets;
-      state[clusterID].data.status = RUNNING_EXECUTION_STATE;
-      state[clusterID].error = null;
+      const lastExecutionState = {
+        ...initialExecutionState,
+        data: {
+          status: RUNNING_EXECUTION_STATE,
+          targets,
+        },
+      };
+      state[clusterID] = lastExecutionState;
     },
     setExecutionRequested: (state, { payload }) => {
       const { clusterID: groupID, hosts, checks } = payload;
--- a/assets/js/state/lastExecutions.js
+++ b/assets/js/state/lastExecutions.js
@@ -54,6 +54,14 @@ export const lastExecutionsSlice = createSlice({
     setExecutionStarted: (state, { payload }) => {
       const { groupID: clusterID, targets } = payload;
 
+      if (typeof state[clusterID] === 'undefined') {
+        state[clusterID] = initialExecutionState;
+      }
+
+      if (state[clusterID].data === null) {
+        state[clusterID].data = {};
+      }
+
       state[clusterID].data.targets = targets;
       state[clusterID].data.status = RUNNING_EXECUTION_STATE;
       state[clusterID].error = null;
--- a/assets/js/state/sagas/lastExecutions.js
+++ b/assets/js/state/sagas/lastExecutions.js
@@ -1,10 +1,14 @@
-import { put, call, takeEvery } from 'redux-saga/effects';
+import { put, call, takeEvery, select } from 'redux-saga/effects';
 
 import {
   UPDATE_LAST_EXECUTION,
   EXECUTION_REQUESTED,
 } from '@state/actions/lastExecutions';
-import { getLastExecutionByGroupID } from '@lib/api/checks';
+import { notify } from '@state/actions/notifications';
+import {
+  getLastExecutionByGroupID,
+  triggerChecksExecution,
+} from '@lib/api/checks';
 import {
   setLastExecutionLoading,
   setLastExecution,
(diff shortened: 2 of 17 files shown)


nelsonkopliku:
@dottorblaster here you go!

assets/js/state/sagas/lastExecutions.js
```diff
@@ -34,7 +40,27 @@ export function* updateLastExecution({ payload }) {
...
+  try {
+    yield call(triggerChecksExecution, clusterID);
+    yield put(setExecutionRequested(payload));
+    yield put(
+      notify({
+        text: `Checks execution requested, cluster: ${clusterName}`,
+        icon: '🐰',
+      })
+    );
+    navigate(`/clusters/${clusterID}/executions/last`);
```
arbulu89, line 55: I don't really think we should do the `navigate` here.
I think we should use a promise chain:
```
dispatch(executionRequested(clusterID, hosts, selectedChecks))
.then(() => {
navigate(`/clusters/${clusterID}/executions/last`)
});
```

I know that this would increase some repeated code, but I think it would be the best.
Reducer functions should be as pure as possible.

nelsonkopliku:
@arbulu89 for reference https://github.com/trento-project/web/pull/1125

assets/js/state/sagas/lastExecutions.js
arbulu89: PD: I know that there was a conversation about this in the previous PR, but it still looks anti-pattern to me.

arbulu89: Anyway, my comment comes most probably late as this was already discussed. Just take it as an external observer comment hehe

assets/js/state/lastExecutions.js
```diff
@@ -54,6 +54,14 @@ export const lastExecutionsSlice = createSlice({
     setExecutionStarted: (state, { payload }) => {
       const { groupID: clusterID, targets } = payload;
 
+      if (typeof state[clusterID] === 'undefined') {
```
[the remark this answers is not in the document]
nelsonkopliku: Ah true! thanks!

assets/js/state/sagas/lastExecutions.js
nelsonkopliku: I am not aware of the possibility to do `.then()` on a dispatch :eyes:, is that an option?

In total honesty, I am not a fan of this approach either (I came with a different solution https://github.com/trento-project/web/pull/1125), but still...

However I'd like this to not get stuck, so I am ok with this middleground.

arbulu89: Quick look at our friend stackoverflow XD
https://stackoverflow.com/a/71582748

assets/js/state/lastExecutions.js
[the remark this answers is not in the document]
nelsonkopliku: done

assets/js/state/selectors/cluster.js
```diff
@@ -7,5 +7,10 @@ export const getClusterHostIDs =
       .filter((host) => host.cluster_id === clusterID)
       .map(({ id: hostID }) => hostID);
 
+export const getClusterName =
```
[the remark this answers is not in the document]
nelsonkopliku: added

assets/js/state/sagas/lastExecutions.js
nelsonkopliku: I see, however I am afraid that's not our case cause if I try I get
```
Uncaught TypeError: dispatch(...).then is not a function
```
maybe we're missing something with `redux-toolkit`, which I would leave out from this PR.

arbulu89: Leave it as it is @nelsonkopliku ,
I'm fine!

arbulu89 (approved):
Green light from my side.
Let's wait for @dottorblaster review as he did the 1st round in the previous PR

assets/js/state/sagas/lastExecutions.js
dottorblaster: @arbulu89 I wish this was so simple :smile: mainly here thing is that we are not using `redux-thunk`, a different middleware that allows dispatching asynchronous functions. (And even in that case, you can't `dispatch().then()` if I recall correctly, because `dispatch` isn't async itself)

When using Saga, navigating programmatically passing the `navigate` through the payload looks like something "socially accepted" in the ecosystem. The best thing would be having something like the `history` object available inside the sagas but it doesn't look like the middleware can be configured that way. Too bad.

If we discover something better, maybe we can pursue another path but for now let's just keep it this way.

dottorblaster (approved):
LGTM

Thanks @arbulu89 for the tag :smile:

PR #1210, opened 2023-02-22
Cleanup Cluster selectors

nelsonkopliku:
# Description

Follows up https://github.com/trento-project/web/pull/1206 by further cleaning up cluster related selectors.

PR #1211, opened 2023-02-22
Cleanup notifications saga

nelsonkopliku:
# Description

Follows up https://github.com/trento-project/web/pull/1206 and https://github.com/trento-project/web/pull/1210 by further cleaning up notifications related actions and saga.


assets/js/state/actions/notifications.js
```diff
@@ -4,3 +4,15 @@ export const notify = ({ text, icon }) => ({
   type: NOTIFICATION,
   payload: { text, icon },
 });
+
+export const notifyChecksExecutionRequested = (clusterName) =>
```
nelsonkopliku, line 8: Would these maybe fit better in custer actions? :thinking:

fabriziosestito: I would probably go with adding a cluster notifications file, so the state notifications actions could be agnostic.

I am afraid that to remove a few lines of repetition we will end up with a "utils" file if you know what I mean.  
Said so, I think there is much need to agree on the folder and project structure for the FE asap.

nelsonkopliku: Sure, I understand the _utils-kinda-file_ feeling, I also had that doubt.
If I get it right, you mean having a file named something like `actions/cluster-notifications.js`, correct?

No need to say I also agree on the agreement note.

dottorblaster: I wouldn't take this far the refactoring as well, I'm not a huge fan of the `doThisAndThatAndOtherThings`, I think we can just keep the `notify` function in place and have the `NOTIFICATION` constant refactored as you did :+1:

dottorblaster (changes requested):
Maybe a bit too far?

fabriziosestito (approved):

dottorblaster (approved):

PR #1212, opened 2023-02-22
Jest table filter tests on the overview pages

arbulu89:
# Description
Move tables filtering tests from cypress to jest. This makes the testing more reliable. Besides, now all filters are covered, and adding new scenarios is pretty simple.
The tests are pretty much identical across all the table views: hosts, clusters, sap systems and databases, but I have decided to add all of them. This is mainly because the filter function for each filter box is specific.

The code is based in scenario tables, which later on are used in the test. Even though the code looks similar, some duplication is better here, to avoid having more complex code.

The most generic table filtering tests are done in the specific `Table.test.jsx` file, as we don't really need specific tables and data for it. (Edit: now done in: https://github.com/trento-project/web/pull/1216)

Finally, it removes the cypress tests (Edit: now done in: https://github.com/trento-project/web/pull/1215)

### DIFF
--- a/assets/js/components/ClustersList.test.jsx
+++ b/assets/js/components/ClustersList.test.jsx
@@ -4,7 +4,7 @@ import 'intersection-observer';
 import '@testing-library/jest-dom';
 import { clusterFactory } from '@lib/test-utils/factories';
 import { filterTable, clearFilter } from '@components/Table/Table.test';
-import { renderWithRouter, withState } from '../lib/test-utils';
+import { renderWithRouter, withState } from '@lib/test-utils';
 
 import ClustersList from './ClustersList';
 
--- a/assets/js/components/ClustersList.test.jsx
+++ b/assets/js/components/ClustersList.test.jsx
@@ -2,14 +2,140 @@ import React from 'react';
 import { screen, fireEvent } from '@testing-library/react';
 import 'intersection-observer';
 import '@testing-library/jest-dom';
+import { clusterFactory } from '@lib/test-utils/factories';
+import { filterTable, clearFilter } from '@components/Table/Table.test';
+import { renderWithRouter, withState } from '../lib/test-utils';
 
 import ClustersList from './ClustersList';
-import { renderWithRouter, withDefaultState } from '../lib/test-utils';
 
 describe('ClustersList component', () => {
-  describe('query string filtering behavior', () => {
+  describe('filtering', () => {
+    const cleanInitialState = {
+      hostsList: {
+        hosts: [],
+      },
+      clustersList: {
+        clusters: [],
+      },
+      sapSystemsList: {
+        sapSystems: [],
+        applicationInstances: [],
+        databaseInstances: [],
+      },
+    };
+
+    const scenarios = [
+      {
+        filter: 'Health',
+        options: ['unknown', 'passing', 'warning', 'critical'],
+        state: {
+          ...cleanInitialState,
+          clustersList: {
+            clusters: [].concat(
+              clusterFactory.buildList(2, { health: 'unknown' }),
+              clusterFactory.buildList(2, { health: 'passing' }),
+              clusterFactory.buildList(2, { health: 'warning' }),
+              clusterFactory.buildList(2, { health: 'critical' })
+            ),
+          },
+        },
+        expectedRows: 2,
+      },
+      {
+        filter: 'Name',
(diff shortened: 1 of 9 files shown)


assets/js/components/ClustersList.test.jsx
```diff
@@ -1,30 +1,127 @@
...
+              clusterFactory.buildList(2, { tags: [{ value: 'Tag2' }] })
+            ),
+          },
+        },
+        expectedRows: 2,
+      },
+    ];
+
+    it.each(scenarios)(
+      'should filter the table content by $filter filter',
```
nelsonkopliku, line 108: What is the intent? display exactly `$filter` or interpolate it with Health/SID/Tags/etc?

nelsonkopliku (approved):
Looks good!

assets/js/components/ClustersList.test.jsx
arbulu89: Interpolate with `Health/SID...`. This is how the interpolation is done for `it.each`

arbulu89: https://jestjs.io/docs/api#1-testeachtablename-fn-timeout

assets/js/components/Table/Table.test.jsx
```diff
@@ -0,0 +1,113 @@
...
+    it('should filter by the chosen filter option', () => {
+      const data = tableFactory.buildList(10);
+      const { column1: value1 } = data[0];
+
+      renderWithRouter(<TestTable config={tableConfig} data={data} />);
+
+      filterTable('Column1', value1);
+
+      const table = screen.getByRole('table');
+      expect(table.querySelectorAll('tbody > tr')).toHaveLength(1);
```
dottorblaster, line 71: You don't need to `renderWithRouter` the `TestTable` here, you could just render the `Table` with empty search params and a noop as `setSearchParams`

assets/js/components/Table/Table.test.jsx
```diff
@@ -0,0 +1,113 @@
...
+      );
+
+      fireEvent.click(container.querySelector('.tn-page-item:nth-child(2)'));
+
+      filterTable('Column3', 'value3');
+
+      const table = screen.getByRole('table');
+      expect(table.querySelectorAll('tbody > tr')).toHaveLength(1);
+    });
+  });
```
dottorblaster, line 112: I don't think you need the TestTable here?

assets/js/components/Table/Table.test.jsx
```diff
@@ -0,0 +1,113 @@
...
+          screen
+            .getByTestId(`filter-${filter}-options`)
+            .querySelector('li > div > span').firstChild
+        );
+      });
+
+      expect(window.location.search).toEqual(
+        `?column1=${value1}&column2=${value2}&column3=${value3}`
+      );
+    });
```
dottorblaster, line 93: You could test this side effect in a lot of different ways. This is the first, but it requires React Router, and we aren't really focusing on the filtering behavior but more on the interaction with the browser.

We could also test that for example filling a list with filters behaves the correct way, using plain functions instead of the ones that come from React Router. What do you think?

dottorblaster (changes requested):
Thanks for the huge contribution! Besides, I don't know if having the cypress tests ported as-they-are adds much value.

@CDimonaco what do you think?

CDimonaco (changes requested):
I totally agree with all the comments made by @dottorblaster.

I think we will gain much value by splitting this process in separate prs, one for the pure table filtering, in which we test the table component and all the configuration, the filtering and the pure interactions with that component, one for the "integration" in other views and one for the removal.

Anyway thanks for the contribution, I think that splitting and addressing the comments made gains a lot of value and makes the review process more easy

assets/js/components/Table/Table.test.jsx
arbulu89: I will add these tests back to the overview tests.
I had it there at the beginning, but they looked exactly the same, that's why I put them here.
But i'm having moving them back

arbulu89:
@CDimonaco @dottorblaster The other PR rebased, so this one ready to review!

nelsonkopliku (approved):

dottorblaster (approved):
Looks good

CDimonaco (approved):

PR #1218, opened 2023-02-24
Refactor AMQP adapters and Check integration service

fabriziosestito:


test/trento/application/integration/checks/checks_test.exs
```diff
@@ -1,17 +1,17 @@
...
     ExecutionRequested,
     Target
   }
 
   describe "event mapper" do
-    test "should map to a ExecutionRequested event" do
+    test "should publish an ExecutionRequested event" do
       execution_id = UUID.uuid4()
       group_id = UUID.uuid4()
       provider = :some_provider
```
nelsonkopliku, line 17: Not a biggie at all, but gh started complaining on unmodified code (`provider` and `hosts` being unused variables).
Up to you whether to make it happy.

nelsonkopliku (approved):
Great job! Very nice cleanup, thanks.

lib/trento_web/controllers/v1/cluster_controller.ex
```diff
@@ -1,20 +1,21 @@
 defmodule TrentoWeb.V1.ClusterController do
   use TrentoWeb, :controller
+  use OpenApiSpex.ControllerSpecs
 
   alias Trento.Clusters
 
   alias TrentoWeb.OpenApi.Schema
 
-  use OpenApiSpex.ControllerSpecs
+  plug(OpenApiSpex.Plug.CastAndValidate, json_render_error_v2: true)
```
arbulu89, line 9: Is `mix format` adding parenthesis here and the next line?
It looked nicer without XD

arbulu89 (approved):
great

lib/trento_web/controllers/v1/cluster_controller.ex
fabriziosestito: argh, this is a formatter quirk that happens on certain conditions, it adds parenthesis even if you have `local_without_parens` enabled for that macro

test/trento/application/integration/checks/checks_test.exs
fabriziosestito: Thanks, we should enforce  "warning as errors" in the CI in the future. We still haven't done it because Elixir v1.14 changed something that causes the compiler to complain in certain files and I still need to understand how to make it happy.

PR #1226, opened 2023-03-01
Fix margin between filter elements

EMaksy:
# Description

Fix margins and alignment  of the drop down filters. 
Adds space between the filters

Before:
Screenshot from 2023-03-01 16-09-38

After:
Screenshot from 2023-03-01 16-07-41


arbulu89 (approved):
Thanks!

nelsonkopliku (approved):

PR #1227, opened 2023-03-01
Update commanded and commanded_ecto_projections

fabriziosestito:
Updates `commanded` and  `commanded_ecto_projections` to the next minor.
This removes warnings in the projectors introduced by Elixir v1.14.x


arbulu89 (approved):
The changes and the description are missleading though. Because as far as I see `commanded` stays in the same version.

fabriziosestito:
@arbulu89 thanks, pinned in the mix file as well

PR #1231, opened 2023-03-03
Re-add checks filtering into the new Checks UI

dottorblaster:
Re-added checks filtering as it was, and restored the two tests that were previously deleted


EMaksy (approved):
Wow, very nice :D LGTM!

nelsonkopliku (approved):
Thanks! :rocket:

PR #1232, opened 2023-03-03
Improve error handling on checks results page

nelsonkopliku:
# Description

Improves error handling on Checks results page.

<details>
  <summary>Click here to see last execution loading failure</summary>
  
loading-last-execution-failure

</details>

<details>
  <summary>Click here to see catalog loading failure</summary>
  
loading-catalog-failure

</details>

<details>
  <summary>Click here to see a composite failure</summary>
  
loading-composite-failure

</details>

Extra consideration: we might want to streamline how we handle failures and make it consistent through the app

/summon @jagabomb 

<details>
  <summary>Click here to see how the catalog fails</summary>
  
image

</details>

## How was this tested?

Visually + automated tests

### DIFF
--- a/assets/js/components/ExecutionResults/ExecutionResults.jsx
+++ b/assets/js/components/ExecutionResults/ExecutionResults.jsx
@@ -3,9 +3,7 @@ import Table from '@components/Table';
 
 import ReactMarkdown from 'react-markdown';
 import remarkGfm from 'remark-gfm';
-import { EOS_ERROR } from 'eos-icons-react';
 import LoadingBox from '@components/LoadingBox';
-import NotificationBox from '@components/NotificationBox';
 
 import { getCheckResults, getCheckDescription } from './checksUtils';
 
@@ -93,7 +91,7 @@ function ExecutionResults({
 
   const hosts = hostnames.map((item) => item.id);
 
-  if (catalogLoading) {
+  if (catalogLoading || executionLoading) {
     return <LoadingBox text="Loading checks execution..." />;
   }
 
@@ -101,27 +99,14 @@ function ExecutionResults({
     return <LoadingBox text="Checks execution starting..." />;
   }
 
-  if (catalogError || executionError) {
-    return (
-      <NotificationBox
-        icon={<EOS_ERROR className="m-auto" color="red" size="xl" />}
-        text={
-          catalogError && executionError
-            ? `${catalogError}\n${executionError}`
-            : catalogError || executionError
-        }
-        buttonText="Try again"
-        buttonOnClick={() => {
-          if (catalogError) {
-            onCatalogRefresh();
-          }
-          if (executionError) {
-            onLastExecutionUpdate();
-          }
-        }}
-      />
-    );
-  }
+  const onContentRefresh = () => {
+    if (catalogError) {
+      onCatalogRefresh();
+    }
+    if (executionError) {
+      onLastExecutionUpdate();
+    }
+  };
 
   const tableData = getCheckResults(executionData)
     .filter((check) => {
@@ -158,12 +143,16 @@ function ExecutionResults({
         onFilterChange={(newPredicates) => setPredicates(newPredicates)}
       />
       <ResultsContainer
-        catalogError={false}
+        error={catalogError || executionError}
+        errorContent={[
+          catalogError ? `Failed loading catalog: ${catalogError}` : null,
+          executionError ? `Failed loading execution: ${executionError}` : null,
+        ]}
         clusterID={clusterID}
         hasAlreadyChecksResults={!!(executionData || executionLoading)}
         selectedChecks={clusterSelectedChecks}
         hosts={hosts}
-        onCatalogRefresh={onCatalogRefresh}
+        onContentRefresh={onContentRefresh}
         onStartExecution={onStartExecution}
       >
         <Table config={resultsTableConfig} data={tableData} />
--- a/assets/js/components/ExecutionResults/ExecutionResults.jsx
+++ b/assets/js/components/ExecutionResults/ExecutionResults.jsx
@@ -80,6 +80,7 @@ function ExecutionResults({
   catalogError,
   executionLoading,
   executionStarted,
+  executionRunning,
   executionData,
   executionError,
   clusterSelectedChecks = [],
@@ -97,6 +98,10 @@ function ExecutionResults({
 
   if (!executionStarted) {
     return <LoadingBox text="Checks execution starting..." />;
+  }
+
+  if (executionRunning) {
+    return <LoadingBox text="Checks execution running..." />;
   }
 
   const onContentRefresh = () => {
--- a/assets/js/components/ExecutionResults/ExecutionResultsPage.jsx
+++ b/assets/js/components/ExecutionResults/ExecutionResultsPage.jsx
@@ -15,6 +15,8 @@ import {
 } from '@state/lastExecutions';
 import ExecutionResults from './ExecutionResults';
 
+const runningStates = [RUNNING_EXECUTION_STATE, REQUESTED_EXECUTION_STATE];
+
 function ExecutionResultsPage() {
   const { clusterID } = useParams();
   const dispatch = useDispatch();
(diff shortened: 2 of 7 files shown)


assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -3,17 +3,21 @@ import Table from '@components/Table';
...
+  RUNNING_EXECUTION_STATE,
+} from '@state/lastExecutions';
 import { getCheckResults, getCheckDescription } from './checksUtils';
 
 import ResultsContainer from './ResultsContainer';
 import { ExecutionIcon } from './ExecutionIcon';
 import CheckResultOutline from './CheckResultOutline';
 import ExecutionHeader from './ExecutionHeader';
 
+const runningStates = [RUNNING_EXECUTION_STATE, REQUESTED_EXECUTION_STATE];
```
arbulu89, line 19: It looks a bit strange that we need to import the same constants we do in the `ExecutionResultsPage` here.
Maybe it would make more sense to add other extra paremeter to `ExecutionResults`, like `executionRunning`, or simply send the `executionData.status` and do the logic only in one file, not in both.
Because I see the `executionStarted={executionData?.status !== REQUESTED_EXECUTION_STATE}` in the other and `  if (runningStates.includes(executionData?.status)) {` here, and both have the same intention

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -71,6 +75,24 @@ function MarkdownContent({ children }) {
   );
 }
 
+function LoadingFailure({ entity, error }) {
```
arbulu89, line 78: Do we need this local components here?
The Loading and Notification boxes are generic, so if they need some improvement, I think we should do them here, to have them tested.
Because I see that we start adding terms like `entity` that nobody really understand in the long run.

Maybe we can convert the `text` prop of the `NotificationBox` in a list or so.

arbulu89 (commented):
Thanks @nelsonkopliku ,
In general it looks good to me.
I added some comments about some tech details

assets/js/components/ExecutionResults/ExecutionResults.jsx
nelsonkopliku: Yes, makes sense. Addressing.

assets/js/components/ExecutionResults/ExecutionResults.jsx
nelsonkopliku: Agreed. I'll take care of changing the `NotificationBox` as suggested.

nelsonkopliku:
Hey @arbulu89, thanks for the review. Addressed the feedbacks.

assets/js/components/ExecutionResults/ExecutionResultsPage.jsx
```diff
@@ -64,6 +66,7 @@ function ExecutionResultsPage() {
       catalog={catalog}
       catalogError={catalogError}
       executionLoading={executionLoading}
+      executionRunning={runningStates.includes(executionData?.status)}
```
arbulu89, line 69: Minor thing, but maybe we could move `executionStarted` paremeter down below here, to group the similar things together

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -158,12 +148,16 @@ function ExecutionResults({
         onFilterChange={(newPredicates) => setPredicates(newPredicates)}
       />
       <ResultsContainer
-        catalogError={false}
+        error={catalogError || executionError}
+        errorContent={[
+          `Failed loading catalog: ${catalogError}`,
```
arbulu89, line 153: This now has the issue that if we only have one error, we would still have 2 messages, right?
Maybe we could have something like this, with the spread operator:
```
errorContent{[
  ... catalogError ? [`Failed loading catalog: ${catalogError}`] : [],
  ... executionError ? [`Failed loading execution: ${executionError}`]: []
]}
```

arbulu89 (approved):
Great @nelsonkopliku ,
Just a couple of tiny comments, but ready to be merged from my side

assets/js/components/ExecutionResults/ExecutionResults.jsx
nelsonkopliku: Not really, the `NotificationBox` filters out falsy entries in that list https://github.com/trento-project/web/pull/1232/files#diff-b5395dece1ad14cf45ccd7203b1761e3a873ea659351067b200654f3e0c7dcd4R8

Would that be enough?

nelsonkopliku: my bad, you're right. There will always be something in the way I did :facepalm: Fixing

nelsonkopliku: ended up with 
```
errorContent={[
  catalogError ? `Failed loading catalog: ${catalogError}` : null,
  executionError ? `Failed loading execution: ${executionError}` : null,
]}
```

assets/js/components/NotificationBox.jsx
```diff
@@ -3,13 +3,23 @@ import React from 'react';
...
 function NotificationBox({ icon, text, buttonText, buttonOnClick }) {
+  const textContent = Array.isArray(text)
+    ? text
+        .filter((textRow) => !!textRow)
+        .map((textRow) => (
+          <span key={textRow} className="block">
+            {textRow}
+          </span>
+        ))
+    : text;
```
dottorblaster, line 14: Could you declare this outside the scope of the component as a function?

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -158,12 +148,16 @@ function ExecutionResults({
         onFilterChange={(newPredicates) => setPredicates(newPredicates)}
       />
       <ResultsContainer
-        catalogError={false}
+        error={catalogError || executionError}
+        errorContent={[
+          catalogError ? `Failed loading catalog: ${catalogError}` : null,
+          executionError ? `Failed loading execution: ${executionError}` : null,
+        ]}
```
dottorblaster, line 155: Maybe this works better?

```suggestion
        errorContent={
          <>
            catalogError && `Failed loading catalog: ${catalogError}`
            executionError && `Failed loading execution: ${executionError}`
          </>
        }
```

assets/js/components/ExecutionResults/ResultsContainer.test.jsx
```diff
@@ -10,22 +10,32 @@ import ResultsContainer from './ResultsContainer';
 
 describe('ChecksResults ResultsContainer component', () => {
   it('should render the notification box', () => {
-    renderWithRouter(<ResultsContainer catalogError />);
+    renderWithRouter(<ResultsContainer error />);
 
     expect(screen.getByRole('button')).toHaveTextContent('Try again');
   });
 
+  it('should render the notification box with a composite error', () => {
```
dottorblaster, line 18:
```suggestion
  it('should render the notification box with a both errors', () => {
```

dottorblaster (changes requested):
Some changes then we are good to merge

assets/js/components/ExecutionResults/ExecutionResults.jsx
dottorblaster: You could also use the `span`s here removing the need for the `Array.isArray` check.

nelsonkopliku: The suggestion does not really work better.
In order to get something workable here's how it would end up
```
errorContent={
  <>
    {catalogError && (
      <span className="block">
        Failed loading catalog: {catalogError}
      </span>
    )}
    {executionError && (
      <span className="block">
        Failed loading execution: {executionError}
      </span>
    )}
  </>
}
```
which is a bit noisy.

I'd rather keep passing the list of messages and extract the function in `NotificationBox` as you suggested.

dottorblaster: Your call .)

assets/js/components/ExecutionResults/ResultsContainer.test.jsx
nelsonkopliku: What about `with several errors` instead of `with a both errors`?

dottorblaster: Looks good!

dottorblaster (approved):

PR #1242, opened 2023-03-07
Refactor API errors

fabriziosestito:
This PR refactors controllers, error view, and fallback controller.

- Add the missing errors (4xx) schema based on the jsonapi/openapispex schema
- Refactor the fallback controller and the error view accordingly
- Add the  missing operations specs in some controllers (tags, discovery, heartbeat)
- Add  the missing cast and validate plugs, and it changes the controllers/contexts accordingly

NOTE: we need to revisit how we handle errors and how we use the fallback controller, this is coming in a follow-up PR, but as for now this is enough to have the endpoint properly documented.


lib/trento_web/views/error_view.ex
```diff
@@ -1,24 +1,97 @@
...
+      errors: [
+        %{
+          title: "Invalid value",
+          detail: error
+        }
+      ]
+    }
+  end
+
+  def render("404.json", %{detail: detail}) do
```
arbulu89, line 54: Could we maybe put this after `401` error?
To have numerical order

lib/trento_web/views/error_view.ex
```diff
@@ -1,24 +1,97 @@
...
+    %{
+      errors: render_validation_error(error, "")
+    }
+  end
+
+  def render("422.json", %{error: error}) do
+    %{
+      errors: [
+        %{
+          title: "Invalid value",
```
arbulu89, line 47: For consistency, this should be `Invalid Value`, right? We have upper case in all the words in every other place

lib/trento_web/views/error_view.ex
```diff
@@ -1,24 +1,97 @@
...
+      errors: [
+        %{
+          title: "Unauthorized",
+          detail: detail
+        }
+      ]
+    }
+  end
+
+  def render("422.json", %{error: error}) when is_map(error) do
```
arbulu89, line 37: I find a bit strange to find `error` as argument some times and `detail` others, but I guess it might have some reason about how phonix and plug handle this errors

lib/trento_web/controllers/fallback_controller.ex
```diff
@@ -3,59 +3,38 @@ defmodule TrentoWeb.FallbackController do
...
-    |> put_status(:unauthorized)
-    |> put_view(TrentoWeb.SessionView)
-    |> render("unauthorized.json")
-  end
-
-  defp render_error(conn, reason) do
-    conn
     |> put_view(ErrorView)
-    |> render("error.json", reason: reason)
+    |> render(:"401")
```
arbulu89, line 31: Shouldn't we set some `detail` here? Like `Invalid credentials` or something.
Otherwise, how is this rendered?
I see in the tests that the used details is ` Invalid crendetials`, but is this coming from here or the next `call` function?
Maybe this `call` is not even used

lib/trento_web/openapi/schema/discovery_event.ex
```diff
@@ -0,0 +1,21 @@
...
+    description: "A discovery event",
+    type: :object,
+    properties: %{
+      agent_id: %Schema{type: :string, format: :uuid},
+      discovery_type: %Schema{type: :string},
+      payload: %Schema{
+        oneOf: [%Schema{type: :object}, %Schema{type: :array}]
+      }
+    },
+    required: [:agent_id, :discovery_type]
```
arbulu89, line 19: The `payload` is not required?

lib/trento_web/controllers/v1/discovery_controller.ex
```diff
@@ -1,21 +1,48 @@
...
   action_fallback TrentoWeb.FallbackController
 
-  @spec collect(Plug.Conn.t(), map) :: Plug.Conn.t()
-  def collect(conn, event) do
+  operation :collect,
+    summary: "Collect data from the target infrastructure",
+    tags: ["Agent"],
+    description:
+      "This endpoint is used by the agents to collect data from the target infrastructure",
+    request_body: {"Checks Selection", "application/json", Schema.DiscoveryEvent},
```
arbulu89, line 17: `Checks Selection`? This should be `Discovery Event` or anything like that, right?

lib/trento_web/controllers/fallback_controller.ex
```diff
@@ -3,59 +3,38 @@ defmodule TrentoWeb.FallbackController do
...
-  end
-
-  def call(conn, {:error, {:bad_request, reason}}) do
-    conn
-    |> put_status(:bad_request)
-    |> render_error(reason)
-  end
-
-  def call(conn, {:error, {:unprocessable_entity, reason}}) do
+  def call(conn, {:error, {:unprocessable_entity, error}}) do
```
arbulu89, line 6: What about the `bad_request` now?
Is it handled by `phoenix` itself?

arbulu89 (approved):
Fantastic @fabriziosestito

lib/trento_web/controllers/v1/tags_controller.ex
```diff
@@ -1,33 +1,85 @@
...
+    request_body:
+      {"Tag", "application/json",
+       %OpenApiSpex.Schema{
+         type: :object,
+         properties: %{
+           value: %OpenApiSpex.Schema{type: :string}
+         }
+       }},
+    responses: [
+      accepted: "The tag has been added to the resource",
```
nelsonkopliku, line 32: This operation returns a 201 Created on success.

lib/trento_web/controllers/v1/tags_controller.ex
```diff
@@ -1,33 +1,85 @@
...
+        type: %OpenApiSpex.Schema{type: :string, format: :uuid}
+      ],
+      value: [
+        in: :path,
+        required: true,
+        type: %OpenApiSpex.Schema{type: :string}
+      ]
+    ],
+    responses: [
+      accepted: "The tag has been removed from the resource",
```
nelsonkopliku, line 75: Currently this operation returns 204 No Content on success and possibly a 404.
Are we planning to update the operation based on this new api spec?

lib/trento_web/controllers/v1/tags_controller.ex
```diff
@@ -1,33 +1,85 @@
...
+      {"Tag", "application/json",
+       %OpenApiSpex.Schema{
+         type: :object,
+         properties: %{
+           value: %OpenApiSpex.Schema{type: :string}
+         }
+       }},
+    responses: [
+      accepted: "The tag has been added to the resource",
+      bad_request: Schema.BadRequest.response(),
```
nelsonkopliku, line 33: Currently this operation never returns a bad request.
However if the plan is to add that case and we want to have that documented in the meantime, ignore the comment.

lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -28,15 +28,35 @@ defmodule TrentoWeb.V1.HostController do
...
+      id: [
+        in: :path,
+        required: true,
+        type: %OpenApiSpex.Schema{type: :string, format: :uuid}
+      ]
+    ],
+    responses: [
+      accepted: "The heartbeat has been updated",
+      not_found: TrentoWeb.OpenApi.Schema.NotFound.response(),
+      bad_request: TrentoWeb.OpenApi.Schema.BadRequest.response(),
```
nelsonkopliku, line 46: Currently this operation returns either 204, instead of 202 and never returns a 400 nor a 422.
Is there any behaviour that I missing that takes care of 400s/422s?

However if the plan is to add that case and we want to have that documented in the meantime, ignore the comment.

test/trento_web/controllers/v1/discovery_controller_test.exs
```diff
@@ -2,13 +2,15 @@ defmodule TrentoWeb.V1.DiscoveryControllerTest do
   use TrentoWeb.ConnCase, async: true
 
   describe "discovery" do
-    test "collect action should return bad request when the payload is invalid", %{conn: conn} do
+    test "collect action should return bad request when the event is unknown", %{conn: conn} do
```
nelsonkopliku, line 5: `collect action should return bad request...` is in conflict with what the api actually returns, a 422.

lib/trento_web/controllers/fallback_controller.ex
fabriziosestito: yes

lib/trento_web/controllers/fallback_controller.ex
fabriziosestito: This actually should be removed, because it never happens. Note that this is a temporary fallback controller refractor and we are not using the fallback controller properly yet.

lib/trento_web/controllers/v1/discovery_controller.ex
fabriziosestito: yep!

nelsonkopliku (commented):
Really thanks for the effort, a well documented API is a great value.

I noted down some discrepancies between what the spec says and what the operation actually returns.

Please point out any phoenix/openapispex behavior that might be automagically taking care of making spec aligned with controller implementation. I might be very well missing such knowledge.

lib/trento_web/openapi/schema/discovery_event.ex
fabriziosestito: good catch

lib/trento_web/views/error_view.ex
fabriziosestito: this is going to be refactored soon as we don't want the fallback controller to act this way so I wouldn't worry for now

lib/trento_web/controllers/v1/discovery_controller.ex
```diff
@@ -1,21 +1,48 @@
...
+    responses: [
+      accepted: "The Event has been accepted and the data is being processed",
```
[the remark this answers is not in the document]
fabriziosestito: rephrased

lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -28,15 +28,35 @@ defmodule TrentoWeb.V1.HostController do
     render(conn, "hosts.json", hosts: hosts)
   end
 
-  operation :heartbeat, false
+  operation :heartbeat,
+    summary: "Update the heartbeat for a host",
```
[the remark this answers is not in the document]
fabriziosestito: rephrased

lib/trento_web/controllers/v1/host_controller.ex
fabriziosestito: The 202 is indeed wrong.

422 and 400 are handled by `OpenApiSpex cast_and_validate` plug and Phoenix respectively.

lib/trento_web/controllers/v1/tags_controller.ex
fabriziosestito: it does, phoenix handles it

lib/trento_web/controllers/v1/tags_controller.ex
fabriziosestito: good catch! thanks

lib/trento_web/controllers/v1/tags_controller.ex
fabriziosestito: as above, the 422 and 400 are gently offered by openapispex and phoenix.
202 is wrong and 404 is missing.

lib/trento_web/views/error_view.ex
fabriziosestito: changed it in "Unprocessable Entity"

lib/trento_web/views/error_view.ex
fabriziosestito: done

nelsonkopliku (approved):
Super!

PR #1244, opened 2023-03-07
Add Checks warning banner when provider is VMware

jamie-suse:
# Description

Displays a warning banner in _Cluster Settings_ and _Execution Results_ views when cluster provider is VMware because no default Checks values have been established at this point in time for VMware.

### Old Cluster Settings layout

image

### New Cluster Settings layout

image

### Execution Results

image

## How was this tested?

Added React render tests for when provider is VMware or `unknown` in _Cluster Settings_ and _Execution Results_ views.

### DIFF
--- a/assets/js/components/ClusterDetails/ClusterSettings.jsx
+++ b/assets/js/components/ClusterDetails/ClusterSettings.jsx
@@ -16,6 +16,23 @@ import { getCluster, getClusterHostIDs } from '@state/selectors/cluster';
 
 export const UNKNOWN_PROVIDER = 'unknown';
 
+const warningBanners = {
+  [UNKNOWN_PROVIDER]: (
+    <WarningBanner>
+      The following catalog is valid for on-premise bare metal platforms.
+      <br />
+      If you are running your HANA cluster on a different platform, please use
+      results with caution
+    </WarningBanner>
+  ),
+  vmware: (
+    <WarningBanner>
+      Configuration checks for HANA scale-up performance optimized clusters on
+      VMware are still in experimental phase. Please use results with caution.
+    </WarningBanner>
+  ),
+};
+
 export function ClusterSettings() {
   const { clusterID } = useParams();
 
@@ -35,14 +52,7 @@ export function ClusterSettings() {
         <span className="font-bold">{getClusterName(cluster)}</span>
       </PageHeader>
       <ClusterInfoBox haScenario={cluster.type} provider={cluster.provider} />
-      {cluster.provider === UNKNOWN_PROVIDER && (
-        <WarningBanner>
-          The following catalog is valid for on-premise bare metal platforms.
-          <br />
-          If you are running your HANA cluster on a different platform, please
-          use results with caution
-        </WarningBanner>
-      )}
+      {warningBanners[cluster.provider] ?? null}
       <ChecksSelection clusterId={clusterID} cluster={cluster} />
     </div>
   );
--- a/assets/js/components/ClusterDetails/ClusterSettings.jsx
+++ b/assets/js/components/ClusterDetails/ClusterSettings.jsx
@@ -16,7 +16,7 @@ import { getCluster, getClusterHostIDs } from '@state/selectors/cluster';
 
 export const UNKNOWN_PROVIDER = 'unknown';
 
-export const warningBanners = {
+export const providerWarningBanners = {
   [UNKNOWN_PROVIDER]: (
     <WarningBanner>
       The following catalog is valid for on-premise bare metal platforms.
@@ -51,8 +51,8 @@ export function ClusterSettings() {
         Cluster Settings for{' '}
         <span className="font-bold">{getClusterName(cluster)}</span>
       </PageHeader>
+      {providerWarningBanners[cluster.provider] ?? null}
       <ClusterInfoBox haScenario={cluster.type} provider={cluster.provider} />
-      {warningBanners[cluster.provider] ?? null}
       <ChecksSelection clusterId={clusterID} cluster={cluster} />
     </div>
   );
--- a/assets/js/components/ClusterDetails/ClusterSettings.jsx
+++ b/assets/js/components/ClusterDetails/ClusterSettings.jsx
@@ -42,6 +42,8 @@ export function ClusterSettings() {
     return <div>Loading...</div>;
   }
 
+  const warning = providerWarningBanners[cluster.provider];
+
   return (
     <div className="w-full px-2 sm:px-0">
       <BackButton url={`/clusters/${clusterID}`}>
@@ -51,7 +53,7 @@ export function ClusterSettings() {
         Cluster Settings for{' '}
         <span className="font-bold">{getClusterName(cluster)}</span>
       </PageHeader>
-      {providerWarningBanners[cluster.provider] ?? null}
+      {warning}
       <ClusterInfoBox haScenario={cluster.type} provider={cluster.provider} />
       <ChecksSelection clusterId={clusterID} cluster={cluster} />
     </div>
--- a/assets/js/components/ClusterDetails/ClusterSettings.jsx
+++ b/assets/js/components/ClusterDetails/ClusterSettings.jsx
@@ -16,7 +16,7 @@ import { getCluster, getClusterHostIDs } from '@state/selectors/cluster';
 
 export const UNKNOWN_PROVIDER = 'unknown';
 
-const warningBanners = {
+export const warningBanners = {
   [UNKNOWN_PROVIDER]: (
     <WarningBanner>
       The following catalog is valid for on-premise bare metal platforms.
--- a/assets/js/components/ClusterDetails/ClusterSettings.test.jsx
+++ b/assets/js/components/ClusterDetails/ClusterSettings.test.jsx
@@ -8,7 +8,7 @@ import { withState, defaultInitialState } from '@lib/test-utils';
 import { catalogCheckFactory } from '@lib/test-utils/factories';
 
 import { Route, Routes, MemoryRouter } from 'react-router-dom';
-import { ClusterSettings } from './ClusterSettings';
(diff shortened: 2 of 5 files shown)


jamie-suse:
Note that the placement of the warning banner is not consistent between _Cluster Settings_ and _Execution Results_ views. @jagabomb could you please advise what the correct layout should be?

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -343,4 +344,98 @@ describe('ExecutionResults', () => {
...
+        executionStarted={executionStarted}
+        catalogError={error}
+        executionLoading={executionLoading}
+        executionData={executionData}
+        executionError={executionError}
+      />,
+      { route: `/clusters/${clusterID}/executions/last?health=passing` }
+    );
+
+    expect(screen.getByText('test-cluster')).toBeTruthy();
```
arbulu89, line 380: I think most of this `expect`a are redundant. Wouldn't be enough with the `VMware` and the warning message expects. To not have so many tests checking the same.
The same goes for the next test.

assets/js/components/ClusterDetails/ClusterSettings.jsx
```diff
@@ -16,6 +16,23 @@ import { getCluster, getClusterHostIDs } from '@state/selectors/cluster';
 
 export const UNKNOWN_PROVIDER = 'unknown';
 
+export const warningBanners = {
```
arbulu89, line 19: Maybe `providerWarningBanners` ?

assets/js/components/ClusterDetails/ClusterSettings.test.jsx
```diff
@@ -42,4 +42,84 @@ describe('ClusterDetails ClusterSettings component', () => {
...
   });
+
+  it('given VMware provider, should render the warning banner', async () => {
+    const group = faker.animal.cat();
+    const catalog = catalogCheckFactory.buildList(2, { group });
+
+    const [StatefulChecksSettings, state] = withState(<ClusterSettings />, {
+      ...defaultInitialState,
+      catalog: { loading: false, data: catalog, error: null },
+      clustersList: {
```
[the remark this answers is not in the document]
arbulu89: You should use the factory.
```
clusters = clusterFactory.buildList(1, {provider: 'vmware'})
```
You don't maybe even need to modify the `defaultInitialState`, but even in this case, just with one cluster, the one realted to the test should be enough

assets/js/components/ClusterDetails/ClusterSettings.test.jsx
```diff
@@ -42,4 +42,84 @@ describe('ClusterDetails ClusterSettings component', () => {
...
+        clusters: defaultInitialState.clustersList.clusters.map((cluster) => ({
+          ...cluster,
+          provider: 'vmware',
+        })),
+      },
+    });
+
+    const {
```
arbulu89, line 61: You don't need this `getState`. You should create the cluster with a factory and get the id from it

jagabomb:
> Note that the placement of the warning banner is not consistent between _Cluster Settings_ and _Execution Results_ views. @jagabomb could you please advise what the correct layout should be?

Hi @jamie-suse, the **Check Results** screen looks correct to me. I would suggest that we move the Warning Banner component on the **Cluster Settings** screen to sit just below the page title. This will help to keep consistency in terms of placement and give the Warning Banner a higher level of priority when reading the page content.

arbulu89 (changes requested):
The implementation looks good.
Let's fix the tests. In fact, we should test that we don't have any warning banner when any other provider is used (unless we already test this)
PD: I would try to align the warning message in this PR, it is as simple as putting both in the same place hehe

assets/js/components/ClusterDetails/ClusterSettings.test.jsx
```diff
@@ -42,4 +42,84 @@ describe('ClusterDetails ClusterSettings component', () => {
...
+      clusters: [, , , { id: clusterID }],
+    } = state.getState().clustersList;
+
+    render(
```
arbulu89, line 65: Can we try to replaces this `render` by our internal `renderWithRouter` and set the route there?
I don't know if it will work, but at least let's try

arbulu89: Anyway, forget about this by now. I found out that it is not that trivial

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -136,14 +135,7 @@ function ExecutionResults({
...
       </div>
-      {cloudProvider === UNKNOWN_PROVIDER && (
-        <WarningBanner>
-          The following results are valid for on-premise bare metal platforms.
-          <br />
-          If you are running your HANA cluster on a different platform, please
-          use results with caution
-        </WarningBanner>
-      )}
+      {warningBanners[cloudProvider] ?? null}
```
nelsonkopliku, line 138: Just a headsup: changes in this file might conflict with what is happening in `new-checks-ui` branch.
We need to be careful when this PR gets merged and `new-checks-ui` gets rebased.

jamie-suse: Thank you for the heads-up @nelsonkopliku! I will keep an eye out for that PR.

jagabomb (approved):
LGTM... I just noticed that we will also need to replace the VMware icon with the logo soon. I will log a new ticket for this request.

assets/js/components/ClusterDetails/ClusterSettings.test.jsx
```diff
@@ -42,4 +42,80 @@ describe('ClusterDetails ClusterSettings component', () => {
...
+    const catalog = catalogCheckFactory.buildList(2, { group });
+    const clusters = clusterFactory.buildList(1, { provider: 'vmware' });
+
+    const state = {
+      catalog: { loading: false, data: catalog, error: null },
+      clustersList: { clusters },
+    };
+    const { id: clusterID } = clusters[0];
+
+    const [StatefulChecksSettings] = withState(<ClusterSettings />, {
```
arbulu89, line 57: I don't think we need still to merge with `defaultInitialState`?
Did you try simply with `withState(<ClusterSettings />, state)` ?

arbulu89 (commented):
Some static analysis errors as well

assets/js/components/ClusterDetails/ClusterSettings.test.jsx
jamie-suse: Yep, I tried that but an error occurs because it's missing some other data required. I can see if I can mock up the minimal set of data required but as it currently is, `state` is not sufficient.

arbulu89: If you really need any value from the `defaultInitialState`, let's put it like:
```
    const state = {
      ...defaultInitialState,
      catalog: { loading: false, data: catalog, error: null },
      clustersList: { clusters },
    };
...
const [StatefulChecksSettings] = withState(<ClusterSettings />, state)
```

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -343,4 +344,84 @@ describe('ExecutionResults', () => {
...
+      loading,
+      catalog,
+      error,
+      executionLoading,
+      executionData,
+      executionError,
+      executionStarted,
+    } = prepareStateData('passing');
+
+    renderWithRouter(
```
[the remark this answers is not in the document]
arbulu89: Allright, we don't need the `MemoryRouter` because this component doesn't have a `useParams` function in it. We can even remove the recently added `route` option from the `renderWithRouter`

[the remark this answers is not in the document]
jamie-suse: We still need to provide a `Router` context as the `BackButton` inside `ExecutionResults` uses `useNavigate()`.

https://github.com/trento-project/web/blob/main/assets/js/components/ExecutionResults/ExecutionResults.jsx#L123

https://github.com/trento-project/web/blob/main/assets/js/components/BackButton/BackButton.jsx#L7

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -343,4 +344,82 @@ describe('ExecutionResults', () => {
...
+        hostnames={hostnames}
+        catalogLoading={loading}
+        catalog={catalog}
+        executionStarted={executionStarted}
+        catalogError={error}
+        executionLoading={executionLoading}
+        executionData={executionData}
+        executionError={executionError}
+      />,
+      { route: `/clusters/${clusterID}/executions/last?health=passing` }
```
arbulu89, line 376: I don't know if this `route` does anything, but well, it is at least descriptive hehe

arbulu89 (approved):
All good from my side!
PD: @nelsonkopliku as far as I have seen this, this warning box was not included yet in the new ux, right?
Unfortunately, I don't see any alternative rather than fixing the conflicts later on :disappointed:

rtorrero (approved):
Looks good @jamie-suse :ok_hand: !

assets/js/components/ClusterDetails/ClusterSettings.jsx
```diff
@@ -34,15 +51,8 @@ export function ClusterSettings() {
         Cluster Settings for{' '}
         <span className="font-bold">{getClusterName(cluster)}</span>
       </PageHeader>
+      {providerWarningBanners[cluster.provider] ?? null}
```
dottorblaster, line 54: You could externalize this outside the `return` statement doing

```js
const warning = providerWarningBanners[cluster.provider];
```

And then using that in the markup, avoiding the `?? null`

dottorblaster (changes requested):
Just a couple things then we can merge :+1:

dottorblaster (approved):

PR #1246, opened 2023-03-09
Fix determining expect expectations

nelsonkopliku:
# Description

Fixes # (issue)

## How was this tested?


assets/js/components/ExecutionResults/CheckResultOutline.jsx
```diff
@@ -8,21 +8,29 @@ const TARGET_CLUSTER = 'TARGET_CLUSTER';
...
-}) {
-  const expectExpectationsCount = expectationResults.filter(isExpect).length;
-  const expectResults = agentsCheckResults.map((agentCheckResult) => {
+const expectExpectations = (expectationList) =>
+  expectationList.filter(isExpect);
+
+const getExpectExpectationsMet = (expectationEvaluations) =>
+  expectExpectations(expectationEvaluations).filter(
+    ({ return_value }) => return_value
+  );
```
dottorblaster, line 17: Same?

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -86,6 +86,20 @@ export const checkResultFactory = Factory.define(() => ({
...
+    result,
+  });
+
+export const expectSameExpectationResult = (expectationName, result) =>
+  expectationResultFactory.build({
+    name: expectationName,
+    type: 'expect_same',
+    result,
+  });
+
```
dottorblaster, line 102: These have to go inside the test itself

assets/js/components/ExecutionResults/CheckResultOutline.jsx
```diff
@@ -36,17 +44,49 @@ function CheckResultOutline({
         isAgentCheckError || metExpectations < expectExpectationsCount,
     };
   });
+};
+
+const expectSameExpectations = (expectationList) =>
+  expectationList.filter(isExpectSame);
+
```
dottorblaster, line 51: Inline this please

assets/js/components/ExecutionResults/CheckResultOutline.jsx
```diff
@@ -36,17 +44,49 @@ function CheckResultOutline({
...
+const getExpectSameExpectationResult = (expectationResults, name) =>
+  expectSameExpectations(expectationResults).find(
+    ({ name: resultExpectationName }) => name === resultExpectationName
+  ) || {};
```
dottorblaster, line 55: You can achieve the same result using an `if` statement

assets/js/components/ExecutionResults/CheckResultOutline.jsx
```diff
@@ -8,21 +8,29 @@ const TARGET_CLUSTER = 'TARGET_CLUSTER';
...
+  return agentsCheckResults.map((agentCheckResult) => {
     const { hostname, expectation_evaluations } = agentCheckResult;
 
     const isAgentCheckError = !!agentCheckResult?.type;
 
-    const metExpectations = expectation_evaluations
-      ?.filter(isExpect)
-      .filter(({ return_value }) => return_value).length;
+    const metExpectations = getExpectExpectationsMet(
+      expectation_evaluations || []
```
dottorblaster, line 32:
```js
const { hostname, expectation_evaluations = [] } = agentCheckResult
```

then you can use it directly without the need or `||`ing it

assets/js/components/ExecutionResults/CheckResultOutline.jsx
```diff
@@ -8,21 +8,29 @@ const TARGET_CLUSTER = 'TARGET_CLUSTER';
...
+  expectExpectations(expectationEvaluations).filter(
+    ({ return_value }) => return_value
+  );
+
+const extractExpectResults = (expectations, agentsCheckResults) => {
+  const expectExpectationsCount = expectExpectations(expectations).length;
+
+  if (expectExpectationsCount === 0) {
+    return [];
+  }
```
dottorblaster, line 24: You can have the array in place and if the length is 0 you can return it directly

assets/js/components/ExecutionResults/CheckResultOutline.jsx
```diff
@@ -8,21 +8,29 @@ const TARGET_CLUSTER = 'TARGET_CLUSTER';
...
-function CheckResultOutline({
-  checkID,
-  agentsCheckResults,
-  expectationResults,
-  clusterName,
-}) {
-  const expectExpectationsCount = expectationResults.filter(isExpect).length;
-  const expectResults = agentsCheckResults.map((agentCheckResult) => {
+const expectExpectations = (expectationList) =>
+  expectationList.filter(isExpect);
```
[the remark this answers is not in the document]
nelsonkopliku: We definitely could inline this (and other functions), however I am afraid that the code would become clunky to read (saying because that's where I started from before extracting functions :smile: )

What about a different naming?
`expectExpectations` -> `getExpectExpectations`|`getExpectExpectationsFromList`, and similarly for `expectSameExpectations`? Suggestions?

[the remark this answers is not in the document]
dottorblaster: Maybe `getExpectStatements`? I'd like to avoid `expectExpectation` :smile:

assets/js/components/ExecutionResults/CheckResultOutline.jsx
nelsonkopliku: Following the other comment and renaming to `getExpectSameStatements`

assets/js/components/ExecutionResults/CheckResultOutline.jsx
nelsonkopliku: Yes I know, and if it helps you better, the best I can do is change it to
```js
const getExpectSameStatementResult = (expectationResults, name) => {
  const expectSameStatement = getExpectSameStatements(expectationResults).find(
    ({ name: resultExpectationName }) => name === resultExpectationName
  );

  if (!expectSameStatement) {
    return {};
  }

  return expectSameStatement;
};
```

assets/js/components/ExecutionResults/CheckResultOutline.jsx
nelsonkopliku: ok, let's inline this. No harm

dottorblaster (approved):

jagabomb:
@nelsonkopliku Sorry to only mention this now... I would like to confirm the font sizes of the Check Row and Expectation Result Table. From what I can tell, it seems like the Check Row (e.g: **BA215C corosync.conf files are identical...**) font size is 14px and the Expectation Result table detail is 16px. I would like the bigger font to be on the **Check Row so that it is 16px** and the **Expectation Result table to have a font size of 14px**.

nelsonkopliku:
@jagabomb here you go

Check rows are now `text-base` while expectation result rows `text-sm`.

I had to do a small change to the Table component to allow overriding font, we need to consider also this aspect while improving the table component.

I also changed the table headers to `bg-gray-50` as you suggested in a previous PR, however the emerald hovering has to wait :smile:   

image

Good to go?

jagabomb (approved):
LGTM, thanks Nelson

PR #1247, opened 2023-03-09
Fix remaining act warning

arbulu89:
Boring story about the npm depedency tree havoc coming:

It turns out that we are having the act warning because the `testing-library/dom` version we use in `testing-library/react`  and `testing-library/user-event` don't match: https://testing-library.com/docs/user-event/install

And this is happening (at least one reason) because `@storybook/testing-library` needs a lower version of it, and due how the dependency tree is created, the user-event uses the incorrect version.
If we remove the `@storybook/testing-library` everything works fine.

Going technical, this happens because `testing-library/user-event` has `testing-library/dom` as `peerDependency`, `storybook/testing-library` has it as main dependency as well as `testing-library/react`. This 2nd installs the `dom` library in its internal tree. This means that the version of the `dom` that is in the main `node_modules` tree comes from `storybook` creating the missmatch.

Other potential solution is to force the version of `testing-library/dom` to a good one, but this is discouraged.

More ref: https://github.com/testing-library/user-event/issues/1104

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -36,7 +36,6 @@
         "@storybook/addon-interactions": "^6.5.16",
         "@storybook/addon-links": "^6.5.16",
         "@storybook/react": "^6.5.16",
-        "@storybook/testing-library": "^0.0.13",
         "@testing-library/jest-dom": "^5.16.5",
         "@testing-library/react": "^14.0.0",
         "@testing-library/user-event": "^14.4.3",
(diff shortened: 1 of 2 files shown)


arbulu89:
Other solution is to add this to `package.json`:
```
"overrides": {
    "@testing-library/dom": "^9.0.1"
  },
```

arbulu89:
At the end I have opted for the last comment option, pinning `testing-library/dom` to a specific version as suggested here: https://github.com/testing-library/user-event/issues/1104#issuecomment-1461658418

Besides that, I have run `npm dedupe` to clean up the dependencies tree

nelsonkopliku (approved):
:eyes:

dottorblaster (approved):
LGTM

PR #1249, opened 2023-03-09
Host Deregistration process manager and aggregate handling

CDimonaco:
# Description

Host deregistration process manager and host aggregate deregistration events/commands handling.

## How was this tested?

Automated tests


test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -0,0 +1,92 @@
...
+      assert {:continue, ^host_id} =
+               DeregistrationProcessManager.interested?(%HostDeregistrationRequested{
+                 host_id: host_id
+               })
+    end
+
+    test "should stop the process manager when HostDeregistered arrives" do
+      host_id = UUID.uuid4()
+
+      assert {:stop, ^host_id} =
```
rtorrero, line 41: I'm not sure about: is the Process Manager stopped because it got never started in the first place or because the `%HostDeregistered` event arrived?

I would expect to see something explicit that causes the registration in the first place, am I missing something?

rtorrero (approved):
Good stuff @CDimonaco

Appart from the question in the tests, I'm wondering about the `%HostRolledUp` and just want to double check that I understand it properly: we need this because once we roll up our events and they get replayed, there will be no longer a `%HostRegistered` event but instead a `%HostRolledUp` one (which is effectively the same for us?)

test/trento/application/process_managers/deregistration_process_manager_test.exs
CDimonaco: Yes, the process manager is stopped when HostDeregistered is emitted, and started when HostRegistered or HostRolledUp is emitted

CDimonaco:
@rtorrero Exactly, if we have some host rolled up we don't have the "original" events of registration available but we have the HostRolledUp event, the event has a snapshot of the host and this is for us a proof that he host exists and it's registered, so we start the process manager for that particular host

test/trento/application/process_managers/deregistration_process_manager_test.exs
rtorrero: But shouldn't the test have something starting the process manager first? (before stopping it with the `%HostDeregistered` event)

CDimonaco: Yes, you can look at the different match of the `interested?` function, we have some cases when we return :start, the process manager itself is started by commanded, it's just another genserver under the hood

fabriziosestito: @rtorrero those are unit tests, here the test is to show that certain functions return expected output given certain inputs. 
The lifecycle of the ProcessManager is managed by commanded, so we don't need to start a process "for real", we just need to test that the business logic is correct.

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -0,0 +1,92 @@
...
+      events = [%HostDeregistrationRequested{host_id: host_id}]
+
+      {commands, state} = reduce_events(events, initial_state)
+
+      assert ^initial_state = state
+      assert %DeregisterHost{host_id: ^host_id, deregistered_at: _deregistered_at} = commands
+    end
+  end
+
+  defp reduce_events(events, initial_state) do
```
fabriziosestito, line 84: @CDimonaco could you please keep track of the possibility to add `ProcessManagerCase` in the backlog?

test/trento/domain/host/host_test.exs
```diff
@@ -644,4 +648,66 @@ defmodule Trento.HostTest do
...
+        [
+          host_registered_event,
+          %HostDeregistered{
+            host_id: host_id,
+            deregistered_at: dat
+          }
+        ],
+        [],
+        fn host ->
+          assert host.deregistered_at == dat
```
fabriziosestito, line 708: Please use `expected == actual` in assertions. This way if it fails we know that the left part of the diff is the expected one.

fabriziosestito (changes requested):
LGTM! just a nitpick

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -0,0 +1,62 @@
+defmodule Trento.DeregistrationProcessManager do
+  @moduledoc """
+    DeregistrationProcessManager is a Commanded ProcessManager, it's the responsible
+    for the deregistration procedure for both hosts and clusters.
```
arbulu89, line 4: I would maybe remove this `hosts and clusters` or in any case include the sapsystems.
Maybe simply put `deregistration procedure for the aggregate, such as hosts, clusters, sap systems, etc

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -0,0 +1,62 @@
...
+
+    This represent a transaction to ensure that the procedure of deregistering hosts and cluster
```
arbulu89, line 6: `This represent"s"`.
And the same for `hosts and clusters`, this excludes sap systems message

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -0,0 +1,62 @@
...
+
+  use Commanded.ProcessManagers.ProcessManager,
+    application: Trento.Commanded,
+    name: "deregistration_process_manager"
+
+  deftype do
+    field :host_id, Ecto.UUID
+  end
+
+  alias Trento.DeregistrationProcessManager
```
arbulu89, line 24: It is not a bit strange to self-alias the same module?
In other parts of the code we have used `__MODULE__`

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -0,0 +1,62 @@
...
+    HostDeregistered,
+    HostDeregistrationRequested,
+    HostRegistered,
+    HostRolledUp
+  }
+
+  alias Trento.Domain.Commands.DeregisterHost
+
+  @doc """
+    The process manager is interested in HostRegistered which starts or join an existing process
```
arbulu89, line 36: `or join"s"`

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -0,0 +1,62 @@
...
+    manager for the host identified with host_id field.
```
arbulu89, line 37: `identified by a host_id`

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -0,0 +1,62 @@
...
+    The process manager is interested in HostDeregistered which stops a process manager for the host identified by host id.
```
arbulu89, line 38: `host id` -> `host_id`

lib/trento/domain/host/commands/deregister_host.ex
```diff
@@ -8,5 +8,6 @@ defmodule Trento.Domain.Commands.DeregisterHost do
 
   defcommand do
     field :host_id, Ecto.UUID
+    field :deregistered_at, :utc_datetime_usec, default: DateTime.utc_now()
```
arbulu89, line 11: I don't think we need this new field. Actually it is no even used in the process manager code.
We can get the datetime during the execute or apply functions of the aggregate.
Or is it go make it easier to test?

lib/trento/domain/host/lifespan.ex
```diff
@@ -9,13 +9,20 @@ defmodule Trento.Domain.Host.Lifespan do
...
+  alias Trento.Domain.Events.{
+    HostDeregistered,
+    HostRollUpRequested
+  }
 
   @doc """
   The host aggregate will be stopped after a HostRollUpRequested event is received.
   This is needed to reset the aggregate version, so the aggregate can start appending events to the new stream.
+
+  The host aggregate will be stopped after a HostDeregisteredEvent is received.
```
arbulu89, line 21: `HostDeregistered event is received`

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -0,0 +1,62 @@
...
+    The process manager starts with a Deregistration request and stops when the host is fully deregistered.
+  """
+  def interested?(%HostRegistered{host_id: host_id}), do: {:start, host_id}
+  def interested?(%HostRolledUp{host_id: host_id}), do: {:start, host_id}
+  def interested?(%HostDeregistrationRequested{host_id: host_id}), do: {:continue, host_id}
+  def interested?(%HostDeregistered{host_id: host_id}), do: {:stop, host_id}
+  def interested?(_event), do: false
+
+  def handle(%DeregistrationProcessManager{}, %HostDeregistrationRequested{host_id: host_id}) do
+    %DeregisterHost{host_id: host_id}
```
arbulu89, line 52: Shouldn't we set here the datetime?

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -0,0 +1,92 @@
...
+    HostDeregistrationRequested,
+    HostRegistered,
+    HostRolledUp
+  }
+
+  alias Trento.DeregistrationProcessManager
+  alias Trento.Domain.Commands.DeregisterHost
+
+  describe "events interested" do
+    test "should start the process manager when HostRegisteredEvent arrives" do
```
arbulu89, line 15: `HostRegistered event`

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -0,0 +1,92 @@
...
+    test "should stop the process manager when HostDeregistered arrives" do
+      host_id = UUID.uuid4()
+
+      assert {:stop, ^host_id} =
+               DeregistrationProcessManager.interested?(%HostDeregistered{host_id: host_id})
+    end
+  end
+
+  describe "host deregistration procedure" do
+    test "should update the state with the proper host id when HostRegistered events is emitted" do
```
arbulu89, line 47: `events` -> `event`

test/trento/domain/host/host_test.exs
```diff
@@ -644,4 +648,66 @@ defmodule Trento.HostTest do
       )
     end
   end
+
+  describe "deregistration" do
+    test "should emit the HostDeregistered Event" do
```
arbulu89, line 653: why `Event` in upper case?

arbulu89 (changes requested):
Thanks @CDimonaco 
Mostly, I have simply commented typos and grammatical thing that we could fix.
@jamie-suse could you have a look on the docstrings and test description messages?
Otherwise we could pass grammarly.

Besides that, I see that we don't pass the datetime when when we run the `DeregisterHost` command. Do I miss something there?

lib/trento/domain/host/commands/deregister_host.ex
fabriziosestito: I like this solution because this makes the process manager and aggregate logic completely pure, since asking for a DateTime is a side effect. doesn't change much if we store this date instead of the actual deregistration event date since they would differ only by a few milliseconds and we don't need that solution. Doing this, everything is pure and testing is easier.

lib/trento/application/process_managers/deregistration_process_manager.ex
fabriziosestito: yes +1

test/trento/application/process_managers/deregistration_process_manager_test.exs
CDimonaco: done!

CDimonaco:
The datetime is now passed by the initial command down to the procedure without using implicit and impure default in the process manager.

The docstrings are correct now, thanks guys

fabriziosestito (approved):

test/trento/domain/host/host_test.exs
```diff
@@ -644,4 +648,69 @@ defmodule Trento.HostTest do
...
+          }
+        ],
+        %HostDeregistrationRequested{
+          host_id: host_id,
+          requested_at: requested_at
+        }
+      )
+    end
+
+    test "should apply the HostDeregistered command and set the deregistration date into the state" do
```
arbulu89, line 695: `HostDeregistered` is an event, not a command

arbulu89 (approved):
Great @CDimonaco ,
All good from my side

PR #1252, opened 2023-03-10
Deregistration side effects

rtorrero:
# Description

This PR handles side effects of the `HostDeregistered` event.

- Host _Read Model_ now includes a `deregistered_at` field.
- The _Projector_ handles the `HostDeregistered` event and sets the `deregistered_at` field in the _Read Model_.
- _Deregistered_ Hosts are filtered in the Hosts _Usecase_, (currently used by the _Hosts Controller_) as well as in the Ecto associations for the _Application Instance Read Model_ and _Database Instance Read Model_.

## How was this tested?

Added units tests for the Host _Usecase_ and _Projector_.

### DIFF
--- a/test/trento/application/integration/discovery/protocol/enrich_register_application_instance_test.exs
+++ b/test/trento/application/integration/discovery/protocol/enrich_register_application_instance_test.exs
@@ -31,7 +31,7 @@ defmodule Trento.EnrichRegisterApplicationInstanceTest do
              Enrichable.enrich(command, %{})
   end
 
-  test "should not return an enriched command if the database instance host has been deregistered" do
+  test "should return a database not found error if the database instance host has been deregistered" do
     deregistered_host = insert(:host, deregistered_at: DateTime.utc_now())
 
     %{
--- a/test/trento/application/integration/discovery/protocol/enrich_register_application_instance_test.exs
+++ b/test/trento/application/integration/discovery/protocol/enrich_register_application_instance_test.exs
@@ -31,6 +31,34 @@ defmodule Trento.EnrichRegisterApplicationInstanceTest do
              Enrichable.enrich(command, %{})
   end
 
+  test "should not return an enriched command if the database instance host has been deregistered" do
+    deregistered_host = insert(:host, deregistered_at: DateTime.utc_now())
+
+    %{
+      tenant: tenant,
+      host: %{ip_addresses: [ip]}
+    } =
+      insert(:database_instance_without_host,
+        host_id: deregistered_host.id,
+        host: deregistered_host
+      )
+
+    command =
+      build(
+        :register_application_instance_command,
+        sap_system_id: nil,
+        sid: Faker.StarWars.planet(),
+        db_host: ip,
+        tenant: tenant,
+        instance_number: "00",
+        features: Faker.Pokemon.name(),
+        host_id: Faker.UUID.v4(),
+        health: :passing
+      )
+
+    assert {:error, :database_not_found} = Enrichable.enrich(command, %{})
+  end
+
   test "should return an error if the database was not found" do
     %{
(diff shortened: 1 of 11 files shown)


test/trento/application/usecases/hosts_test.exs
```diff
@@ -25,4 +25,16 @@ defmodule Trento.HostsTest do
       assert 6 = Hosts.get_all_sles_subscriptions()
     end
   end
+
+  describe "Hosts listing" do
+    test "should filter unregistered hosts" do
```
fabriziosestito, line 30: could we change this to "should list all hosts except the deregistered ones"

test/trento/application/usecases/hosts_test.exs
```diff
@@ -25,4 +25,16 @@ defmodule Trento.HostsTest do
...
+      %{id: host_id} = insert(:host)
```
fabriziosestito, line 31: maybe `insert_list(10, :host)`

test/trento/application/usecases/hosts_test.exs
```diff
@@ -25,4 +25,16 @@ defmodule Trento.HostsTest do
...
+      insert(:host, deregistered_at: DateTime.utc_now())
+
+      hosts = Hosts.get_all_hosts()
+
+      assert length(hosts) == 1
```
fabriziosestito, line 36: this assertion is redundant given the assertion below

fabriziosestito (changes requested):
Overall great work!
A test that shows that the application/database instance  `has_many` relation is filtering deregistered hosts correctly is missing.

test/trento/application/usecases/hosts_test.exs
```diff
@@ -25,4 +25,16 @@ defmodule Trento.HostsTest do
       assert 6 = Hosts.get_all_sles_subscriptions()
     end
   end
+
+  describe "Hosts listing" do
```
fabriziosestito, line 29: could we change this in `get_all_hosts/0`

test/trento/application/projectors/host_projector_test.exs
```diff
@@ -344,4 +345,29 @@ defmodule Trento.HostProjectorTest do
...
+       } do
+    timestamp = DateTime.utc_now()
+
+    event = %HostDeregistered{
+      host_id: host_id,
+      deregistered_at: timestamp
+    }
+
+    ProjectorTestHelper.project(HostProjector, event, "host_projector")
+    host_projection = Repo.get!(HostReadModel, event.host_id)
```
arbulu89, line 362: You don't need `event.host_id` here, you could simply use `host_id`

arbulu89 (commented):
Looks great!
Nothing to add from my side. We could merge once @fabriziosestito 's recommendations about the tests are done

test/trento/application/usecases/hosts_test.exs
rtorrero: I used an Enum.map so that the resulting `registered_hosts` was sorted alphabetically (otherwise the assertion below fails due to different ordering as `get_all_hosts` uses `sort`)

test/trento/application/integration/discovery/protocol/enrich_register_application_instance_test.exs
```diff
@@ -31,6 +31,34 @@ defmodule Trento.EnrichRegisterApplicationInstanceTest do
              Enrichable.enrich(command, %{})
   end
 
+  test "should not return an enriched command if the database instance host has been deregistered" do
```
fabriziosestito, line 34: nitpick: better say "should return a database not found error if....
so it's cleared what is doing in terms of domain

fabriziosestito (approved):
just one comment then greenlight from my side

PR #1256, opened 2023-03-13
Open remediation modal when check id is clicked in the ExecutionResults view

EMaksy:
# Description

This pull request adds a new feature to the ExecutionResults view, where clicking on the check ID will now open the remediation modal

The modal will not open if the user clicks outside the check ID but inside the same row, preventing unintended opening of the modal.

[remediationModalOnClick.webm](https://user-images.githubusercontent.com/54111255/224759477-e05a9137-58fd-4723-a373-8af6127fe6f4.webm)

## How was this tested?
A test to check if the remediation modal opens when clicking on the check ID.
A test to check if the remediation modal does not open when clicking outside the check ID but within the same row.


assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -4,22 +4,57 @@ import Table from '@components/Table';
...
+    const { agent_id } = target;
+
+    const { hostname } = hostnames.find(({ id }) => agent_id === id);
+    return {
+      ...target,
+      hostname,
+    };
+  });
+
+function MarkdownContent({ children }) {
```
arbulu89, line 31: I know that you are simply moving this code from place, but...
Do we need this markdown wrapper?
It doesn't do anything, we could simply use the one inside, don't we?

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -159,9 +181,23 @@ function ExecutionResults({
         hosts={hosts}
         onContentRefresh={onContentRefresh}
         onStartExecution={onStartExecution}
+        selectedCheck={selectedCheck}
```
arbulu89, line 184: The `ResultsContainer` doesn't have any `selectedCheck` field. 
I think you can remove this addition safely

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -135,6 +153,10 @@ function ExecutionResults({
         description: getCheckDescription(catalog, checkID),
         expectationResults,
         agentsCheckResults: addHostnameToTargets(agentsCheckResults, hostnames),
+        onClickRemediation: () => {
```
arbulu89, line 156: `onClickRemediation` looks a bit confusing?
Maybe simply `onClick` or `onClickCheckID`.
First of all, because we don't only show the remediation and seconds `Remediation` doesn't match what the user does, which is to click in the check id field

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -411,4 +410,74 @@ describe('ExecutionResults', () => {
...
+    );
+
+    const { id: checkID, remediation } = catalog[0];
+    const clickableID = screen.getByText(checkID);
+    expect(clickableID.textContent).toBe(checks[0]);
+    let remediationModalElement = screen.queryByText(remediation);
+    expect(remediationModalElement).not.toBeInTheDocument();
+    fireEvent.click(clickableID);
+    remediationModalElement = screen.getByText(remediation);
+    expect(remediationModalElement).toBeInTheDocument();
```
arbulu89, line 447: It would be nice to have an additional expect to check that the modal is gone once the user clicks outside the modal.

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -411,4 +410,74 @@ describe('ExecutionResults', () => {
...
+        executionError={executionError}
+        clusterSelectedChecks={checks}
+      />
+    );
+
+    const { remediation, description } = catalog[0];
+    const checkDescriptionText = screen.getByText(description);
+    expect(description).toBe(checkDescriptionText.textContent);
+    fireEvent.click(checkDescriptionText);
+    const remediationModalElement = screen.queryByText(remediation);
```
arbulu89, line 480: Most probably you don't need this constant.
It is really common to do:
`expect(screen.queryByText(remediation)).not.toBeInTheDocument();`

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -411,4 +410,74 @@ describe('ExecutionResults', () => {
...
+        executionStarted={executionStarted}
+        executionData={executionResult}
+        executionError={executionError}
+        clusterSelectedChecks={checks}
+      />
+    );
+
+    const { remediation, description } = catalog[0];
+    const checkDescriptionText = screen.getByText(description);
+    expect(description).toBe(checkDescriptionText.textContent);
```
arbulu89, line 478: This looks inverted.
`expect(checkDescriptionText.textContent).toBe(description)`

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -411,4 +410,74 @@ describe('ExecutionResults', () => {
...
+        executionData={executionResult}
+        executionError={executionError}
+        clusterSelectedChecks={checks}
+      />
+    );
+
+    const { id: checkID, remediation } = catalog[0];
+    const clickableID = screen.getByText(checkID);
+    expect(clickableID.textContent).toBe(checks[0]);
+    let remediationModalElement = screen.queryByText(remediation);
```
arbulu89, line 443: You don't need this `let` or even the variable.
Simply do: 
`expect(screen.queryByText(remediation)).not.toBeInDocument()`

arbulu89 (changes requested):
Hey @EMaksy ,
Some adjustments and ready to go from my side

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -4,22 +4,57 @@ import Table from '@components/Table';
...
-      render: (checkID) => (
-        <div className="whitespace-nowrap text-jungle-green-500">{checkID}</div>
+      render: (checkID, item) => (
+        <div className="whitespace-nowrap text-jungle-green-500">
+          <span
+            className="inline-block"
+            aria-hidden="true"
+            onClick={(e) => {
+              e.stopPropagation();
+              item.onClickRemediation();
```
dottorblaster, line 52: You've got the `checkID` here, so you can directly open the modal with the selected ID here without the need of a function bound to the item

dottorblaster (changes requested):

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
arbulu89: mmm, now I think of this, it might not work this way :sweat_smile: 
and you need it in 2 steps as you have. Give a try hehe

assets/js/components/ExecutionResults/ExecutionResults.jsx
EMaksy: Fair point, naming is always hard hehe :D

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
EMaksy: Sure, I can remove it. 
For me I just prefer to use named variables because it helps the reader to clarify what is happening.

assets/js/components/ExecutionResults/ExecutionResults.jsx
nelsonkopliku: My fault :sweat_smile: just to reduce a bit of boiler... ocd you konw...

However nothing against removing.

EMaksy: Both fine by me, but i remove them.

arbulu89 (approved):
Thanks @EMaksy

nelsonkopliku (approved):

dottorblaster (approved):

PR #1263, opened 2023-03-15
Add Openapi spec to prometheus controller

CDimonaco:
# Description

As the title

## How was this tested?

Automated tests


lib/trento_web/controllers/v1/prometheus_controller.ex
```diff
@@ -1,18 +1,42 @@
...
 
+  operation :exporters_status,
+    summary: "Get prometheus exporters status",
+    tags: ["Target Infrastructure"],
+    description: "Get Prometheus exporters status for a host identified by host id",
+    responses: [
+      ok:
+        {"The status for the prometheus exporter", "application/json",
+         TrentoWeb.OpenApi.Schema.Prometheus.ExporterStatus},
+      not_found: Schema.NotFound.response()
```
arbulu89, line 37: We should maybe have other error besides not found?
For example, if the prometheus server doesn't exists or something like that (in this case we return 500 right now).
The thing is that we don't check if the coming host id even exists and maybe we should do it. Otherwise we depend on parsing the prometheus server response

arbulu89 (commented):
Hey @CDimonaco ,
I think that the `exporters_status` endpoint needs some more work.
I'm afraid that we might need to touch what the controller does to have a proper `not_found`.
Otherwise, you can change this by an `internal error`, which is what we return now in case of error (prometheus server not available). 
We never return not found now

lib/trento_web/controllers/v1/prometheus_controller.ex
arbulu89: I have just checked, and if we use a host id that is not registered in our server, we will receive an empty map with 200 return code

lib/trento_web/openapi/schema/prometheus.ex
```diff
@@ -0,0 +1,23 @@
...
+    OpenApiSpex.schema(%{
+      title: "PrometheusExporterStatus",
+      type: :object,
+      example: %{
+        "Node exporter" => "running"
+      },
+      additionalProperties: %Schema{
+        description:
+          "Status of the exporter, the value could be one of passing, critical, unknown",
+        type: :string
```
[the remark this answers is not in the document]
CDimonaco: Unfortunately seems not possibile to do it with openapi spec, the spec means that the key could be any string and the value any string, seems there is no way to express a object with the keys as string value and the map value to be restricted to a particular type, like an enum :/

lib/trento_web/controllers/v1/prometheus_controller.ex
CDimonaco: I think I could add a generic error, and file a bug for the not found behavior and then ad the not found after the fix, what do you think?

arbulu89: Yes, it would be at least what we really return

CDimonaco: kk

lib/trento_web/openapi/schema/prometheus.ex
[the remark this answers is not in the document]
arbulu89: mmm, i thought that you could do:
```
"additionalProperties": {
        "description":
          "Status of the exporter, the value could be one of passing, critical, unknown",
        "type": "string",
        "enum": ["passing", "critical"]
	}
```
At least in https://www.jsonschemavalidator.net/ catches the failures.
Anyway, don't spend time on this

lib/trento_web/openapi/schema/prometheus.ex
```diff
@@ -0,0 +1,23 @@
...
+  alias OpenApiSpex.Schema
+
+  defmodule ExporterStatus do
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
+      title: "PrometheusExporterStatus",
+      type: :object,
+      example: %{
+        "Node exporter" => "running"
```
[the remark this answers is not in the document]
CDimonaco: fixed

lib/trento_web/controllers/v1/prometheus_controller.ex
CDimonaco: done

lib/trento_web/openapi/schema/prometheus.ex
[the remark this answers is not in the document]
CDimonaco: You're right, I was too tired yesterday, I was looking at other things, sry :/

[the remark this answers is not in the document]
CDimonaco: done

arbulu89:
Thanks @CDimonaco super green light from my side

PR #1237, opened 2023-03-06
Host rollup

CDimonaco:
# Description

Rollup procedure for host aggregate.
Similar to the rollup procedure for Cluster aggregate

## How was this tested?

Automated tests

PR #1009, opened 2022-11-24
Rollup cluster aggregate stream after X events

fabriziosestito:

PR #1272, opened 2023-03-21
Sapsystem rollup

rtorrero:
# Description

This PR continues the effort started by PRs  #1009 and  #1237 to implement rollup for the sap system. Check the description of PR #544 for more information about what exactly rollups are.

## How was this tested?
Tests have been added to:
 - ensure sap system aggregate correctly handles all rollup scenarios
 - rollup is triggered when it should

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -61,13 +61,7 @@ config :trento, Trento.Commanded,
     event_store: Trento.EventStore
   ],
   pubsub: :local,
-  registry: :local,
-  snapshotting: %{
-    Trento.Domain.SapSystem => [
-      snapshot_every: 200,
-      snapshot_version: 1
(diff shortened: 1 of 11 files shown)


arbulu89 (commented):
Thanks @rtorrero 
Some minor comments.
Besides that, we should disable the snapshoting as done here: https://github.com/trento-project/web/pull/1265
It should be a matter of removing it from the config file I think

CDimonaco (changes requested):
I agree with the minor comments of @arbulu89, the snapshotting should be disabled

After that, we can merge

fabriziosestito (approved):

CDimonaco (approved):
All good, nice work, let's merge!

PR #1274, opened 2023-03-22
Cluster and cluster host deregistration events/commands

rtorrero:
# Description

Cluster and cluster host deregistration events and commands

Note: I rebased `deregistration` on `main` before splitting the branch for this PR to keep the development close with current `main`

### DIFF
--- a/lib/trento/domain/cluster/commands/deregister_cluster_host.ex
+++ b/lib/trento/domain/cluster/commands/deregister_cluster_host.ex
@@ -1,0 +1,13 @@
+defmodule Trento.Domain.Commands.DeregisterClusterHost do
+  @moduledoc """
+    Deregister a cluster host
+  """
+  @required_fields :all
+
+  use Trento.Command
+
+  defcommand do
+    field :host_id, Ecto.UUID
+    field :cluster_id, Ecto.UUID
+  end
+end
--- a/lib/trento/domain/cluster/commands/deregister_cluster_host.ex
+++ b/lib/trento/domain/cluster/commands/deregister_cluster_host.ex
@@ -1,6 +1,6 @@
 defmodule Trento.Domain.Commands.DeregisterClusterHost do
   @moduledoc """
-    Deregister a cluster host
+  Deregister a host from a cluster
   """
   @required_fields :all
 
@@ -9,5 +9,6 @@ defmodule Trento.Domain.Commands.DeregisterClusterHost do
   defcommand do
     field :host_id, Ecto.UUID
     field :cluster_id, Ecto.UUID
+    field :deregistered_at, :utc_datetime_usec
   end
 end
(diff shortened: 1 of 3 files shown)


lib/trento/domain/cluster/commands/deregister_cluster_host.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Trento.Domain.Commands.DeregisterClusterHost do
+  @moduledoc """
+    Deregister a cluster host
```
jamie-suse, line 3: Personally, I always found this wording a bit confusing as it sounds as if a "_cluster host_" is a special type of host.

What about something like "Unregister a host from a cluster" or "Remove a host from a cluster"?

lib/trento/domain/cluster/events/cluster_deregistered.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Trento.Domain.Events.ClusterDeregistered do
+  @moduledoc """
+    This event is emitted when a cluster is decommissioned.
```
arbulu89, line 3: Why the word `decommissioned`?
We use `Deregisted` all over the code. If you want to have the same approach than in host, you could put in parenthesis

lib/trento/domain/cluster/commands/deregister_cluster_host.ex
```diff
@@ -0,0 +1,13 @@
...
+  """
+  @required_fields :all
+
+  use Trento.Command
+
+  defcommand do
+    field :host_id, Ecto.UUID
+    field :cluster_id, Ecto.UUID
```
arbulu89, line 11: Shouldn't we have here `deregistered_at` field?
At the end we will cascade this value from events to commands in the process manager

CDimonaco: +1

lib/trento/domain/cluster/commands/deregister_cluster_host.ex
rtorrero: Though I'm not a huge fan of the "deregister" word myself, I'm all about commitment :smile: ... Let's go for now with that word and change it later. I'll make the second part of the sentence more clear though, I think you are right.

lib/trento/domain/cluster/events/host_removed_from_cluster.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Trento.Domain.Events.HostRemovedFromCluster do
+  @moduledoc """
+    This event is emitted when a host is removed from a cluster.
```
arbulu89, line 3: I would remove this initial trailling tab, as we don't have it in others

arbulu89 (approved):
:rocket:

CDimonaco (approved):

PR #1275, opened 2023-03-22
Cluster registration also for nodes not DC

CDimonaco:
# Description

As the title, we change the cluster registration process to register a new cluster with empty details when a node register itself without being a designated controller node.

The details of the cluster will be updated as usual only when a message is received from a designated controller node.

The registration of dc nodes remains as is, with full details.

## How was this tested?

Automated tests


test/trento/domain/cluster/cluster_test.exs
```diff
@@ -116,25 +116,165 @@ defmodule Trento.ClusterTest do
       )
     end
 
-    test "should return an error if the cluster was not registered yet and a command from a non-DC is received" do
-      assert_error(
-        [],
+    test "should add a host the cluster when the cluster has empty/unknown details" do
```
rtorrero, line 119: shouldn't we  keep (in the title)  a reference to the fact that the command is from a non-dc?

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -81,7 +81,7 @@ defmodule Trento.ClusterTest do
       )
     end
 
-    test "should add a host to the cluster" do
+    test "should add a host to the cluster when the cluster has full details" do
```
arbulu89, line 84: I think the added new info doesn't match the reality of the test, as for the cluster is indifferent whether it had full details or not, as new hosts will be added in all cases

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -116,25 +116,165 @@ defmodule Trento.ClusterTest do
...
+          assert :unknown == cluster.type
+          assert :unknown == cluster.provider
+          assert 2 == cluster.hosts_number
+          assert [host_id_two, host_id] == cluster.hosts
+          assert :unknown == cluster.health
+        end
+      )
+    end
+
+    test "should create a cluster with empty details when the cluster was not registered yet and a command from a non-DC is received" do
```
arbulu89, line 168: `should register a cluster...`
And it would make sense to put this test after the 1st one, as they are totally related. Like 1st we test the cluster registration from scratch, and them the adding of hosts

test/trento/domain/cluster/cluster_test.exs
arbulu89, line 119: maybe `should add a new host to the cluster and keep the details as empty when a non-DC node message is received`

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -116,25 +116,165 @@ defmodule Trento.ClusterTest do
...
+          hosts: [host_id],
+          discovered_health: :unknown,
+          health: :unknown
+        }
       )
     end
   end
 
   describe "cluster details" do
+    test "should update cluster details when the cluster has empty details" do
```
arbulu89, line 214: Do we really need this test? It doesn't test anything new compared with the next test in reality.

arbulu89 (commented):
Some suggestions in the docstrings and test descriptions

lib/trento/domain/cluster/cluster.ex
```diff
@@ -7,11 +7,19 @@ defmodule Trento.Domain.Cluster do
   SAP workloads.
 
   Each deployed cluster is registered as a new aggregate entry, meaning that all the hosts belonging
-  to the same cluster are part of the same stream. A cluster is registered first time/details updated afterwards
-  only by cluster discovery messages coming from the **designated controller** node. Once a cluster is
-  registered other hosts can be added receiving discovery messages coming from other nodes. All the hosts
+  to the same cluster are part of the same stream.
+  A cluster is registered first time by cluster discovery messages coming from all the nodes of the cluster.
```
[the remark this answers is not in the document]
CDimonaco: done

test/trento/domain/cluster/cluster_test.exs
CDimonaco: done

lib/trento/domain/cluster/cluster.ex
```diff
@@ -149,9 +158,32 @@ defmodule Trento.Domain.Cluster do
     ]
   end
 
-  # If no DC node was received yet, no cluster was registered.
-  def execute(%Cluster{cluster_id: nil}, %RegisterClusterHost{designated_controller: false}),
-    do: {:error, :cluster_not_found}
+  # If no DC node was received yet, the cluster is registered with empty details
```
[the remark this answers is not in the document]
CDimonaco: done

test/trento/domain/cluster/cluster_test.exs
CDimonaco: should be done in the @arbulu89 comments

lib/trento/domain/cluster/cluster.ex
```diff
@@ -112,8 +120,9 @@ defmodule Trento.Domain.Cluster do
 
   def execute(%Cluster{rolling_up: true}, _), do: {:error, :cluster_rolling_up}
 
-  # When a DC cluster node is registered for the first time, a cluster is registered
-  # and the host of the node is added to the cluster
+  # When a DC cluster node is registered for the first time,
+  # a cluster is registered and the host of the node is added to the cluster,
```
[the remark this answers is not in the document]
arbulu89: I still don't like it.
Term liks `host of the node` and `we` are things that I wouldn't put in the comments

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -81,6 +81,50 @@ defmodule Trento.ClusterTest do
       )
     end
 
+    test "should create a cluster with empty details when the cluster was not registered yet and a command from a non-DC is received" do
```
arbulu89, line 84: - `should create` -> `should register`
- `a command` -> `a message`

test/trento/domain/cluster/cluster_test.exs
arbulu89: Allright!
Up to you. It looks like we are testing a black box hehe

arbulu89 (approved):
Thanks @CDimonaco 
Some minor comments yet in the texts, but it is good to merge anyway

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -116,25 +160,121 @@ defmodule Trento.ClusterTest do
       )
     end
 
-    test "should return an error if the cluster was not registered yet and a command from a non-DC is received" do
-      assert_error(
-        [],
+    test "should add a new host to the cluster and keep the details as empty when a non-DC node message is received" do
```
fabriziosestito, line 163: Do we need this one? the rule is that if details do not change, we do not emit a DetailsUpdated event. This is a consequence of that domain rule, which is already formalized in tests, so I do not think we need to maintain this test.

fabriziosestito (changes requested):

test/trento/domain/cluster/cluster_test.exs
CDimonaco: done

CDimonaco:
Bug with the cluster host add when the host is a DC, fixed.

Tests fixed and new test cases added. Hopefully ready to merge @fabriziosestito

lib/trento/domain/cluster/cluster.ex
```diff
@@ -163,14 +194,17 @@ defmodule Trento.Domain.Cluster do
     maybe_emit_host_added_to_cluster_event(cluster, host_id)
   end
 
+  # When a message arrives from a non DC host who belongs to a cluster, we add that host to the cluster
```
arbulu89, line 197: This comment makes little sense, as it talks about a `non DC` but the code is explicitily for a DC host

lib/trento/domain/cluster/cluster.ex
```diff
@@ -163,14 +193,19 @@ defmodule Trento.Domain.Cluster do
     maybe_emit_host_added_to_cluster_event(cluster, host_id)
   end
 
+  # When a DC node is discovered, if the cluster is already regiestered,
```
rtorrero, line 196: regiestered -> registered

rtorrero (approved):
Just noticed a small typo, everything else looks good :+1:

lib/trento/domain/cluster/cluster.ex
fabriziosestito: thanks!

arbulu89:
Thanks @fabriziosestito 
I like the new comments!

PR #1278, opened 2023-03-23
Cluster deregistration process manager & aggregate changes

rtorrero:
# Description

This PR adds the required changes to the 
 - deregistration process manager
 - cluster aggregate
to handle the removal of a host from a cluster.

In addition to this, we've also removed the host_id from the state of the process manager as it's basically 1:1 with the `process_uuid` that identifies each instance of the process manager and is thus not needed.

### DIFF
--- a/lib/trento/application/process_managers/deregistration_process_manager.ex
+++ b/lib/trento/application/process_managers/deregistration_process_manager.ex
@@ -29,6 +29,7 @@ defmodule Trento.DeregistrationProcessManager do
     HostDeregistered,
     HostDeregistrationRequested,
     HostRegistered,
+    HostRemovedFromCluster,
     HostRolledUp
   }
 
@@ -53,6 +54,7 @@ defmodule Trento.DeregistrationProcessManager do
   def interested?(%ClusterRolledUp{snapshot: %{hosts: hosts}}), do: {:start, hosts}
 
   def interested?(%HostDeregistrationRequested{host_id: host_id}), do: {:continue, host_id}
+  def interested?(%HostRemovedFromCluster{host_id: host_id}), do: {:continue, host_id}
 
   def interested?(%HostDeregistered{host_id: host_id}), do: {:stop, host_id}
 
@@ -91,5 +93,9 @@ defmodule Trento.DeregistrationProcessManager do
         cluster_id: cluster_id
       }) do
     %DeregistrationProcessManager{state | cluster_id: cluster_id}
+  end
+
+  def apply(%DeregistrationProcessManager{} = state, %HostRemovedFromCluster{}) do
+    %DeregistrationProcessManager{state | cluster_id: nil}
   end
 end
--- a/lib/trento/application/process_managers/deregistration_process_manager.ex
+++ b/lib/trento/application/process_managers/deregistration_process_manager.ex
@@ -18,19 +18,24 @@ defmodule Trento.DeregistrationProcessManager do
     name: "deregistration_process_manager"
 
   deftype do
-    field :host_id, Ecto.UUID
+    field :cluster_id, Ecto.UUID
   end
 
   alias Trento.DeregistrationProcessManager
 
   alias Trento.Domain.Events.{
+    ClusterRolledUp,
+    HostAddedToCluster,
     HostDeregistered,
     HostDeregistrationRequested,
     HostRegistered,
     HostRolledUp
   }
 
-  alias Trento.Domain.Commands.DeregisterHost
+  alias Trento.Domain.Commands.{
+    DeregisterClusterHost,
+    DeregisterHost
+  }
 
   @doc """
     The process manager is interested in HostRegistered which starts or joins an existing process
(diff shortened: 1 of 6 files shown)


rtorrero:
Setting it back to draft as I realized by testing manually that it's not working as expected (crashes when it attempts to execute `%DeregisterClusterHost` with `:unregistered_command`

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -44,22 +49,47 @@ defmodule Trento.DeregistrationProcessManager do
   """
   def interested?(%HostRegistered{host_id: host_id}), do: {:start, host_id}
   def interested?(%HostRolledUp{host_id: host_id}), do: {:start, host_id}
+  def interested?(%HostAddedToCluster{host_id: host_id}), do: {:start, host_id}
+  def interested?(%ClusterRolledUp{snapshot: %{hosts: hosts}}), do: {:start, hosts}
+
   def interested?(%HostDeregistrationRequested{host_id: host_id}), do: {:continue, host_id}
+
   def interested?(%HostDeregistered{host_id: host_id}), do: {:stop, host_id}
+
```
arbulu89, line 58: Shouldn't the code listen to `HostRemovedFromCluster` to run the `apply` and remove the cluster from the state?
Otherwise, the process manager is ignorant of this process and would run a non needed `DeregisterClusterHost`. Keep in mind that, the commands could be launched somewhere else

lib/trento/domain/cluster/cluster.ex
```diff
@@ -262,6 +267,26 @@ defmodule Trento.Domain.Cluster do
     }
   end
 
+  def execute(
```
arbulu89, line 270: We should consider the scenario when a `DeregisterClusterHost` is triggered for a cluster that doesn't exist, doing nothing and returning an error in this case.
You might tell me that this won't happen as by now the Process manager is the unique one that runs this commands, but I can go myself and run the command in the console, or in the future the command might be executed by some automatic de-registration, right?

lib/trento/domain/cluster/cluster.ex
```diff
@@ -262,6 +267,26 @@ defmodule Trento.Domain.Cluster do
...
+        %Cluster{cluster_id: cluster_id} = cluster,
+        %DeregisterClusterHost{
+          host_id: host_id,
+          cluster_id: cluster_id
+        } = command
+      ) do
+    cluster
+    |> Multi.new()
+    |> Multi.execute(fn _ ->
+      [
```
arbulu89, line 280: Do we need to wrap the `HostRemovedFromCluster` event in a list?
I think we could simply send it by itself, right?

lib/trento/domain/cluster/cluster.ex
```diff
@@ -469,6 +505,18 @@ defmodule Trento.Domain.Cluster do
     }
   end
 
+  defp maybe_emit_cluster_deregistered_event(
+         %Cluster{cluster_id: cluster_id, hosts_number: hosts_number},
```
arbulu89, line 509: I don't think we can use the `hosts_number` here. This value is coming directly from the agent (parsed from crmmon or cibadmin) and it is not something we make up.
Here we need to get the list of `hosts` and get its length to do the comparison, which is the list we use to register the hosts in our aggregate.
We would need to check if we only have one host, and in that case trigger the deregistered event

arbulu89 (changes requested):
Good progress!
Some comments from my side

lib/trento/domain/cluster/cluster.ex
fabriziosestito: This is already handled by line 221, every command except the registration ones are rejected if the cluster does not exist

lib/trento/domain/cluster/cluster.ex
fabriziosestito: +1

lib/trento/domain/cluster/cluster.ex
fabriziosestito: yes those are CIB details you can skip it

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -44,22 +49,47 @@ defmodule Trento.DeregistrationProcessManager do
   """
   def interested?(%HostRegistered{host_id: host_id}), do: {:start, host_id}
   def interested?(%HostRolledUp{host_id: host_id}), do: {:start, host_id}
+  def interested?(%HostAddedToCluster{host_id: host_id}), do: {:start, host_id}
+  def interested?(%ClusterRolledUp{snapshot: %{hosts: hosts}}), do: {:start, hosts}
+
```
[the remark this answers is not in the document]
fabriziosestito: Whoops, I asked for this to increase the readability of start/stop/continue since it was becoming a big chunk. Maybe we can use comments in the empty lines, not a strong preference tho

lib/trento/application/process_managers/deregistration_process_manager.ex
fabriziosestito: Good catch!

lib/trento/domain/cluster/cluster.ex
```diff
@@ -371,6 +396,17 @@ defmodule Trento.Domain.Cluster do
     snapshot
   end
 
+  def apply(%Cluster{hosts: hosts, hosts_number: hosts_number} = cluster, %HostRemovedFromCluster{
+        host_id: host_id
+      }) do
+    %Cluster{cluster | hosts: List.delete(hosts, host_id), hosts_number: hosts_number - 1}
```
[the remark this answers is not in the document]
fabriziosestito: +1

lib/trento/domain/cluster/cluster.ex
arbulu89: Totally right! I missed that piece of code

lib/trento/application/process_managers/deregistration_process_manager.ex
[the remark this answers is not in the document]
arbulu89: Allright! as you like folks. Fine for me, let's not extend this hehe

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -3,10 +3,12 @@ defmodule Trento.ClusterTest do
 
   import Trento.Factory
 
+  alias Trento.Domain.Events.HostRemovedFromCluster
```
arbulu89, line 6: Let's put this together with the other `Trento.Domain.Events`

arbulu89 (approved):
Great work!
Added some minor suggestions. Feel free to ask a new review if you need, but it is not totally required

PR #1279, opened 2023-03-23
Add empty state subcomponent for the table component

EMaksy:
# Description

Add a more friendly feedback when there are no results for the Table component.

## Host SAP Table
hostSAPEmptyState

## Dashboard Table
[dashBoardEmptyState.webm](https://user-images.githubusercontent.com/54111255/227236076-54823630-f568-4bab-a28a-2cff7cd66378.webm)

## Checks Results Table
[checksResultsEmptyState.webm](https://user-images.githubusercontent.com/54111255/227236090-1fe25e63-291f-4823-a288-f55749ebf942.webm)

## How was this tested?
Added test to check if the error message is rendered
Adjusted HomeHealthSummary test.


assets/js/components/Table/EmptyTable.jsx
```diff
@@ -0,0 +1,21 @@
...
+          className
+        )}
+      >
+        No data available
+      </td>
+    </tr>
+  );
+}
+
+export default EmptyTable;
```
dottorblaster, line 21: Could you rename this just `EmptyState`?

assets/js/components/Table/EmptyTable.stories.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import EmptyTable from './EmptyTable';
+
+export default {
+  title: 'EmptyTable',
+  component: EmptyTable,
+};
+
+export function Default(args) {
```
dottorblaster, line 9: Same :D

assets/js/components/Table/EmptyTable.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import classNames from 'classnames';
+
+function EmptyTable({ colSpan = 1, className = '', fontSize = 'text-sm' }) {
```
dottorblaster, line 4: Since we are not using this prop externally and there's no way to use this explicitly, you could just remove the `fontSize` and `className` props

assets/js/components/Table/EmptyTable.stories.jsx
```diff
@@ -0,0 +1,11 @@
...
+  return <EmptyTable {...args} />;
```
dottorblaster, line 10: Instead of having a story for `EmptyTable`, you could have a story for the `Table` component rendering the empty state :+1:

assets/js/components/Table/Table.test.jsx
```diff
@@ -119,5 +119,20 @@ describe('Table component', () => {
         expect(table.querySelectorAll('tbody > tr')).toHaveLength(1);
       });
     });
+
+    it('should return empty state message when data is empty', async () => {
+      const data = [];
+      const emptyStateMessage = 'No data available';
+
+      render(
+        <Table config={tableConfig} data={data} setSearchParams={() => {}} />
```
dottorblaster, line 128:
```suggestion
      const emptyStateMessage = 'No data available';

      render(
        <Table config={tableConfig} data={[]} setSearchParams={() => {}} />
```

dottorblaster (changes requested):
Just a couple changes then LGTM, nice approach

dottorblaster:
Also keep in mind that we could have an `emptyState` prop for the table, or at least the possibility to customize the empty state text.

jagabomb (approved):
This works for me in terms of a basic solution. Thanks @EMaksy.

EMaksy:
Good idea @dottorblaster!
I added a emptyStateText prop, so the EmptyState component is more reusable and adaptable as the text can be adjusted.

assets/js/components/Table/EmptyTable.jsx
EMaksy: I deleted the EmptyTable story as the Table component already has it covered!

assets/js/components/Table/EmptyTable.stories.jsx
EMaksy: I deleted the EmptyTable story as the Table component already has it covered!

assets/js/components/Table/Table.test.jsx
EMaksy: I adjusted the emptyState component.
Now it accepts a emptyStateText prop so we can have different messages and the test now proves it.

assets/js/components/Table/Table.test.jsx
```diff
@@ -119,5 +119,24 @@ describe('Table component', () => {
         expect(table.querySelectorAll('tbody > tr')).toHaveLength(1);
       });
     });
+
+    it('should return empty state message when data is empty', async () => {
```
arbulu89, line 123: Why do you need `async` here? I don't see any `await` or any async operation

arbulu89 (commented):
Hey @EMaksy ,
I don't see the storybook story you mention

assets/js/components/Table/Table.test.jsx
EMaksy: You are fully right! i will remove it.

EMaksy:
I removed the story, as  it was suggested in a comment before.
The table component includes a story with an empty state, so there is no real need for this extra story.

PR #1280, opened 2023-03-23
Cluster deregistered side effects

CDimonaco:
# Description

ClusterReadModel soft delete side effects.

We should exclude from the query in which is involved the `ClusterReadModel`, the soft-deleted entities.

This Schema is not involved in any association.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/application/projectors/cluster_projector.ex
+++ b/lib/trento/application/projectors/cluster_projector.ex
@@ -12,6 +12,7 @@ defmodule Trento.ClusterProjector do
 
   alias Trento.Domain.Events.{
     ChecksSelected,
+    ClusterDeregistered,
     ClusterDetailsUpdated,
     ClusterHealthChanged,
     ClusterRegistered
@@ -22,6 +23,21 @@ defmodule Trento.ClusterProjector do
   alias Trento.Repo
 
   import Trento.Clusters, only: [enrich_cluster_model: 1]
+
+  project(
+    %ClusterDeregistered{
+      cluster_id: cluster_id,
+      deregistered_at: deregistered_at
+    },
+    fn multi ->
+      changeset =
+        ClusterReadModel.changeset(%ClusterReadModel{id: cluster_id}, %{
+          deregistered_at: deregistered_at
+        })
+
+      Ecto.Multi.update(multi, :cluster, changeset)
+    end
+  )
 
   project(
     %ClusterRegistered{
@@ -147,6 +163,16 @@ defmodule Trento.ClusterProjector do
     TrentoWeb.Endpoint.broadcast("monitoring:clusters", "cluster_health_changed", %{
       cluster_id: cluster_id,
       health: health
+    })
+  end
+
+  @impl true
+  def after_update(%ClusterDeregistered{cluster_id: cluster_id}, _, _) do
+    %ClusterReadModel{name: name} = Repo.get!(ClusterReadModel, cluster_id)
+
+    TrentoWeb.Endpoint.broadcast("monitoring:clusters", "cluster_deregistered", %{
+      cluster_id: cluster_id,
+      name: name
     })
   end
 
--- a/lib/trento/application/projectors/cluster_projector.ex
+++ b/lib/trento/application/projectors/cluster_projector.ex
@@ -25,21 +25,6 @@ defmodule Trento.ClusterProjector do
   import Trento.Clusters, only: [enrich_cluster_model: 1]
 
   project(
-    %ClusterDeregistered{
-      cluster_id: cluster_id,
-      deregistered_at: deregistered_at
-    },
-    fn multi ->
-      changeset =
-        ClusterReadModel.changeset(%ClusterReadModel{id: cluster_id}, %{
-          deregistered_at: deregistered_at
-        })
-
-      Ecto.Multi.update(multi, :cluster, changeset)
-    end
-  )
-
-  project(
     %ClusterRegistered{
       cluster_id: id,
       name: name,
@@ -66,6 +51,21 @@ defmodule Trento.ClusterProjector do
         })
 
       Ecto.Multi.insert(multi, :cluster, changeset)
+    end
+  )
+
+  project(
+    %ClusterDeregistered{
+      cluster_id: cluster_id,
+      deregistered_at: deregistered_at
+    },
+    fn multi ->
+      changeset =
+        ClusterReadModel.changeset(%ClusterReadModel{id: cluster_id}, %{
+          deregistered_at: deregistered_at
+        })
+
+      Ecto.Multi.update(multi, :cluster, changeset)
     end
   )
 
--- a/lib/trento/application/projectors/host_projector.ex
+++ b/lib/trento/application/projectors/host_projector.ex
@@ -19,6 +19,7 @@ defmodule Trento.HostProjector do
     HostDeregistered,
     HostDetailsUpdated,
     HostRegistered,
+    HostRemovedFromCluster,
     ProviderUpdated
   }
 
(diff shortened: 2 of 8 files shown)


test/trento/application/projectors/cluster_projector_test.exs
```diff
@@ -187,6 +188,18 @@ defmodule Trento.ClusterProjectorTest do
...
+  test "should update the deregistered_at field when ClusterDeregistered is received" do
+    insert(:cluster, id: cluster_id = Faker.UUID.v4())
+    deregistered_at = DateTime.utc_now()
+
+    event = ClusterDeregistered.new!(%{cluster_id: cluster_id, deregistered_at: deregistered_at})
+
+    ProjectorTestHelper.project(ClusterProjector, event, "cluster_projector")
+    cluster_projection = Repo.get!(ClusterReadModel, event.cluster_id)
+
+    assert event.deregistered_at == cluster_projection.deregistered_at
```
arbulu89, line 200: Could we test the broadcasted message?

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -29,16 +29,23 @@ defmodule Trento.Clusters do
 
   @spec request_checks_execution(String.t()) :: :ok | {:error, any}
   def request_checks_execution(cluster_id) do
-    ClusterReadModel
-    |> Repo.get(cluster_id)
+    query =
```
arbulu89, line 32: We don't need this assignment, right?
Cannot we pipe this with `Repo.one` already?
What is the `first` usage there by the way?
Anyway, I guess it comes from other functions in the file that do the same

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -60,7 +67,9 @@ defmodule Trento.Clusters do
     query =
       from(c in ClusterReadModel,
         select: c.id,
-        where: c.type == ^ClusterType.hana_scale_up() or c.type == ^ClusterType.hana_scale_out()
+        where:
+          c.type == ^ClusterType.hana_scale_up() or
```
arbulu89, line 71: This parenthesis look strange.
Shouldn't be?
`(c.type == ^ClusterType.hana_scale_up() or c.type == ^ClusterType.hana_scale_out()) and is_nil(c.deregistered_at)`
Otherwise scale up clusters are always requested, independently of the `deregisterd_at` field

test/trento/application/usecases/clusters_test.exs
```diff
@@ -64,6 +64,7 @@ defmodule Trento.ClustersTest do
       cluster_id = Faker.UUID.v4()
 
       insert(:cluster, id: cluster_id)
+      insert(:cluster, deregistered_at: DateTime.utc_now())
```
arbulu89, line 67: Personally, i would have added a new test for this. The feature is important enough to have some meaningful test. This looks pretty much hidden here

arbulu89 (commented):
I miss some more thorough test for the clusters use case honestly. For example, i think that the `request_clusters_checks_execution` condition is wrong, so the test would be helpful there.

Besides that, I think the `HostRemovedFromCluster` in the hosts projection is missing.
There, we should put the `cluster_id` of the host to nil.
I know that this cannot happen when you run the `HostDeregistrationRequest`, but if you deregister the cluster from the cli (or in the future with the delta deregistration), this scenario must be considered.

lib/trento/application/usecases/clusters/clusters.ex
CDimonaco: Done, nice catch!

test/trento/application/projectors/cluster_projector_test.exs
CDimonaco: done

test/trento/application/usecases/clusters_test.exs
CDimonaco: You are right, done

lib/trento/application/projectors/cluster_projector.ex
```diff
@@ -23,6 +24,21 @@ defmodule Trento.ClusterProjector do
 
   import Trento.Clusters, only: [enrich_cluster_model: 1]
 
+  project(
```
fabriziosestito, line 27: could we move this after the ClusterRegistered clause?

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -29,16 +29,22 @@ defmodule Trento.Clusters do
 
   @spec request_checks_execution(String.t()) :: :ok | {:error, any}
   def request_checks_execution(cluster_id) do
-    ClusterReadModel
-    |> Repo.get(cluster_id)
+    query =
+      from(c in ClusterReadModel,
```
fabriziosestito, line 33: a style nitpick, we do not use parenthesis here (locals w/o parens) if you remove them it should still work

fabriziosestito (changes requested):
Overall great work! Some nitpick and I agree with @arbulu89 comment about the `HostRemovedFromCluster` projection.

arbulu89:
> @arbulu89 I addressed the feedbacks but I don't really understand what do you mean here
> 
> > Besides that, I think the HostRemovedFromCluster in the hosts projection is missing.
> > There, we should put the cluster_id of the host to nil.
> > I know that this cannot happen when you run the HostDeregistrationRequest, but if you deregister the cluster from the cli (or in the future with the delta deregistration), this scenario must be considered.

In one of the latest https://github.com/trento-project/web/pull/1274 we have included the `HostRemovedFromCluster` event. This event happens when a host is removed from the cluster, that can lead to a total cluster deregistration, if it was the last host in the cluster. To host/cluster relationship in the frontend read-models is done with the `cluster_id` field in the host read model. So as far as we have removed the host, we should put this field to `nil`, so it reflects the real state of the host.

This is important if we have a standalone cluster de-registration (without de-registering the host itself). 

At the end we need to revert what we do here: https://github.com/trento-project/web/blob/deregistration/lib/trento/application/projectors/host_projector.ex#L66

PD: I don't know why we do this in the host projector if the event belongs to the cluster...

rtorrero (commented):
Besides from what's already been commented, it looks good @CDimonaco :+1:

CDimonaco:
@arbulu89 @fabriziosestito Done, please review again

arbulu89 (approved):
Great!
I proposed other alternative, but it should do the same (as the test passes hehe)

lib/trento/application/projectors/host_projector.ex
```diff
@@ -81,6 +82,20 @@ defmodule Trento.HostProjector do
     end
   )
 
+  project(
+    %HostRemovedFromCluster{
+      host_id: id
+    },
+    fn multi ->
+      query =
+        Ecto.Query.from(h in HostReadModel,
```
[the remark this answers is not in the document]
CDimonaco: Yes but no, I mean we don't know the order of the events, and we want to make sure to not update anything that is already soft delete, to mantain consistency and logic.

This guard will just make that happen. Right know I know that this could potentially be a no-op, but I don't really like the idea to pollute deleted entities :D 

Anyway if you like it we can merge.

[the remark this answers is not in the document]
arbulu89: I don't really think this could ever happen. You cannot de-register a host without de-registering it first from a cluster (and in any case, what you "say" about polluting would be reflecting the source of truth in that strange scenario XD)
I guess that the cost of doing an extra query is low.
Anyway, you already had my green light hehe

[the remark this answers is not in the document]
fabriziosestito: @arbulu89 @CDimonaco I think it's my fault on this one.
We cannot guarantee the order of the events, since HostRemovedFromCluster is emitted by the Cluster aggregate and might arrive after the HostDeregistered event which is emitted concurrently by the Host aggregate. I was confused and thought we were deleting the entry for good, therefore my suggestion was to use `update_all`.

I still prefer Xabi's solution since the host is just soft deleted so the read model can be found. It doesn't matter if we update a soft deleted read model, actually, we must update it to have a consistent entry if we then plan to roll back the deregistration.

CDimonaco:
I strongly disagree on updating a soft delete model, if this is the only way we have to rollback I will do it but it’s maybe the symptom of other design problems.

Anyway, for the sake of pr I will update the code so we can close it.

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

arbulu89 (approved):

PR #1290, opened 2023-03-29
Refactor controllers and contexts error handling

fabriziosestito:
Refactor web error handling:
- Make controllers agnostic of domain error
- Make fallback controller aware of domain error
- Fix error handling in certain contexts, improve logging
- Add changeset rendering in ErrorView
- use a tagged tuple to return validation errors

### DIFF
--- a/lib/trento/application/usecases/clusters/clusters.ex
+++ b/lib/trento/application/usecases/clusters/clusters.ex
@@ -22,16 +22,27 @@ defmodule Trento.Clusters do
 
   @spec select_checks(String.t(), [String.t()]) :: :ok | {:error, any}
   def select_checks(cluster_id, checks) do
+    Logger.debug("Selecting checks, cluster: #{cluster_id}")
+
     with {:ok, command} <- SelectChecks.new(%{cluster_id: cluster_id, checks: checks}) do
       commanded().dispatch(command)
     end
   end
 
-  @spec request_checks_execution(String.t()) :: :ok | {:error, any}
+  @spec request_checks_execution(String.t()) ::
+          :ok | {:error, any}
   def request_checks_execution(cluster_id) do
-    ClusterReadModel
-    |> Repo.get(cluster_id)
-    |> maybe_request_checks_execution()
+    case Repo.get(ClusterReadModel, cluster_id) do
+      %ClusterReadModel{} = cluster ->
+        Logger.debug("Requesting checks execution, cluster: #{cluster_id}")
+
+        maybe_request_checks_execution(cluster)
+
+      nil ->
+        Logger.error("Requested checks execution for a non-existing cluster: #{cluster_id}")
+
+        {:error, :not_found}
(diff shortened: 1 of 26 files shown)


lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -71,7 +82,9 @@ defmodule Trento.Clusters do
           :ok
 
         {:error, reason} ->
-          Logger.error("Failed to request checks execution for cluster #{cluster_id}: #{reason}")
+          Logger.error(
```
arbulu89, line 85: Not something you should do in this PR, but we should reduce the errors in this `each` now and return error if any of them fails.
Or this return the error if fails?

lib/trento/application/usecases/hosts/heartbeats.ex
```diff
@@ -28,15 +28,29 @@ defmodule Trento.Heartbeats do
...
+
+    case result do
+      {:ok, _} ->
+        :ok
+
+      {:error, :command, :host_not_registered, _} ->
+        {:error, :not_found}
+
+      error ->
+        Logger.error("Error while updating heartbeat for agent #{agent_id}", error: inspect(error))
```
arbulu89, line 50: Nothing wrong here, but i see that we use the error metadata here, but in every other place we include the error in the main message text

lib/trento_web/controllers/fallback_controller.ex
```diff
@@ -3,31 +3,60 @@ defmodule TrentoWeb.FallbackController do
...
     conn
     |> put_status(:not_found)
     |> put_view(ErrorView)
-    |> render(:"404", detail: detail)
+    |> render(:"404")
   end
 
-  def call(conn, {:error, {:unauthorized, detail}}) do
+  def call(conn, {:error, reason})
+      when reason in [:host_not_registered, :cluster_not_registered, :sap_system_not_registered] do
```
arbulu89, line 28: I hope this list doesn't convert on a root for bugs hehe

arbulu89 (approved):
Cool!

As you have added/updated error logging, did you check if we are capturing them in the tests? Just to have a clean tests output, and not filled with "expected" error messages

lib/trento/application/usecases/hosts/heartbeats.ex
fabriziosestito: ah yes, we did not decided yet where and when to use metadata so I will add the error to the message and defer this decision

lib/trento/application/usecases/clusters/clusters.ex
fabriziosestito: this is used by the job scheduler in a best-effort manner, this is why we just use the each loop and we return nothing.
If something went wrong while scheduling the execution we log it

nelsonkopliku (approved):
Nice job, thanks!

PR #1303, opened 2023-04-05
Button stories

dottorblaster:
Just some Storybook stories for the `Button` component.

immagine


assets/js/components/Button/Button.stories.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+
+import Button from '.';
+
+export default {
+  /* 👇 The title prop is optional.
```
EMaksy, line 6: do we want to have the comments?

EMaksy (approved):
Looks good to me :D

assets/js/components/Button/Button.stories.jsx
dottorblaster: No, you're right :+1:

PR #1307, opened 2023-04-06
Added Tooltip story

jagabomb:
# Description

Added Tooltip story to enhance Storybook. I also noticed some issues with Tooltip component which is solved by wrapping the component into additional `divs` so that it will to display properly.

## How was this tested?

Visually

Screenshot 2023-04-06 at 11 07 55


EMaksy (approved):
A great start, maybe we can add more cases on a later iteration.
Other than that, LGTM!

PR #1308, opened 2023-04-06
Add PremiumPill directory with a story

EMaksy:
# Description

Create a new directory for the PremiumPill component and add a story for the storybook

## How was this tested?
Visually


assets/js/components/PremiumPill/PremiumPill.stories.jsx
```diff
@@ -0,0 +1,23 @@
...
+export function PremiumPillDefault() {
+  return <PremiumPill />;
+}
+
+export function PremiumPillRed() {
+  return <PremiumPill className="bg-red-100" />;
+}
+
+export function PremiumPillYellow() {
+  return <PremiumPill className="bg-yellow-100 text-black" />;
```
jagabomb, line 18: The text colour class here has no effect... with that being said I would suggest a dark yellow text class like `text-yellow-600`. That way the text and background colours are related. See below for a reference screenshot which includes a red version too.
Screenshot 2023-04-06 at 15 07 10

assets/js/components/PremiumPill/PremiumPill.stories.jsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export function PremiumPillWithMarginLeft() {
+  return <PremiumPill className="ml-6" />;
```
jagabomb, line 22: I would question the relevance of this variation at an atomical level as it seems redundant. I think the scenario is more applicable in a full page layout. @dottorblaster What do you think?

EMaksy: I have a similar opinion, but why did I put it in ? I put it in because we use it in two places in the project.
The main question is, should we cover everything how it is used in the project or how you suggest @jagabomb  and leave it out?
@dottorblaster  what do you think ?

jagabomb: @EMaksy In this instance it is so minimal and the margin is something that can change dependent on the page application. This why I feel it is stronger if you show a full page with the application of the PremiumPill but once again it is out of scope for this story in my opinion. So I would say leave it out. Let's see what @dottorblaster advises.

assets/js/components/PremiumPill/PremiumPill.stories.jsx
EMaksy: Yeah good idea :D

assets/js/components/PremiumPill/PremiumPill.stories.jsx
dottorblaster: I see you took your decision without my advice :D anyway yes it was pretty redundant, @jagabomb was right, don't worry @EMaksy!

jagabomb (approved):
LGTM

PR #1326, opened 2023-04-13
Database registration changes

CDimonaco:
# Description

This PR aims to change the database registration process in sap systems.
We want to ignore database instances that are Secondary in a system replicated scenario.

We accept only database instances that are **not replicated** or **Primary** instances in a replicated scenario.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/application/usecases/clusters/clusters.ex
+++ b/lib/trento/application/usecases/clusters/clusters.ex
@@ -126,10 +126,9 @@ defmodule Trento.Clusters do
        }) do
     hosts_data =
       Repo.all(
-        from(h in HostReadModel,
+        from h in HostReadModel,
           select: %{host_id: h.id},
           where: h.cluster_id == ^cluster_id and is_nil(h.deregistered_at)
-        )
       )
 
     Checks.request_execution(
--- a/lib/trento/application/usecases/clusters/clusters.ex
+++ b/lib/trento/application/usecases/clusters/clusters.ex
@@ -126,9 +126,10 @@ defmodule Trento.Clusters do
        }) do
     hosts_data =
       Repo.all(
-        from h in HostReadModel,
+        from(h in HostReadModel,
           select: %{host_id: h.id},
           where: h.cluster_id == ^cluster_id and is_nil(h.deregistered_at)
+        )
       )
 
     Checks.request_execution(
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -90,10 +90,9 @@ defmodule Trento.Domain.SapSystem do
   end
 
   def execute(
-        %SapSystem{},
+        %SapSystem{sap_system_id: nil},
         %RegisterDatabaseInstance{
-          system_replication: "Secondary",
-          system_replication_status: "ACTIVE"
+          system_replication: "Secondary"
         }
       ),
       do: {:error, :sap_system_not_registered}
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -96,7 +96,7 @@ defmodule Trento.Domain.SapSystem do
           system_replication_status: "ACTIVE"
         }
       ),
-      do: {:error, :database_instance_not_primary}
+      do: {:error, :sap_system_not_registered}
 
   # First time that a Database instance is registered, the SAP System starts its registration process.
   # We accept the database instance when the system replication is disabled or when enabled, only if the database
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -98,7 +98,7 @@ defmodule Trento.Domain.SapSystem do
       do: {:error, :sap_system_not_registered}
 
   # First time that a Database instance is registered, the SAP System starts its registration process.
-  # Database instance are accepted when the system replication is disabled or when enabled, only if the database
+  # Database instances are accepted when the system replication is disabled or when enabled, only if the database
   # has a primary role
   # When an Application is discovered, the SAP System completes the registration process.
   def execute(
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -28,8 +28,9 @@ defmodule Trento.Domain.SapSystem do
 
   That being said, this is the logical order of events in order to register a full system:
 
-  1. A SAP system discovery message with a new database instance is received. At this point, the
-     registration process starts and the database is registered.
+  1. A SAP system discovery message with a new database instance is received.
+     We ignore database with Secondary role in a system replicated scenarios.
+     At this point, the registration process starts and the database is registered.
      Any application instance discovery message without an associated database is ignored.
   2. New database instances/updates coming from already registered database instances are registered/applied.
   3. A SAP system discovery with a new application instance is received, and the database associated to
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -29,7 +29,7 @@ defmodule Trento.Domain.SapSystem do
   That being said, this is the logical order of events in order to register a full system:
 
   1. A SAP system discovery message with a new database instance is received.
-     We ignore database with Secondary role in a system replicated scenarios.
+     Database instances with Secondary role in a system replication scenario are discarded.
      At this point, the registration process starts and the database is registered.
      Any application instance discovery message without an associated database is ignored.
   2. New database instances/updates coming from already registered database instances are registered/applied.
@@ -99,7 +99,7 @@ defmodule Trento.Domain.SapSystem do
       do: {:error, :sap_system_not_registered}
 
   # First time that a Database instance is registered, the SAP System starts its registration process.
-  # We accept the database instance when the system replication is disabled or when enabled, only if the database
+  # Database instance are accepted when the system replication is disabled or when enabled, only if the database
   # has a primary role
   # When an Application is discovered, the SAP System completes the registration process.
   def execute(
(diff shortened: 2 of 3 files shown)


lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -88,7 +88,18 @@ defmodule Trento.Domain.SapSystem do
...
   end
 
+  def execute(
+        %SapSystem{},
+        %RegisterDatabaseInstance{
+          system_replication: "Secondary",
+          system_replication_status: "ACTIVE"
+        }
+      ),
+      do: {:error, :database_instance_not_primary}
```
fabriziosestito, line 99: The error should be `:sap_system_not_registered` to indicate that an instance was discovered but no sap system was registered yet. Keep in mind that those domain errors are then bubbling up to the fallback controller, in this case we return a 404.

fabriziosestito (changes requested):

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: Right, fix coming asap

CDimonaco: done

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -28,8 +28,9 @@ defmodule Trento.Domain.SapSystem do
 
   That being said, this is the logical order of events in order to register a full system:
 
-  1. A SAP system discovery message with a new database instance is received. At this point, the
-     registration process starts and the database is registered.
+  1. A SAP system discovery message with a new database instance is received.
+     We ignore database with Secondary role in a system replicated scenarios.
```
arbulu89, line 32: I would avoid using `we` term. `Database instances with Secondary role in a system replication scenario are discarded`

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -88,7 +89,18 @@ defmodule Trento.Domain.SapSystem do
     embeds_one :application, Application
   end
 
+  def execute(
+        %SapSystem{},
+        %RegisterDatabaseInstance{
+          system_replication: "Secondary",
+          system_replication_status: "ACTIVE"
```
arbulu89, line 96: I wouldn't honestly check the status field. I would only check what the sap admin configured, which is the role, `Primary/Secondary`. 
Otherwise, we make everything much more complicated, even in the deregistration, as we would need to handle more scenarios.

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -88,7 +89,18 @@ defmodule Trento.Domain.SapSystem do
     embeds_one :application, Application
   end
 
+  def execute(
+        %SapSystem{},
```
arbulu89, line 93: Shouldn't we match with `%SapSystem{sap_system_id: nil},` ? otherwise, what happens with sap systems with primary instances registered?

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -88,7 +89,18 @@ defmodule Trento.Domain.SapSystem do
...
+        %RegisterDatabaseInstance{
+          system_replication: "Secondary",
+          system_replication_status: "ACTIVE"
+        }
+      ),
+      do: {:error, :sap_system_not_registered}
+
   # First time that a Database instance is registered, the SAP System starts its registration process.
+  # We accept the database instance when the system replication is disabled or when enabled, only if the database
```
arbulu89, line 102: Let's update the `we`

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -126,9 +126,10 @@ defmodule Trento.Clusters do
        }) do
     hosts_data =
       Repo.all(
-        from h in HostReadModel,
+        from(h in HostReadModel,
```
arbulu89, line 129: why this change? we have a lot of entries without parenthesis

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -26,7 +26,138 @@ defmodule Trento.SapSystemTest do
   alias Trento.Domain.SapSystem
 
   describe "SAP System registration" do
-    test "should create an incomplete SAP system aggregate and register a database instance" do
+    test "should fail when a sap system not exists and the database instance has Secondary role with system replication ACTIVE" do
```
arbulu89, line 29: `should discard sap system discovery messages from a Secondary role host if the system is not registered`

or at least change `system not exists` by `system which does not exist`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -26,7 +26,138 @@ defmodule Trento.SapSystemTest do
...
+          system_replication_status: "ACTIVE"
+        )
+
+      assert_error(
+        command,
+        {:error, :sap_system_not_registered}
+      )
+    end
+
+    test "should fail when a sap system already exists and the database instance has Secondary role with system replication ACTIVE" do
```
arbulu89, line 42: I don't understand this scenario. Once we have a sap system with a primary, other secondary nodes should be added to the instances list

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -26,7 +26,138 @@ defmodule Trento.SapSystemTest do
...
+          sid: sid,
+          tenant: tenant,
+          instance_number: instance_number,
+          instance_hostname: instance_hostname,
+          features: features,
+          http_port: http_port,
+          https_port: https_port,
+          start_priority: start_priority,
+          host_id: host_id,
+          system_replication: "Primary",
```
arbulu89, line 110: In this case the system replication is enabled, as you have the `Primary` thing.
In a non SR setup, both `system_replication` and `system_replication_status` should be set to nil

arbulu89 (changes requested):
Many comments from my side. I find the logic wrong.
Besides that, I wouldn't use the `status` field, as this is the runtime status of the system replication. With the `system_replication` field, which is what the sap admin sets should be enough

lib/trento/application/usecases/clusters/clusters.ex
CDimonaco: Idk, I think the rebase did some magic, here

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: set accordingly to the test description

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: yes, removed

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: fixed

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: done

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: done man thanks

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: rixed

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: done

CDimonaco:
Changed, thanks guys, ready to re-review

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -88,7 +89,17 @@ defmodule Trento.Domain.SapSystem do
...
+  def execute(
+        %SapSystem{sap_system_id: nil},
+        %RegisterDatabaseInstance{
+          system_replication: "Secondary"
+        }
+      ),
+      do: {:error, :sap_system_not_registered}
+
   # First time that a Database instance is registered, the SAP System starts its registration process.
+  # Database instance are accepted when the system replication is disabled or when enabled, only if the database
```
arbulu89, line 101: `Database instance"s"` add s

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -26,7 +26,93 @@ defmodule Trento.SapSystemTest do
   alias Trento.Domain.SapSystem
 
   describe "SAP System registration" do
-    test "should create an incomplete SAP system aggregate and register a database instance" do
+    test "should fail when a sap system not exists and the database instance has Secondary role" do
```
arbulu89, line 29: I would add a new test case to see that Secondary nodes are registered once the primary is there.
I think we miss this part, and it would be good to check

lib/trento/application/usecases/clusters/clusters.ex
fabriziosestito: please remove the parenthesis, the formatter did something wrong

test/trento/domain/sap_system/sap_system_test.exs
fabriziosestito: we already have this test: https://github.com/trento-project/web/blob/9f8e805d08433d0b53a584f7ba3802467840f709/test/trento/domain/sap_system/sap_system_test.exs#L190 

I wouldn't add an extra test since it will be redundant

arbulu89: Alright! It is not explicit, but good enough. I buy your point hehe

arbulu89 (approved):
Approving. Please fix the test description before merging

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: done

test/trento/domain/sap_system/sap_system_test.exs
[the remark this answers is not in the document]
CDimonaco: done

lib/trento/application/usecases/clusters/clusters.ex
CDimonaco: done

fabriziosestito (approved):

PR #1335, opened 2023-04-17
Make api key enablement configurable in runtime

arbulu89:
Make `api_key` usage enabled option configurable in runtime. It will come handy for e2e testing scenarios.

PD:
The other option was to provide the option to use a configurable api_key base value, but we decided to not implement this, as the current api_key implementation is not that good, and we might change it at some point, so going with the easy `disable` option was preferred. In any case, the api_key depends on the `secret_key_base`, so both this and the custom `api_key` should be fixed to have a unique key

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -79,8 +79,6 @@ config :trento, :pow,
   extensions: [PowPersistentSession],
   controller_callbacks: Pow.Extension.Phoenix.ControllerCallbacks
 
-config :trento, :api_key_authentication, enabled: true
-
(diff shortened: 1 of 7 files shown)


fabriziosestito (approved):

arbulu89:
@fabriziosestito At the end, I have changed strategy :sweat_smile: 
We cannot change a variable that was fetched during compilation (`compile_env`).
I have opted to go with `unplug` as we do with wanda so we can enable/disable phoenix plugs on demand with runtime changes.
This time I tested properly and it works :rabbit:

fabriziosestito (approved):

PR #1336, opened 2023-04-17
Add some margin between cluster-wide facts

nelsonkopliku:
Nothing fancy. Just re-adding some space between cluster wide gathered facts.

From
image

To
image


EMaksy (approved):
LGTM :D

PR #1337, opened 2023-04-17
[POC] HANA scale-up e2e tests with a real cluster 

fabriziosestito:
This PR is a PoC that adds an automated test that spins up a HANA scale-up cluster.
The test involves running checks and comparing the output with the expected results.

It can be used as a base to script other automated tests or scenarios.

It uses Terraform to deploy a real infrastructure on AWS, ansible to deploy trento server and the agents, and cypress to run the e2e tests.
Destroying the infrastructure requires a workaround since the VPC is created with some dependencies that Terraform doesn't seem to be able to manage.
I've added a Cypress env variable to skip photofinish scenario loading at the beginning of the test, and to make the test skippable in the current e2e tests. Maybe in the future we should separate the test folders.
As for now, the action can be triggered by a `workflow_dispatch` event  (manually) and only works with the rolling version of trento. 
In the future, we could consider deploying a specific commit, generalize the version for another scenario, and use [composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action).

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,0 +1,135 @@
+# This workflow requires the following secrets:
+# `SSH_PUBLIC_KEY`, `SSH_PRIVATE_KEY` ssh keys to access the e2e machines
+# `AWS_SECRET_ACCESS_KEY`, `AWS_ACCESS_KEY_ID` aws credentials, 
+# used by terraform to deploy the infrastructure.
+# 
+# An artifact containing the terraform state is uploaded in case something goes wrong, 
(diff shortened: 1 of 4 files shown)


arbulu89 (commented):
Great job @fabriziosestito 
Really valuable
PD: I think unfortunately already got outdated with the latest changes on the `trento-ansible` project

arbulu89 (approved):
Great!
So this is only dispatched on demand, right?
Super cool

nelsonkopliku (approved):
Cool! :rocket:

PR #1343, opened 2023-04-18
SAP system deregistration aggregate changes

rtorrero:
# Description

This PR adds the deregistration logic to the sap_system aggregate. Upon deregistering a host, it should remove the relevant application and/or database instances and eventually sap systems, depending on the role of the deregistered host.

## How was this tested?
Based on the deregistration flow char, all of the possible scenarios have been added to the test cases.

### DIFF
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -54,19 +54,25 @@ defmodule Trento.Domain.SapSystem do
   }
 
   alias Trento.Domain.Commands.{
+    DeregisterApplicationInstance,
+    DeregisterDatabaseInstance,
     RegisterApplicationInstance,
     RegisterDatabaseInstance,
     RollUpSapSystem
   }
 
   alias Trento.Domain.Events.{
+    ApplicationInstanceDeregistered,
     ApplicationInstanceHealthChanged,
     ApplicationInstanceRegistered,
+    DatabaseDeregistered,
     DatabaseHealthChanged,
+    DatabaseInstanceDeregistered,
     DatabaseInstanceHealthChanged,
     DatabaseInstanceRegistered,
     DatabaseInstanceSystemReplicationChanged,
     DatabaseRegistered,
+    SapSystemDeregistered,
     SapSystemHealthChanged,
     SapSystemRegistered,
     SapSystemRolledUp,
@@ -80,13 +86,14 @@ defmodule Trento.Domain.SapSystem do
   use Trento.Type
 
   deftype do
-    field :sap_system_id, Ecto.UUID
-    field :sid, :string
-    field :health, Ecto.Enum, values: Health.values()
-    field :rolling_up, :boolean, default: false
+    field(:sap_system_id, Ecto.UUID)
+    field(:sid, :string)
+    field(:health, Ecto.Enum, values: Health.values())
+    field(:rolling_up, :boolean, default: false)
+    field(:deregistered_at, :utc_datetime_usec, default: nil)
 
-    embeds_one :database, Database
-    embeds_one :application, Application
+    embeds_one(:database, Database)
+    embeds_one(:application, Application)
   end
 
   def execute(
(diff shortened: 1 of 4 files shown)


lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +297,188 @@ defmodule Trento.Domain.SapSystem do
...
+    end)
+    |> Multi.execute(fn sap_system ->
+      maybe_emit_application_instances_deregistered_events(
+        sap_system,
+        command,
+        deregistered_instance_role
+      )
+    end)
+    |> Multi.execute(fn sap_system ->
+      maybe_emit_sap_system_deregistered_event(sap_system, command)
```
arbulu89, line 340: Do we send the whole command for some reason? Only the `deregistered_at` field is used in the command.
The same for the `maybe_` function above

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +297,188 @@ defmodule Trento.Domain.SapSystem do
...
+          application: %Application{instances: instances}
+        } = sap_system,
+        %DeregisterApplicationInstance{
+          instance_number: instance_number,
+          sap_system_id: sap_system_id,
+          host_id: host_id,
+          deregistered_at: deregistered_at
+        } = command
+      ) do
+    instance = get_instance(instances, host_id, instance_number)
```
arbulu89, line 314: For me personally this function looks really non declarative and confusing.

In my view this should look like:
```
    sap_system
    |> Multi.new()
    |> Multi.execute(fn _ ->
      %ApplicationInstanceDeregistered{
        sap_system_id: sap_system_id,
        ...
      }
    |> Multi.execute(fn sap_system ->
      maybe_emit_sap_system_deregistered_events(
        sap_system,
        deregistered_at
      )
```

And that's all. 
You check all the logic about what to do with the sap system at `maybe_emit_sap_system_deregistered_events`, where you check if instances of application or database are nil, or there is not any instance with `ascs/pas` type.
It is much easier to check what's in the current state rather than in the incoming command.

It would simplify a lot and make the code meaningful

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +297,188 @@ defmodule Trento.Domain.SapSystem do
...
+  def apply(
+        %SapSystem{database: %Database{instances: []}, application: %Application{instances: []}} =
+          sap_system,
+        %DatabaseDeregistered{
+          sap_system_id: sap_system_id
+        }
+      ) do
+    %SapSystem{
+      sap_system
+      | sap_system_id: sap_system_id,
```
arbulu89, line 448: You don't need to add the `sap_system_id` again, it already had this id

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +297,188 @@ defmodule Trento.Domain.SapSystem do
...
+
+    %SapSystem{
+      sap_system
+      | application: %Application{
+          instances: instances
+        }
+    }
+  end
+
+  def apply(
```
arbulu89, line 439: Do we really need this apply?
Cannot we use the next 2 to set database and application to nil?

arbulu89 (commented):
Hey @rtorrero and @jamie-suse ,
I did an initial (superficial) review, and I have to confess that I found the code overly complicated. I see too many `maybe_` functions, to many `imperative` things (instead of using the state to check what's going on).

I will wait until other reviews, but I think that we should sit and revisit the whole logic, to make the code cleaner

fabriziosestito (changes requested):
I strongly agree with @arbulu89, the domain logic needs a rework as it can be simplified a lot.
I find the tests good enough to drive this refactor, maybe after the rework we could think of removing some redundancy there as well.

rtorrero:
Closing this PR as the code has been merged in two separate PRs: #1357 and #1372

PR #1348, opened 2023-04-19
Add single-hana-single-app dumped scenario

arbulu89:
Add single-hana-single-app dumped scenario, to be used by `photofinish`.

The scenario contains:
- 1 machine with a single HANA (no system replication)
- 1 machine with ASCS and PAS instances (no HA)
- Both machines are running on aws

To run:
```
./photofinish run --url "http://localhost:4000/api/v1/collect" single-hana-single-app
```

> Sending the PR to `main` but you can rebase in the deregistration branch if you want to use it there

### DIFF
--- a/.photofinish.toml
+++ b/.photofinish.toml
@@ -2,6 +2,10 @@
 
 directories = ["./test/fixtures/scenarios/healthy-27-node-SAP-cluster"]
 
+[single-hana-single-app]
+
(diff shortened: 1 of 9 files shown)


fabriziosestito (approved):
lovely!

dottorblaster (approved):
AW YEEEEAH

PR #1356, opened 2023-04-21
Add SAP System de/registration logic to Process Manager

jamie-suse:
# Description

Adds behaviour for processing `DatabaseInstanceRegistered`, `ApplicationInstanceRegistered`, `SapSystemRolledUp`, `DatabaseInstanceDeregistered` & `ApplicationInstanceDeregistered` events.

Adds `application_instances` & `database_instances` lists to the Process Manager state.

## How was this tested?

Added unit tests for each scenario.

### DIFF
--- a/lib/trento/application/process_managers/deregistration_process_manager.ex
+++ b/lib/trento/application/process_managers/deregistration_process_manager.ex
@@ -4,12 +4,38 @@ defmodule Trento.DeregistrationProcessManager do
     for the deregistration procedure for the aggregates
 
     This represents a transaction to ensure that the procedure of deregistering domain aggregates
-    follows a certain path and satisfies some requisities.
+    follows a certain path and satisfies some requisites.
 
     For more information see https://hexdocs.pm/commanded/process-managers.html
   """
 
-  @required_fields []
+  defmodule Instance do
+    @moduledoc """
+    An application or database instance and which SAP System it belongs to.
+    """
+    @required_fields :all
+    use Trento.Type
+
+    deftype do
+      field :sap_system_id, Ecto.UUID
+      field :instance_number, :string
+    end
+  end
+
+  defmodule Instance do
+    @moduledoc """
+    An application or database instance and which SAP System it belongs to.
+    """
+    @required_fields :all
+    use Trento.Type
+
+    deftype do
+      field :sap_system_id, Ecto.UUID
+      field :instance_number, :string
+    end
+  end
+
+  @required_fields :all
 
   use Trento.Type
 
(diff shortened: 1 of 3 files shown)


lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -84,12 +138,147 @@ defmodule Trento.DeregistrationProcessManager do
...
+          sap_system_id: sap_system_id,
+          snapshot: %SapSystem{
+            database: %SapSystem.Database{instances: database_instances},
+            application: %SapSystem.Application{instances: application_instances}
+          }
+        }
+      ) do
+    %DeregistrationProcessManager{
+      state
+      | database_instances:
```
arbulu89, line 271: This logic looks wrong. You are replacing completely the database and application instances, but be aware that one host can have applications/databases from different sap systems.
So you would need to get other instances from other sap systems and concat with the ones coming in the rollup

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -84,12 +138,147 @@ defmodule Trento.DeregistrationProcessManager do
...
+          instance_number: instance_number,
+          host_id: host_id,
+          deregistered_at: requested_at
+        }
+      end),
+      %DeregisterHost{host_id: host_id, deregistered_at: requested_at}
+    ])
+  end
+
+  def apply(
```
arbulu89, line 185: I got a bit lost with the order of the applies.
We should still continue having the same order like: host - cluster - sap system function. Or grouped some other way, but I found that now they are just completely unsorted

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -2,24 +2,55 @@ defmodule Trento.DeregistrationProcessManagerTest do
...
 
   alias Trento.Domain.Commands.{
+    DeregisterApplicationInstance,
     DeregisterClusterHost,
+    DeregisterDatabaseInstance,
     DeregisterHost
   }
 
   describe "events interested" do
+    test "should start the process manager when DatabaseInstanceRegistered event arrives" do
```
arbulu89, line 36: Let's try to put a correct order. This tests are last to add, so put them at the botton of this `describe` please

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -41,6 +72,47 @@ defmodule Trento.DeregistrationProcessManagerTest do
                DeregistrationProcessManager.interested?(%HostAddedToCluster{host_id: host_id})
     end
 
+    test "should start process managers when SapSystemRolledUp arrives" do
+      db_host_id = UUID.uuid4()
+      app_host_id = UUID.uuid4()
+
+      database_instances = [
+        %SapSystem.Instance{
```
arbulu89, line 80: Could be use factories to create this elements?

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -74,13 +146,83 @@ defmodule Trento.DeregistrationProcessManagerTest do
     end
   end
 
+  describe "host registration procedure" do
+    test "should add database instance with the proper host id when DatabaseInstanceRegistered event is emitted" do
```
arbulu89, line 150: I miss a bit the fact the we don't test that the new instance is _added_ to the list. We always start with an empty list, which could somehow hide issues

arbulu89 (changes requested):
Hey @jamie-suse ,
Some initial comments. I think that we have some holes in our logic.
PD: I didn't finish reviewing the tests code. I will once the prod code is ready. Anyway, I would like to have a more consistent order of code additions. I wouldn't put new code in top of already existing code if there is not a logical reason.

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -19,61 +32,108 @@ defmodule Trento.DeregistrationProcessManager do
...
-    The process manager is interested in HostRegistered which starts or joins an existing process
-    manager for the host identified by a host_id field.
-    The process manager is interested in HostDeregistered which stops a process manager for the host identified by host_id.
+  alias Trento.Domain.SapSystem
 
-    We consider also the host rollup case, starting a process manager when the host rolled up arrived, so we identify a registered host
-    without the HostRegistered event, because that event could be rolled up and then we have to consider also the rolled up host as registered.
+  @doc """
+    The Process Manager is started by the following events (provided the instance hasn't been started already):
+    - HostRegistered starts a Process Manager for a newly registered host.
```
arbulu89, line 67: The new doc looks great.
So many `starts a Process Manager` look redundant.
What about:
```
- HostRegistered for a newly registered host/
- HostAddedToCluster when a Host gets added to a cluster, as this event may arrive prior to ...
- DatabaseInstanceRegistered when an instance gets...
```

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -84,10 +144,47 @@ defmodule Trento.DeregistrationProcessManager do
...
+          database_instances: database_instances,
+          application_instances: application_instances
+        },
+        %HostDeregistrationRequested{
+          host_id: host_id,
+          requested_at: requested_at
+        }
+      ) do
+    List.flatten([
+      %DeregisterClusterHost{
```
arbulu89, line 159: We still have the scenario where this `cluster_id` could be nil. We need to take care of that.

test/support/factory.ex
```diff
@@ -349,6 +350,16 @@ defmodule Trento.Factory do
     build(:application_instance_without_host, host_id: host.id, host: host)
   end
 
+  def sap_system_instance_factory do
+    %SapSystem.Instance{
+      sid: Faker.UUID.v4(),
+      instance_number: String.pad_leading(sequence(:instance_number, &"#{&1}"), 2, "0"),
+      features: Faker.Pokemon.name(),
+      host_id: Faker.UUID.v4(),
+      health: :passing
```
arbulu89, line 359: `Health.passing()`

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -50,6 +65,49 @@ defmodule Trento.DeregistrationProcessManagerTest do
...
+                 host_id: host_id
+               })
+    end
+
+    test "should start process managers when SapSystemRolledUp arrives" do
+      db_host_id = UUID.uuid4()
+      app_host_id = UUID.uuid4()
+
+      database_instances = [
+        build(:sap_system_instance, host_id: db_host_id)
```
arbulu89, line 91: Let's be more generous here and create a list of instances. At the end, we do operations with them

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -59,48 +117,200 @@ defmodule Trento.DeregistrationProcessManagerTest do
...
+        %ClusterRolledUp{cluster_id: cluster_id, snapshot: %Cluster{hosts: cluster_hosts}}
+      ]
 
       {commands, state} = reduce_events(events, initial_state)
 
       assert [] == commands
       assert %DeregistrationProcessManager{cluster_id: ^cluster_id} = state
     end
 
+    test "should update state after when SapSystemRolledUp event received" do
```
arbulu89, line 260: remove `after`

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -59,48 +117,200 @@ defmodule Trento.DeregistrationProcessManagerTest do
...
+      sap_system_id = UUID.uuid4()
+      database_instance_number = "00"
+      application_instance_number = "01"
+
+      initial_state = %DeregistrationProcessManager{
+        database_instances: [],
```
arbulu89, line 266: Maybe add at least one instance to database_instance or application_instance, to see that it adds the values properly

arbulu89 (changes requested):
Hey @jamie-suse ,
in general terms it look great.
Only missing to handle properly some other scenario when the cluster is not there but app/database are

rtorrero (approved):
LGTM

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -19,61 +32,108 @@ defmodule Trento.DeregistrationProcessManager do
...
+      ) do
     %DeregisterHost{host_id: host_id, deregistered_at: requested_at}
   end
 
-  # First step in host deregistration when host belongs to a cluster
-  def handle(%DeregistrationProcessManager{cluster_id: cluster_id}, %HostDeregistrationRequested{
-        host_id: host_id,
-        requested_at: requested_at
-      }) do
+  def handle(
```
arbulu89, line 126: With the next function this one is not needed anymore, right?

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -84,10 +144,56 @@ defmodule Trento.DeregistrationProcessManager do
...
+        %DeregistrationProcessManager{
+          cluster_id: cluster_id,
+          database_instances: database_instances,
+          application_instances: application_instances
+        },
+        %HostDeregistrationRequested{
+          host_id: host_id,
+          requested_at: requested_at
+        }
+      ) do
```
arbulu89, line 157: What about:
```
List.flatten([
    Enum.map(database_instances, fn %Instance{
    ...
    end),
    Enum.map(application_instances, fn %Instance{
    ...
    end),
    if cluster_id do
        %DeregisterClusterHost{...
    end,
    %DeregisterHost{...
])
```
This way we respect the logical order of commands as well
The `if` condition smells a bit and maybe there is a better way of doing this with pattern matching.

test/trento/application/process_managers/deregistration_process_manager_test.exs
```diff
@@ -50,6 +65,53 @@ defmodule Trento.DeregistrationProcessManagerTest do
...
+               })
+    end
+
+    test "should start process managers when SapSystemRolledUp arrives" do
+      db_host_id_1 = UUID.uuid4()
+      db_host_id_2 = UUID.uuid4()
+      app_host_id_1 = UUID.uuid4()
+      app_host_id_2 = UUID.uuid4()
+
+      database_instances = [
```
arbulu89, line 92: This looks an antipattern:
Maybe:
```
[%{host_id: db_host_id_1}, %{host_id: db_host_id_2}] = database_instances = build_list(2, :sap_system_instance)
```
And so on. In this case (and almost always when you use factories), giving pre-created unique id-s to the created objects doesn't make sense. Let the factory do that work for you. You can capture later on these values.

arbulu89 (changes requested):
Some final improvements from my side

lib/trento/application/process_managers/deregistration_process_manager.ex
jamie-suse: 🧠 Ah yes, I missed that

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -19,84 +32,295 @@ defmodule Trento.DeregistrationProcessManager do
...
 
-  # First step in host deregistration when host belongs to a cluster
-  def handle(%DeregistrationProcessManager{cluster_id: cluster_id}, %HostDeregistrationRequested{
-        host_id: host_id,
-        requested_at: requested_at
-      }) do
-    [
+  def handle(
+        %DeregistrationProcessManager{
+          cluster_id: nil,
```
arbulu89, line 128: So now we have 40 lines function duplicated to avoid that `if` XD
I even preferred the `if` option honestly

arbulu89 (approved):
Green from my light.
I have to confess that i might miss something, as following the history after so many reviews with overriding commits is hard.  At the end, I don't know what was changed after every commit.
I would recommend to wait until receiving other approve at least

lib/trento/application/process_managers/deregistration_process_manager.ex
arbulu89: Or maybe some auxiliar function to flatten the app/db instances so it can be reused in the current code. At the end, that part of the code is duplicated

fabriziosestito (changes requested):

lib/trento/application/process_managers/deregistration_process_manager.ex
```diff
@@ -4,12 +4,25 @@ defmodule Trento.DeregistrationProcessManager do
...
+    @required_fields :all
+    use Trento.Type
+
+    deftype do
+      field :sap_system_id, Ecto.UUID
+      field :instance_number, :string
+    end
+  end
+
+  @required_fields :all
```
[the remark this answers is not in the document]
fabriziosestito: @CDimonaco could you confirm?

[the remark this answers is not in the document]
CDimonaco: Yes this should be empty, I confirm

CDimonaco (approved):

arbulu89 (approved):

PR #1357, opened 2023-04-21
Add application instance deregistration logic

rtorrero:
# Description

This PR adds the application instance deregistration logic to the sap_system aggregate. Upon deregistering a host, it should remove the relevant application instances and eventually sap systems, depending on the role of the deregistered host.

Note: This PR derives from https://github.com/trento-project/web/pull/1343 and is a subset of it. A follow-up PR will contain the database instance deregistration logic.

## How was this tested?
Based on the deregistration flow chart, all of the possible scenarios have been added to the test cases.

### DIFF
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -702,9 +702,14 @@ defmodule Trento.Domain.SapSystem do
   end
 
   defp maybe_emit_sap_system_deregistered_event(
+         %SapSystem{sid: nil},
+         _deregistered_at
+       ),
+       do: []
+
+  defp maybe_emit_sap_system_deregistered_event(
          %SapSystem{
            sap_system_id: sap_system_id,
-           sid: sid,
            application: %Application{
              instances: instances
            }
@@ -716,7 +721,7 @@ defmodule Trento.Domain.SapSystem do
     has_messageserver? =
       Enum.any?(instances, fn %{features: features} -> features =~ "MESSAGESERVER" end)
 
-    unless has_abap? and has_messageserver? and !is_nil(sid) do
+    unless has_abap? and has_messageserver? do
       %SapSystemDeregistered{sap_system_id: sap_system_id, deregistered_at: deregistered_at}
     end
   end
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -305,7 +305,7 @@ defmodule Trento.Domain.SapSystem do
           sap_system_id: sap_system_id,
           host_id: host_id,
           deregistered_at: deregistered_at
-        } = command
+        }
       ) do
     sap_system
     |> Multi.new()
@@ -320,7 +320,7 @@ defmodule Trento.Domain.SapSystem do
     |> Multi.execute(fn sap_system ->
       maybe_emit_sap_system_deregistration_events(
         sap_system,
-        command
+        deregistered_at
       )
     end)
   end
@@ -347,16 +347,14 @@ defmodule Trento.Domain.SapSystem do
   end
 
   def apply(
-        %SapSystem{application: %Application{instances: []}} = sap_system,
+        %SapSystem{} = sap_system,
         %SapSystemDeregistered{
-          sap_system_id: sap_system_id,
           deregistered_at: deregistered_at
         }
       ) do
     %SapSystem{
       sap_system
-      | sap_system_id: sap_system_id,
-        deregistered_at: deregistered_at
+      | deregistered_at: deregistered_at
     }
   end
 
@@ -710,10 +708,7 @@ defmodule Trento.Domain.SapSystem do
              instances: instances
            }
          },
-         %DeregisterApplicationInstance{
-           sap_system_id: sap_system_id,
-           deregistered_at: deregistered_at
-         }
+         deregistered_at
        ) do
     if !Enum.any?(instances, fn %Instance{features: features} ->
          String.contains?(features, "MESSAGESERVER")
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -710,12 +710,12 @@ defmodule Trento.Domain.SapSystem do
          },
          deregistered_at
        ) do
-    if !Enum.any?(instances, fn %Instance{features: features} ->
-         String.contains?(features, "MESSAGESERVER")
-       end) ||
-         !Enum.any?(instances, fn %Instance{features: features} ->
-           String.contains?(features, "ABAP")
-         end) do
+    has_abap? = Enum.any?(instances, fn instance -> instance.features =~ "ABAP" end)
+
+    has_messageserver? =
+      Enum.any?(instances, fn instance -> instance.features =~ "MESSAGESERVER" end)
+
+    unless has_abap? and has_messageserver? do
       Enum.map(instances, fn %Instance{instance_number: instance_number, host_id: host_id} ->
         %ApplicationInstanceDeregistered{
           sap_system_id: sap_system_id,
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -354,7 +354,8 @@ defmodule Trento.Domain.SapSystem do
       ) do
     %SapSystem{
       sap_system
-      | deregistered_at: deregistered_at
+      | deregistered_at: deregistered_at,
+        sid: nil
     }
   end
 
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -84,7 +84,7 @@ defmodule Trento.Domain.SapSystem do
 
   deftype do
     field :sap_system_id, Ecto.UUID
-    field :sid, :string
+    field :sid, :string, default: nil
     field :health, Ecto.Enum, values: Health.values()
     field :rolling_up, :boolean, default: false
     field :deregistered_at, :utc_datetime_usec, default: nil
@@ -318,7 +318,7 @@ defmodule Trento.Domain.SapSystem do
       }
     end)
     |> Multi.execute(fn sap_system ->
-      maybe_emit_sap_system_deregistration_events(
+      maybe_emit_sap_system_deregistered_event(
         sap_system,
         deregistered_at
       )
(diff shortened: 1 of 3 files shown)


lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +294,72 @@ defmodule Trento.Domain.SapSystem do
...
+    %SapSystem{
+      sap_system
+      | application: %Application{
+          instances: instances
+        }
+    }
+  end
+
+  def apply(
+        %SapSystem{application: %Application{instances: []}} = sap_system,
```
fabriziosestito, line 350: no need for this match, just match `%SapSystem{} = sap_system`

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +294,72 @@ defmodule Trento.Domain.SapSystem do
...
+        %SapSystemDeregistered{
+          sap_system_id: sap_system_id,
```
fabriziosestito, line 352: no need for this

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +294,72 @@ defmodule Trento.Domain.SapSystem do
...
+          deregistered_at: deregistered_at
+        }
+      ) do
+    %SapSystem{
+      sap_system
+      | sap_system_id: sap_system_id,
```
fabriziosestito, line 358: why update the sap_system_id here? no need for this

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -633,6 +703,36 @@ defmodule Trento.Domain.SapSystem do
...
+           sap_system_id: sap_system_id,
+           deregistered_at: deregistered_at
+         }
+       ) do
+    if !Enum.any?(instances, fn %Instance{features: features} ->
+         String.contains?(features, "MESSAGESERVER")
+       end) ||
+         !Enum.any?(instances, fn %Instance{features: features} ->
+           String.contains?(features, "ABAP")
+         end) do
```
fabriziosestito, line 723: Maybe a bit cleaner with the help of variables, `unless` and text-match operator:

```
    has_abap? = Enum.any?(instances, fn instance -> instance.features =~ "ABAP" end)

    has_messageserver? =
      Enum.any?(instances, fn instance -> instance.features =~ "MESSAGESERVER" end)

    unless has_abap? && has_messageserver? do
    ....
```

fabriziosestito:
@rtorrero @jamie-suse @arbulu89 I sent some preliminary comments as I will be off till Wednesday. It is a very superficial review, so feel free to have a deeper discussion next week.

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +294,72 @@ defmodule Trento.Domain.SapSystem do
...
+        sap_system_id: sap_system_id,
+        instance_number: instance_number,
+        host_id: host_id,
+        deregistered_at: deregistered_at
+      }
+    end)
+    |> Multi.execute(fn sap_system ->
+      maybe_emit_sap_system_deregistration_events(
+        sap_system,
+        command
```
fabriziosestito, line 323: why command here? we just need deregistered_at

lib/trento/domain/sap_system/sap_system.ex
arbulu89: Other solution could be:
```
# We map all the available features
instances
|> Enum.map(fn %{features: features} -> features end)
|> Enum.uniq()

# Check if message server and abap ara available in our list checking if the subset is there (https://stackoverflow.com/questions/63143363/elixir-check-array-contains-all-the-values-of-another-array)
unless Enum.empty?(["MESSAGESERVER", "ABAP"] -- features) do
   ...
end
```

arbulu89 (commented):
Looks great @rtorrero 

I would still work in the `maybe_..` function. Both solutions proposed by me and Fabrizio look easier to understand in my opinion

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -633,6 +701,33 @@ defmodule Trento.Domain.SapSystem do
...
+  defp maybe_emit_sap_system_deregistration_events(
+         %SapSystem{
+           sap_system_id: sap_system_id,
+           application: %Application{
+             instances: instances
+           }
+         },
+         deregistered_at
+       ) do
+    has_abap? = Enum.any?(instances, fn instance -> instance.features =~ "ABAP" end)
```
arbulu89, line 713: Maybe `fn %{features: features} -> ` better

arbulu89 (approved):
Alright @rtorrero 
It looks good in any case

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -633,6 +701,33 @@ defmodule Trento.Domain.SapSystem do
...
+    unless has_abap? and has_messageserver? do
+      Enum.map(instances, fn %Instance{instance_number: instance_number, host_id: host_id} ->
+        %ApplicationInstanceDeregistered{
+          sap_system_id: sap_system_id,
+          instance_number: instance_number,
+          host_id: host_id,
+          deregistered_at: deregistered_at
+        }
+      end) ++
+        [%SapSystemDeregistered{sap_system_id: sap_system_id, deregistered_at: deregistered_at}]
```
arbulu89, line 727: Hey @rtorrero 
I have noticed that this event should be conditional.
We cannot send this event always as the sap system might not even be registered.
Example:
- We register only message server
- We deregister this same host
At this point, we shouldn't be sending the sap system deregistered event

arbulu89 (changes requested):
Hey @rtorrero @jamie-suse 
I have realized that we might have some issue. Could you check my last comment below?
PD: Besides that, i'm having doubts yet again with the operation of deregistering all the instances at once

rtorrero:
@arbulu89 Thanks for the comments, I've:
 - Applied the additional style improvements you suggested
 - Updated the aggregate logic to not deregister all instances when the last ABAP / messageserver instance gets deregistered
 - Updated the logic to only emit SapSystemDeregistered if it was previously registered. Note that this is currently causing the test to fail as we need pr#1358 from @CDimonaco  to be merged first
 
 As the last point indicates, we need another PR to land first so I'm setting this one to draft meanwhile.

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -701,30 +701,23 @@ defmodule Trento.Domain.SapSystem do
...
-      Enum.map(instances, fn %Instance{instance_number: instance_number, host_id: host_id} ->
-        %ApplicationInstanceDeregistered{
-          sap_system_id: sap_system_id,
-          instance_number: instance_number,
-          host_id: host_id,
-          deregistered_at: deregistered_at
-        }
-      end) ++
-        [%SapSystemDeregistered{sap_system_id: sap_system_id, deregistered_at: deregistered_at}]
+    unless has_abap? and has_messageserver? and sid != nil do
```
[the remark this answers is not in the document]
rtorrero: I used `!is_nil(sid)` which I think is more appropriate here, thanks!

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -633,6 +701,26 @@ defmodule Trento.Domain.SapSystem do
...
+  defp maybe_emit_sap_system_deregistered_event(
+         %SapSystem{
+           sap_system_id: sap_system_id,
+           sid: sid,
+           application: %Application{
+             instances: instances
+           }
+         },
+         deregistered_at
+       ) do
```
arbulu89, line 713: Come on @rtorrero ,
Don't fall in the old habits already :smiley_cat: 
```
defp maybe_emit_sap_system_deregistered_event(%SapSystem{sid: nil}), do: []
defp maybe_emit_sap_system_deregistered_event(
         %SapSystem{
...
#what you had before
```

lib/trento/domain/sap_system/sap_system.ex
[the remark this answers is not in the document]
arbulu89: https://github.com/trento-project/web/pull/1357#discussion_r1176564351

lib/trento/domain/sap_system/sap_system.ex
rtorrero: I've added a test for this particular scenario and fixed the code to handle this situation properly. Thanks!

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -290,6 +294,70 @@ defmodule Trento.Domain.SapSystem do
...
+
+  def apply(
+        %SapSystem{} = sap_system,
+        %SapSystemDeregistered{
+          deregistered_at: deregistered_at
+        }
+      ) do
+    %SapSystem{
+      sap_system
+      | deregistered_at: deregistered_at
```
arbulu89, line 357: Maybe we should put `sid: nil` here. 
Otherwise, this instance cannot be de-registered never again.
Anyway, let's simply keep an eye on this for the restoration work

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1064,4 +1073,466 @@ defmodule Trento.SapSystemTest do
...
+                         instance_number: ^message_server_instance_number
+                       }
+                     ]
+                   }
+                 } = sap_system
+        end
+      )
+    end
+
+    test "should deregister an ABAP Application Instance" do
```
arbulu89, line 1171: maybe add `without deregistering the SAP system`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1064,4 +1073,466 @@ defmodule Trento.SapSystemTest do
...
+                       }
+                     ]
+                   },
+                   deregistered_at: ^deregistered_at
+                 } = sap_system
+        end
+      )
+    end
+
+    test "should deregister just a Message Server from a not-fully-registered SAP system" do
```
arbulu89, line 1375: I would remove this test, it does exactly the same than the first. There is no any distinction in reality in the code (we know that elixis'r pattern matching works)

arbulu89 (approved):
Thanks @rtorrero 
Good to go from my side. Added some minor comments about the tests

test/trento/domain/sap_system/sap_system_test.exs
jamie-suse: We checked this code path/scenario for coverage using _Coveralls_; this test covers the scenario where a SAP System is **not** registered, and _then_ an instance is deregistered - testing that a `SapSystemDeregistered` event is **not** emitted in this case. This is not the same as the first test - removing an ENQREP server, as it would never emit a `SapSystemDeregistered` event.

PR #1358, opened 2023-04-21
Application instances registration changes

CDimonaco:
# Description

Implement changes in application instance registration and complete Sap System registration.

In order to register a Sap System, ABAP and MESSAGESERVER are the minimum sets of features.

Without them we cannot register a sap system.
ABAP and MESSAGESERVER instances are added as application instances when the sap system is not completely registered, other types of instances are discarded.

Once a Sap system is proper registered, all the types of instances are added to the sapsystem.

## How was this tested?

AUTOMATED TESTS

### DIFF
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -216,86 +216,23 @@ defmodule Trento.Domain.SapSystem do
   # Otherwise if the istance we want register together with already present istances
   # have one MESSAGESERVER and one ABAP, we register the istance and the sap system
   def execute(
-        %SapSystem{sid: nil, application: %Application{instances: instances}} = sap_system,
+        %SapSystem{sid: nil, application: %Application{}} = sap_system,
         %RegisterApplicationInstance{
-          sap_system_id: sap_system_id,
-          sid: sid,
-          instance_number: instance_number,
-          instance_hostname: instance_hostname,
-          tenant: tenant,
-          db_host: db_host,
-          features: features,
-          http_port: http_port,
-          https_port: https_port,
-          start_priority: start_priority,
-          host_id: host_id,
-          health: health
-        }
+          features: features
+        } = instance
       ) do
     if abap_or_messageserver?(features) do
-      instances_features =
-        instances
-        |> Enum.map(& &1.features)
-        |> Kernel.++([features])
-        |> Enum.join()
-
-      events =
-        if abap_and_messageserver?(instances_features) do
-          [
-            %SapSystemRegistered{
-              sap_system_id: sap_system_id,
-              sid: sid,
-              tenant: tenant,
-              db_host: db_host,
-              health: health
-            }
-          ]
-        else
-          []
-        end
-
-      instance =
-        Enum.find(instances, fn
-          %Instance{host_id: ^host_id, instance_number: ^instance_number} ->
-            true
-
-          _ ->
-            false
-        end)
-
-      event =
-        case instance do
-          %Instance{health: ^health} ->
-            nil
-
-          %Instance{host_id: host_id, instance_number: instance_number} ->
-            %ApplicationInstanceHealthChanged{
-              sap_system_id: sap_system_id,
-              host_id: host_id,
-              instance_number: instance_number,
-              health: health
-            }
-
-          nil ->
-            %ApplicationInstanceRegistered{
-              sap_system_id: sap_system_id,
-              sid: sid,
-              instance_number: instance_number,
-              instance_hostname: instance_hostname,
-              features: features,
-              http_port: http_port,
-              https_port: https_port,
-              start_priority: start_priority,
-              host_id: host_id,
-              health: health
-            }
-        end
-
-      events = events ++ [event]
-
       sap_system
       |> Multi.new()
-      |> Multi.execute(fn _ -> events end)
+      |> Multi.execute(fn sap_system ->
+        maybe_emit_sap_system_registered_event(sap_system, instance)
+      end)
+      |> Multi.execute(fn sap_system ->
+        emit_application_instance_registered_or_application_instance_health_changed(
+          sap_system,
+          instance
+        )
+      end)
       |> Multi.execute(&maybe_emit_sap_system_health_changed_event/1)
     else
       # TODO: Check error
@@ -307,60 +244,17 @@ defmodule Trento.Domain.SapSystem do
   # the SAP System aggregate registers the Application instance if it is not already registered
   # and updates the health when needed.
   def execute(
-        %SapSystem{application: %Application{instances: instances}} = sap_system,
-        %RegisterApplicationInstance{
-          sap_system_id: sap_system_id,
-          sid: sid,
-          instance_number: instance_number,
-          instance_hostname: instance_hostname,
-          features: features,
-          http_port: http_port,
-          https_port: https_port,
-          start_priority: start_priority,
-          host_id: host_id,
-          health: health
-        }
+        %SapSystem{application: %Application{instances: _instances}} = sap_system,
+        %RegisterApplicationInstance{} = instance
       ) do
-    instance =
-      Enum.find(instances, fn
-        %Instance{host_id: ^host_id, instance_number: ^instance_number} ->
-          true
-
-        _ ->
-          false
-      end)
-
-    event =
-      case instance do
-        %Instance{health: ^health} ->
-          nil
-
-        %Instance{host_id: host_id, instance_number: instance_number} ->
-          %ApplicationInstanceHealthChanged{
-            sap_system_id: sap_system_id,
-            host_id: host_id,
-            instance_number: instance_number,
-            health: health
-          }
-
-        nil ->
-          %ApplicationInstanceRegistered{
-            sap_system_id: sap_system_id,
-            sid: sid,
-            instance_number: instance_number,
-            instance_hostname: instance_hostname,
-            features: features,
-            http_port: http_port,
-            https_port: https_port,
-            start_priority: start_priority,
-            host_id: host_id,
-            health: health
-          }
-      end
-
     sap_system
     |> Multi.new()
-    |> Multi.execute(fn _ -> event end)
+    |> Multi.execute(fn sap_system ->
+      emit_application_instance_registered_or_application_instance_health_changed(
+        sap_system,
+        instance
+      )
+    end)
     |> Multi.execute(&maybe_emit_sap_system_health_changed_event/1)
   end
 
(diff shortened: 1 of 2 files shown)


lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -208,7 +201,102 @@ defmodule Trento.Domain.SapSystem do
...
+      # TODO: Check error
+      {:error, :sap_system_not_registered}
+    end
+  end
+
+  # Sap system not registered, application already present
+  # If the istance is not one of MESSAGESERVER or ABAP we discard.
+  # Otherwise if the istance we want register together with already present istances
+  # have one MESSAGESERVER and one ABAP, we register the istance and the sap system
+  def execute(
```
arbulu89, line 214: I know that this code is based in the old one, but I find it difficult to digest (even 2 nested levels of `if`s look to much now hehe).
I find easier the strategy we talked about the de-registration and using `multis` as needed.
So I would do:
- register application instance if it is not registered and if it is message or abap server (or always, if we decide to do so)
- in the next multi, check the sap system conditions, if abap and message server are present we register the sap system
- next multi we check if the health much be changed

This way, we don't need to have an almost duplicated new function maybe

arbulu89 (commented):
I think following the new strategy of using the state to see if the sap system needs to be registered is cleaner and easier to understand.
(and changing the strategy to register always the instance without discarding would even simply it, but let's discuss this offline once we are all)

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: thanks, done :D

arbulu89 (changes requested):
@CDimonaco after our last discussion about what to the with the instances, I think we can change strategy.
We could simply register all the instance independently of their type, and once the conditions to have a full SAP system are there, we could send that event.
It makes the code and the logic easier and cleaner.
What do you think?

This way we would have:
- Any instance is registered
- Change health if needed
- If SAP system conditions are met, register SAP system
- Change SAP system health if needed 

I see this as:
```
def execute(%SapSystem{...} = sap_system. %RegisterApplicationInstace{...} do
    sap_system 
    |> Multi.new()
    |> Multi.execute(&maybe_emit_application_registered_event) # here we register or update the instance
    |> Multi.execute(&maybe_emit_sap_system_registered) # we check list of instances and emit the event if it is not registered
    |> Multi.execute(&maybe_emit_sap_system_health_changed_event/1)
```

CDimonaco:
> @CDimonaco after our last discussion about what to the with the instances, I think we can change strategy. We could simply register all the instance independently of their type, and once the conditions to have a full SAP system are there, we could send that event. It makes the code and the logic easier and cleaner. What do you think?
> 
> This way we would have:
> 
>     * Any instance is registered
> 
>     * Change health if needed
> 
>     * If SAP system conditions are met, register SAP system
> 
>     * Change SAP system health if needed
> 
> 
> I see this as:
> 
> ```
> def execute(%SapSystem{...} = sap_system. %RegisterApplicationInstace{...} do
>     sap_system 
>     |> Multi.new()
>     |> Multi.execute(&maybe_emit_application_registered_event) # here we register or update the instance
>     |> Multi.execute(&maybe_emit_sap_system_registered) # we check list of instances and emit the event if it is not registered
>     |> Multi.execute(&maybe_emit_sap_system_health_changed_event/1)
> ```

I'm just looking at it now, there are some infos like db_host, which is present only on the commands, like RegisterApplicationInstance, and this is needed because we need in order to register a sap system, so we need to carry this informations to the maybe_emit_sap_system_registered_event, so we end up in the same situation as now, we can't rely only on the state.  

Carrying this type of information leads to a situation which is similar to what we have now, of course we change the ordering of things, I will commit the update and we can decide if we want to mantain the approach

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -170,105 +176,29 @@ defmodule Trento.Domain.SapSystem do
...
-        http_port: http_port,
-        https_port: https_port,
-        start_priority: start_priority,
-        host_id: host_id,
-        health: health
-      }
-    ]
-  end
-
+  # Sap system not registered, application already present
```
arbulu89, line 179: `Sap` -> `SAP`

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -642,4 +572,129 @@ defmodule Trento.Domain.SapSystem do
...
+
+  defp abap_or_messageserver?(features) do
+    String.contains?(features, ["MESSAGESERVER", "ABAP"])
+  end
+
+  defp abap_and_messageserver?(features) do
+    String.contains?(features, "ABAP") and String.contains?(features, "MESSAGESERVER")
+  end
+
+  defp maybe_emit_sap_system_registered_event(
```
arbulu89, line 584: Let's respect the order of the functions by precedence.
So:
1. `emit_application_instance_registered_or_application_instance_health_changed`
2. `maybe_emit_sap_system_registered_event`
etc

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -642,4 +572,129 @@ defmodule Trento.Domain.SapSystem do
         false
     end)
   end
+
+  defp abap_or_messageserver?(features) do
```
arbulu89, line 576: Do we still need these 2 functions? I think the de-registration code added exactly the same functionality:
https://github.com/trento-project/web/blob/deregistration/lib/trento/domain/sap_system/sap_system.ex#L720

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -323,13 +323,12 @@ defmodule Trento.SapSystemTest do
       )
     end
 
-    test "should register a SAP System and add an application instance" do
+    test "should register a SAP System and add an application instance when is already present a MESSAGESERVER instance and one ABAP instance is added" do
```
arbulu89, line 326: `... when a MESSAGESERVER instance is already present and a new ABAP instance is added`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -358,32 +362,122 @@ defmodule Trento.SapSystemTest do
...
+                         features: "MESSAGESERVER"
+                       }
+                     ]
+                   }
+                 } = state
+        end
+      )
+    end
+
+    test "should register a SAP System and add an application instance when is already present a ABAP instance and one MESSAGESERVER instance is added" do
```
arbulu89, line 416: `when an ABAP instance is already present and a new MESSAGESERVER instance is added`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -395,7 +489,162 @@ defmodule Trento.SapSystemTest do
...
+                         features: "ABAP"
+                       }
+                     ]
+                   }
+                 } = state
+        end
+      )
+    end
+
+    test "should add an application instance to a non registered sap system when the instance is ABAP without complete a sap system registration" do
```
arbulu89, line 506: `sap` -> `SAP`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -395,7 +489,162 @@ defmodule Trento.SapSystemTest do
...
+                         health: :passing
+                       }
+                     ]
+                   }
+                 } = state
+        end
+      )
+    end
+
+    test "should app an application instance to a non registered sap system when the instance is MESSAGESERVER without complete a sap system registration" do
```
arbulu89, line 582: `should add`, `sap` -> `SAP`, `without a completing a SAP`...

arbulu89 (changes requested):
@CDimonaco 
- Let's respect the order of functions by precedence of how the appear
- Revisit texts, specially in the test
- If we are concerned about the SAP system health, we could change the current code to only apply the health if the SAP system is fully registered (`sid` is present)

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -642,4 +572,129 @@ defmodule Trento.Domain.SapSystem do
...
+           sid: sid,
+           tenant: tenant,
+           db_host: db_host,
+           features: features,
+           health: health
+         }
+       ) do
+    instances_features =
+      instances
+      |> Enum.map(& &1.features)
```
rtorrero, line 597: Hey @CDimonaco , I think if you use the suggestion by @arbulu89 in https://github.com/trento-project/web/pull/1358/files#r1178679263 you wouldn't need to do this in the first place

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: done

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: done

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: I want to leave this for now and then maybe remove when they are merged together

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: done

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: done

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: done

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: done

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -5,9 +5,11 @@ defmodule Trento.Domain.SapSystem do
   **The HANA database is the only supported database type.**
 
   In order to have a fully registered SAP system, both the database and application
-  composing this system must be registered. And each of the two layers might be composed
-  by multiple instances altogether. This means that a SAP system aggregate state can have
-  multiple application/database instances.
+  composing this system must be registered.
+  The minimum set of application features is ABAP and MESSAGESERVER, otherwise, a complete Sap system cannot exist.
```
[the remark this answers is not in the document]
CDimonaco: done

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -33,9 +35,13 @@ defmodule Trento.Domain.SapSystem do
...
-  3. A SAP system discovery with a new application instance is received, and the database associated to
-     this application exists, the application instance is registered together with the complete
-     SAP system. The SAP system is fully registered now.
+  3. When a SAP system discovery with a new application instance is received, and the database associated to
+     this application exists:
+      - If the application instance does not have ABAP or MESSAGESERVER as features, will be rejected
+      - If the application instance has ABAP or MESSAGESERVER as features, and is already present an ABAP or MESSAGESERVER application instance,
+        the application instance is registered together with the complete SAP system. The SAP system is fully registered now.
+      - If the application instance has ABAP or MESSAGESERVER as features, but there are no other instances with ABAP or MESSAGESERVER as features,
+        the application instance is registered without the complete sap system registration.
```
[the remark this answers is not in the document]
CDimonaco: rewrited to the current logic

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -80,13 +86,13 @@ defmodule Trento.Domain.SapSystem do
   use Trento.Type
 
   deftype do
-    field :sap_system_id, Ecto.UUID
-    field :sid, :string
-    field :health, Ecto.Enum, values: Health.values()
-    field :rolling_up, :boolean, default: false
+    field(:sap_system_id, Ecto.UUID)
```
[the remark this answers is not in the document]
CDimonaco: done

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -170,16 +176,16 @@ defmodule Trento.Domain.SapSystem do
     |> Multi.execute(&maybe_emit_sap_system_health_changed_event/1)
   end
 
-  # When an Application is discovered, the SAP System completes the registration process.
+  # Sap system not registered, no applications.
```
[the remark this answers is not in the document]
CDimonaco: done

fabriziosestito (approved):

arbulu89 (approved):
Great @CDimonaco ,
It looks great

PR #1369, opened 2023-04-27
Collapsible checks catalog

nelsonkopliku:
# Description

Fixes # (issue)

## How was this tested?


assets/js/components/Accordion/Accordion.stories.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+
+import PremiumPill from '@components/PremiumPill';
+import Accordion from '.';
+
+export default {
+  title: 'Accordion',
+  component: Accordion,
```
EMaksy, line 8: Hey the story works fine as it is, but what do you think about [using storybook argTypes](https://storybook.js.org/docs/react/api/argtypes) ?
We could have something like this which can be passed as args to a Template
```
argTypes: {
    header: {
      type: 'string',
      defaultValue: 'Accordion Header',
      description: 'The title of the accordion component',
      control: {
        type: 'text',
      },
    }
```
```
 export function Default(args) {
  return (
    <Accordion {...args}>
      <div className="p-6">Accordion content</div>
    </Accordion>
  );
}
```   

This would have the nice effect that we can change the text inside storybook, playing with it without touching  the component :D
The rest of the parameters could be added too.

nelsonkopliku: sure, will do!

assets/js/components/Accordion/Accordion.jsx
```diff
@@ -0,0 +1,81 @@
...
+import { Disclosure, Transition } from '@headlessui/react';
+import { EOS_KEYBOARD_ARROW_DOWN } from 'eos-icons-react';
+
+import classNames from 'classnames';
+
+function Accordion({
+  className,
+  header,
+  headerClassnames,
+  withHeader = false,
```
dottorblaster, line 12: Can you rename this just to `header`? Since this accepts a JSX node I think it's better that way. Default can be `false` or just `undefined`

dottorblaster (changes requested):
Just these two then we can merge

EMaksy (commented):
Hey nice job man, besides the comments nothing to add :+1:

assets/js/components/Accordion/Accordion.stories.jsx
nelsonkopliku: @EMaksy I changed as per the documentation of the last storybook's version

assets/js/components/Accordion/Accordion.stories.jsx
```diff
@@ -0,0 +1,92 @@
...
+  argTypes: {
+    header: {
+      type: 'string',
+      description:
+        'The content of the accordion header. It can be a plain string or a Component',
+      control: {
+        type: null,
```
EMaksy, line 15: Thanks for adding this, just a small nitpick. 
Could you change 
`type: null`  to` type: 'string'`
Screenshot from 2023-04-28 11-35-26

assets/js/components/Accordion/Accordion.stories.jsx
```diff
@@ -0,0 +1,92 @@
...
+      },
+    },
+    children: {
+      description:
+        'The content of the accordion panel. It should be a Component',
+      control: {
+        type: null,
```
EMaksy, line 22: Also here, as it has the same effect
`type: null `to `type: 'string'`

EMaksy (changes requested):
Hey man everything looks good, just two small comments about the story then LGTM :D

assets/js/components/Accordion/Accordion.stories.jsx
nelsonkopliku: as discussed we're keeping the header editable, but not the children at the moment

EMaksy (approved):

PR #1372, opened 2023-05-02
Add database instance deregistration logic

jamie-suse:
# Description

This PR adds the _database instance_ deregistration logic to the `sap_system` aggregate.

## How was this tested?

Based on the deregistration flow chart, all of the possible scenarios have been added to the test cases.

Note: This PR derives from https://github.com/trento-project/web/pull/1343

### DIFF
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -814,7 +814,7 @@ defmodule Trento.Domain.SapSystem do
          %SapSystem{
            sap_system_id: sap_system_id,
            database: %Database{
-             instances: []
+             sid: nil
            }
          },
          deregistered_at
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -99,6 +99,9 @@ defmodule Trento.Domain.SapSystem do
     embeds_one :application, Application
   end
 
+  # Stop everything during the rollup process
+  def execute(%SapSystem{rolling_up: true}, _), do: {:error, :sap_system_rolling_up}
+
   def execute(
         %SapSystem{sap_system_id: nil},
         %RegisterDatabaseInstance{
@@ -152,9 +155,6 @@ defmodule Trento.Domain.SapSystem do
       }
     ]
   end
-
-  # Stop everything during the rollup process
-  def execute(%SapSystem{rolling_up: true}, _), do: {:error, :sap_system_rolling_up}
 
   # When a RegisterDatabaseInstance command is received by an existing SAP System aggregate,
   # the SAP System aggregate registers the Database instance if it is not already registered
@@ -224,37 +224,6 @@ defmodule Trento.Domain.SapSystem do
     }
   end
 
-  # Deregister an application instance and emit a ApplicationInstanceDeregistered
-  # also emit SapSystemDeregistered event if this was the last application instance
-  def execute(
-        %SapSystem{
-          sap_system_id: sap_system_id
-        } = sap_system,
-        %DeregisterApplicationInstance{
-          instance_number: instance_number,
-          sap_system_id: sap_system_id,
-          host_id: host_id,
-          deregistered_at: deregistered_at
-        }
-      ) do
-    sap_system
-    |> Multi.new()
-    |> Multi.execute(fn _ ->
-      %ApplicationInstanceDeregistered{
-        sap_system_id: sap_system_id,
-        instance_number: instance_number,
-        host_id: host_id,
-        deregistered_at: deregistered_at
-      }
-    end)
-    |> Multi.execute(fn sap_system ->
-      maybe_emit_sap_system_deregistered_event(
-        sap_system,
-        deregistered_at
-      )
-    end)
-  end
-
   # Deregister a database instance and emit a DatabaseInstanceDeregistered
   # also potentially emit SapSystemDeregistered and DatabaseDeregistered events
   def execute(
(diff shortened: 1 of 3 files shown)


lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -618,6 +702,18 @@ defmodule Trento.Domain.SapSystem do
        ),
        do: []
 
+  defp maybe_emit_sap_system_deregistered_event(
+         %SapSystem{
+           sap_system_id: sap_system_id,
+           database: %Database{
+             instances: nil
```
arbulu89, line 709: Is this `instances: nil` or `instances: []` (or empty list)?
The same in other parts of the code.
Do we have any tests on this?

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1339,6 +1342,242 @@ defmodule Trento.SapSystemTest do
   end
 
   describe "deregistration" do
+    test "should deregister a single DB instance if no SR enabled" do
```
arbulu89, line 1345: I'm totally lost with the order of things.
Why this goes first in the `deregistration` describe?
Let's have some logic or time based order please. I see total random additions

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1339,6 +1342,242 @@ defmodule Trento.SapSystemTest do
...
+        build_list(
+          2,
+          :database_instance_registered_event,
+          sap_system_id: sap_system_id,
+          system_replication: nil
+        )
+
+      assert_events_and_state(
+        [
+          build(
```
arbulu89, line 1368: Is this the correct order of things? Or do we have `instance`, `database`, `instance` order?
Just to be more precise

arbulu89 (changes requested):
Hey @jamie-suse 
I'm not sure with the approach of setting `nil` to the whole `database` field.

Please have a look on the order of the new functions and order of precendence

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -273,6 +340,23 @@ defmodule Trento.Domain.SapSystem do
     }
   end
 
+  def apply(
```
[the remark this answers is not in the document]
fabriziosestito: this shouldn't be possible in fact

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -632,6 +728,67 @@ defmodule Trento.Domain.SapSystem do
...
+      end)
+
+    if has_secondary? and !has_primary? do
+      [
+        %DatabaseDeregistered{
+          sap_system_id: sap_system_id,
+          deregistered_at: deregistered_at
+        }
+      ] ++
+        Enum.map(instances, fn %Instance{instance_number: instance_number, host_id: host_id} ->
```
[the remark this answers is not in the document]
rtorrero: We talked about *not* deregistering the application instances, but the database instances. If we consider the deregistration the reverse of the registration process, we currently don't register database instances before the primary arrives.

Adding to this thought, we might actually want to also deregister all application instances on DatabaseDeregistered (not on SapSystemDeregistered though) ir order to properly "reverse" the registration.

test/trento/domain/sap_system/sap_system_test.exs
rtorrero: I did notice this and you are totally right. Putting it in the right order requires some changes to the factory (or at least to how we use it). Will give it another try.

lib/trento/domain/sap_system/sap_system.ex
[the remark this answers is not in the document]
arbulu89: @rtorrero Let's forget about an exact mirror of processes. 
Here the main events are `SapSystemRegistered` and `DatabaseRegistered`. The instance registrations are `value/objects` kind of information registration, and we use them to composed the main 2.

Having 2 different actions where you can deregister instances can lead to concurrency issues (imagine a future button where we can deregister 2 instances at once, we would send the same database instance deregistration twice), and mostly, they don't add any value, as having the instances there doesn't affect to the general sap system/database domain logic.

Let's keep the logic simple

test/trento/domain/sap_system/sap_system_test.exs
rtorrero: Actually @arbulu89, looking at database registration tests, we first emit DatabaseRegistered and then we emit the instances registered events. The flow chart confirms this.

rtorrero (approved):
Nothing to add @jamie-suse, LGTM

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -221,6 +224,35 @@ defmodule Trento.Domain.SapSystem do
     }
   end
 
+  # Deregister a database instance and emit a DatabaseInstanceDeregistered
```
arbulu89, line 227: Something new that came to my mind.
We shouldn't send any event if the instance is not registered. This shouldn't happen if the commands are not launched from the cli though, but a good to have thing.
So we should have 2 extra conditions:
- SAP system is not even registered
- Instance is not in the registered SAP system

Anyway, we can work on that in a next PR.

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -775,4 +803,95 @@ defmodule Trento.Domain.SapSystem do
...
     end
   end
+
+  defp maybe_emit_sap_system_deregistered_event(
+         %SapSystem{sid: nil},
+         _deregistered_at
+       ),
+       do: []
+
+  defp maybe_emit_sap_system_deregistered_event(
```
arbulu89, line 813: I think we are missing one scenario, the `database: %{sid: nil}`.
In fact, we shouldn't check for instances to be an empty list, we should check for the above condition.
We can have one secondary instance but a deregistered database.
Right?

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1339,6 +1342,248 @@ defmodule Trento.SapSystemTest do
...
+            system_replication: "Primary"
+          ),
+          build(
+            :database_instance_registered_event,
+            sap_system_id: sap_system_id,
+            host_id: secondary_database_host_id,
+            instance_number: instance_number_2,
+            system_replication: "Secondary"
+          ),
+          build(
```
arbulu89, line 1375: I would say this test has a "mistake".
It passes, because (emitting the SapSystemDeregistered), because you don't have messageserver/abap combo, not because you remove the database instance

arbulu89 (changes requested):
@jamie-suse we are almost there.
A couple of more changes and we can merge.
Take the comment on having safe pattern matching functions out of the equation, for future work

arbulu89 (approved):
Thank you @jamie-suse 
All fine from my side

PR #1375, opened 2023-05-03
Fix wrong filtered catalog being used in execution results page

nelsonkopliku:
# Description

This PR fixes a bug where a previously filtered catalog state by a given provider is not what is needed in the execution results page where the cluster provider might differ from the provider used for the previous filtering in the catalog.

Example: filtering by GCP in the catalog loads a state that is not what is needed for the execution results of a cluster discovered on Azure.

This fix comes with a cost: landing on the execution results page triggers a refresh of the catalog.
Note that navigating from the execution results to the check detail does not retrigger the refresh, however navigating back from check detail to execution results does trigger a refresh.

fix-catalog-loading-in-check-results

## How was this tested?

Manually.

### DIFF
--- a/assets/js/components/ExecutionResults/ExecutionResultsPage.jsx
+++ b/assets/js/components/ExecutionResults/ExecutionResultsPage.jsx
@@ -28,14 +28,16 @@ function ExecutionResultsPage() {
     },
   } = useSelector(getLastExecutionData(clusterID));
 
+  const cloudProvider = cluster?.provider;
+
   useEffect(() => {
-    if (catalog.length === 0) {
-      dispatch(updateCatalog());
+    if (cloudProvider) {
+      dispatch(updateCatalog({ provider: cloudProvider }));
     }
     if (!executionData) {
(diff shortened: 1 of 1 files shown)


assets/js/components/ExecutionResults/ExecutionResultsPage.jsx
```diff
@@ -29,13 +29,13 @@ function ExecutionResultsPage() {
   } = useSelector(getLastExecutionData(clusterID));
 
   useEffect(() => {
-    if (catalog.length === 0) {
-      dispatch(updateCatalog());
+    if (cluster?.provider) {
```
dottorblaster, line 32: Would you declare this like `const cloudProvider = cluster?.provider;` and just reference this instead of accessing two times the field?

dottorblaster (changes requested):
Just a tiny modification then we can merge :+1:

EMaksy (commented):
Besides @dottorblaster  comment, lgtm :D

dottorblaster (approved):
:shipit:

PR #1386, opened 2023-05-08
Cluster delta deregistration

fabriziosestito:
This PR introduces the "delta deregistration" process for clusters.
When a discovery payload is received, the current state is considered and extra deregistration commands are dispatched alongside the registration ones. 
This is done with the help of the read models.
Eventual consistency is acceptable in this case because the agent will emit the same event after a while, so eventually the desired state will be reached.

### DIFF
--- a/test/trento/application/integration/discovery/policies/cluster_policy_test.exs
+++ b/test/trento/application/integration/discovery/policies/cluster_policy_test.exs
@@ -10,7 +10,7 @@ defmodule Trento.Integration.Discovery.ClusterPolicyTest do
 
   alias Trento.Integration.Discovery.ClusterPolicy
 
-  alias Trento.Domain.Commands.RegisterClusterHost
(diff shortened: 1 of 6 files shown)


CDimonaco (approved):
lgtm

test/trento/application/integration/discovery/policies/cluster_policy_test.exs
```diff
@@ -499,497 +515,528 @@ defmodule Trento.Integration.Discovery.ClusterPolicyTest do
...
+                resources_number: 8,
+                discovered_health: :passing,
+                provider: Provider.azure()
+              }
+            ]} ==
              "ha_cluster_discovery_unnamed"
              |> load_discovery_event_fixture()
-             |> ClusterPolicy.handle()
+             |> ClusterPolicy.handle(nil)
+  end
```
fabriziosestito, line 1014: new tests starting below this comment

arbulu89 (approved):
Super!!

PR #1392, opened 2023-05-09
Detect ascs/ers cluster type

arbulu89:
# Description

Detect ASCS/ERS clusters. 
It checks if the `SAPInstance` resource appears in the `crm_mon` output of the `ha_discovery` message, even number of times (for multi sid environments).

This is the most basic and less invasive way of detecting it. There are more things we could check to see if the cluster is properly configured, but let's leave those for actual "checks".

**Sending the PR to `deregistration` branch, as I don't want to add this yet to `main` to not hijack the new checks ui potential release** 

Frontend coming next. Easy things, but I wanted to add more tests.

## How was this tested?
Automatic tests added.
I have tested in a real env as well

### DIFF
--- a/lib/trento/application/integration/discovery/payloads/cluster/cluster_discovery_payload.ex
+++ b/lib/trento/application/integration/discovery/payloads/cluster/cluster_discovery_payload.ex
@@ -57,7 +57,18 @@ defmodule Trento.Integration.Discovery.ClusterDiscoveryPayload do
   defp parse_cluster_type(%{"crmmon" => %{"clones" => nil, "groups" => nil}}),
     do: ClusterType.unknown()
 
-  defp parse_cluster_type(%{"crmmon" => %{"clones" => clones}}) when not is_nil(clones) do
+  defp parse_cluster_type(%{"crmmon" => %{"clones" => nil, "groups" => groups}}) do
+    sap_instance_count =
+      Enum.count(groups, fn %{"resources" => resources} ->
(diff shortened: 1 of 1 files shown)


nelsonkopliku:
Not directly in tower radar but nice job :eyes:

lib/trento/application/integration/discovery/payloads/cluster/cluster_discovery_payload.ex
```diff
@@ -77,10 +78,28 @@ defmodule Trento.Integration.Discovery.ClusterDiscoveryPayload do
     do_detect_cluster_type(has_sap_hana_topology, has_sap_hana, has_sap_hana_controller)
   end
 
+  defp parse_cluster_type(%{"crmmon" => %{"groups" => groups}}) do
```
fabriziosestito, line 81: This clause matches if the above one is not `nil`.  Maybe it is clearer if we invert the order and add `"clones" => nil` to this one, wdyt?

fabriziosestito (approved):
LGTM! Just a nitpick added. 

Love the simplicity.

lib/trento/application/integration/discovery/payloads/cluster/cluster_discovery_payload.ex
arbulu89: Right! It looks even better. Thank you!
Changed!

PR #1398, opened 2023-05-10
Ascs ers type frontend

arbulu89:
# Description

Frontend visualization of the `ASCS/ERS` cluster type.

image

I have taken the advantage to improve the e2e tests of this page, to check if the cluster type and sid are properly updated.
It is more complete that the backend/frontend UT tests, as this tests the complete flow.

## How was this tested?

Unit and E2E tests


abravosuse (approved):

fabriziosestito (approved):

dottorblaster (approved):
LGTM

PR #1399, opened 2023-05-10
Automatically accept changes on Chromatic

dottorblaster:
Since we use factories and data coming from faker inside stories, every rendering gets new data. This way, Chromatic's visual changes tooling becomes super noisy in the CI. Disabling that for the moment.


nelsonkopliku (approved):
thank you :rocket:

PR #1400, opened 2023-05-10
Discover ASCS/ERS cluster SIDs

arbulu89:
# Description

Discover ASCS/ERS cluster SIDs.
I have added a new `additional_sids` for that. The main reasons are 
- to not add breaking changes changing the `sid` field
- we still need to know the main sid of HANA clusters, so this way this value continues being not ambiguous
In the future, we could use this field to add HANA clusters cost-optimized SID, for example.

The SID discovery is simple. Find `SAPInstance` entries in the cib configuration, and from its `instance_attributes`, get the `InstanceName`, which always has a `SID_INSTANCE_VIRTUALHOSTNAME` pattern. **This is always the case in correctly configured clusters, otherwise the cluster doesn't even work**

## How was this tested?

Tests added


fabriziosestito (approved):
Super!

PR #1403, opened 2023-05-11
Display ASCS/ERS clusters SID in clusters view

arbulu89:
# Description
Show `additional_sids` in the frontend.

image

## How was this tested?

Tests added

### DIFF
--- a/lib/trento/application/usecases/clusters/clusters.ex
+++ b/lib/trento/application/usecases/clusters/clusters.ex
@@ -53,7 +53,7 @@ defmodule Trento.Clusters do
   @spec get_all_clusters :: [ClusterReadModel.t()]
   def get_all_clusters do
     from(c in ClusterReadModel,
-      order_by: [asc: c.name],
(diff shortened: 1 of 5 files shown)


lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -53,7 +53,7 @@ defmodule Trento.Clusters do
   @spec get_all_clusters :: [ClusterReadModel.t()]
   def get_all_clusters do
     from(c in ClusterReadModel,
-      order_by: [asc: c.name],
+      order_by: [asc: c.name, asc: c.id],
```
arbulu89, line 56: This is required in order to have a predictable order of clusters in the view, as the clusters might have the same name sometimes (as in our demo)

abravosuse (approved):

jagabomb (approved):
LGTM

fabriziosestito (approved):

PR #1405, opened 2023-05-12
Handle `:database_not_registered` errors

jamie-suse:
# Description

Fixes issues encountered when running [Photofinish](https://github.com/trento-project/photofinish). Received status `500` responses because of unhandled `{:error, :database_not_found}` errors.

Renamed error `:database_not_found` -> `:database_not_registered`

## How was this tested?

Current unit tests enforce current behaviour

PR #1413, opened 2023-05-15
Properly handle multiple errors returned

rtorrero:
# Description
PR #1405 allowed the fallback controller to handle a list of errors being returned. We where stripping away the error and just returning the reason. It now returns the `call(conn, {:error, reason}`.

## How was this tested?
Manually, using photofinish.


jamie-suse:
LGTM! 🚀

jamie-suse (approved):

arbulu89 (commented):
mmm, it looks like only the 1st error in the list is handled, right?
Not that it was better before.

PR #1415, opened 2023-05-15
Do not render Values section when no values are available

nelsonkopliku:
# Description

This PR makes sure that the **Values** section in a check result detail is not rendered if there were no values to be rendered (ie the check has no `values` defined)

Scenarios are:
- cluster wide checks -> no values section
- target host + no values defined -> no Values section
- target host + values defined -> Values section + values
- target host + values defined + error -> Values section + error message

no-values-detail

### DIFF
--- a/assets/js/components/ExecutionResults/CheckResultDetail/CheckResultDetail.jsx
+++ b/assets/js/components/ExecutionResults/CheckResultDetail/CheckResultDetail.jsx
@@ -53,9 +53,11 @@ function CheckResultDetail({
         isError={isError}
         errorMessage={message}
       />
-      {targetHost && (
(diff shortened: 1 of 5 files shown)


nelsonkopliku:
@abravosuse @jagabomb the alternative to this is to explicitly say: hey there are no values defined for the current check.
Whatever you guys prefer :wink:

abravosuse (approved):
I personally like it as it is, specially for checks with target cluster: if there is no value for this type of check, it makes no sense to show it. Then, by extension, same run for checks with target host.

dottorblaster (approved):
Just a nitpick but LGTM

EMaksy (approved):
Hey man nice job, LGTM!

PR #1422, opened 2023-05-17
Discover ASCS/ERS clusters health and initial details

arbulu89:
# Description

Discovery of the ASCS/ERS clusters health.
In order to achieve that, we need to start building the cluster details.
In this case, the details is composed by a list of SAP systems (for a multi-sid environment) where each of the systems has its own details. Together with them, some common fields like `sbd` and `stopped_resources` are used.

Each sap system has the `distributed` field (check in the code docstring what this means).
If any of the sap systems in the cluster is **not** distributed, the health is set to `critical`.

PD: I have excluded the API schema updates in general. We will work on them once we have the whole details structure, to not overwhelm these PRs.

## How was this tested?

In order to test it properly, i have avoided continuing adding more and more fixtures files, as each single tests need new values. I have combined some factories together with the `put_in` function to update already existing fixtures for my purpose.

### DIFF
--- a/lib/trento/application/integration/discovery/payloads/cluster/crmmon_discovery_payload.ex
+++ b/lib/trento/application/integration/discovery/payloads/cluster/crmmon_discovery_payload.ex
@@ -165,6 +165,13 @@ defmodule Trento.Integration.Discovery.ClusterDiscoveryPayload.Crmmon do
 
     embeds_one :summary, Summary
 
+    embeds_many :nodes, Node, primary_key: false do
+      field :id, :string
+      field :name, :string
+      field :online, :boolean
+      field :unclean, :boolean
+    end
+
     embeds_many :resources, CrmmonResource
 
     embeds_many :groups, CrmmonGroup, primary_key: false do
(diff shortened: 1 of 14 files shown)


lib/trento/application/integration/discovery/payloads/cluster/crmmon_discovery_payload.ex
```diff
@@ -213,6 +221,12 @@ defmodule Trento.Integration.Discovery.ClusterDiscoveryPayload.Crmmon do
     |> validate_required_fields(@required_fields)
   end
 
+  defp nodes_changeset(nodes, attrs) do
+    nodes
+    |> cast(attrs, [:id, :name, :online, :unclean])
+    |> validate_required([:id, :name])
```
arbulu89, line 227: This is `validate_required` and not `validate_required_fields` in purpose, as the 2nd doesn't work properly in this case, as it filters out :id` as the main schema doesn't have this field

fabriziosestito (approved):

PR #1424, opened 2023-05-17
Deregistration UI websocket events

rtorrero:
# Description
This PR adds the code to handle the state changes in react upon the arrival of the `*_deregistered` events from the host, cluster and sap_system aggregates.

## How was this tested?
Manually, issuing Host deregistration requests in the REPL. End-to-end tests will be added once the UI changes are completed.


arbulu89 (commented):
@rtorrero I think this is a good chance to start adding jest unit tests to this changes

assets/js/state/clusters.test.js
```diff
@@ -0,0 +1,18 @@
+import clustersReducer, { removeCluster } from '@state/clusters';
+
+describe('Clusters reducer', () => {
+  it('should remove cluster from state', () => {
+    const cluster = { id: 'test-cluster-id' };
```
arbulu89, line 5: Most probably it would be better to use the `clusterFactory`.
```
const { id } = clusterFactory.build()
# or
const clusters = cluster.Factory.buildList(5) # with some further work, which would do a more robust test, just removing 1 from the list, instead of looking for an empty list
```

The same goes for the other tests

assets/js/state/sapSystems.js
```diff
@@ -100,6 +100,11 @@ export const sapSystemsListSlice = createSlice({
         return sapSystem;
       });
     },
+    removeSAPSystem: (state, { payload: { id } }) => {
+      state.sapSystems = state.sapSystems.filter(
+        (sapSystem) => id !== sapSystem.id
```
arbulu89, line 105: Following the others you would do `sapSystem.id !== id`

assets/js/state/sagas/clusters.js
```diff
@@ -0,0 +1,20 @@
...
+  yield put(removeCluster(payload));
+  yield put(
+    appendEntryToLiveFeed({
+      source: payload.name,
+      message: 'Cluster deregistered.',
+    })
+  );
+  yield put(
+    notify({
+      text: `The cluster ${payload.name} has been deregistered.`,
```
arbulu89, line 16: Keep in mind that not all clusters have a name, so this could end up empty.
Maybe something like `${payload.name || payload.id}` ?

assets/js/state/sagas/clusters.test.js
```diff
@@ -0,0 +1,19 @@
+import { recordSaga } from '@lib/test-utils';
+import { clusterDeregistered } from '@state/sagas/clusters';
+import { removeCluster } from '@state/clusters';
+
+describe('Clusters sagas', () => {
+  it('should trigger reducer to remove cluster', async () => {
```
arbulu89, line 6: Maybe `should remove the cluster` simply?

assets/js/state/sagas/clusters.test.js
```diff
@@ -0,0 +1,19 @@
...
+    const payload = {
```
arbulu89, line 7: Use factories

assets/js/state/sagas/index.js
```diff
@@ -205,6 +208,10 @@ function* watchHeartbeatFailed() {
   yield takeEvery('HEARTBEAT_FAILED', heartbeatFailed);
 }
 
+function* watchHostDeregistered() {
+  yield takeEvery('HOST_DEREGISTERED', hostDeregistered);
```
arbulu89, line 212: Let's move this `watch` to the saga files (as we do with the catalog and lastExecutions).
Besides that, `HOST_DEREGISTERED` should be a constant.
By now, these are stored in the `actions` file, but with the latest ADR i'm not sure where this should go. Reducer or saga? @dottorblaster wdyt?

lib/trento/application/projectors/sap_system_projector.ex
```diff
@@ -207,11 +209,14 @@ defmodule Trento.SapSystemProjector do
         _,
         _
       ) do
+    %SapSystemReadModel{sid: sid} = Repo.get!(SapSystemReadModel, sap_system_id)
```
arbulu89, line 212: Doesn't this raise some any failure in tests?
We should have a look, or at least try to include this change in the tests. 
Maybe the assertion looks for partial objects

arbulu89 (changes requested):
In general it looks good.
I think that we should use more the factories in the tests if possible.
Besides that, I commented some topics about the placement of functions and usage of constants.

Finally, What about the `databases`?

assets/js/state/sagas/clusters.test.js
jamie-suse: This test, is testing that the saga dispatches the correct action to the reducer. i.e. it does not validate the state change (because we test this in the reducer test).

arbulu89: But that's going technical...
At the end this saga aims to update our state internally, in our domain terminology this is to `remove a cluster`.
Anyway, I don't want to force anything, it is just my point of view

jamie-suse: Okay no worries, changed it. Was just making sure we are on the same page 👍

assets/js/state/clusters.test.js
```diff
@@ -0,0 +1,19 @@
...
+  it('should remove cluster from state', () => {
+    const cluster = clusterFactory.build();
+    const initialState = {
+      clusters: [cluster],
+    };
+
+    const action = removeCluster(cluster);
+
+    const expectedState = {
+      clusters: [],
```
arbulu89, line 14: I still think that this test is weak. I would prefer having a removal from a list where you have more than one element

arbulu89 (commented):
Hey @jamie-suse 
It looks much better.
Waiting to confirm where to put the new `action` constants.
And please, have a look on the comments I wrote about the `id/cluster_id` possible difference.
I think we would be fine to merge once that is checked

assets/js/state/actions/cluster.js
```diff
@@ -1,6 +0,0 @@
-export const CHECKS_SELECTED = 'CHECKS_SELECTED';
```
arbulu89, line 1: I don't understand why you do these changes in this PR :sweat_smile: 
I would create tech-debt tasks and work on them. 
Otherwise the reviews got out of hand.
(And I know that I asked you to not use action files, but you could have only do things belonging to the actual change there)

assets/js/state/sagas/clusters.js
```diff
@@ -0,0 +1,24 @@
+import { put, takeEvery } from 'redux-saga/effects';
+import { CLUSTER_DEREGISTERED, removeCluster } from '@state/clusters';
+import { appendEntryToLiveFeed } from '@state/liveFeed';
+import { notify } from '@state/actions/notifications';
+
+export function* clusterDeregistered({ payload: { name, cluster_id } }) {
+  yield put(removeCluster({ id: cluster_id }));
```
arbulu89, line 7: It was not simply better to change the projection code to use `id` instead of `cluster_id`?
To avoid transformations that we might have issues finding in the future

assets/js/state/sagas/hosts.test.js
```diff
@@ -0,0 +1,18 @@
+import { recordSaga } from '@lib/test-utils';
+import { hostDeregistered } from '@state/sagas/hosts';
+import { HOST_DEREGISTERED, removeHost } from '@state/hosts';
+import { hostFactory } from '@lib/test-utils/factories';
+
+describe('Hosts sagas', () => {
+  it('should remove the host', async () => {
+    const { id, hostname } = hostFactory.build();
```
arbulu89, line 8: This doesn't make much sense, right?
Why not simply do `const payload = hostFactory.build()` ?
Having more fields from the factory doesn't affect anything as far as I see

arbulu89 (approved):
@jamie-suse I'm approving this because I see it is stalling.
But I still think that there are things to be improved.

Other thing that now I miss is the instances live update.
We don't listen to any message to remove the sap app/database instances, so what happens when we remove them?
I would expect that they are removed from many views (host, sap system, database).
Could you track this missing thing and still work on it?

assets/js/state/sagas/hosts.test.js
jamie-suse: The extra fields cause the check using `toContainEqual()` to fail. Trying to see if there's an easy way to get it to work in Jest.

assets/js/state/sagas/clusters.test.js
```diff
@@ -0,0 +1,17 @@
...
+import { clusterDeregistered } from '@state/sagas/clusters';
+import { CLUSTER_DEREGISTERED, removeCluster } from '@state/clusters';
+import { clusterFactory } from '@lib/test-utils/factories';
+
+describe('Clusters sagas', () => {
+  it('should remove the cluster', async () => {
+    const { id, name } = clusterFactory.build();
+
+    const dispatched = await recordSaga(clusterDeregistered, {
+      type: CLUSTER_DEREGISTERED,
```
CDimonaco, line 11: We should use the value of CLUSTER_DEREGISTERED as expected value, not the const imported,

assets/js/state/sagas/hosts.test.js
```diff
@@ -0,0 +1,18 @@
...
+    const payload = { id, hostname };
+
+    const dispatched = await recordSaga(hostDeregistered, {
+      type: HOST_DEREGISTERED,
```
CDimonaco, line 12: Why we need the type field here>
I think we don't need the type value here, we could just easily add the payload like the other tests

CDimonaco (changes requested):

assets/js/state/sagas/clusters.js
```diff
@@ -0,0 +1,24 @@
...
+  yield put(
+    appendEntryToLiveFeed({
```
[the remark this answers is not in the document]
jamie-suse: Okay, will do 👍
Also other sagas in the codebase use `appendEntryToLiveFeed()` though, shouldn't we be removing them also? I can create a tech-debt ticket for this

stefanotorresi:
@jamie-suse I understood this was still undergoing review, why did you self-merge it?

PR #1425, opened 2023-05-17
Deregistration tombstoning rollup

CDimonaco:
# Description

After a successful rollup procedure, our aggregates should emit a tombstone event.

The tombstone event will be responsible for the lifespan stopping of aggregate and triggering the rollup for the tombstoned aggregate.

## How was this tested?

Automated tests


lib/trento/domain/host/events/host_tombstoned.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Trento.Domain.Events.HostTombstoned do
+  @moduledoc """
+    This event is emitted after a successful host deregistration, to tombstone and stop the host aggregate
+  """
+
+  use Trento.Event
+
+  defevent do
+    field(:host_id, Ecto.UUID)
```
fabriziosestito, line 9: formatter :(

fabriziosestito (commented):
@CDimonaco I have second thoughts about the Lifespan. Since the tombstone will always cause a cluster roll-up, do we need to stop the aggregate when tombstoning? Also, consider that after stopping the aggregate, it will start again since it will receive a rollup command.

arbulu89:
@CDimonaco @fabriziosestito 

I have shared feelings with the `Tombstone` term here. Tombstoning would mean deleting data, right?
But here we don't delete anything actually, it is more like archiving, as the data is still there. How are we gonna remove this "temporary" tombstones otherwise? Would we have some other policy for that?
Sorry for the philosophical question hehe

About @fabriziosestito lifespan comment, indeed it becomes a tricky thing, as we wouldn't stop actually the aggregate never, as the rollup would start it again. But you cannot stop neither on a roll up event, as this might be caused simply for having a lot of events. So it converts a chicken/egg loop as far as I see. Right?

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -809,7 +810,7 @@ defmodule Trento.ClusterTest do
       )
     end
 
-    test "should emit the ClusterDeregistered event when the last ClusterHost is deregistered and set the deregistration date into the state" do
+    test "should emit the ClusterDeregistered event and ClusterTombstoned event when the last ClusterHost is deregistered and set the deregistration date into the state" do
```
arbulu89, line 813: shorter `ClusterDeregisterd and ClusterTombstoned events`

test/trento/domain/host/host_test.exs
```diff
@@ -650,7 +651,7 @@ defmodule Trento.HostTest do
   end
 
   describe "deregistration" do
-    test "should emit the HostDeregistered event" do
+    test "should emit the HostDeregistered and HostTombstoned event" do
```
arbulu89, line 654: `event` -> `events`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1341,6 +1342,152 @@ defmodule Trento.SapSystemTest do
     end
   end
 
+  describe "tombstoning" do
+    test "should tombstone a deregistered SAP system when no applications and no databases instances are left" do
```
arbulu89, line 1346: `when no application and database instances are left`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1341,6 +1342,152 @@ defmodule Trento.SapSystemTest do
     end
   end
 
+  describe "tombstoning" do
+    test "should tombstone a deregistered SAP system when no application and no database instances are left" do
```
arbulu89, line 1346: i think that `no application and database instances are left` looks better

arbulu89 (approved):
Perfect!!!

PR #1428, opened 2023-05-18
Refactor cluster details

arbulu89:
# Description

Refactor of the `ClusterDetails` page to make it more pure. This enables its reusability (for future ASCS/ERS cluster details) and storybook stories addition. It includes the `ClusterDetailsPage` which does all the data management and `ClusterDetails` is now just a visualization component.

The unique real change it the usage of already existing selectors to create the `rendererNodes` constant.

I have extracted some sub-components (SBD and stopped resources sections) as they are required in ASCS/ERS cluster details as well.

cluster_details_story

## How was this tested?

Storybook added. Previous e2e tests should test the same.
I didn't add any new jest test as the content is exactly as it was before, but if you think I should add them, I will.

### DIFF
--- a/assets/js/components/ClusterDetails/ClusterDetails.stories.jsx
+++ b/assets/js/components/ClusterDetails/ClusterDetails.stories.jsx
@@ -1,0 +1,80 @@
+import React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+
+import {
+  clusterFactory,
+  checksExecutionCompletedFactory,
+  hostFactory,
+} from '@lib/test-utils/factories';
+
+import ClusterDetails from './ClusterDetails';
+
+export default {
+  title: 'ClusterDetails',
+  components: ClusterDetails,
+  decorators: [
+    (Story) => (
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
+    ),
+  ],
+};
+
+function ContainerWrapper({ children }) {
+  return (
+    <div className="max-w-7xl mx-auto px-4 sm:px-6 md:px-8">{children}</div>
+  );
+}
+
+export function Hana() {
+  const {
+    id,
+    name,
+    sid,
+    type,
+    selected_checks,
+    provider,
+    cib_last_written,
+    details,
+  } = clusterFactory.build({ type: 'hana_scale_up' });
+
+  const lastExecution = {
+    data: checksExecutionCompletedFactory.build({
+      result: 'passing',
+      passing_count: 3,
+      warning_count: 2,
+      critical_count: 1,
+    }),
+  };
+
+  const hosts = [
+    hostFactory.build({ hostname: details.nodes[0].name }),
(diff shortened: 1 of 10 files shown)


assets/js/components/ClusterDetails/ClusterDetailsPage.jsx
```diff
@@ -0,0 +1,66 @@
...
+    dispatch(updateLastExecution(clusterID));
+  }, [dispatch]);
+
+  const clusterHosts = useSelector(getClusterHosts(clusterID));
+
+  if (!cluster) {
+    return <div>Loading...</div>;
+  }
+
+  const renderedNodes = cluster.details?.nodes?.map((node) => ({
```
arbulu89, line 37: This is the major change in the refactoring. Beforehand, this conversion was handle with much more code.
I think it is enough to "merge" the hosts information with the cluster node details

assets/js/components/ClusterDetails/ClusterDetails.stories.jsx
```diff
@@ -0,0 +1,80 @@
...
+  ],
+};
+
+function ContainerWrapper({ children }) {
+  return (
+    <div className="max-w-7xl mx-auto px-4 sm:px-6 md:px-8">{children}</div>
+  );
+}
+
+export function Hana() {
```
nelsonkopliku, line 30: In the latest versions of storybook the way of writing stories has changed a bit https://storybook.js.org/docs/react/writing-stories/introduction

What do you think of having new stories written that way?
```js
export const Hana = {
  args: {
    clusterID: ...,
    clusterName: ...,
  },
  render: () => <ContainerWrapper><ClusterDetails ... /</ContainerWrapper>
}
```
[Accordion](https://github.com/trento-project/web/blob/main/assets/js/components/Accordion/Accordion.stories.jsx) and [CheckResultDetail](https://github.com/trento-project/web/blob/main/assets/js/components/ExecutionResults/CheckResultDetail/CheckResultDetail.stories.jsx) try to follow that way.

No strong opinions.

nelsonkopliku (commented):
Nice Xabi! Thanks!

Only doubt is about the storybook and the updated way of writing those suggested by the doc.

I'll leave it to your judgement to decide about adding or not jest tests for `ClusterDetails`. I know there wasn't any before.

dottorblaster (approved):
Amazing! Just as Nelson pointed out maybe we can try following the new format for storybook stories, but not strong opinions as well.

Thanks for this actually, was much needed

assets/js/components/ClusterDetails/ClusterDetails.stories.jsx
arbulu89: Hey @nelsonkopliku ,
Updated how story book story was written.
Indeed, i prefer this way, as it lets you customize the view super easily from storbook's web. Pretty cool!

nelsonkopliku (approved):

PR #1441, opened 2023-05-23
Set default value to additional_sids in cluster events

arbulu89:
Only affects when the event is created without this field (which is not really our case, as the values are coming from the commands) setting these values to empty list. The already existing events are not neither changed.

It doesn't really change how things work right now, but well, better safe than sorry.


dottorblaster (approved):
LGTM

nelsonkopliku (approved):

PR #1445, opened 2023-05-23
Persist selected filters in checks results inside Redux

dottorblaster:
Saving applied filters into a redux slice so when the user comes back we just restore previously applied state

## How was this tested?
Jest test added

### DIFF
--- a/assets/js/components/ExecutionResults/ChecksResultFilters.jsx
+++ b/assets/js/components/ExecutionResults/ChecksResultFilters.jsx
@@ -4,14 +4,25 @@ import Filter from '@components/Table/Filter';
 
 export const RESULT_FILTER_FIELD = 'result';
 
-function ChecksResultFilters({ onChange }) {
+const defaultSavedFilters = [];
+
+const getFilters = (savedFilters, searchParams) =>
+  savedFilters.length >= 0 && searchParams.getAll('health').length === 0
+    ? savedFilters
+    : searchParams.getAll('health');
+
+function ChecksResultFilters({
+  savedFilters = defaultSavedFilters,
+  onChange = () => {},
+  onSave = () => {},
+}) {
   const [filtersForField, setFiltersForField] = useState({});
   const [searchParams, setSearchParams] = useSearchParams();
   // This structure is the foundation for a multi field filters
   // we can reuse later this structure in other parts of the application
 
   useEffect(() => {
-    const selectedFilters = searchParams.getAll('health');
+    const selectedFilters = getFilters(savedFilters, searchParams);
 
     setFiltersForField({
       RESULT_FILTER_FIELD: {
@@ -21,7 +32,7 @@ function ChecksResultFilters({ onChange }) {
         values: selectedFilters,
       },
     });
-  }, [searchParams]);
+  }, [searchParams, savedFilters]);
 
   useEffect(() => {
     if (Object.keys(filtersForField).length >= 0) {
@@ -40,8 +51,9 @@ function ChecksResultFilters({ onChange }) {
         key={RESULT_FILTER_FIELD}
         title="checks result"
         options={['passing', 'warning', 'critical', 'unknown']}
-        value={searchParams.getAll('health')}
+        value={getFilters(savedFilters, searchParams)}
         onChange={(list) => {
+          onSave(list);
           setSearchParams(createSearchParams({ health: list }));
         }}
       />
--- a/assets/js/components/ExecutionResults/ExecutionResults.jsx
+++ b/assets/js/components/ExecutionResults/ExecutionResults.jsx
@@ -23,6 +23,13 @@ import CheckResultOutline from './CheckResultOutline';
 import ExecutionHeader from './ExecutionHeader';
 import ExecutionContainer from './ExecutionContainer';
 
+// To have an array as a default prop that is also used in a useEffect's dependency
+// array we need to declare it outside the scope as a `const`, in order to prevent
+// rerendering loops.
+//
+// https://github.com/facebook/react/issues/18123
+const defaultSavedFilters = [];
+
 const resultsTableConfig = {
   usePadding: false,
   rowClassName: 'tn-check-result-row',
@@ -101,9 +108,11 @@ function ExecutionResults({
   executionData,
   executionError,
   clusterSelectedChecks = [],
+  savedFilters = defaultSavedFilters,
   onCatalogRefresh = () => {},
   onLastExecutionUpdate = () => {},
   onStartExecution = () => {},
+  onSaveFilters = () => {},
 }) {
   const [predicates, setPredicates] = useState([]);
   const [selectedCheck, setSelectedCheck] = useState(null);
@@ -117,6 +126,7 @@ function ExecutionResults({
       onLastExecutionUpdate();
     }
   };
+
   const checksResults = getCheckResults(executionData);
   const catalogCategoryList = getCatalogCategoryList(catalog, checksResults);
   const tableData = checksResults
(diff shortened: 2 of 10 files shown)


assets/js/components/ExecutionResults/ChecksResultFilters.jsx
```diff
@@ -4,14 +4,25 @@ import Filter from '@components/Table/Filter';
 
 export const RESULT_FILTER_FIELD = 'result';
 
-function ChecksResultFilters({ onChange }) {
+const getFilters = (savedFilters, searchParams) =>
+  savedFilters &&
```
nelsonkopliku, line 8: If I am getting it right in the worst case `savedFilters` comes as an empty array, so this initial check might be not necessary.

assets/js/components/ExecutionResults/ChecksResultFilters.jsx
```diff
@@ -4,14 +4,25 @@ import Filter from '@components/Table/Filter';
...
+  savedFilters.length >= 0 &&
+  searchParams.getAll('health').length === 0
```
nelsonkopliku, line 10: We could extract this `searchParams.getAll('health')` outside of `getFilters` function. 
I am aware it wouldn't change much :sweat_smile:, so not a biggie. However if you want to give it a go here's the changes it would require.

```js
function ChecksResultFilters() {
  //...
  const qsFilters = searchParams.getAll('health');

  useEffect(() => {
    const selectedFilters = getFilters(savedFilters, qsFilters);
    
    setFiltersForField({
        //...
    });
  }, [searchParams, savedFilters]);

  //...
  return (
    <div className="flex">
      <Filter
        //...
        value={getFilters(savedFilters, qsFilters)}
      />
    </div>
  );
}
```

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
```diff
@@ -313,6 +313,47 @@ describe('ExecutionResults', () => {
...
+        executionStarted={executionStarted}
+        catalogError={error}
+        executionLoading={executionLoading}
+        executionData={executionData}
+        executionError={executionError}
+        savedFilters={['passing']}
+      />
+    );
+
+    expect(screen.getAllByText('test-cluster')).toHaveLength(2);
```
nelsonkopliku, line 348: What about squashing this test and the previous one in a `it.each` version with two scenarios? 
One scenario that does not pass a `savedFilters` (or an empty one) but it provides the route with querystring, another scenario that does provide the `savedFilters` without querystring.

At the end the expectations are the same for the two tests :thinking:

nelsonkopliku (commented):
Great job!

Just a couple of suggestions if they help.

Additionally, as we discussed, just noting that the filter in the query string gets wiped out when navigating among the pages.
However this does not impact the feature which works just fine. I am ok with the tradeoff if it's too much of a hassle re-applying the qs when we have saved filters and we navigate back to the executions results.

assets/js/state/clusterChecksResultsFilters.js
```diff
@@ -0,0 +1,17 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+const initialState = {};
+
+export const clusterChecksResultsFiltersSlice = createSlice({
+  name: 'clusterChecksResultsFilters',
+  initialState,
+  reducers: {
+    setClusterFilters: (state, { payload: { clusterID, filters } }) => {
```
arbulu89, line 9: What about making this generic `setChecksResultsFilters`.
Getting the `{ id: filters }` payload. I'm just thinking for future uses where we don't only have checks for clusters.
All of our resouces (clusters, hosts, sap systesms) are identified by uuid, so there wouldn't be any collision.

assets/js/components/ExecutionResults/ExecutionResults.jsx
```diff
@@ -101,9 +101,11 @@ function ExecutionResults({
   executionData,
   executionError,
   clusterSelectedChecks = [],
+  savedFilters,
```
arbulu89, line 104: Could we maybe add an empty list as default?
To make things a bit more secure, and in fact it would even help to story book

arbulu89 (commented):
@dottorblaster The code looks good.
I only commented the naming part (our beloved naming).
I think we could generalize this. We will have sooner than later checks that are not only cluster based, so what about making the state already generic?
The usage would be exactly the same, we just need to build the payload correctly, that's all.

assets/js/components/ExecutionResults/ExecutionResults.jsx
dottorblaster: Let me check if nothing explodes, sure

assets/js/state/clusterChecksResultsFilters.js
```diff
@@ -0,0 +1,17 @@
+import { createSlice } from '@reduxjs/toolkit';
```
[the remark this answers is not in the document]
dottorblaster: Yes, you're right

assets/js/state/clusterChecksResultsFilters.js
dottorblaster: Right again :D

assets/js/components/ExecutionResults/ExecutionResults.jsx
dottorblaster: I managed to assign a default prop but I had to declare it outside the inline scope. Left a comment about that, hope it's ok

assets/js/components/ExecutionResults/ChecksResultFilters.jsx
dottorblaster: I tried and the component exploded in my face :sweat_smile:

assets/js/components/ExecutionResults/ChecksResultFilters.jsx
dottorblaster: Could come as undefined, updated it to have a default prop :+1:

assets/js/components/ExecutionResults/ExecutionResults.test.jsx
nelsonkopliku: If it calls for too much refactor that you don't feel like addressing, keep it as it is.

nelsonkopliku (approved):

assets/js/state/selectors/checksResultsFilters.js
```diff
@@ -0,0 +1,4 @@
+export const getSelectedFilters =
+  (clusterID) =>
```
arbulu89, line 2: I would change `clusterID` by `groupID/resourceID` once we started making it generic

assets/js/state/checksResultsFilters.js
```diff
@@ -0,0 +1,17 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+const initialState = {};
+
+export const checksResultsFiltersSlice = createSlice({
+  name: 'checksResultsFilters',
+  initialState,
+  reducers: {
+    setSelectedFilters: (state, { payload: { clusterID, filters } }) => {
```
arbulu89, line 9: `clusterID` by `resourceID` maybe?

assets/js/components/ExecutionResults/ExecutionResultsPage.jsx
```diff
@@ -62,9 +66,13 @@ function ExecutionResultsPage() {
       executionData={executionData}
       executionError={executionError}
       clusterSelectedChecks={cluster?.selected_checks}
+      savedFilters={savedFilters}
       onStartExecution={(clusterId, hosts, selectedChecks, navigate) =>
         dispatch(executionRequested(clusterId, hosts, selectedChecks, navigate))
       }
+      onSaveFilters={(filters) =>
+        dispatch(setSelectedFilters({ clusterID, filters }))
```
arbulu89, line 74: `setSelectedFilters({ resourceID: clusterID, filters })

arbulu89 (approved):
@dottorblaster It looks great!
I would simply change `clusterID` by `resourceID/groupID` to finish making it generic now that we changed the files/state names.
(Maybe run `find all` thing in the changes hehe, as `clusterID` is everywhere, specially on the tests)

PR #1450, opened 2023-05-24
Add endpoint to deregister a host

jamie-suse:
# Description

Adds endpoint `DELETE /api/v1/hosts/{host_id}` to request the server to deregister a host.

Response behaviour:

- If last heartbeat is after timeout period: `204`
- If last heartbeat is within timeout period: `422`
- If host with ID `host_id` doesn't exist: `404`
- If attempting to deregister a host that is already deregistered: `404`

## How was this tested?

By adding unit tests

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -162,6 +162,9 @@ config :trento,
   api_key_authentication_enabled: true,
   jwt_authentication_enabled: true
 
+config :trento,
+  deregistration_debounce: :timer.seconds(0)
+
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
 import_config "#{config_env()}.exs"
--- a/config/config.exs
+++ b/config/config.exs
@@ -82,6 +82,10 @@ config :trento, :pow,
 # Agent heartbeat interval. Adding one extra second to the agent 5s interval to avoid glitches
 config :trento, Trento.Heartbeats, interval: :timer.seconds(6)
 
+# This is passed to the frontend as the time after the last heartbeat
+# to wait before displaying the deregistration button
+config :trento, deregistration_debounce: :timer.seconds(0)
+
 config :trento, Trento.Scheduler,
   jobs: [
     heartbeat_check: [
@@ -161,9 +165,6 @@ config :trento, :jwt_authentication,
 config :trento,
   api_key_authentication_enabled: true,
   jwt_authentication_enabled: true
-
-config :trento,
-  deregistration_debounce: :timer.seconds(0)
 
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
--- a/config/test.exs
+++ b/config/test.exs
@@ -41,6 +41,13 @@ config :logger, level: :warn
 # Initialize plugs at runtime for faster test compilation
 config :phoenix, :plug_init_mode, :runtime
 
+# Agent heartbeat interval. Adding one extra second to the agent 5s interval to avoid glitches
+config :trento, Trento.Heartbeats, interval: :timer.seconds(6)
+
+# This is passed to the frontend as the time after the last heartbeat
+# to wait before displaying the deregistration button
+config :trento, deregistration_debounce: :timer.seconds(5)
+
 config :trento,
   api_key_authentication_enabled: false,
   jwt_authentication_enabled: false
--- a/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
+++ b/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
@@ -3,11 +3,7 @@ defimpl Trento.Support.Middleware.Enrichable,
   alias Trento.Domain.Commands.RequestHostDeregistration
 
   alias Trento.Repo
-
-  alias Trento.Heartbeat
   alias Trento.HostReadModel
-
-  import Ecto.Query
 
   @heartbeat_interval Application.compile_env!(:trento, Trento.Heartbeats)[:interval]
   @deregistration_debounce Application.compile_env!(
@@ -21,25 +17,29 @@ defimpl Trento.Support.Middleware.Enrichable,
           | {:error, :host_alive}
           | {:error, :host_not_registered}
   def enrich(%RequestHostDeregistration{host_id: host_id} = command, _) do
-    with true <- Repo.exists?(HostReadModel, id: host_id),
-         :ok <- host_deregisterable(host_id) do
-      {:ok, command}
-    else
-      {:error, :host_alive} -> {:error, :host_alive}
-      _ -> {:error, :host_not_registered}
-    end
+    Repo.get(HostReadModel, host_id)
+    |> Repo.preload([:heartbeat_timestamp])
+    |> host_deregisterable(command)
   end
 
-  @spec host_deregisterable(Ecto.UUID) :: :ok | {:error, :host_alive}
-  defp host_deregisterable(host_id) do
-    query =
-      from(h in Heartbeat,
-        where:
-          ^DateTime.utc_now() <
-            datetime_add(h.timestamp, @total_deregistration_debounce, "millisecond") and
-            h.agent_id == ^host_id
-      )
+  defp host_deregisterable(
+         %HostReadModel{heartbeat_timestamp: nil},
+         %RequestHostDeregistration{} = command
+       ),
+       do: {:ok, command}
 
-    if Repo.exists?(query), do: {:error, :host_alive}, else: :ok
+  defp host_deregisterable(
+         %HostReadModel{heartbeat_timestamp: %Trento.Heartbeat{timestamp: timestamp}},
+         %RequestHostDeregistration{} = command
+       ) do
+    if :lt ==
+         DateTime.compare(
+           DateTime.utc_now(),
+           DateTime.add(timestamp, @total_deregistration_debounce, :millisecond)
+         ),
+       do: {:error, :host_alive},
+       else: {:ok, command}
   end
+
+  defp host_deregisterable(_, _), do: {:error, :host_not_registered}
 end
--- a/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
+++ b/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
@@ -21,27 +21,25 @@ defimpl Trento.Support.Middleware.Enrichable,
           | {:error, :host_alive}
           | {:error, :host_not_registered}
   def enrich(%RequestHostDeregistration{host_id: host_id} = command, _) do
-    host_exists = Repo.exists?(HostReadModel, id: host_id)
+    with true <- Repo.exists?(HostReadModel, id: host_id),
+         :ok <- host_deregisterable(host_id) do
+      {:ok, command}
+    else
+      {:error, :host_alive} -> {:error, :host_alive}
+      _ -> {:error, :host_not_registered}
+    end
+  end
 
+  @spec host_deregisterable(Ecto.UUID) :: :ok | {:error, :host_alive}
+  defp host_deregisterable(host_id) do
     query =
       from(h in Heartbeat,
         where:
-          h.timestamp >
-            ^DateTime.add(DateTime.utc_now(), -@total_deregistration_debounce, :millisecond) and
+          ^DateTime.utc_now() <
+            datetime_add(h.timestamp, @total_deregistration_debounce, "millisecond") and
             h.agent_id == ^host_id
       )
 
-    heartbeat_invalid = Repo.exists?(query)
-
-    cond do
-      host_exists and heartbeat_invalid ->
-        {:error, :host_alive}
-
-      host_exists and not heartbeat_invalid ->
-        {:ok, command}
-
-      true ->
-        {:error, :host_not_registered}
-    end
+    if Repo.exists?(query), do: {:error, :host_alive}, else: :ok
   end
 end
--- a/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
+++ b/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
@@ -1,0 +1,47 @@
+defimpl Trento.Support.Middleware.Enrichable,
+  for: Trento.Domain.Commands.RequestHostDeregistration do
+  alias Trento.Domain.Commands.RequestHostDeregistration
+
+  alias Trento.Repo
+
+  alias Trento.Heartbeat
+  alias Trento.HostReadModel
+
+  import Ecto.Query
+
+  @heartbeat_interval Application.compile_env!(:trento, Trento.Heartbeats)[:interval]
+  @deregistration_debounce Application.compile_env!(
+                             :trento,
+                             :deregistration_debounce
+                           )
+  @total_deregistration_debounce @heartbeat_interval + @deregistration_debounce
+
+  @spec enrich(RequestHostDeregistration.t(), map) ::
+          {:ok, RequestHostDeregistration.t()}
+          | {:error, :host_alive}
+          | {:error, :host_not_registered}
+  def enrich(%RequestHostDeregistration{host_id: host_id} = command, _) do
+    host_exists = Repo.exists?(HostReadModel, id: host_id)
+
+    query =
+      from(h in Heartbeat,
+        where:
+          h.timestamp >
+            ^DateTime.add(DateTime.utc_now(), -@total_deregistration_debounce, :millisecond) and
+            h.agent_id == ^host_id
+      )
+
+    heartbeat_invalid = Repo.exists?(query)
+
+    cond do
+      host_exists and heartbeat_invalid ->
+        {:error, :host_alive}
+
+      host_exists and not heartbeat_invalid ->
+        {:ok, command}
+
+      true ->
+        {:error, :host_not_registered}
+    end
+  end
+end
--- a/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
+++ b/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
@@ -23,13 +23,16 @@ defimpl Trento.Support.Middleware.Enrichable,
   end
 
   defp host_deregisterable(
-         %HostReadModel{heartbeat_timestamp: nil},
+         %HostReadModel{heartbeat_timestamp: nil, deregistered_at: nil},
          %RequestHostDeregistration{} = command
        ),
        do: {:ok, command}
 
   defp host_deregisterable(
-         %HostReadModel{heartbeat_timestamp: %Trento.Heartbeat{timestamp: timestamp}},
+         %HostReadModel{
+           heartbeat_timestamp: %Trento.Heartbeat{timestamp: timestamp},
+           deregistered_at: nil
+         },
          %RequestHostDeregistration{} = command
        ) do
     if :lt ==
--- a/lib/trento/application/read_models/host_read_model.ex
+++ b/lib/trento/application/read_models/host_read_model.ex
@@ -32,6 +32,10 @@ defmodule Trento.HostReadModel do
       foreign_key: :host_id,
       preload_order: [desc: :identifier]
 
+    has_one :heartbeat_timestamp, Trento.Heartbeat,
+      references: :id,
+      foreign_key: :agent_id
+
     field :deregistered_at, :utc_datetime_usec
   end
 
--- a/lib/trento/application/usecases/hosts/hosts.ex
+++ b/lib/trento/application/usecases/hosts/hosts.ex
@@ -6,11 +6,21 @@ defmodule Trento.Hosts do
   import Ecto.Query
 
   alias Trento.{
+    Heartbeat,
     HostReadModel,
+    Repo,
     SlesSubscriptionReadModel
   }
 
-  alias Trento.Repo
+  alias Trento.Support.DateService
+
+  alias Trento.Domain.Commands.RequestHostDeregistration
+
+  @heartbeat_interval Application.compile_env!(:trento, Trento.Heartbeats)[:interval]
+  @deregistration_debounce Application.compile_env!(
+                             :trento,
+                             :deregistration_debounce
+                           )
 
   @spec get_all_hosts :: [HostReadModel.t()]
   def get_all_hosts do
@@ -18,7 +28,7 @@ defmodule Trento.Hosts do
     |> where([h], not is_nil(h.hostname) and is_nil(h.deregistered_at))
     |> order_by(asc: :hostname)
     |> Repo.all()
-    |> Repo.preload([:sles_subscriptions, :tags])
+    |> Repo.preload([:sles_subscriptions, :tags, :heartbeat_timestamp])
   end
 
   @spec get_all_sles_subscriptions :: non_neg_integer()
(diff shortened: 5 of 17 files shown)


config/config.exs
```diff
@@ -162,6 +162,9 @@ config :trento,
   api_key_authentication_enabled: true,
   jwt_authentication_enabled: true
 
+config :trento,
+  deregistration_debounce_seconds: :timer.seconds(0)
```
jamie-suse, line 166: As discussed, the default value for the deregistration debounce period should be set to 0, but remain configurable.

config/config.exs
```diff
@@ -162,6 +162,9 @@ config :trento,
   api_key_authentication_enabled: true,
   jwt_authentication_enabled: true
 
+config :trento,
+  deregistration_debounce: :timer.seconds(0)
```
jamie-suse, line 166: As discussed, the default value for the deregistration debounce period should be set to 0, but remain configurable.

test/trento/application/usecases/hosts_test.exs
```diff
@@ -38,4 +41,70 @@ defmodule Trento.HostsTest do
...
+        fn -> last_heartbeat end
+      )
+
+      %{id: id} = insert(:host, heartbeat: Health.critical())
+      insert(:heartbeat, agent_id: id, timestamp: Trento.Support.DateService.Mock.utc_now())
+
+      expect(
+        Trento.Support.DateService.Mock,
+        :utc_now,
+        fn -> DateTime.add(last_heartbeat, 500, :millisecond) end
```
jamie-suse, line 104: Should we document that the default value of `Trento.Heartbeats.interval` is set to 6 seconds? Otherwise seeing this deregistration attempt within 500 ms of the last heartbeat seems a bit mysterious in my opinion.

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -6,27 +6,38 @@ defmodule Trento.Hosts do
...
     |> Repo.all()
-    |> Repo.preload([:sles_subscriptions, :tags])
+    |> Repo.preload([:sles_subscriptions, :tags, :heartbeat_timestamp])
   end
 
   @spec get_all_sles_subscriptions :: non_neg_integer()
   def get_all_sles_subscriptions do
     query =
-      from s in SlesSubscriptionReadModel,
+      from(s in SlesSubscriptionReadModel,
```
CDimonaco, line 37: Remove parenthesis pls, the editor formatter could lead to this situations :/

lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -28,6 +33,31 @@ defmodule TrentoWeb.V1.HostController do
...
+      not_found: TrentoWeb.OpenApi.Schema.NotFound.response(),
+      unprocessable_entity: OpenApiSpex.JsonErrorResponse.response()
+    ]
+
+  @spec delete(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def delete(conn, %{id: host_id}) do
+    case Hosts.deregister_host(host_id) do
+      :ok -> send_resp(conn, 204, "")
+      {:error, :host_alive} -> send_resp(conn, 422, "")
+      {:error, _} -> send_resp(conn, 404, "")
```
CDimonaco, line 57: I would catch the real error, here, bubbling up from the context, we could end up treating an unrelated error as 404 without proper logging and handling, set up a proper error and catch here, the unexpected error will be catch by the fallback controller with a proper status code

lib/trento_web/router.ex
```diff
@@ -75,6 +75,8 @@ defmodule TrentoWeb.Router do
       get "/sap_systems/health", HealthOverviewController, :overview
       get "/databases", SapSystemController, :list_databases
 
+      delete "/hosts/:id", HostController, :delete
```
CDimonaco, line 78: Can you move it under the /hosts/id/tags? so we can have visually grouped routes

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -36,4 +47,41 @@ defmodule Trento.Hosts do
...
+
+      _ ->
+        maybe_dispatch_host_deregistration_request(host_id, date_service)
+    end
+  end
+
+  defp commanded,
+    do: Application.fetch_env!(:trento, Trento.Commanded)[:adapter]
+
+  defp maybe_dispatch_host_deregistration_request(host_id, date_service) do
```
CDimonaco, line 65: I think we could have a discussion about that, not on the logic, but on the place in which we should put the logic, we could add it to a a enrichable middleware, so we protect the command dispatching, making sure that the command will fail with that rule, right know we are making this rule only in the context, and the command could be dispatched without this rule.

If you take a look at others enrichable middleware, like enrich_application_instance, you can find that we are enforcing a business rule into the command dispatching.

This is more safe and aligned with the rest of codebase, it's a matter to add a new defimpl and copy the code, but I think it's more safe.

Ofcourse cc @arbulu89 @fabriziosestito

CDimonaco (changes requested):

test/trento/application/usecases/hosts_test.exs
CDimonaco: yes!

arbulu89: This is already documented in the `config` file. Where or what do you plan to document?

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -36,4 +47,41 @@ defmodule Trento.Hosts do
...
+    now = date_service.utc_now()
+    total_deregistration_debounce = @heartbeat_interval + @deregistration_debounce
```
arbulu89, line 67: Could we maybe move this to a constant above? No need to compute it every time.
Wdyt?

config/config.exs
arbulu89: Could be add a comment above? The name doesn't say me much hehe

lib/trento/application/usecases/hosts/hosts.ex
arbulu89: That's a really good point @CDimonaco 
Putting in the middleware moves the logic closer to the domain, so less chances to have unwanted behaviours.
I don't think of any scenario where we don't want to enforce this logic, so I would agree that the middleware is a better place.

arbulu89, line 65: Let's move this function above of `defp commanded` and respect the order of precedence

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -36,4 +47,41 @@ defmodule Trento.Hosts do
...
+
+    query =
+      from(h in Heartbeat,
+        where:
+          h.timestamp >
+            ^DateTime.add(now, -total_deregistration_debounce, :millisecond) and
```
arbulu89, line 73: Going back in time with the `-` here hehe
I find this "negative" logic a bit confusing, but well, I guess it works.

I found the `h.timestamp + total_debounce < now` and sending the command in `true` easier.
But not a hard opinion, leave it as you best think

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -36,4 +47,41 @@ defmodule Trento.Hosts do
         subscription_count
     end
   end
+
+  @spec deregister_host(Ecto.UUID.t(), DateService) :: :ok | {:error, any}
```
arbulu89, line 51: Could be have something more specific for `{:error, any}`. `{:error, :host_alive} | {:error, :host_not_found}`

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -36,4 +47,41 @@ defmodule Trento.Hosts do
...
+  def deregister_host(host_id, date_service \\ DateService) do
+    case Repo.get_by(HostReadModel, id: host_id) do
+      nil ->
+        {:error, :host_not_found}
```
arbulu89, line 55: Maybe better `host_not_registered` ?
This is what we use in the fallback controller as well.
This way you don't need to handle this in the controller itself

lib/trento_web/openapi/schema/host.ex
```diff
@@ -69,6 +69,20 @@ defmodule TrentoWeb.OpenApi.Schema.Host do
...
           items: SlesSubscription
+        },
+        deregistered_at: %Schema{
+          title: "DeregisteredAt",
+          description: "Timestamp of the last deregistration of the host",
+          type: :string,
+          nullable: true,
+          format: :"date-time"
+        },
+        heartbeat_timestamp: %Schema{
```
arbulu89, line 80: Is this correct?
I though we were sending a struct here with `agent_id/timestamp` combo

test/trento/application/usecases/hosts_test.exs
```diff
@@ -38,4 +41,70 @@ defmodule Trento.HostsTest do
       refute deregistered_host.id in hosts_ids
     end
   end
+
+  describe "deregister_host" do
+    test "should deregister a host with unknown health" do
```
arbulu89, line 46: I'm totally disoriented with these tests.
As far as I see the code, I don't see any place where the heartbeat state is even used.
How does this work?
Please, have a look on the tests. I don't find any logic on this.

And maybe add `setup :verify_on_exit!` on top of the test file. I think almost all (if not all) the expects are wrong.

arbulu89 (changes requested):
@jamie-suse @rtorrero 
Some comments from my side.
Please have a look on the tests.

lib/trento/application/usecases/hosts/hosts.ex
rtorrero: @CDimonaco Very good point, I like the fact that this also protects the command from any other use outside of the usecase. We added it, but we are a bit unsure about the location we chose. It seems wrong to put it inside the `discovery` structure, but it's where the most similar code currently lives at (because other registration middleware comes from the discovery.

Do you have a better suggestion as where to place this?

arbulu89: @rtorrero I think the suggestion remains the same. To implement the logic in an enrichment middleware protocol

config/config.exs
```diff
@@ -82,6 +82,10 @@ config :trento, :pow,
 # Agent heartbeat interval. Adding one extra second to the agent 5s interval to avoid glitches
 config :trento, Trento.Heartbeats, interval: :timer.seconds(6)
 
+# This is passed to the frontend as the time after the last heartbeat
+# to wait before displaying the deregistration button
+config :trento, deregistration_debounce: :timer.seconds(0)
```
CDimonaco, line 87: Why this is zero? I don't think this a good value, but anyway this is not really used in this pr other then passed to the frontend, which is a no-op without the frontend, pls remove it and we'll discuss this when we will make the frontend.

I know that we add it in the middleware but right know is zero and this a no-op, let's rediscuss it on the frontend

config/test.exs
```diff
@@ -41,6 +41,13 @@ config :logger, level: :warn
 # Initialize plugs at runtime for faster test compilation
 config :phoenix, :plug_init_mode, :runtime
 
+# Agent heartbeat interval. Adding one extra second to the agent 5s interval to avoid glitches
+config :trento, Trento.Heartbeats, interval: :timer.seconds(6)
+
+# This is passed to the frontend as the time after the last heartbeat
+# to wait before displaying the deregistration button
+config :trento, deregistration_debounce: :timer.seconds(5)
```
CDimonaco, line 49: Same consideration as above

lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
```diff
@@ -0,0 +1,47 @@
...
+                             :deregistration_debounce
+                           )
+  @total_deregistration_debounce @heartbeat_interval + @deregistration_debounce
+
+  @spec enrich(RequestHostDeregistration.t(), map) ::
+          {:ok, RequestHostDeregistration.t()}
+          | {:error, :host_alive}
+          | {:error, :host_not_registered}
+  def enrich(%RequestHostDeregistration{host_id: host_id} = command, _) do
+    host_exists = Repo.exists?(HostReadModel, id: host_id)
```
CDimonaco, line 24: Please short circuit here, we don't need to perform the query below, if the host does not exists

You can use different functions, or maybe a with - else block, but it's better if we avoid to perform useless query, not perfomance wise obvs, the query is really simple, but for general correctness

lib/trento/application/read_models/host_read_model.ex
```diff
@@ -32,6 +32,10 @@ defmodule Trento.HostReadModel do
       foreign_key: :host_id,
       preload_order: [desc: :identifier]
 
+    has_one :heartbeat_timestamp, Trento.Heartbeat,
```
CDimonaco, line 35: I will replace :heartbeat_timestamp with :hearbeat as association name, when defining association, it's better to stick with the entity name, in this case hearbeat, we know that's is only one because is a 1:1 association, and it does not appear that we perform some select or manipulation on it, so it' s not a timestamp, it's an heartbeat.

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -7,18 +7,21 @@ defmodule Trento.Hosts do
...
+  alias Trento.Domain.Commands.RequestHostDeregistration
 
   @spec get_all_hosts :: [HostReadModel.t()]
   def get_all_hosts do
     HostReadModel
     |> where([h], not is_nil(h.hostname) and is_nil(h.deregistered_at))
     |> order_by(asc: :hostname)
     |> Repo.all()
-    |> Repo.preload([:sles_subscriptions, :tags])
+    |> Repo.preload([:sles_subscriptions, :tags, :heartbeat_timestamp])
```
CDimonaco, line 24: Same consideration as above, use hearbeat as association name

lib/trento_web/openapi/schema/host.ex
```diff
@@ -69,6 +69,23 @@ defmodule TrentoWeb.OpenApi.Schema.Host do
...
```
CDimonaco, line 80: Same consideration as above, the association and perhaps the entity field should be hearbeat, this is not a timestamp and is misleading

lib/trento_web/templates/page/index.html.heex
```diff
@@ -3,7 +3,8 @@
 <script>
 const config = {
     grafanaPublicUrl: '<%= @grafana_public_url %>',
-    checksServiceBaseUrl: '<%= @check_service_base_url %>'
+    checksServiceBaseUrl: '<%= @check_service_base_url %>',
+    deregistrationDebounce: '<%= @deregistration_debounce %>',
```
CDimonaco, line 7: Same consideration as above for the debounce

test/trento/application/integration/discovery/protocol/enrich_request_host_deregistration_test.exs
```diff
@@ -0,0 +1,63 @@
...
+  use ExUnit.Case
+  use Trento.DataCase
+
+  import Trento.Factory
+
+  alias Trento.Domain.Commands.RequestHostDeregistration
+  alias Trento.Support.Middleware.Enrichable
+
+  describe "enrich RequestHostDeregistration" do
+    test "should return the original command if a host with critical health is requested to be deregistered" do
```
CDimonaco, line 11: please, specify more the condition, critical health could be misleading, we don't pass health critical in these tests, we specify a "old hearbeat", rephrase the tests with this in mind

test/trento/application/integration/discovery/protocol/enrich_request_host_deregistration_test.exs
```diff
@@ -0,0 +1,63 @@
...
+      %{id: id} = insert(:host)
+      insert(:heartbeat, agent_id: id, timestamp: DateTime.add(DateTime.utc_now(), -30, :second))
+
+      command = RequestHostDeregistration.new!(%{host_id: id, requested_at: now})
+
+      assert {:ok, %RequestHostDeregistration{host_id: id, requested_at: now}} ==
+               Enrichable.enrich(command, %{})
+    end
+
+    test "should return the original command if a host with unknown health is requested to be deregistered after debounce" do
```
CDimonaco, line 23: Same as before but for unknown health

test/trento/application/integration/discovery/protocol/enrich_request_host_deregistration_test.exs
```diff
@@ -0,0 +1,63 @@
...
+
+      %{id: id} = insert(:host)
+
+      command = RequestHostDeregistration.new!(%{host_id: id, requested_at: now})
+
+      assert {:ok, %RequestHostDeregistration{host_id: id, requested_at: now}} ==
+               Enrichable.enrich(command, %{})
+    end
+
+    test "should return an error if a host with critical health is requested to be deregistered before debounce timer" do
```
CDimonaco, line 34: same as before

test/trento/application/integration/discovery/protocol/enrich_request_host_deregistration_test.exs
```diff
@@ -0,0 +1,63 @@
...
+
+      %{id: id} = insert(:host)
+      insert(:heartbeat, agent_id: id, timestamp: DateTime.add(DateTime.utc_now(), -6, :second))
+
+      command = RequestHostDeregistration.new!(%{host_id: id, requested_at: now})
+
+      assert {:error, :host_alive} == Enrichable.enrich(command, %{})
+    end
+
+    test "should return an error if a host with passing health is requested to be deregistered" do
```
CDimonaco, line 45: same

lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -28,6 +33,31 @@ defmodule TrentoWeb.V1.HostController do
...
+      no_content: "The host has been deregistered",
+      not_found: TrentoWeb.OpenApi.Schema.NotFound.response(),
+      unprocessable_entity: OpenApiSpex.JsonErrorResponse.response()
+    ]
+
+  @spec delete(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def delete(conn, %{id: host_id}) do
+    case Hosts.deregister_host(host_id) do
+      :ok -> send_resp(conn, 204, "")
+      {:error, :host_alive} -> send_resp(conn, 422, "")
```
CDimonaco, line 56: We are seding a no content without a reason? We should have the appropriate response for the 422 as described in OpenApiSpex.JsonErrorResponse.response(), please check the schema, you are describing it

test/trento_web/controllers/v1/host_controller_test.exs
```diff
@@ -48,4 +48,54 @@ defmodule TrentoWeb.V1.HostControllerTest do
...
+        Trento.Commanded.Mock,
+        :dispatch,
+        fn %Trento.Domain.Commands.RequestHostDeregistration{host_id: ^host_id} ->
+          {:error, :host_alive}
+        end
+      )
+
+      conn
+      |> delete("/api/v1/hosts/#{host_id}")
+      |> response(422)
```
CDimonaco, line 82: Same as the controller comment, we should use the correct schema for 422 as described in open api and we should check here with openapi spex test helper that the schema is respected

test/trento_web/controllers/v1/host_controller_test.exs
```diff
@@ -48,4 +48,54 @@ defmodule TrentoWeb.V1.HostControllerTest do
...
+        Trento.Commanded.Mock,
+        :dispatch,
+        fn %Trento.Domain.Commands.RequestHostDeregistration{host_id: ^host_id} ->
+          {:error, :host_not_registered}
+        end
+      )
+
+      conn
+      |> delete("/api/v1/hosts/#{host_id}")
+      |> response(:not_found)
```
CDimonaco, line 98: Not found is a json schema object, described by our openapi spec for that controller, please use the appropriate test helper from the openapi library

CDimonaco (changes requested):
Guys just some comments,

- Rename the association to hearbeat, updating all the preloading etc..
- When we deal with controllers and controller tests, use the openapi spex test helper when needed, the inconsistency between what the controller returns and the schema could be spotted easily
- Rephrase the tests, we are not dealing explicitely with health, so clarify it or change the phrasing

config/config.exs
jamie-suse: This was previously discussed to be set to 0, but as a configurable value.
We do still need it for this PR, it's used in the [middleware](https://github.com/trento-project/web/pull/1450/files/06491f0da5b7c449b717342efa62f4d3ec0eb43c#diff-5513fdb4d4dc67a57b45ef3b29c62825cd3c9a11f3ca22c5485067c9d0945e9dR13) in order to validate the timing on the backend.

arbulu89: @CDimonaco It is actually used. With the value set as zero. This is not a frontend operation, it simply makes the backend more robust against the usage of this command. That the frontend could use this value for its own work doesn't make it a no-op.

config/test.exs
```diff
@@ -41,6 +41,13 @@ config :logger, level: :warn
 # Initialize plugs at runtime for faster test compilation
 config :phoenix, :plug_init_mode, :runtime
 
+# Agent heartbeat interval. Adding one extra second to the agent 5s interval to avoid glitches
+config :trento, Trento.Heartbeats, interval: :timer.seconds(6)
```
arbulu89, line 45: This has the same exact value that in `config.exs`. What's the purpose?

lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -28,6 +33,31 @@ defmodule TrentoWeb.V1.HostController do
...
+    ],
+    responses: [
+      no_content: "The host has been deregistered",
+      not_found: TrentoWeb.OpenApi.Schema.NotFound.response(),
+      unprocessable_entity: OpenApiSpex.JsonErrorResponse.response()
+    ]
+
+  @spec delete(Plug.Conn.t(), map) :: Plug.Conn.t()
+  def delete(conn, %{id: host_id}) do
+    case Hosts.deregister_host(host_id) do
```
arbulu89, line 54: I have some doubts if this `case` wouldn't allow to use the fallback controller for the 404 scenario.
Does this send the expected `Not found` schema? https://github.com/trento-project/web/blob/2.0.0/lib/trento_web/views/error_view.ex#L26

If not, we would need to have something like:
```
with :ok Hosts.deregister_host(host_id) do
    send_resp(conn, 204, "")
else
    {:error, :host_alive} -> send_resp(conn, 422, "")
end
```
And the fallback controller would take care of the not found

test/trento/application/integration/discovery/protocol/enrich_request_host_deregistration_test.exs
```diff
@@ -0,0 +1,52 @@
...
+
+      assert {:ok, %RequestHostDeregistration{host_id: id, requested_at: now}} ==
+               Enrichable.enrich(command, %{})
+    end
+
+    test "should return an error if deregistration request is within debounce period" do
+      now = DateTime.utc_now()
+
+      %{id: id} = insert(:host)
+      insert(:heartbeat, agent_id: id, timestamp: DateTime.add(DateTime.utc_now(), -6, :second))
```
arbulu89, line 38: Is this value related to the values you put in the `test.exs` file.
Maybe you could fetch the value from the config file, so if you change it, the test doesn't break

arbulu89 (commented):
It looks good in general!
Once @CDimonaco comments are addresses feel free to merge from my side

config/test.exs
jamie-suse: To decouple the test environment from the production environment. If this didn't exist in `config/test.exs` and someone were to update the value in `config/config.exs`, then the test might start failing.

arbulu89: Okey, defensive measure I guess

lib/trento_web/controllers/v1/host_controller.ex
jamie-suse: Just ran this manually and get the JSON schema:
```
curl -v -X DELETE http://localhost:4000/api/v1/hosts/240f96b1-8d26-53b7-9e99-ffb0f2e735b6
*   Trying 127.0.0.1:4000...
* Connected to localhost (127.0.0.1) port 4000 (#0)
> DELETE /api/v1/hosts/240f96b1-8d26-53b7-9e99-ffb0f2e735b6 HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/7.82.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< cache-control: max-age=0, private, must-revalidate
< content-length: 85
< content-type: application/json; charset=utf-8
< date: Tue, 06 Jun 2023 07:10:29 GMT
< server: Cowboy
< x-request-id: F2XVe8nL9JnA_gsAAMgG
< 
* Connection #0 to host localhost left intact
{"errors":[{"detail":"The requested resource cannot be found.","title":"Not Found"}]}
```

arbulu89 (approved):
Ready to merge once you address @CDimonaco comments

lib/trento_web/controllers/v1/host_controller.ex
arbulu89: It breaks a bit how other controllers functions are done (using the `with`), as you explicitily return those errors, but well, it works hehe

CDimonaco (approved):

PR #1451, opened 2023-05-24
Display check failure message

nelsonkopliku:
# Description

This PR adds rendering of a check expectation's `failure_message` returned by wanda on a failing check.

Took the chance to tidy up a bit factories. More might come as we go.

@jagabomb I went for option 1 in figma just out of personal taste, however let me know if option two should be the way.

image

## How was this tested?

Automated tests/storybook

### DIFF
--- a/assets/js/lib/test-utils/factories/executions.js
+++ b/assets/js/lib/test-utils/factories/executions.js
@@ -16,14 +16,29 @@ export const executionValueFactory = Factory.define(({ sequence }) => ({
 
 export const executionExpectationEvaluationFactory = Factory.define(
   ({ sequence, params }) => {
-    const name = params.name || `expectation_${sequence}`;
+    const {
+      name = `expectation_${sequence}`,
+      failure_message = params.failure_message
+        ? { failure_message: params.failure_message }
+        : {},
+    } = params;
 
     return {
       name,
       return_value: faker.datatype.number(),
       type: expectationReturnTypeEnum(),
+      ...failure_message,
     };
   }
+);
+
+export const failingExpectEvaluationFactory = Factory.define(({ params }) =>
+  executionExpectationEvaluationFactory.build({
+    ...params,
+    return_value: false,
+    type: 'expect',
+    failure_message: faker.lorem.sentence(),
+  })
 );
 
 export const executionExpectationEvaluationErrorFactory = Factory.define(
@@ -40,14 +55,29 @@ export const executionExpectationEvaluationErrorFactory = Factory.define(
 
 export const expectationResultFactory = Factory.define(
   ({ sequence, params }) => {
-    const name = params.name || `expectation_${sequence}`;
+    const {
+      name = `expectation_${sequence}`,
+      failure_message = params.failure_message
+        ? { failure_message: params.failure_message }
+        : {},
+    } = params;
 
     return {
       name,
       result: faker.datatype.boolean(),
       type: expectationReturnTypeEnum(),
+      ...failure_message,
     };
   }
+);
+
+export const failingExpectationResultFactory = Factory.define(({ params }) =>
+  expectationResultFactory.build({
+    ...params,
+    result: false,
+    type: expectationReturnTypeEnum(),
+    failure_message: faker.lorem.sentence(),
+  })
 );
 
 export const executionFactFactory = Factory.define(({ sequence }) => ({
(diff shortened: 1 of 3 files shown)


abravosuse (approved):
Code apart, that looks awesome to me.

jagabomb:
> @jagabomb I went for option 1 in figma just out of personal taste, however let me know if option two should be the way.

Thanks for this @nelsonkopliku, Option 1 is a good option to proceed with. The only question I have is, is it possible to emphasis certain values in the filing message, like the works **Failing** and values (**3** and **1**)?

nelsonkopliku:
> is it possible to emphasis certain values in the failing message, like the words **Failing** and values (**3** and **1**)?

@jagabomb I was waiting for this :smile: 
So, we could make `Failing` bold/semibold without much hassle, however consider that in the same result there might also be `Passing` evaluations. What about those?

In regards to highlighting specific parts of the failure message I am afraid that would be trickier.
It would mean making `failure_message` field exposed by wanda a markdown possibly, which would bring in the problem of interpolation inside a markdown and refactoring all the checks.

Honestly, not an option for me at the moment :sweat_smile:

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -17,15 +17,30 @@ export const executionValueFactory = Factory.define(({ sequence }) => ({
 export const executionExpectationEvaluationFactory = Factory.define(
   ({ sequence, params }) => {
     const name = params.name || `expectation_${sequence}`;
+    const type = params.type || expectationReturnTypeEnum();
```
arbulu89, line 20: I don't think you need these changes (except the failure message one).
The returned object values take what it comes from `params` automatically, even if you don't define them here, using the default value you define there.

I mean, leaving:
```
return {
  name,
  return_value: faker.datatype.number(),
  type: expectationReturnTypeEnum()
...
```
Achieves the same without adding the top lines with the param.

And if you want to have to coming from params, you can do something like this:
```
{
   name = `expectation_${sequence}`,
   failure_message = params.failure_message ? { failure_message: params.failure_message } : {}
} = params;
```

https://github.com/thoughtbot/fishery#use-params-to-access-passed-in-properties

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -17,15 +17,30 @@ export const executionValueFactory = Factory.define(({ sequence }) => ({
...
     };
   }
 );
 
+export const failingExpectEvaluationFactory = Factory.define(({ params }) =>
+  executionExpectationEvaluationFactory.build({
+    ...params,
+    return_value: false,
+    type: 'expect',
+    failure_message: params.failure_message || faker.lorem.sentence(),
```
arbulu89, line 40: I think no need for this:
```
failure_message: faker.lorem.sentence(),
```
would be enough and it would take the `faker` value only as default.
I think `fishery` is much more smarter than what we think hehe

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -41,15 +56,30 @@ export const executionExpectationEvaluationErrorFactory = Factory.define(
 export const expectationResultFactory = Factory.define(
   ({ sequence, params }) => {
     const name = params.name || `expectation_${sequence}`;
+    const type = params.type || expectationReturnTypeEnum();
```
arbulu89, line 59: The same for above

arbulu89 (approved):
@nelsonkopliku Good!
I just proposed some changes in the factory itself.
We don't need to do such a default values handling, `fishery` is pretty smart with those things.
Don't need a new review anyway

assets/js/lib/test-utils/factories/executions.js
nelsonkopliku: Thanks for the reminder :smile:

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -40,16 +55,31 @@ export const executionExpectationEvaluationErrorFactory = Factory.define(
...
+      ...failure_message,
     };
   }
 );
 
+export const failingExpectationResultFactory = Factory.define(({ params }) =>
+  expectationResultFactory.build({
+    ...params,
+    result: false,
+    type: params.type || expectationReturnTypeEnum(),
```
arbulu89, line 78: This still unchanged :see_no_evil:

nelsonkopliku: whoops :sweat_smile:

arbulu89 (approved):
Great!

jagabomb (approved):
Approved as per comments.

EMaksy (approved):
Hey man nice!
LGTM

PR #1461, opened 2023-05-26
Basic deregistration leftovers

CDimonaco:
# Description

Basic deregistration leftovers, we need to address this issues

- Ensure that the application and database are updated with fields and not overwritten with new fields and old fields nil.
- Ensure that the database/application deregistered events, are emitted once.

   Currently the database deregistered event is emitted more than once

    When the primary instance is deregistered
    After the last database instance is deregistered

- Fix faulty tests in the sap system aggregate
- Ensure no sid at nil is set on database when deregistered
- Add deregistered_at timestamp when the database is deregistered
- Remove sid to nil in sap system aggregate when the sap system is deregistered

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/domain/sap_system/database.ex
+++ b/lib/trento/domain/sap_system/database.ex
@@ -14,7 +14,7 @@ defmodule Trento.Domain.SapSystem.Database do
   deftype do
     field :sid, :string
     embeds_many :instances, Instance
-    field :deregistered_at, :utc_datetime_usec, default: nil
+    field :deregistered_at, :utc_datetime_usec
     field :health, Ecto.Enum, values: Health.values()
   end
 end
--- a/lib/trento/domain/sap_system/database.ex
+++ b/lib/trento/domain/sap_system/database.ex
@@ -14,6 +14,7 @@ defmodule Trento.Domain.SapSystem.Database do
   deftype do
     field :sid, :string
     embeds_many :instances, Instance
+    field :deregistered_at, :utc_datetime_usec, default: nil
     field :health, Ecto.Enum, values: Health.values()
   end
 end
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -814,12 +814,10 @@ defmodule Trento.Domain.SapSystem do
   defp maybe_emit_sap_system_deregistered_event(
          %SapSystem{
            sap_system_id: sap_system_id,
-           database: %Database{
-             sid: nil
-           }
+           database: %Database{deregistered_at: database_deregistered_at}
          },
          deregistered_at
-       ) do
+       ) when not is_nil(database_deregistered_at) do
     %SapSystemDeregistered{sap_system_id: sap_system_id, deregistered_at: deregistered_at}
   end
 
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -853,7 +853,8 @@ defmodule Trento.Domain.SapSystem do
          %SapSystem{
            sap_system_id: sap_system_id,
            database: %Database{
-             instances: instances
+             instances: instances,
+             deregistered_at: nil
            }
          },
          deregistered_at
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -499,7 +499,7 @@ defmodule Trento.Domain.SapSystem do
   end
 
   def apply(
-        %SapSystem{database: %Database{instances: instances}} = sap_system,
+        %SapSystem{database: %Database{instances: instances} = database} = sap_system,
         %DatabaseInstanceDeregistered{
           instance_number: instance_number,
           host_id: host_id
@@ -517,13 +517,14 @@ defmodule Trento.Domain.SapSystem do
     %SapSystem{
       sap_system
       | database: %Database{
-          instances: instances
+          database
+          | instances: instances
         }
     }
   end
 
   def apply(
-        %SapSystem{application: %Application{instances: instances}} = sap_system,
+        %SapSystem{application: %Application{instances: instances} = application} = sap_system,
         %ApplicationInstanceDeregistered{instance_number: instance_number, host_id: host_id}
       ) do
     instances =
@@ -538,7 +539,8 @@ defmodule Trento.Domain.SapSystem do
     %SapSystem{
       sap_system
       | application: %Application{
-          instances: instances
+          application
+          | instances: instances
         }
     }
   end
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -817,7 +817,8 @@ defmodule Trento.Domain.SapSystem do
            database: %Database{deregistered_at: database_deregistered_at}
          },
          deregistered_at
-       ) when not is_nil(database_deregistered_at) do
+       )
+       when not is_nil(database_deregistered_at) do
     %SapSystemDeregistered{sap_system_id: sap_system_id, deregistered_at: deregistered_at}
   end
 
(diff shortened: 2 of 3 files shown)


lib/trento/domain/sap_system/database.ex
```diff
@@ -14,6 +14,7 @@ defmodule Trento.Domain.SapSystem.Database do
   deftype do
     field :sid, :string
     embeds_many :instances, Instance
+    field :deregistered_at, :utc_datetime_usec, default: nil
```
arbulu89, line 17: I know that this is coming from other parts of the code, but the `default` option strucked me out.
I would say all fields are `nil` by default. Just for curiosity I wonder if having this default again does something hehe

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -840,6 +840,7 @@ defmodule Trento.Domain.SapSystem do
          %SapSystem{
            sap_system_id: sap_system_id,
            database: %Database{
+             deregistered_at: nil,
```
arbulu89, line 843: I think we can have some unsupported scenario with this clause. With this conditions, we only check for `deregistered_at` when the list of `instances` is empty. 
What if we have:
- Scale-out deployment with 2 primary and 2 secondary
- Already deregistered database (all primaries were deregistered)
- 1 of the secondaries is deregistered now
At this point the function would go for the next function, emitting the `DatabaseDeregistered` event again, when it shouldn't.

I think we should do the same with do above in `maybe_emit_sap_system_deregistered_event` and `when not is_nil(database_deregistered_at) do`.

Or you add this same clause in the next `maybe_emit_database_deregistered_event` function. I would bet for the 1st option, so we don't need this `when` twice

What do you think?

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1344,12 +1344,18 @@ defmodule Trento.SapSystemTest do
...
+
+      primary_database_host_id = UUID.uuid4()
       secondary_database_host_id = UUID.uuid4()
+
       deregistered_at = DateTime.utc_now()
+
       db_instance_number_1 = "00"
       db_instance_number_2 = "01"
 
+      db_sid = UUID.uuid4()
```
arbulu89, line 1356: Not asking any change, but using uuid for sids has been always wrong...
The sid is a 3 charts/numbers string...
Not that it matters for tests, but it always raises my ocd detector

arbulu89 (changes requested):
Hi @CDimonaco 
I think we still have at least one scenario where we would raise the database deregistered event incorrectly.

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: I will write a test for it so we track this scenario, thanks for pointing me out this thing!

I would go for the first scenario, fixing it asap :D

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: You're right, my laziness did this, I will write a custom factory for sid so we can have a dedicated tech debt, my goal as you can see was to have the tests close as possibile to the real scenario, I tacked all sids, ids etc, but I agree we need to tackle this right know. Thanks for the suggestion man!

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: Add a test for this scenario, and added the clause in the guard, we not emit the database deregistered event if it's already registered, IN ANY CASE, with or without instances.

The test passes, and event the new test for this scenario pass (was ofcourse not passing before)

lib/trento/domain/sap_system/database.ex
CDimonaco: It's redundant, I removed it :D

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: I added a custom function to the sap system testing suite for faking sids and add only to the deregistration test, if we like this approach we can merge it, and I'll have a tech deb to wire properly in all the factories

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -813,12 +814,11 @@ defmodule Trento.Domain.SapSystem do
   defp maybe_emit_sap_system_deregistered_event(
          %SapSystem{
            sap_system_id: sap_system_id,
-           database: %Database{
-             sid: nil
-           }
+           database: %Database{deregistered_at: database_deregistered_at}
```
arbulu89, line 817: I'm missing something here. We should check both, database `deregistered_at` and sap system `deregistered_at` . Something like:
```
when not is_nil(database_deregistered_at) or not is_nil(sapsystem_deregistered_at)
```
Or add this condition to the next `maybe_emit_sap_system_deregistered_event` clause. I'm not sure.

Otherwise we can have `SapSystemDeregistered` emitted twice, one when the database is deregistered, and 2nd when the applications are deregistered.

At the end, we need to check if the SAP system was deregistered somewhere to avoid emitting this event.

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1344,12 +1344,18 @@ defmodule Trento.SapSystemTest do
...
+      primary_database_host_id = UUID.uuid4()
       secondary_database_host_id = UUID.uuid4()
+
       deregistered_at = DateTime.utc_now()
+
       db_instance_number_1 = "00"
       db_instance_number_2 = "01"
 
+      db_sid = fake_sid()
+      application_sid = UUID.uuid4()
```
arbulu89, line 1357: `application_sid` should use `fake_sid` as well

arbulu89 (changes requested):
@CDimonaco I found out other scenario where we could emit `SapSystemDeregistered` incorrectly. Could you have a look?
Otherwise all good!

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: Nice catch, adding a test and shipping the fix asap

CDimonaco: Done!

CDimonaco: The sap system is deregistered only if it's not already deregistered, following the database deregistered emit strategy, I just added the deregistered_at nil pattern match in function signatures and a black hole signature for capturing already deregistered sap systems with a no-op

arbulu89 (approved):
Thanks mate!
Huge effort!

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -1547,58 +1577,251 @@ defmodule Trento.SapSystemTest do
...
+                     deregistered_at: nil,
+                     instances: [%Instance{instance_number: ^db_instance_number_2}]
                    }
                  } = sap_system
         end
       )
     end
 
-    test "should deregister an ENQREP Application Instance" do
+    test "should deregister the primary instance, the entire database and then the secondary instance without deregistering the database a second time, SAP system not registered" do
```
fabriziosestito, line 1638: Are we sure we need to keep the "a second-time" wording? I think we will  soon forget about the mistake we made but this will stay in the codebase

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -2052,10 +2332,118 @@ defmodule Trento.SapSystemTest do
...
+            instance_number: message_server_instance_number,
+            deregistered_at: deregistered_at
+          }
+        ]
+      )
+    end
   end
+
+  defp fake_sid,
+    do: Enum.join([Faker.Util.letter(), Faker.Util.letter(), Faker.Util.letter()])
```
fabriziosestito, line 2448: what about moving this to the factory?

fabriziosestito (approved):
Some nitpicks but overall great work!

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: You're right, I will rephrase it

PR #1466, opened 2023-05-29
ASCS/ERS cluster details component

arbulu89:
# Description

Add ASCS/ERS details component.
By now, it excludes the nodes table. I will create other PR with it to make it more manageable to review.

Find story book in the CI executions link below (`Storybook publish`). In `AscsErsClusterDetails` section

## How was this tested?

Storybook stories and tests

### DIFF
--- a/assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
+++ b/assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
@@ -38,7 +38,7 @@ function AscsErsClusterDetails({
         </div>
       </div>
       <div className="flex xl:flex-row flex-col">
-        <div className="mt-4 bg-white shadow rounded-lg py-8 px-8 xl:w-1/3 mr-4">
+        <div className="mt-4 bg-white shadow rounded-lg py-8 px-8 xl:w-2/5 mr-4">
           <ListView
             className="grid-rows-2"
             titleClassName="text-lg"
@@ -65,7 +65,7 @@ function AscsErsClusterDetails({
             ]}
           />
         </div>
-        <div className="mt-4 bg-white shadow rounded-lg py-8 px-8 xl:w-1/3 mr-4">
+        <div className="mt-4 bg-white shadow rounded-lg py-8 px-8 xl:w-2/5 mr-4">
           <ListView
             className="grid-rows-2 mb-10"
             titleClassName="text-lg"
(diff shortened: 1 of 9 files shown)


assets/js/components/DottedPagination/Arrow.jsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 
-function Arrow({ children, onClick }) {
+function Arrow({ children, onClick, datatestID }) {
   return (
     <div
+      data-testid={datatestID}
```
arbulu89, line 6: Added this to test properly the change of SAP system in the details view. 
Otherwise, the tests looked super dependant on number of icons, styles, etc

dottorblaster: You did good, totally plus one on this :+1:

assets/js/components/DottedPagination/Arrow.jsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 
-function Arrow({ children, onClick }) {
+function Arrow({ children, onClick, datatestID }) {
```
dottorblaster, line 3: You can use `...rest` here, spreading the `...rest` onto the `div` returned by the element, and pass `data-testid` from the outside without worrying about the casing

assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
```diff
@@ -0,0 +1,120 @@
...
+  clusterName,
+  cibLastWritten,
+  provider,
+  details,
+}) {
+  const [sapSystems, setSapSystems] = useState([]);
+  const [currentSapSystem, setSapSystem] = useState(null);
+
+  useEffect(() => {
+    setSapSystems(details?.sap_systems);
```
dottorblaster, line 25: I think you don't need a state variable for the `sapSystems`

dottorblaster (changes requested):
Just a tiny change then we can merge

assets/js/components/DottedPagination/Arrow.jsx
arbulu89: Nice! I didn't think about it and it looks way better!

assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
arbulu89: You are right!
Backported from other commit that will need (I think so XD). But it will come in a next PR.
Removing by now.

dottorblaster (approved):
LGTM

nelsonkopliku (approved):
Well done!

PR #1469, opened 2023-05-29
Improve tooltip styling

nelsonkopliku:
# Description

Fixes a tiny inconsistency of the tooltip suggesting to select some checks before being able to trigger an execution.

Before
image

After
image

I removed the gradient bg colouring out of personal taste.

If we want to consider using a better Tooltip component and not a custom one let's track it and consider available options:
- https://popper.js.org/react-popper/
- https://github.com/atomiks/tippyjs-react
- https://github.com/ReactTooltip/react-tooltip
- others?

## How was this tested?
No need for test change

### DIFF
--- a/assets/js/components/ClusterDetails/HanaClusterDetails.jsx
+++ b/assets/js/components/ClusterDetails/HanaClusterDetails.jsx
@@ -121,7 +121,10 @@ function HanaClusterDetails({
               />{' '}
               <span>Start Execution</span>
               {!hasSelectedChecks && (
-                <Tooltip tooltipText="Select some Checks first!" />
+                <Tooltip
+                  tooltipText="Select some Checks first!"
+                  width="-translate-x-1/4"
+                />
               )}
             </TriggerChecksExecutionRequest>
           </div>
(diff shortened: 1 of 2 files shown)


arbulu89 (commented):
+1 for the style change on the gradient.
About the width thing. Let's be careful, the tooltips are used in other places as well.
Did you check if removing the default affects them?
In any case, the fix being the width smells a bit to me, but i'm an ignorant on this css/styling world hehe

jagabomb (approved):
Thanks @nelsonkopliku, removing the gradient for me great because now it is consistent with the design proposal (screenshot attached below). I am aware of some of the limitations of the current tooltip component and I would be open to another implementation if it gives us more flexibility. For example, I know we currently can't position the tooltip at the top of a component. It would be nice to be able to change the position of the tooltip.

Screenshot 2023-05-30 at 10 12 49

EMaksy:
Nice job,  like @arbulu89  mentioned, the Tooltip is also used in HostsList component, but it seems to look good :D

The current tooltip component could be improved but at the moment we just use it in two  places [HostsList, HanaClusterDetails] for this it works, when we add more tool-tips to the application then we should head to use a custom component, or what do you think ? :D

dottorblaster (approved):

PR #1472, opened 2023-05-30
Add cluster nodes details to ASCS/ERS cluster details view

arbulu89:
# Description

Last episode in `ASCS/ERS` frontend series!
Include the nodes table in the details view.

Have a look in the published storybook to see the changes. `AscsErsClusterDetails` section.
New story for failover added, to see how multiple role, virtual ip, filesystem look in the nodes table.

## How was this tested?

Tests and storybook updated

### DIFF
--- a/assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
+++ b/assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
@@ -2,27 +2,78 @@ import React, { useState, useEffect } from 'react';
 
 import PageHeader from '@components/PageHeader';
 import BackButton from '@components/BackButton';
-
+import Table from '@components/Table';
 import ListView from '@components/ListView';
 import ProviderLabel from '@components/ProviderLabel';
 import DottedPagination from '@components/DottedPagination';
+import HostLink from '@components/HostLink';
 
 import ChecksComingSoon from '@static/checks-coming-soon.svg';
 
 import SBDDetails from './SBDDetails';
+import SiteDetails from './SiteDetails';
 import StoppedResources from './StoppedResources';
+
+const nodeDetailsConfig = {
+  usePadding: false,
+  columns: [
+    {
+      title: 'Hostname',
+      key: '',
+      render: (_, { id, name }) => <HostLink hostId={id}>{name}</HostLink>,
+    },
+    {
+      title: 'Role',
+      key: 'roles',
+      render: (content) =>
+        content?.map((role) => role.toUpperCase()).join(', '),
+    },
+    {
+      title: 'Virtual IP',
+      key: 'virtual_ips',
+      className: 'table-col-m',
+      render: (content) => content?.join(', '),
+    },
+    {
+      title: 'Filesystem',
+      key: 'filesystems',
+      className: 'table-col-m',
+      render: (content) => content?.join(', '),
+    },
+    {
+      title: '',
+      key: '',
+      className: 'table-col-xs',
+      render: (_, item) => {
+        const { attributes, resources } = item;
+        return <SiteDetails attributes={attributes} resources={resources} />;
+      },
+    },
+  ],
+};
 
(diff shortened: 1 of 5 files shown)


jagabomb:
@arbulu89 Sounds good to me. I am happy to tackle what we can now and then do the other changes in another PR. I was just looking at the layout as a whole so I understand how some of the requests are outside of this PR. 😄

assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
```diff
@@ -2,27 +2,78 @@ import React, { useState, useEffect } from 'react';
...
+
 function AscsErsClusterDetails({
   clusterName,
   cibLastWritten,
   provider,
+  hosts,
   details,
 }) {
+  const [sapSystems, setSapSystems] = useState([]);
   const [currentSapSystem, setSapSystem] = useState(null);
```
nelsonkopliku, line 63: I know this was already here, however what do you think of streamlining to `setCurrentSapSystem`?
`setSapSystems`/`setSapSystem` look too similar and might cause some headaches.

assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
```diff
@@ -2,27 +2,78 @@ import React, { useState, useEffect } from 'react';
...
+    const enrichedSapSystems = details?.sap_systems.map((sapSystem) => ({
+      ...sapSystem,
+      nodes: sapSystem?.nodes.map((node) => ({
+        ...node,
+        ...hosts.find(({ hostname }) => hostname === node.name),
+      })),
+    }));
+
+    setSapSystems(enrichedSapSystems);
+    setSapSystem(enrichedSapSystems[0]);
```
nelsonkopliku, line 75: Do we always have at least one element in `enrichedSapSystems`? 
Or it can be an empty list? If so, we should consider the case.

I didn't try this myself, though.

nelsonkopliku (commented):
Nice! Only real doubt is about the assumption that there is always at least on element in `enrichedSapSystems`  in `AscsErsClusterDetails`.

assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
arbulu89: Yes, we should always have one sap system. This is a hard requirement in the backend. 
Otherwise, the cluster is not identified as `ASCS/ERS` cluster. It would be an unknown cluster, which shouldn't have this page, and we should go to the `Unknown cluster type` view.

Anyway, in a pure way, if the list is empty, we simply have the same view with many empty fields. Nothing would break

assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
```diff
@@ -2,27 +2,78 @@ import React, { useState, useEffect } from 'react';
...
 }) {
+  const [sapSystems, setSapSystems] = useState([]);
   const [currentSapSystem, setSapSystem] = useState(null);
 
   useEffect(() => {
-    setSapSystem(details?.sap_systems[0]);
-  }, [details]);
+    const enrichedSapSystems = details?.sap_systems.map((sapSystem) => ({
+      ...sapSystem,
+      nodes: sapSystem?.nodes.map((node) => ({
```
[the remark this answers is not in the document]
arbulu89: Let's see what I can do on this

jagabomb (approved):
Approved as per discussion to postpone UI requests.

dottorblaster (approved):
I think you still don't need a state for the sapSystems and the enrichNodes function could have its own place, but I don't mind having room for future optimization. LGTM

arbulu89:
@dottorblaster 
> I think you still don't need a state for the sapSystems 

You were right again. I just tested it outside of the `useEffect` and it does not make any difference.
I will remove it and merge afterwards

PR #1473, opened 2023-05-30
Deregistration process manager retry the rollup errors from aggregates

CDimonaco:
# Description

Deregistration process manager needs to retry the `rolling up` errors from aggregates.

The other errors will be captured by the default implementation, so the process manager process will stop.

### DIFF
--- a/lib/trento/application/process_managers/deregistration_process_manager.ex
+++ b/lib/trento/application/process_managers/deregistration_process_manager.ex
@@ -289,6 +289,17 @@ defmodule Trento.DeregistrationProcessManager do
     }
   end
 
+  # Retry the rollup errors, stop the process on other errors
+
+  def error({:error, :host_rolling_up}, _command_or_event, %{context: context}),
+    do: {:retry, context}
+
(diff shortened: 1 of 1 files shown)


fabriziosestito (approved):

CDimonaco:
> > Nice @CDimonaco
> > Checking this reminded me that we might want to change the `stream_roll_up_event_handler.ex` and include the "deregistration` events there as well
> 
> @arbulu89 I don't think deregistration should cause a rollup, I think those events should be skipped. A roll-up will be triggered by the tombstone event if any.

Yep, this is in the other pr #1425, currently is in draft because is blocked by #1461, but you can see the logic done for clusters and hosts, only sap system are blocked

arbulu89:
> > Nice @CDimonaco
> > Checking this reminded me that we might want to change the `stream_roll_up_event_handler.ex` and include the "deregistration` events there as well
> 
> @arbulu89 I don't think deregistration should cause a rollup, I think those events should be skipped. A roll-up will be triggered by the tombstone event if any.

Actually, i was thinking more if getting those events could create some pattern matching error, as we don't have them in the list. But the events handlers ignore not matched `handle`s, so we are fine

PR #1474, opened 2023-05-30
Fix host removed from cluster concurrency issues

fabriziosestito:
This PR fixes possible concurrency issues when a host is removed from a cluster (delta-deregistration flow).
The issue is caused by two aggregates acting concurrently, so the `HostAddedToCluster` event and `HostRemovedFromCluster` event ordering is not guaranteed.

The solution is to change the projector logic to skip the event if the host is not part of the cluster anymore.

We still need to conditionally broadcast to  the FE the event, but this will come once the FE PR is merged.


arbulu89 (approved):
kudos!

CDimonaco:
lgtm

PR #1480, opened 2023-06-01
Send cluster type to wanda

arbulu89:
# Description
Send the `cluster_type` to wanda in the `env` map.
It can be used as `env.cluster_type` in the checks.

I have made the checks module `request_execution` function generic, so it already receives the env map, so we can have different envs for different executions (for example, host executions might not need the cluster_type value).

This should already work as it is in wanda, but I will send a change to be able to map the env values with the correct map.
(even though we send strings by now)

@abravosuse Where should we document the available options for this env value? (`hana_scale_up`, `hana_scale_out`, `ascs_ers` and `unknonw` are the current values). I found out that we don't have any documentation for the available providers neither.

## How was this tested?

Tested

### DIFF
--- a/lib/trento/application/integration/checks/checks.ex
+++ b/lib/trento/application/integration/checks/checks.ex
@@ -10,9 +10,11 @@ defmodule Trento.Integration.Checks do
     Target
   }
 
+  alias Trento.Integration.Checks.ClusterExecutionEnv
+
   require Logger
 
-  @spec request_execution(String.t(), String.t(), map, [map], [String.t()]) ::
+  @spec request_execution(String.t(), String.t(), ClusterExecutionEnv.t(), [map], [String.t()]) ::
           :ok | {:error, :any}
   def request_execution(execution_id, cluster_id, env, hosts, selected_checks) do
     execution_requested =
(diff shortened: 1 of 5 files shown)


nelsonkopliku (approved):
LGTM

lib/trento/application/integration/checks/checks.ex
```diff
@@ -13,9 +12,9 @@ defmodule Trento.Integration.Checks do
 
   require Logger
 
-  @spec request_execution(String.t(), String.t(), Provider.t(), [map], [String.t()]) ::
+  @spec request_execution(String.t(), String.t(), map, [map], [String.t()]) ::
```
dottorblaster, line 15: The only thing, and I'm really thinking out loud here, is if we should maybe treat this as a union type and encode the `env` into a struct that can be `ClusterChecksExecutionEnv | HostChecksExecutionEnv`

Too much? Too verbose? Just shooting the idea

dottorblaster (approved):
LGTM with a tiny nitpick

lib/trento/application/integration/checks/checks.ex
arbulu89: Alright, it makes sense, even if its almost only for documentation, as this data is not coming from outside, and we most probably don't need such validation

arbulu89: @dottorblaster I did the change. With this approach, the execution is better defined but more rigid at the same time. The end result is the same, and I don't expect the need to have flexible envs (now it is better documented), so this approach is fine for me.
What do you think?

PR #1486, opened 2023-06-05
Handle evaluation errors

nelsonkopliku:
# Description

This PR fixes a bug that doesn't correctly allow for rendering of evaluation errors (ie. rhai script fails)

Before
image

After
image

Side effect: check [373DB8](https://github.com/trento-project/wanda/blob/main/priv/catalog/373DB8.yaml#L69) needs to be fixed.

## How was this tested?

Tests and stories updated.

### DIFF
--- a/assets/js/components/ExecutionResults/checksUtils.js
+++ b/assets/js/components/ExecutionResults/checksUtils.js
@@ -90,10 +90,21 @@ export const getExpectStatements = (expectationList) =>
 export const getExpectSameStatements = (expectationList) =>
   expectationList.filter(isExpectSame);
 
-export const getExpectSameStatementResult = (expectationResults, name) => {
-  const expectSameStatement = getExpectSameStatements(expectationResults).find(
-    ({ name: resultExpectationName }) => name === resultExpectationName
+const expectationByName =
+  (expectationName) =>
+  ({ name }) =>
+    name === expectationName;
+
+export const getExpectStatementsResults = (
+  expectations,
+  expectationEvaluations
+) =>
+  getExpectStatements(expectations).map(
+    ({ name }) => expectationEvaluations.find(expectationByName(name)) || {}
   );
+
+export const getExpectSameStatementResult = (expectationResults, name) => {
+  const expectSameStatement = expectationResults.find(expectationByName(name));
 
   if (!expectSameStatement) {
     return { name, result: null };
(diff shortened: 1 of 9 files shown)


abravosuse (commented):
Based on the screenshots, it looks good to me. Is this something that I could review on the Storybook and, if so, where in the Storybook? I'm just trying to familiarize myself with the tool.

Also, you mention that as a side effect check 373DB8 needs to be fixed. Are you engaging the Checks Team for this or is it something that you will address yourself? One way or another we will need, I guess, a ticket and I'm trying to figure out which is the right board for it (ours or the   Checks Team's).

assets/js/components/ExecutionResults/checksUtils.js
```diff
@@ -90,11 +90,22 @@ export const getExpectStatements = (expectationList) =>
 export const getExpectSameStatements = (expectationList) =>
   expectationList.filter(isExpectSame);
 
-export const getExpectSameStatementResult = (expectationResults, name) => {
-  const expectSameStatement = getExpectSameStatements(expectationResults).find(
-    ({ name: resultExpectationName }) => name === resultExpectationName
+export const byName =
```
arbulu89, line 93: Is this something we need to export?

arbulu89 (approved):
Great @nelsonkopliku

assets/js/components/ExecutionResults/checksUtils.js
nelsonkopliku: nope, thanks!

arbulu89:
@abravosuse I couple of new stories where added under the `ExpectationResults` story. 
I don't think you will get many value, as the displayed error message is the same as before for the user, and the content is the story is faked data.

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -16,12 +16,11 @@ export const executionValueFactory = Factory.define(({ sequence }) => ({
 
 export const executionExpectationEvaluationFactory = Factory.define(
   ({ sequence, params }) => {
-    const {
-      name = `expectation_${sequence}`,
-      failure_message = params.failure_message
-        ? { failure_message: params.failure_message }
-        : {},
-    } = params;
+    const { name = `expectation_${sequence}` } = params;
```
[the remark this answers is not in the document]
arbulu89: Didn't understand much, but I guess you fought against it hehe
It is fine though!
Thank you for explaining what happens

PR #1487, opened 2023-06-05
Change redirect plug to use a list of available versions

arbulu89:
# Description

Update the current api redirect plug, to be more versitale and accept a list of available api version.

Once this is done, we can move forward and version the openapi schemas

So now, instead of having only the latest version for all the endpoints, each endpoint can be versioned individually.

## How was this tested?

Tests updated

### DIFF
--- a/lib/trento_web/plugs/api_redirector.ex
+++ b/lib/trento_web/plugs/api_redirector.ex
@@ -31,15 +31,15 @@ defmodule TrentoWeb.Plugs.ApiRedirector do
     if Enum.empty?(available_api_versions),
       do: raise(ArgumentError, ":available_api_versions must have 1 element at least")
 
-    opts[:router] || raise ArgumentError, "expected :router option"
+    Keyword.get(opts, :router) || raise ArgumentError, "expected :router option"
 
     opts
   end
 
   @impl true
   def call(%Plug.Conn{path_info: [_ | path_parts], method: method} = conn, opts) do
-    router = opts[:router]
-    available_api_versions = opts[:available_api_versions]
+    router = Keyword.get(opts, :router)
+    available_api_versions = Keyword.get(opts, :available_api_versions)
 
     case find_versioned_path(router, available_api_versions, path_parts, method) do
       nil ->
--- a/lib/trento_web/plugs/api_redirector.ex
+++ b/lib/trento_web/plugs/api_redirector.ex
@@ -23,20 +23,23 @@ defmodule TrentoWeb.Plugs.ApiRedirector do
   import Plug.Conn
 
   @impl true
-  def init(opts), do: opts
+  def init(opts) do
+    available_api_versions =
+      Keyword.get(opts, :available_api_versions) ||
+        raise ArgumentError, "expected :available_api_versions option"
+
+    if Enum.empty?(available_api_versions),
+      do: raise(ArgumentError, ":available_api_versions must have 1 element at least")
+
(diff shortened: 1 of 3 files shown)


fabriziosestito (approved):

lib/trento_web/plugs/api_redirector.ex
```diff
@@ -23,34 +27,43 @@ defmodule TrentoWeb.Plugs.ApiRedirector do
 
   @impl true
   def call(%Plug.Conn{path_info: [_ | path_parts], method: method} = conn, opts) do
-    latest_version = Keyword.get(opts, :latest_version)
+    available_api_versions = Keyword.get(opts, :available_api_versions)
     router = Keyword.get(opts, :router)
 
-    unless latest_version do
-      raise ArgumentError, "expected :latest_version option"
+    unless available_api_versions do
```
CDimonaco, line 33: Maybe we can add the same sanity check even if the list is present but empty, to avoid strange behaviour like getting all 404

Wdyt?

CDimonaco (approved):
I left a comment but it's totally optional, good job

test/trento_web/plugs/api_redirector_test.exs
```diff
@@ -72,12 +74,36 @@ defmodule TrentoWeb.Plugs.ApiRedirectorTest do
...
+        |> ApiRedirector.call(available_api_versions: ["v2", "v1"], router: FoundRouter)
+
+      assert 307 == conn.status
+
+      location_header = get_resp_header(conn, "location")
+
+      assert ["/api/v2/test"] == location_header
+    end
+
+    test "should redirect to the next newest path when the route when it is the first available version",
```
nelsonkopliku, line 91: I am not getting this, maybe a typo.
I guess you meant something like: should redirect to the latest available version of an endpoint :thinking:

arbulu89: yes, typo...

arbulu89: Fixed

lib/trento_web/plugs/api_redirector.ex
arbulu89: @CDimonaco sanity check added.
Besides that, I have moved the argument validation to the `init` function which looks the more appropriate place

nelsonkopliku (approved):

lib/trento_web/plugs/api_redirector.ex
```diff
@@ -23,20 +23,23 @@ defmodule TrentoWeb.Plugs.ApiRedirector do
...
-  def call(%Plug.Conn{path_info: [_ | path_parts], method: method} = conn, opts) do
-    available_api_versions = Keyword.get(opts, :available_api_versions)
-    router = Keyword.get(opts, :router)
+    if Enum.empty?(available_api_versions),
+      do: raise(ArgumentError, ":available_api_versions must have 1 element at least")
 
-    unless available_api_versions do
-      raise ArgumentError, "expected :available_api_versions option"
-    end
+    opts[:router] || raise ArgumentError, "expected :router option"
```
CDimonaco, line 34: Why we don't use `Keyword.get` for this option parameter? Maybe we could align the two methods of retrieving the options

arbulu89: Alright! I will use `Keyword.get` in all of them and merge.
Some of the `Plug` examples I saw don't use it, but let's have our code using same patterns

PR #1488, opened 2023-06-05
Enable openapi schema versioning

arbulu89:
# Description

Enable openapi schemas versioning using different pipelines.

## How was this tested?

Tests updated.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -134,7 +134,7 @@ jobs:
       - name: Build docs
         uses: lee-dohm/generate-elixir-docs@v1
       - name: Generate openapi.json
-        run: mix openapi.spec.json --start-app=false --spec TrentoWeb.OpenApi.ApiSpec
+        run: mix openapi.spec.json --start-app=false --spec TrentoWeb.OpenApi.V1.ApiSpec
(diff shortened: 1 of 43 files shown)


.github/workflows/ci.yaml
```diff
@@ -134,7 +134,7 @@ jobs:
       - name: Build docs
         uses: lee-dohm/generate-elixir-docs@v1
       - name: Generate openapi.json
-        run: mix openapi.spec.json --start-app=false --spec TrentoWeb.OpenApi.ApiSpec
+        run: mix openapi.spec.json --start-app=false --spec TrentoWeb.OpenApi.V1.ApiSpec
```
arbulu89, line 137: I will tackle this issue in a next PR

arbulu89:
> I'm doing the review, but maybe we want to merge it against main instead of deregistration?

I'm pushing against deregistration because the v2 changes belong to that branch.

arbulu89:
> Cool! I am curious to see how this evolves. I guess this would allow to bump whole controllers or even specific actions, is it like that?

Yes, initially I plan to only bump one single endpoint, but we can add all the endpoints that we want to any new version. It is as flexible as we want

PR #1491, opened 2023-06-06
Sap system delta deregistration

fabriziosestito:
The PR adds delta deregistration for SAP systems.
This logic is similar to #1386, except that we need the instances list to calculate the difference.
Also, it fixes the cluster deregistration logic, which was not computing the ID correctly.

### DIFF
--- a/test/trento/application/integration/discovery/policies/sap_system_policy_test.exs
+++ b/test/trento/application/integration/discovery/policies/sap_system_policy_test.exs
@@ -2,11 +2,14 @@ defmodule Trento.Integration.Discovery.SapSystemPolicyTest do
   use ExUnit.Case
   use Trento.DataCase
 
+  import Trento.Factory
+
   import Trento.Integration.DiscoveryFixturesHelper
 
   alias Trento.Integration.Discovery.SapSystemPolicy
 
   alias Trento.Domain.Commands.{
+    DeregisterDatabaseInstance,
     RegisterApplicationInstance,
     RegisterDatabaseInstance
   }
@@ -28,7 +31,7 @@ defmodule Trento.Integration.Discovery.SapSystemPolicyTest do
             ]} =
              "sap_system_discovery_database"
              |> load_discovery_event_fixture()
-             |> SapSystemPolicy.handle()
+             |> SapSystemPolicy.handle([])
   end
 
   test "should return the expected commands when a sap_system payload of type database is handled in the event of a stopped instance" do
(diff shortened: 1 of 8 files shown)


CDimonaco (approved):

test/trento/application/integration/discovery/policies/sap_system_policy_test.exs
```diff
@@ -86,6 +89,58 @@ defmodule Trento.Integration.Discovery.SapSystemPolicyTest do
             ]} =
              "sap_system_discovery_application_diagnostics"
              |> load_discovery_event_fixture()
-             |> SapSystemPolicy.handle()
+             |> SapSystemPolicy.handle([])
+  end
+
+  describe "delta deregistration" do
+    test "should deregister the old instances and register the new ones" do
```
arbulu89, line 96: Could we have some application instances here?
To see if the application deregister command is sent.
At the end, one possible scenario is to have app and database instances in the same host (this is a real case)

test/trento/application/integration/discovery/policies/sap_system_policy_test.exs
```diff
@@ -86,6 +89,58 @@ defmodule Trento.Integration.Discovery.SapSystemPolicyTest do
             ]} =
              "sap_system_discovery_application_diagnostics"
              |> load_discovery_event_fixture()
-             |> SapSystemPolicy.handle()
+             |> SapSystemPolicy.handle([])
+  end
+
+  describe "delta deregistration" do
```
arbulu89, line 95: Other possible test would be to check that no commands are sent if we don't have currently instances and the coming payload is empty of sap instances as well.

arbulu89 (approved):
Thank you @fabriziosestito 
I proposed some additional tests you could add.
Maybe they could be redundant with the current ones, I'm not sure. They are simply some real scenarios that came to my mind

test/trento/application/integration/discovery/policies/sap_system_policy_test.exs
fabriziosestito: Sure!

test/trento/application/integration/discovery/policies/sap_system_policy_test.exs
arbulu89: indeed, it doesn't need to be part of the delta group of tests.
Yes, it would look like the same than in the cluster

arbulu89 (approved):
great! thank you for adding the tests

PR #1492, opened 2023-06-06
Add V2 scope to /api/clusters endpoint

arbulu89:
# Description

Follow up for: https://github.com/trento-project/web/pull/1468

Include the `v2` scope for the `/api/clusters` endpoint.
The `v1` has been updated to be backward compatible.

## How was this tested?

Tested

### DIFF
--- a/lib/trento_web/openapi/v1/schema/cluster.ex
+++ b/lib/trento_web/openapi/v1/schema/cluster.ex
@@ -2,7 +2,6 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
   @moduledoc false
 
   require OpenApiSpex
-  require Trento.Domain.Enums.ClusterType, as: ClusterType
 
   alias OpenApiSpex.Schema
 
@@ -37,13 +36,11 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
         site: %Schema{type: :string},
         hana_status: %Schema{type: :string},
         attributes: %Schema{
-          title: "ClusterNodeAttributes",
-          type: :array,
-          items: %Schema{type: :string}
+          type: :object,
+          description: "Node attributes"
         },
         virtual_ip: %Schema{type: :string},
         resources: %Schema{
-          title: "ClustrNodeResources",
           description: "A list of Cluster resources",
           type: :array,
           items: ClusterResource
@@ -57,7 +54,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
 
     OpenApiSpex.schema(%{
       title: "SbdDevice",
-      description: "Ad Sbd Device",
+      description: "SBD Device",
       type: :object,
       properties: %{
         device: %Schema{type: :string},
@@ -83,22 +80,20 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
         sr_health_state: %Schema{type: :string, description: "SR health state"},
         fencing_type: %Schema{type: :string, description: "Fencing Type"},
         stopped_resources: %Schema{
-          title: "ClusterResource",
           description: "A list of the stopped resources on this HANA Cluster",
           type: :array,
           items: ClusterResource
         },
         nodes: %Schema{
-          title: "HanaClusterNodes",
           type: :array,
           items: HanaClusterNode
         },
         sbd_devices: %Schema{
-          title: "SbdDevice",
           type: :array,
           items: SbdDevice
         }
-      }
+      },
+      required: [:nodes]
     })
   end
 
@@ -106,7 +101,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
     @moduledoc false
 
     OpenApiSpex.schema(%{
-      title: "PacemakerClusterDetail",
+      title: "PacemakerClusterDetails",
       description: "Details of the detected PacemakerCluster",
       nullable: true,
       oneOf: [
@@ -135,7 +130,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
         type: %Schema{
           type: :string,
           description: "Detected type of the cluster",
-          enum: ClusterType.values()
+          enum: [:hana_scale_up, :hana_scale_out, :unknown]
         },
         selected_checks: %Schema{
           title: "SelectedChecks",
--- a/lib/trento_web/openapi/v2/schema/cluster.ex
+++ b/lib/trento_web/openapi/v2/schema/cluster.ex
@@ -1,0 +1,175 @@
+defmodule TrentoWeb.OpenApi.V2.Schema.Cluster do
+  @moduledoc false
+
+  require OpenApiSpex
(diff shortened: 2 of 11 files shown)


lib/trento_web/openapi/v1/schema/cluster.ex
```diff
@@ -135,7 +130,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
         type: %Schema{
           type: :string,
           description: "Detected type of the cluster",
-          enum: ClusterType.values()
+          enum: [:hana_scale_up, :hana_scale_out, :unknown]
```
arbulu89, line 133: Using hardcoded values as they cannot be changed without adding non backward compatible changes.
This schema won't be updated anymore

lib/trento_web/views/v1/cluster_view.ex
```diff
@@ -22,30 +23,12 @@ defmodule TrentoWeb.V1.ClusterView do
     |> Map.put(:id, data.cluster_id)
   end
 
-  def render("settings.json", %{settings: settings}) do
```
arbulu89, line 25: Taking advantage to remove old unused code

assets/js/lib/network/index.js
```diff
@@ -68,36 +68,36 @@ function handleResponseStatus(response) {
   return response;
 }
 
-export const post = function post(url, data) {
+export const post = function post(url, data, config) {
```
arbulu89, line 71: We can pass the config to overwrite the existing one, in our case the `baseURL`. 
If an empty config is provided, the default one is used:
https://axios-http.com/docs/instance

lib/trento_web/openapi/v2/schema/cluster.ex
```diff
@@ -0,0 +1,175 @@
...
+  defmodule AscsErsClusterNode do
+    @moduledoc false
+
+    OpenApiSpex.schema(%{
+      title: "AscsErsClusterNode",
+      description: "ASCS/ERS Cluster Node",
+      type: :object,
+      properties: %{
+        attributes: %Schema{
+          type: :object,
```
nelsonkopliku, line 20: As before: is this a `key => value` of type `string => string`? Maybe `additionalProperties` could help.

lib/trento_web/openapi/v2/schema/cluster.ex
```diff
@@ -0,0 +1,175 @@
...
+          description: "Node name"
+        },
+        resources: %Schema{
+          type: :array,
+          items: Cluster.ClusterResource,
+          description: "A list of Cluster resources"
+        },
+        roles: %Schema{
+          type: :array,
+          items: %Schema{type: :string, enum: ["ascs", "ers"]},
```
nelsonkopliku, line 39: This mean that we can have at most two roles, right?

lib/trento_web/openapi/v2/schema/cluster.ex
```diff
@@ -0,0 +1,175 @@
...
+      type: :object,
+      properties: %{
+        distributed: %Schema{
+          type: :boolean,
+          description: "ASCS and ERS instances are distributed and running in different nodes"
+        },
+        filesystem_resource_based: %Schema{
+          type: :boolean,
+          description:
+            "ACSS and ERS filesystems are handled by the cluster with the Filesystem resource agent"
```
nelsonkopliku, line 66: is `ACSS` a typo? did you mean `ASCS`?

lib/trento_web/openapi/v1/schema/cluster.ex
```diff
@@ -37,13 +36,11 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
         site: %Schema{type: :string},
         hana_status: %Schema{type: :string},
         attributes: %Schema{
-          title: "ClusterNodeAttributes",
-          type: :array,
-          items: %Schema{type: :string}
+          type: :object,
```
nelsonkopliku, line 39: what kind of object can this be? I guess a `key => value` of type `string => string`.
Could er document it via the `additionalProperties` or something like that?

lib/trento_web/openapi/v2/schema/cluster.ex
```diff
@@ -0,0 +1,175 @@
...
+        },
+        provider: Provider.SupportedProviders,
+        type: %Schema{
+          type: :string,
+          description: "Detected type of the cluster",
+          enum: ClusterType.values()
+        },
+        selected_checks: %Schema{
+          title: "SelectedChecks",
+          description: "A list ids of the checks selected for execution on this cluster",
```
nelsonkopliku, line 147: `A list of check ids selected...` might flow smoother.

assets/js/state/sagas/index.js
```diff
@@ -119,7 +119,9 @@ function* initialDataFetch() {
   yield put(stopHostsLoading());
 
   yield put(startClustersLoading());
-  const { data: clusters } = yield call(get, '/clusters');
+  const { data: clusters } = yield call(get, '/clusters', {
```
nelsonkopliku, line 122: I guess for now this is enough. 
However at a certain point I'd find it of great help having something like `yield call(getV2, '/clusters');`
Feel free to keep it as you did for now.

nelsonkopliku (commented):
Wow, great job man! :rocket: 
I just noted down a few adjustments that might be helpful.

lib/trento_web/openapi/v2/schema/cluster.ex
arbulu89: Adding: `additionalProperties: %Schema{type: :string}`

lib/trento_web/openapi/v2/schema/cluster.ex
arbulu89: Yes, one node can have from 0 to 2 roles for each SAP system.
The normal scenario is when one node is only `ascs` and the other `ers`, but during a failover, one nodes takes both roles and the other none.
PD: Creating a tech-debt ticket to put the enum values in an `Enum` object, so we don't need to spread them all around the code

lib/trento_web/openapi/v2/schema/cluster.ex
arbulu89: Typo....
Thank you

assets/js/state/sagas/index.js
arbulu89: Fair enough. I wanted to keep the current changes minimalist.
This wouldn't require many changes, simply to add 4 new functions (get, post, del, put) with that naming.
Let's keep it on the radar

nelsonkopliku (approved):
Thanks @arbulu89!

assets/js/lib/network/index.js
CDimonaco: Do you think that for readability we could have the config with a default argument of null or something at least compatible with the "empty config" just to match the arity of the function without explicitly ignore the argument

CDimonaco (approved):
Just a minor comment but LGTM

PR #1496, opened 2023-06-07
Identify sap systems ensa version

arbulu89:
# Description

Identify the ENSA version of the SAP system application from the sap system discovery payloads.
The ENSA version belongs to the Application part of the SAP system (even though one sap system have one ENSA version).

Being ENSA1 or ENSA2 depends on the processes running on the application.

After this I need to move this field to the domain part and the persistence layer to send it to the frontend finally.

## How was this tested?
Tested using already existing fixture files with some "basic" modifications.

### DIFF
--- a/lib/trento/domain/enums/ensa_version.ex
+++ b/lib/trento/domain/enums/ensa_version.ex
@@ -1,0 +1,7 @@
+defmodule Trento.Domain.Enums.EnsaVersion do
+  @moduledoc """
+  Type that represents the supported ENSA versions.
+  """
+
+  use Trento.Support.Enum, values: [:ensa1, :ensa2, nil]
+end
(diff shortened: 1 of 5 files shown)


lib/trento/domain/enums/ensa_version.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Trento.Domain.Enums.EnsaVersion do
+  @moduledoc """
+  Type that represents the supported ENSA versions.
+  """
+
+  use Trento.Support.Enum, values: [:ensa1, :ensa2, nil]
```
fabriziosestito, line 6: I wouldn't use nil but something more domain-specific here, such as "no_ensa"

dottorblaster (approved):
Once you fix what @fabriziosestito already pointed out, no other blockers from me

nelsonkopliku (commented):
Late but yes :smile:

PR #1504, opened 2023-06-08
DatabaseDeregistered side effects

rtorrero:
# Description

This PR 
 - adds a deregistered_at field to the DatabaseReadModel
 - handles the side effects for the DatabaseDeregistered events
 - filters deregistered databases for the `get_all_databases` usecase

## How was this tested?

Tests where added on the projector and the usecases.

### DIFF
--- a/lib/trento/application/read_models/database_read_model.ex
+++ b/lib/trento/application/read_models/database_read_model.ex
@@ -18,7 +18,6 @@ defmodule Trento.DatabaseReadModel do
   schema "databases" do
     field :sid, :string
     field :health, Ecto.Enum, values: Health.values()
-    field :deregistered_at, :utc_datetime_usec
 
     has_many :tags, Trento.Tag, foreign_key: :resource_id
 
@@ -26,6 +25,8 @@ defmodule Trento.DatabaseReadModel do
       references: :id,
       foreign_key: :sap_system_id,
       preload_order: [asc: :instance_number, asc: :host_id]
+
+    field :deregistered_at, :utc_datetime_usec
   end
 
   @spec changeset(t() | Ecto.Changeset.t(), map) :: Ecto.Changeset.t()
--- a/lib/trento/application/read_models/database_read_model.ex
+++ b/lib/trento/application/read_models/database_read_model.ex
@@ -18,6 +18,7 @@ defmodule Trento.DatabaseReadModel do
   schema "databases" do
     field :sid, :string
     field :health, Ecto.Enum, values: Health.values()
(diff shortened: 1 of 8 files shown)


lib/trento/application/read_models/database_read_model.ex
```diff
@@ -18,6 +18,7 @@ defmodule Trento.DatabaseReadModel do
   schema "databases" do
     field :sid, :string
     field :health, Ecto.Enum, values: Health.values()
+    field :deregistered_at, :utc_datetime_usec
```
arbulu89, line 21: Just to be consistent, we could put this at the bottom of the file, as we have in the sap systems projector

priv/repo/migrations/20230608105745_add_deregistered_at_field_db_rm.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Trento.Repo.Migrations.AddDeregisteredAtFieldDbRm do
```
arbulu89, line 1: Change filename to: `xxx_add_deregistered_at_to_database_read_model` and class name `AddDeregisteredAtToDatabaseReadModel`

lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -125,6 +125,12 @@ defmodule TrentoWeb.V1.SapSystemView do
 
   def render("sap_system_deregistered.json", %{id: id, sid: sid}), do: %{id: id, sid: sid}
 
+  def render("database_deregistered.json", %{
```
arbulu89, line 128: This could be one liner, right? It could even be `def render("database_deregistered.json", database), do: database`.
But it is alright to leave as the one above

arbulu89 (approved):
Good!
Some nitpick comments

test/trento/application/usecases/sap_systems_test.exs
```diff
@@ -48,12 +48,14 @@ defmodule Trento.SapSystemsTest do
...
     end
 
-    test "should retrieve all the existing databases and the related instances" do
+    test "should retrieve all the currently registered existing databases and the related instances" do
       %DatabaseReadModel{
         id: sap_system_id,
         sid: sid
       } = insert(:database)
 
+      insert(:database, deregistered_at: DateTime.utc_now())
```
jamie-suse, line 57: Shouldn't this cause a failure in the test?

jamie-suse: At the very least, shouldn't we verify that this database is NOT returned from `SapSystems.get_all_databases()`

PR #1505, opened 2023-06-08
Make `HostReadModel.last_heartbeat_timestamp` a virtual field

jamie-suse:
# Description

Renames `HostReadModel.heartbeat_timestamp` to `HostReadModel.last_heartbeat_timestamp`.

Changes `HostReadModel.last_heartbeat_timestamp` to use a virtual field instead of an association with the `Heartbeat` table.

## How was this tested?

Updated test for `get_all_hosts/0`, the existing unit tests should validate that the rest of the behaviour remains the same.

Added tests for new function `get_host_by_id/1`

### DIFF
--- a/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
+++ b/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
@@ -18,9 +18,8 @@ defimpl Trento.Support.Middleware.Enrichable,
           {:ok, RequestHostDeregistration.t()}
           | {:error, :host_alive}
           | {:error, :host_not_registered}
-  def enrich(%RequestHostDeregistration{host_id: host_id} = command, _) do
-    host_deregisterable(Hosts.get_host_by_id(host_id), command)
-  end
+  def enrich(%RequestHostDeregistration{host_id: host_id} = command, _),
+    do: host_deregisterable(Hosts.get_host_by_id(host_id), command)
 
   defp host_deregisterable(
          %HostReadModel{last_heartbeat_timestamp: nil, deregistered_at: nil},
--- a/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
+++ b/lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
@@ -2,8 +2,13 @@ defimpl Trento.Support.Middleware.Enrichable,
   for: Trento.Domain.Commands.RequestHostDeregistration do
   alias Trento.Domain.Commands.RequestHostDeregistration
 
-  alias Trento.Repo
-  alias Trento.HostReadModel
+  import Ecto.Query
+
+  alias Trento.{
+    Heartbeat,
+    HostReadModel,
+    Repo
+  }
 
   @heartbeat_interval Application.compile_env!(:trento, Trento.Heartbeats)[:interval]
   @deregistration_debounce Application.compile_env!(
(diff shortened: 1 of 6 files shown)


lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
```diff
@@ -17,32 +22,41 @@ defimpl Trento.Support.Middleware.Enrichable,
...
-           DateTime.add(timestamp, @total_deregistration_debounce, :millisecond)
+           DateTime.add(last_heartbeat_timestamp, @total_deregistration_debounce, :millisecond)
          ),
        do: {:error, :host_alive},
        else: {:ok, command}
   end
 
   defp host_deregisterable(_, _), do: {:error, :host_not_registered}
+
+  @spec enrich_host_read_model_query(Ecto.Query.t()) :: Ecto.Query.t()
```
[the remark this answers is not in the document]
arbulu89: Why don't you add a `get_host_by_id` or similar in the usecases file?
You would reuse the enrichment function there, and here you could use this instead of running the query here.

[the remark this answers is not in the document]
CDimonaco: It's just query related, I think we could safely place it in the Readmodel itself and call it from other parts

So you can pass enrich readmodel related query as you need

lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
```diff
@@ -17,32 +22,41 @@ defimpl Trento.Support.Middleware.Enrichable,
           | {:error, :host_alive}
           | {:error, :host_not_registered}
   def enrich(%RequestHostDeregistration{host_id: host_id} = command, _) do
-    Repo.get(HostReadModel, host_id)
-    |> Repo.preload([:heartbeat_timestamp])
+    HostReadModel
+    |> where([h], h.id == ^host_id)
```
arbulu89, line 26: Should we check here if the host was already deregistered?
Or are we going to "reject" these deregistration commands in the domain aggregate itself.

arbulu89, line 26: `get_by`

arbulu89 (dismissed):
Let's move the logic of getting a particular enriched host to the `usecases`.

PD: Doesn't the change in the protocol require to change the tests?

lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
jamie-suse: We are already checking if the host is already deregistered in the `host_deregisterable()` function.

jamie-suse: (There is also a test for this)

jamie-suse:
> Let's move the logic of getting a particular enriched host to the `usecases`.
> 
> PD: Doesn't the change in the protocol require to change the tests?

From the test's perspective, nothing is changing though, the tests don't care *how* we get the Heartbeat timestamp into the Host Read Model - it's just an implementation detail.

lib/trento/application/usecases/hosts/hosts.ex
```diff
@@ -20,8 +21,18 @@ defmodule Trento.Hosts do
     HostReadModel
     |> where([h], not is_nil(h.hostname) and is_nil(h.deregistered_at))
     |> order_by(asc: :hostname)
+    |> enrich_host_read_model_query()
     |> Repo.all()
-    |> Repo.preload([:sles_subscriptions, :tags, :heartbeat_timestamp])
+    |> Repo.preload([:sles_subscriptions, :tags])
+  end
+
+  @spec get_host_by_id(Ecto.UUID.t()) :: HostReadModel.t()
```
arbulu89, line 29: Tests?

arbulu89, line 29: Is the signature okey?
What does it return if the id is not found? maybe nil?
just asking

arbulu89 (commented):
Add tests to the new function.
Looks good besides that.
Just putting a simple comment so others review this as I will be off tomorrow

CDimonaco (approved):

lib/trento/application/integration/discovery/protocol/enrich_request_host_deregistration.ex
```diff
@@ -17,32 +22,41 @@ defimpl Trento.Support.Middleware.Enrichable,
...
        do: {:error, :host_alive},
        else: {:ok, command}
   end
 
   defp host_deregisterable(_, _), do: {:error, :host_not_registered}
+
+  @spec enrich_host_read_model_query(Ecto.Query.t()) :: Ecto.Query.t()
+  defp enrich_host_read_model_query(query) do
+    query
+    |> join(:left, [h], hb in Heartbeat, on: type(h.id, :string) == hb.agent_id)
```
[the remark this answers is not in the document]
jamie-suse: As discussed: since we reuse this helper function in both `enrich_host_read_model_query()` in `get_all_hosts()` and `get_host_by_id()`, using a `JOIN` like this offers a more flexible and reusable solution.

nelsonkopliku (commented):
Nice job! Just leaving a couple of comments aiming to be helpful

CDimonaco (changes requested):
Thanks @nelsonkopliku for finding it, we need to exclude the deregistered host in get_host_by_id so we can avoid to fetch a deregistered hosts, and instead returning a `nil` value.

CDimonaco (approved):

nelsonkopliku (approved):

PR #1507, opened 2023-06-08
Project removal of db/app instances to the read models

rtorrero:
# Description
Project the removal of db and application instances to the relevant read models

## How was this tested?

Automated tests added

### DIFF
--- a/lib/trento/application/projectors/database_projector.ex
+++ b/lib/trento/application/projectors/database_projector.ex
@@ -18,6 +18,7 @@ defmodule Trento.DatabaseProjector do
   alias Trento.Domain.Events.{
     DatabaseDeregistered,
     DatabaseHealthChanged,
+    DatabaseInstanceDeregistered,
     DatabaseInstanceHealthChanged,
     DatabaseInstanceRegistered,
     DatabaseInstanceSystemReplicationChanged,
@@ -159,6 +160,25 @@ defmodule Trento.DatabaseProjector do
     end
   )
 
+  project(
+    %DatabaseInstanceDeregistered{
+      instance_number: instance_number,
+      host_id: host_id,
+      sap_system_id: sap_system_id,
+      deregistered_at: _deregistered_at
+    },
+    fn multi ->
+      deregistered_instance =
+        Repo.get_by(DatabaseInstanceReadModel,
+          sap_system_id: sap_system_id,
+          instance_number: instance_number,
+          host_id: host_id
+        )
+
+      Ecto.Multi.delete(multi, :database_instance, deregistered_instance)
+    end
+  )
+
   @impl true
   def after_update(
         %DatabaseRegistered{},
@@ -279,6 +299,27 @@ defmodule Trento.DatabaseProjector do
       SapSystemView.render("database_deregistered.json",
         id: sap_system_id,
         sid: sid
+      )
+    )
+  end
+
+  @impl true
+  def after_update(
+        %DatabaseInstanceDeregistered{
+          instance_number: instance_number,
+          host_id: host_id,
+          sap_system_id: sap_system_id
+        },
+        _,
+        _
+      ) do
+    TrentoWeb.Endpoint.broadcast(
+      @databases_topic,
+      "database_instance_deregistered",
+      SapSystemView.render("database_instance_deregistered.json",
+        id: sap_system_id,
+        instance_number: instance_number,
+        host_id: host_id
       )
     )
   end
--- a/lib/trento/application/projectors/database_projector.ex
+++ b/lib/trento/application/projectors/database_projector.ex
@@ -164,16 +164,14 @@ defmodule Trento.DatabaseProjector do
     %DatabaseInstanceDeregistered{
       instance_number: instance_number,
       host_id: host_id,
-      sap_system_id: sap_system_id,
-      deregistered_at: _deregistered_at
+      sap_system_id: sap_system_id
     },
     fn multi ->
-      deregistered_instance =
-        Repo.get_by(DatabaseInstanceReadModel,
-          sap_system_id: sap_system_id,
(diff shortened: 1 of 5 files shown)


lib/trento/application/projectors/database_projector.ex
```diff
@@ -159,6 +160,25 @@ defmodule Trento.DatabaseProjector do
...
+  project(
+    %DatabaseInstanceDeregistered{
+      instance_number: instance_number,
+      host_id: host_id,
+      sap_system_id: sap_system_id,
+      deregistered_at: _deregistered_at
+    },
+    fn multi ->
+      deregistered_instance =
+        Repo.get_by(DatabaseInstanceReadModel,
```
arbulu89, line 172: I don't think you need a `get_by`. Just creating the read model with the primary keys should be enough.
https://hexdocs.pm/ecto/Ecto.Repo.html#c:delete/2

lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -127,6 +127,20 @@ defmodule TrentoWeb.V1.SapSystemView do
 
   def render("database_deregistered.json", %{id: id, sid: sid}), do: %{id: id, sid: sid}
 
+  def render("database_instance_deregistered.json", %{
```
arbulu89, line 130: Any reason to have two views instead of simply `instante_deregistered`?

lib/trento/application/projectors/sap_system_projector.ex
```diff
@@ -221,6 +241,27 @@ defmodule Trento.SapSystemProjector do
...
+          sap_system_id: sap_system_id
+        },
+        _,
+        _
+      ) do
+    TrentoWeb.Endpoint.broadcast(
+      @sap_systems_topic,
+      "application_instance_deregistered",
+      SapSystemView.render("application_instance_deregistered.json",
+        id: sap_system_id,
```
arbulu89, line 258: Let's be careful with the payload.
I think we use `sap_system_id` to identify the sap system in the app/db instances.
If we send it with the same field name, updating the entries in redux will be easier as well.
You can do the transformation in the view if you want

test/trento/application/projectors/database_projector_test.exs
```diff
@@ -264,4 +265,33 @@ defmodule Trento.DatabaseProjectorTest do
...
     assert deregistered_at == projection.deregistered_at
   end
+
+  test "should remove a database instance from the read model after a deregistration" do
+    deregistered_at = DateTime.utc_now()
+
+    insert(:database, id: sap_system_id = Faker.UUID.v4())
+
+    %{instance_number: instance_number, host_id: host_id} =
+      insert(:database_instance, sap_system_id: sap_system_id)
```
arbulu89, line 275: I would be a bit more exhaustive here, add more instances, and check that they are not removed.

lib/trento_web/views/v1/sap_system_view.ex
arbulu89, line 130: I'm just finding this views redundant (together with some added like those). They don't do anything, and they are not even tested.
I don't know how worthy is to have them...
Anyway, just thinking aloud

arbulu89 (commented):
Some nitpick/improvements comment

lib/trento/application/projectors/database_projector.ex
```diff
@@ -159,6 +160,25 @@ defmodule Trento.DatabaseProjector do
     end
   )
 
+  project(
+    %DatabaseInstanceDeregistered{
+      instance_number: instance_number,
+      host_id: host_id,
+      sap_system_id: sap_system_id,
+      deregistered_at: _deregistered_at
```
jamie-suse, line 168: If you don't care about `deregistered_at`, do you even need to pattern match on it here?

lib/trento/application/projectors/sap_system_projector.ex
```diff
@@ -130,6 +131,25 @@ defmodule Trento.SapSystemProjector do
     end
   )
 
+  project(
+    %ApplicationInstanceDeregistered{
+      instance_number: instance_number,
+      host_id: host_id,
+      sap_system_id: sap_system_id,
+      deregistered_at: _deregistered_at
```
jamie-suse, line 139: Same as previous comment

test/trento/application/projectors/database_projector_test.exs
rtorrero: Agreed. I added 4 additional instances of both tests. Hope you don't mind that I'm checking the count instead of comparing the whole content as this seems simpler. I can change it otherwise.

test/trento/application/projectors/database_projector_test.exs
```diff
@@ -264,4 +265,41 @@ defmodule Trento.DatabaseProjectorTest do
...
+    ProjectorTestHelper.project(DatabaseProjector, event, "database_projector")
+
+    assert nil ==
+             Repo.get_by(DatabaseInstanceReadModel,
+               sap_system_id: sap_system_id,
+               instance_number: instance_number,
+               host_id: host_id
+             )
+
+    query = from(d in DatabaseInstanceReadModel, select: fragment("count(?)", d.sap_system_id))
```
arbulu89, line 294: Simpler this way maybe?
Why not simply 
```
assert 4 ==
    DatabaseInstanceReadModel
    |> Replo.all()
    |> Enum.count()
```

PD: Please, be strict with the order of the asserts: `assert expected == real`

arbulu89 (approved):
Please, be strict with the order of the asserts: `assert expected == real`.

CDimonaco (approved):
I agree with @arbulu89 comments.

Anyway good job! LGTM

jamie-suse (approved):

PR #1509, opened 2023-06-09
Add resource type when filtering the catalog

nelsonkopliku:
# Description

Adds `resource_type=cluster` when querying the catalog.
Defaults now to `cluster` to keep compatibility.

Will be subject to change when we actually add host checks and we revisit Catalog's UI as we should be able to:
- get the whole catalog
- filter by specific properties (ie provider|resource type)
- filter by combinations of properties (ie provider&&resource type)

## How was this tested?

Tests updated.


assets/js/components/ChecksCatalog/ChecksCatalog.jsx
```diff
@@ -24,7 +24,9 @@ const updatedProvider = {
 };
 
 const buildUpdateCatalogAction = (provider) => {
-  const payload = checkProviderExists(provider) ? { provider } : {};
+  const payload = checkProviderExists(provider)
```
arbulu89, line 27: What if we choose `All` as provider filtering. This would return `false` in `checkProviderExists` and we wouldn't filter by `resource_type`, right?
It looks like unfinished. 
I don't know if we should start filtering here behind the scenes before knowing how the user will filter

nelsonkopliku: Correct, when filtering by `All` we don't get anything in the qs, meaning give me everything (wanda does not apply any filtering if an empty `env` was provided)

I believe this is how far we can get by now.

nelsonkopliku: >  why the filtering doesn't work without the resource_type?
Because the when expression is just one expression ie `when: env.resource_type == "cluster" && env.provider !in ["gcp", "aws"]` and if no resource type is in the env or no provider you see it fails.

> If I were a user and would like to get all Azure checks, why I'm forced to filter by resource type?
I agree. That's the limitation of having everything in the when condition.

arbulu89 (approved):
Let's see how we progress on this to filter by the resource type, but this should allow to continue working, so good from my side

PR #1510, opened 2023-06-09
Add text to loading state cluster details

EMaksy:
# Description
This pull request updates the CheckResultsOverview component to implement the design provided by @jagabomb. As part of this update, the Spinner component was enhanced to improve customizability and accessibility.

### Before
beforeChangeDemo

### Now:
afterChangeDemo

## How was this tested?

Test was added for Spinner  component
Story was added for Spinner component

### DIFF
--- a/assets/js/components/Spinner.jsx
+++ b/assets/js/components/Spinner.jsx
@@ -1,19 +1,10 @@
 import React from 'react';
 import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
 
-function Spinner({
-  centered = false,
-  wrapperClassName = '',
-  size = 'm',
-  spinnerColor = 'fill-jungle-green-500',
-}) {
+function Spinner({ className = '', size = 'm' }) {
   return (
-    <div role="alert" aria-label="Loading" className={wrapperClassName}>
-      <EOS_LOADING_ANIMATED
-        size={size}
-        centered={centered}
-        className={spinnerColor}
-      />
+    <div role="alert" aria-label="Loading" className={className}>
+      <EOS_LOADING_ANIMATED size={size} className={'fill-jungle-green-500'} />
     </div>
   );
 }
--- a/assets/js/components/Spinner.jsx
+++ b/assets/js/components/Spinner.jsx
@@ -1,13 +1,18 @@
 import React from 'react';
-import { computedIconCssClass } from '@lib/icon';
-
 import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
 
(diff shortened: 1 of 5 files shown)


jagabomb:
@EMaksy Thanks for this. A last request is to make the text gray for, **Check execution running...** Other than that it looks good to me.

assets/js/components/Spinner.jsx
```diff
@@ -1,13 +1,18 @@
...
-import { computedIconCssClass } from '@lib/icon';
-
 import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
 
-function Spinner({ centered = false }) {
+function Spinner({
+  centered = false,
+  wrapperClassName = '',
+  size = 'm',
+  spinnerColor = 'fill-jungle-green-500',
```
nelsonkopliku, line 8: When do we need a custom color? :thinking:

dottorblaster: Agree

assets/js/components/Spinner.jsx
```diff
@@ -1,13 +1,18 @@
 import React from 'react';
-import { computedIconCssClass } from '@lib/icon';
-
 import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
 
-function Spinner({ centered = false }) {
+function Spinner({
+  centered = false,
+  wrapperClassName = '',
```
dottorblaster, line 6: this can be just `className`

dottorblaster (changes requested):
Just a thing then we can merge :shipit:

assets/js/components/Spinner.jsx
EMaksy: Yes and also i don't think we need centered any more :D 
I checked out, it's not used in the codebase or did i miss something?

assets/js/components/Spinner.stories.jsx
```diff
@@ -0,0 +1,28 @@
...
+};
+
+export const Default = {
+  argTypes: {
+    className: {
+      control: 'text',
+      description: 'Add padding or margin',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: '""' },
```
nelsonkopliku, line 15: what happens with `{ summary: '' },`?

nelsonkopliku (approved):

assets/js/components/Spinner.stories.jsx
EMaksy: Ah well done, nice catch. This is a left over! 
Storybook detects the default value. No need to add it here.

PR #1512, opened 2023-06-12
Ensa version domain

arbulu89:
# Description
Add `ensa_version` detection and logic to the SAP system domain.

Some context:
I'm forced to add this new `SapSystemUpdated` event as we need to update already existing sap systems. If it was in a clean state, we wouldn't need it, as the ensa_version doesn't really change in a life-cycle of a SAP system.
The current update is **eventual consistent**. This means that if the SAP system is registered with non Message server or Enqueue replicator instances (primary application, for example), the `ensa_version` wouldn't update, so it would update once we receive a message from those application instances.

PD1: I have done a lot of changes in the tests, as they were running the order of events incorrectly. Since we changed the SAP systems registration part, the instances are registered before the SAP system itself. Without changing this order, all the tests were failing as the `Application` field was not being initialized.

PD2:  Frontend work is missing

## How was this tested?

Tests added

### DIFF
--- a/lib/trento/application/projectors/sap_system_projector.ex
+++ b/lib/trento/application/projectors/sap_system_projector.ex
@@ -14,7 +14,8 @@ defmodule Trento.SapSystemProjector do
     ApplicationInstanceRegistered,
     SapSystemDeregistered,
     SapSystemHealthChanged,
-    SapSystemRegistered
+    SapSystemRegistered,
+    SapSystemUpdated
   }
 
   alias TrentoWeb.V1.SapSystemView
@@ -32,7 +33,8 @@ defmodule Trento.SapSystemProjector do
       sid: sid,
       tenant: tenant,
       db_host: db_host,
-      health: health
+      health: health,
+      ensa_version: ensa_version
     },
     fn multi ->
       changeset =
@@ -41,7 +43,8 @@ defmodule Trento.SapSystemProjector do
           sid: sid,
           tenant: tenant,
           db_host: db_host,
-          health: health
+          health: health,
+          ensa_version: ensa_version
         })
 
       Ecto.Multi.insert(multi, :sap_system, changeset)
@@ -146,6 +149,21 @@ defmodule Trento.SapSystemProjector do
         )
 
       Ecto.Multi.delete(multi, :application_instance, deregistered_instance)
+    end
+  )
+
+  project(
+    %SapSystemUpdated{
+      sap_system_id: sap_system_id,
+      ensa_version: ensa_version
+    },
+    fn multi ->
+      changeset =
+        SapSystemReadModel.changeset(%SapSystemReadModel{id: sap_system_id}, %{
+          ensa_version: ensa_version
+        })
+
+      Ecto.Multi.update(multi, :sap_system, changeset)
     end
   )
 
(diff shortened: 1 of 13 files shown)


lib/trento/domain/sap_system/application.ex
```diff
@@ -11,6 +13,7 @@ defmodule Trento.Domain.SapSystem.Application do
 
   deftype do
     field :sid, :string
+    field :ensa_version, Ecto.Enum, values: EnsaVersion.values(), default: EnsaVersion.no_ensa()
```
arbulu89, line 16: Setting the value in the `Application` field, as this term is related to the Application part.
It could very well go in the SAP system aggregate as well, as a SAP system only has one application, but it looked more appropriate to put it here to have better context

priv/repo/migrations/20230606152133_add_ensa_version_to_sap_system_read_model.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Trento.Repo.Migrations.AddEnsaVersionToSapSystemReadModel do
+  use Ecto.Migration
+
+  def change do
+    alter table(:sap_systems) do
+      add :ensa_version, :string, default: "no_ensa"
```
arbulu89, line 6: Setting already a default correct value to have the read models correctly populated

CDimonaco: I'm not a really fan of settings the defaults in the schema migration, this looks more like a hybrid within a data migration and a schema migration.

If we really need that, and this could no be enforced at application level, I will suggest another migration which will take care only of migrating data, setting the default value

Wdyt?

arbulu89: How can I enforce this in the application?
I simply need a value that is within the new `EnsaVersion` enum for the scenario that it was not discovered.
Otherwise the database is initialized with `nil` until we get the event that changes this breaking the openapi schema, and in general the usage of the Enum

CDimonaco: You are talking about the already present sap systems when the migration will run?

CDimonaco: Oh okay, so you can create another migration, just with the data migration and leave the schema migration you already created without the default value.

That way you don't alter the schema with the default value (we could talk about having something default if we switch to postgres enums, but I don't recommend it, they are a pain to mantain and evolve), and leave a separate migration which migrates only data

arbulu89: Done!
I have added the update in a new file to make it more obvious.
Ecto doesn't have any `update` command, so the `execute` with the basic SQL command looks the best solution

CDimonaco (approved):
Awesome, LGTM

lib/trento/application/projectors/sap_system_projector.ex
```diff
@@ -149,6 +152,21 @@ defmodule Trento.SapSystemProjector do
     end
   )
 
+  project(
+    %SapSystemUpdated{
```
nelsonkopliku, line 156: Does this need an after update broadcast?

nelsonkopliku (approved):
Cool! I only have a doubt whether we should also broadcast upon `SapSystemUpdated` event.

lib/trento/application/projectors/sap_system_projector.ex
arbulu89: Yes, coming in a new PR with frontend work

PR #1513, opened 2023-06-12
Reject commands when aggregate deregistered

CDimonaco:
# Description

The aggregates, when deregistered, should reject all commands except the registration ones, which could potentially trigger a restore of the aggregate.

The SAP system aggregate has a particular behavior, it should reject all the commands except for:

- Database/Application instance registration, which could trigger a restore
- Database/Application instance deregistration, because even if the aggregate is deregistered, I could have some application/database instance dangling, and I want to remove that before tombstoning the aggregate

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/domain/cluster/cluster.ex
+++ b/lib/trento/domain/cluster/cluster.ex
@@ -233,7 +233,10 @@ defmodule Trento.Domain.Cluster do
   end
 
   def execute(%Cluster{cluster_id: nil}, _),
-    do: {:error, :cluster_not_found}
+    do: {:error, :cluster_not_registered}
+
+  def execute(%Cluster{deregistered_at: deregistered_at}, _) when not is_nil(deregistered_at),
+    do: {:error, :cluster_not_registered}
 
   # Checks selected
   def execute(
--- a/lib/trento/domain/cluster/cluster.ex
+++ b/lib/trento/domain/cluster/cluster.ex
@@ -201,6 +201,12 @@ defmodule Trento.Domain.Cluster do
     ]
   end
 
+  def execute(%Cluster{cluster_id: nil}, _),
+    do: {:error, :cluster_not_registered}
+
+  def execute(%Cluster{deregistered_at: deregistered_at}, _) when not is_nil(deregistered_at),
+    do: {:error, :cluster_not_registered}
+
   # If the cluster is already registered, and the host was never discovered before, it is added to the cluster.
   def execute(
         %Cluster{} = cluster,
@@ -231,12 +237,6 @@ defmodule Trento.Domain.Cluster do
     end)
     |> Multi.execute(fn cluster -> maybe_emit_cluster_health_changed_event(cluster) end)
   end
-
-  def execute(%Cluster{cluster_id: nil}, _),
-    do: {:error, :cluster_not_registered}
-
-  def execute(%Cluster{deregistered_at: deregistered_at}, _) when not is_nil(deregistered_at),
-    do: {:error, :cluster_not_registered}
 
   # Checks selected
   def execute(
--- a/lib/trento/domain/host/host.ex
+++ b/lib/trento/domain/host/host.ex
@@ -148,6 +148,22 @@ defmodule Trento.Domain.Host do
     []
   end
 
+  # Reject all the commands, except for the registration ones when the host_id does not exists
+  def execute(
+        %Host{host_id: nil},
+        _
+      ) do
+    {:error, :host_not_registered}
+  end
+
+  def execute(
+        %Host{deregistered_at: deregistered_at},
+        _
+      )
+      when not is_nil(deregistered_at) do
+    {:error, :host_not_registered}
+  end
+
   def execute(
         %Host{},
         %RegisterHost{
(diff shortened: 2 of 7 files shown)


lib/trento/domain/host/host.ex
```diff
@@ -183,6 +183,14 @@ defmodule Trento.Domain.Host do
     {:error, :host_not_registered}
   end
 
+  def execute(
```
arbulu89, line 186: This has still some events on top, like `HostDetailsUpdated`.
Shouldn't we exclude these events as well?
I mean to updated something that "doesn't exist"

lib/trento/domain/cluster/cluster.ex
```diff
@@ -233,7 +233,10 @@ defmodule Trento.Domain.Cluster do
   end
 
   def execute(%Cluster{cluster_id: nil}, _),
-    do: {:error, :cluster_not_found}
+    do: {:error, :cluster_not_registered}
```
arbulu89, line 236: The same here.
There are some events (details updated, health changed...) on top of this.
We should exclude them.
I think we only need to leave the first 2 `execute`s (and later on the restoration part, which is to come)

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -287,6 +276,24 @@ defmodule Trento.Domain.SapSystem do
     |> Multi.execute(&maybe_emit_sap_system_tombstoned_event/1)
   end
 
+  def execute(
```
arbulu89, line 279: The same as others.
I think we should only leave the database and application registration `execute`s on top of this.
For the application is not really possible, as we cascade multiple events together.
But we could simply leave those 2 functions on top

test/trento/domain/host/host_test.exs
```diff
@@ -651,6 +651,48 @@ defmodule Trento.HostTest do
   end
 
   describe "deregistration" do
+    test "should reject all the commands except the registration ones when the host is deregistered" do
+      host_id = Faker.UUID.v4()
+      dat = DateTime.utc_now()
```
arbulu89, line 656: You don't really need a variable for this (`dat`), you could run this in the event itself, as it is not used anywhere else

test/trento/domain/host/host_test.exs
```diff
@@ -651,6 +651,48 @@ defmodule Trento.HostTest do
...
+      ]
+
+      for command <- commands_to_reject do
+        assert_error(initial_events, command, {:error, :host_not_registered})
+      end
+
+      register_host_command = build(:register_host_command, host_id: host_id)
+
+      assert_events(initial_events, [register_host_command], [
+        %HostDetailsUpdated{
```
arbulu89, line 682: I'm not sure why we allow this

arbulu89 (commented):
Hey @CDimonaco ,
Before finishing the PR and dropping some repeated (more yet...) comments, I have doubts we the code permits the `Updated` events for deregistered aggregates.
I don't think we should allow them. If the aggregate is deregistered, meaning "deleted", we cannot change it in any way.
Imagine that we had a normal deletion instead of a soft deletion, would we allow those events to happen? No, because the projection will break

CDimonaco:
You're right about the update of aggregates, reverting to draft, I will trigger a re-review @arbulu89

test/trento/domain/host/host_test.exs
```diff
@@ -651,6 +651,33 @@ defmodule Trento.HostTest do
...
+      ]
+
+      commands_to_reject = [
+        %DeregisterHost{},
+        %RequestHostDeregistration{},
+        %RollUpHost{},
+        %UpdateHeartbeat{},
+        %UpdateProvider{},
+        %UpdateSlesSubscriptions{},
+        %HostDetailsUpdated{}
```
arbulu89, line 673: This is wrong, `HostDetailsUpdated` is an event, not a command.
You should have `RegisterHost`, and you would need to update it after the restoration code.
Up to you, you can remove that entry and work on it during in the restoration changes

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -777,6 +777,39 @@ defmodule Trento.ClusterTest do
...
+          deregistered_at: deregistered_at
+        )
+      ]
+
+      commands_to_reject = [
+        %CompleteChecksExecution{},
+        %DeregisterClusterHost{},
+        %RollUpCluster{},
+        %SelectChecks{},
+        %RegisterClusterHost{}
```
arbulu89, line 804: Nice! you will need to adapt it during restoration, but so far so good hehe

lib/trento/domain/sap_system/sap_system.ex
arbulu89, line 279: Should we have other `execute` for the `deregistered_at` of the database?
The reality is that we don't have commands in the sap system aggregate outside of the registration and deregistration, so nothing can go wrong in reality.
So by now, it will simply work hehe

arbulu89 (approved):
Nice @CDimonaco !
I'm just "ignoring" a bit the sap system part, because it is really hard, with so many `maybe_` things. I thing during the restoration we will see changes here and there for this aggregate.
Simply fix the test for the `host`

lib/trento/domain/sap_system/sap_system.ex
CDimonaco: Other than rollup, we don't have commands right know, but anyway I think the clause is sufficient

fabriziosestito (approved):
Great!

PR #1515, opened 2023-06-13
Display ensa version frontend

arbulu89:
# Description
Frontend part for ENSA replication. Next for: https://github.com/trento-project/web/pull/1512
Basically, get and use the `ensa_version` field. For that, we need to get with the selector the SAP systems belong to this cluster, which could be multiple in a multi-sid cluster.

## How was this tested?

Tested


assets/js/lib/test-utils/factories/sapSystems.js
```diff
@@ -18,31 +20,36 @@ const roles = () =>
 export const sapSystemApplicationInstanceFactory = Factory.define(() => ({
   features: roles().join('|'),
   health: healthEnum(),
-  hostId: faker.datatype.uuid(),
-  httpPort: faker.internet.port(),
-  httpsPort: faker.internet.port(),
-  instanceHostname: faker.hacker.noun(),
-  instanceNumber: faker.datatype.number({ min: 10, max: 99 }).toString(),
+  host_id: faker.datatype.uuid(),
```
arbulu89, line 23: I have changed this factory, as it was before, was not creating data as we store in the `redux` store. It was using as it comes the `GenericSystemDetails` view page.
Just to make it more generic, now it replicates the data coming from the web.

assets/js/components/SapSystemDetails/GenericSystemDetails.test.jsx
```diff
@@ -2,19 +2,21 @@ import React from 'react';
...
-import { sapSystemFactory } from '@lib/test-utils/factories';
+import { hostFactory, sapSystemFactory } from '@lib/test-utils/factories';
 
 import { GenericSystemDetails } from './GenericSystemDetails';
 
 describe('GenericSystemDetails', () => {
   it('should render correctly', () => {
     const title = faker.datatype.uuid();
-    const sapSystem = sapSystemFactory.build();
+    const sapSystem = keysToCamel(sapSystemFactory.build());
```
arbulu89, line 15: The change on the factory forces me to adapt this a little.

assets/js/lib/test-utils/factories/clusters.js
```diff
@@ -112,7 +117,10 @@ export const clusterFactory = Factory.define(({ sequence, params }) => {
...
 });
 
-export const addHostsToAscsErsClusterDetails = (details) =>
+export const buildHostsFromAscsErsClusterDetails = (details) =>
   details.sap_systems
     .flatMap(({ nodes }) => nodes)
     .map(({ name }) => hostFactory.build({ hostname: name }));
+
+export const buildSapSystemsFromAscsErsClusterDetails = (details) =>
+  details.sap_systems.map(({ sid }) => sapSystemFactory.build({ sid }));
```
dottorblaster, line 126: The only super nitpicky thing I can comment is that I'm not a huge fan of utility functions inside the factories file, as creating the right dataset is responsibility (in theory) of the test itself.

Maybe we could move this to the test directly? Your call :smile:

dottorblaster (approved):
Just one nitpick then we can merge :+1:

rtorrero (approved):
Good stuff @arbulu89. I think changing the factory to settle for `snake_case` instead of having a mix of both `snake_case` and `camelCase` was the right call. For the rest of the PR, no comments on my side :+1:

assets/js/lib/test-utils/factories/clusters.js
arbulu89: The thing is that these functions are not only used in the tests. They are needed for storybook as well.
In fact, I consider them part of the factory, as they complement the data creation.
We have multiple utilities like this in other factories, specially in the `execution.js` with the `with...` syntax.

In any case, if I move them to the tests, I would need to import them to storybook files from the tests, or duplicate the function.
What is the best solution for you? I would vote for keeping as it is, but I want to know your opinion hehe

dottorblaster: Ouch, you are right. Then I think this is the most viable tradeoff. LGTM :smile:

nelsonkopliku (approved):

PR #1523, opened 2023-06-14
Cluster restoration

CDimonaco:
# Description

Cluster aggregate restoration.

When a deregistered cluster receives a RegisterClusterHost command:

- If the host is not a dc, the cluster is restored and the host added to cluster
- If the host is a dc, the cluster is restored, the host is added to cluster and the update cluster procedure is performed, similar to the registration

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/domain/cluster/cluster.ex
+++ b/lib/trento/domain/cluster/cluster.ex
@@ -207,7 +207,7 @@ defmodule Trento.Domain.Cluster do
 
   # Restoration, when a RegisterClusterHost command is received for a deregistered Cluster
   # the cluster is restored, the host is added to cluster and if the host is a DC
-  # cluster details will be updated
+  # cluster details are be updated
   def execute(
         %Cluster{deregistered_at: deregistered_at, cluster_id: cluster_id},
         %RegisterClusterHost{
--- a/lib/trento/domain/cluster/cluster.ex
+++ b/lib/trento/domain/cluster/cluster.ex
@@ -207,7 +207,7 @@ defmodule Trento.Domain.Cluster do
 
   # Restoration, when a RegisterClusterHost command is received for a deregistered Cluster
   # the cluster is restored, the host is added to cluster and if the host is a DC
-  # cluster details are be updated
+  # cluster details are updated
   def execute(
         %Cluster{deregistered_at: deregistered_at, cluster_id: cluster_id},
         %RegisterClusterHost{
--- a/lib/trento/domain/cluster/cluster.ex
+++ b/lib/trento/domain/cluster/cluster.ex
@@ -86,6 +86,7 @@ defmodule Trento.Domain.Cluster do
     ClusterDiscoveredHealthChanged,
     ClusterHealthChanged,
     ClusterRegistered,
+    ClusterRestored,
     ClusterRolledUp,
     ClusterRollUpRequested,
     ClusterTombstoned,
@@ -203,6 +204,48 @@ defmodule Trento.Domain.Cluster do
 
   def execute(%Cluster{cluster_id: nil}, _),
     do: {:error, :cluster_not_registered}
+
+  # Restoration, when a RegisterClusterHost command is received for a deregistered Cluster
+  # the cluster is restored, the host is added to cluster and if the host is a DC
+  # cluster details will be updated
+  def execute(
+        %Cluster{deregistered_at: deregistered_at, cluster_id: cluster_id},
+        %RegisterClusterHost{
+          host_id: host_id,
+          designated_controller: false
+        }
+      )
+      when not is_nil(deregistered_at) do
+    [
+      %ClusterRestored{cluster_id: cluster_id},
+      %HostAddedToCluster{
+        cluster_id: cluster_id,
+        host_id: host_id
+      }
+    ]
+  end
+
+  def execute(
+        %Cluster{deregistered_at: deregistered_at, cluster_id: cluster_id} = cluster,
+        %RegisterClusterHost{
+          host_id: host_id,
+          designated_controller: true
+        } = command
+      )
+      when not is_nil(deregistered_at) do
+    cluster
+    |> Multi.new()
+    |> Multi.execute(fn _ ->
+      %ClusterRestored{cluster_id: cluster_id}
+    end)
+    |> Multi.execute(fn _ ->
+      %HostAddedToCluster{
+        cluster_id: cluster_id,
+        host_id: host_id
+      }
+    end)
+    |> maybe_update_cluster(command)
+  end
 
   def execute(%Cluster{deregistered_at: deregistered_at}, _) when not is_nil(deregistered_at),
     do: {:error, :cluster_not_registered}
@@ -418,6 +461,11 @@ defmodule Trento.Domain.Cluster do
   # Deregistration
   def apply(%Cluster{} = cluster, %ClusterDeregistered{deregistered_at: deregistered_at}) do
     %Cluster{cluster | deregistered_at: deregistered_at}
+  end
+
+  # Restoration
+  def apply(%Cluster{} = cluster, %ClusterRestored{}) do
+    %Cluster{cluster | deregistered_at: nil}
   end
 
   def apply(cluster, %legacy_event{}) when legacy_event in @legacy_events, do: cluster
--- a/lib/trento/domain/cluster/cluster.ex
+++ b/lib/trento/domain/cluster/cluster.ex
@@ -224,18 +224,12 @@ defmodule Trento.Domain.Cluster do
   def execute(
         %Cluster{} = cluster,
         %RegisterClusterHost{
-          designated_controller: true,
-          host_id: host_id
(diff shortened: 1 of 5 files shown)


test/trento/application/projectors/cluster_projector_test.exs
```diff
@@ -206,6 +207,31 @@ defmodule Trento.ClusterProjectorTest do
                      1000
   end
 
+  test "should set deregistered_at field to nil when ClusterRestored is received" do
+    insert(:cluster,
```
arbulu89, line 211: This is a bit strange. What about:
```
%{id: cluster_id, name: name} = insert(:cluster, deregistered_at: DateTime.utc_now())
```

test/trento/application/projectors/cluster_projector_test.exs
```diff
@@ -206,6 +207,31 @@ defmodule Trento.ClusterProjectorTest do
...
+    event = ClusterRestored.new!(%{cluster_id: cluster_id})
+
+    ProjectorTestHelper.project(ClusterProjector, event, "cluster_projector")
+    cluster_projection = Repo.get!(ClusterReadModel, event.cluster_id)
+
+    assert nil == cluster_projection.deregistered_at
+
+    assert_broadcast "cluster_registered",
+                     %{
+                       cib_last_written: nil,
```
arbulu89, line 227: You could add an entry in the `cluster_enrichment`  table for this cluster to see that the enrichment is done properly, and not check just `nil`

test/trento/application/projectors/cluster_projector_test.exs
```diff
@@ -206,6 +207,31 @@ defmodule Trento.ClusterProjectorTest do
...
+                       id: ^cluster_id,
+                       name: ^name,
+                       type: :hana_scale_up
```
arbulu89, line 230: You should get the `type` from the `insert` so you don't need to hardcode this

lib/trento/domain/cluster/cluster.ex
```diff
@@ -204,6 +205,48 @@ defmodule Trento.Domain.Cluster do
   def execute(%Cluster{cluster_id: nil}, _),
     do: {:error, :cluster_not_registered}
 
+  # Restoration, when a RegisterClusterHost command is received for a deregistered Cluster
+  # the cluster is restored, the host is added to cluster and if the host is a DC
+  # cluster details will be updated
```
arbulu89, line 210: `cluster details will be updated` -> `cluster details are updated`

lib/trento/domain/cluster/cluster.ex
```diff
@@ -204,6 +205,48 @@ defmodule Trento.Domain.Cluster do
...
+          designated_controller: true
+        } = command
+      )
+      when not is_nil(deregistered_at) do
+    cluster
+    |> Multi.new()
+    |> Multi.execute(fn _ ->
+      %ClusterRestored{cluster_id: cluster_id}
+    end)
+    |> Multi.execute(fn _ ->
```
arbulu89, line 241: The `HostAddedToCluster` is already considered in the `maybe_update_cluster`.
I think that if you remove this, the tests will even pass the same way

lib/trento/domain/cluster/cluster.ex
```diff
@@ -204,6 +205,48 @@ defmodule Trento.Domain.Cluster do
...
+      %HostAddedToCluster{
```
nelsonkopliku, line 242: Could this event `HostAddedToCluster` be emitted twice?
Asking because later in `maybe_update_cluster` there is also a `maybe_emit_host_added_to_cluster_event`
:thinking:

CDimonaco: Actually no, because the multi is involved, the maybe check if the host is already present, and if yes, does not emit event twice :D, this function is idempotent on both the flows restoration/registration

CDimonaco: You can see the list of emitted events in the test, when we add a dc node, and the host added to cluster event is emitted once in the entire restoring/update flow

lib/trento/domain/cluster/cluster.ex
CDimonaco: Yes absolutely, I left on purpose to be consistent with the non dc restoration, I don't want to hide this is the maybe update, for the purpose of the flow is a no-op no double events emitted, for the reader it's immediate to understand the flow by looking at both clauses, wdyt?

lib/trento/domain/cluster/cluster.ex
arbulu89: @CDimonaco I think the tests would pass without this explicit event as well.
I think the multi in fact is doing nothing, as the `HostRestored` simply changes the `deregistered_at` field, which doesn't affect the `maybe_update_cluster`.
Could you at least give a try?
Remove that event and see if the tests pass

CDimonaco: the test passes of course, I answered in your comment below 👀

arbulu89 (approved):
Looks nice @CDimonaco

lib/trento/domain/cluster/cluster.ex
CDimonaco: I want to explicit say that the host is always added to cluster and not hide the details in the maybe for the reader (the final output is the same) just for the sake of readability

test/trento/application/projectors/cluster_projector_test.exs
CDimonaco: yeah you're right done :D

test/trento/application/projectors/cluster_projector_test.exs
CDimonaco: you're right!

lib/trento/domain/cluster/cluster.ex
arbulu89: Alright, no strong opinions.

lib/trento/domain/cluster/cluster.ex
```diff
@@ -204,6 +205,48 @@ defmodule Trento.Domain.Cluster do
   def execute(%Cluster{cluster_id: nil}, _),
     do: {:error, :cluster_not_registered}
 
+  # Restoration, when a RegisterClusterHost command is received for a deregistered Cluster
+  # the cluster is restored, the host is added to cluster and if the host is a DC
+  # cluster details are be updated
```
nelsonkopliku, line 210: just a typo: `cluster details are updated`

nelsonkopliku (approved):
Cool! I just noticed a couple of typos.

lib/trento/domain/cluster/cluster.ex
nelsonkopliku: Ah-ha! Now I recall: having multi in place makes sure that the state is up to date at each step so at the first `HostAddedToCluster` we'll have the aggregate state updated so it won't be re-added in the following `maybe_update_cluster`

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -777,6 +778,120 @@ defmodule Trento.ClusterTest do
   end
 
   describe "deregistration" do
+    test "should restore a deregistered cluster when a RegisterHost command from a non DC host is received" do
```
rtorrero, line 781: Same here, don't we mean `RegisterClusterHost`?

rtorrero (commented):
Looks good @CDimonaco, just a minor nitpick :+1:

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -777,6 +778,120 @@ defmodule Trento.ClusterTest do
...
+        build(:host_removed_from_cluster_event, cluster_id: cluster_id, host_id: host_two_id),
+        build(:cluster_deregistered_event,
+          cluster_id: cluster_id,
+          deregistered_at: deregistered_at
+        )
+      ]
+
+      new_host_id = UUID.uuid4()
+
+      restoration_event =
```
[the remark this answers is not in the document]
CDimonaco: yup thx

test/trento/domain/cluster/cluster_test.exs
CDimonaco: thx!

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -777,6 +778,120 @@ defmodule Trento.ClusterTest do
...
+            host_id: new_host_id
+          }
+        ],
+        fn cluster ->
+          assert nil == cluster.deregistered_at
+        end
+      )
+    end
+
+    test "should restore a deregistered cluster and perform the cluster update procedure when a RegisterHost command from a DC host is received" do
```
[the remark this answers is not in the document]
CDimonaco: yes thx!

lib/trento/domain/cluster/cluster.ex
CDimonaco: thx

PR #1524, opened 2023-06-14
Add Host Check Selection Domain logic

nelsonkopliku:
# Description

This PR adds a new command `Trento.Domain.Commands.SelectHostChecks`, which allows the selection of host checks within the host aggregate. Upon execution, the command triggers the emission of the `Trento.Domain.Events.HostChecksSelected` event, leading to a change in the state.

We're deferring updating cluster related command/event pair related to check selection
`Trento.Domain.Commands.SelectChecks`/`Trento.Domain.Events.ChecksSelected`.
Not much for the command, but mainly for the event.

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/domain/host/host.ex
+++ b/lib/trento/domain/host/host.ex
@@ -31,6 +31,7 @@ defmodule Trento.Domain.Host do
   alias Trento.Domain.Commands.{
     RegisterHost,
     RollUpHost,
+    SelectHostChecks,
     UpdateHeartbeat,
     UpdateProvider,
     UpdateSlesSubscriptions
@@ -39,6 +40,7 @@ defmodule Trento.Domain.Host do
   alias Trento.Domain.Events.{
     HeartbeatFailed,
     HeartbeatSucceded,
+    HostChecksSelected,
     HostDetailsUpdated,
     HostRegistered,
     HostRolledUp,
@@ -66,6 +68,7 @@ defmodule Trento.Domain.Host do
     field :installation_source, Ecto.Enum, values: [:community, :suse, :unknown]
     field :heartbeat, Ecto.Enum, values: [:passing, :critical, :unknown]
     field :rolling_up, :boolean, default: false
+    field :selected_checks, {:array, :string}, default: []
 
     embeds_many :subscriptions, SlesSubscription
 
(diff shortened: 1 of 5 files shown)


lib/trento/infrastructure/router.ex
```diff
@@ -19,6 +19,7 @@ defmodule Trento.Router do
     RollUpHost,
     RollUpSapSystem,
     SelectChecks,
```
EMaksy, line 21: As we have now a checks selection for hosts and clusters, should we maybe rename SelectChecks to SelectClusterChecks ? :D

nelsonkopliku: Yes, we might want to align naming, however let's defer it.

test/trento/domain/host/host_test.exs
```diff
@@ -154,6 +156,32 @@ defmodule Trento.HostTest do
     end
   end
 
+  describe "host checks selection" do
+    test "should select desired checks for host" do
+      host_id = Faker.UUID.v4()
+      selected_host_checks = Enum.map(0..4, fn _ -> Faker.Cat.name() end)
+
+      assert_events_and_state(
+        [build(:host_registered_event, host_id: host_id)],
```
dottorblaster, line 165: I think you can pass the event without a list, and declare it separately

dottorblaster (changes requested):
One change then we can merge :+1:

EMaksy (approved):
Hey thanks for all  --> LGTM :D

lib/trento/domain/host/host.ex
```diff
@@ -267,6 +270,27 @@ defmodule Trento.Domain.Host do
     }
   end
 
+  def execute(
+        %Host{host_id: nil},
+        %SelectHostChecks{}
+      ) do
+    {:error, :host_not_registered}
```
nelsonkopliku, line 277: I added this clause to keep the domain consistent. 
However, this will be likely deleted with deregistration https://github.com/trento-project/web/blob/deregistration/lib/trento/domain/host/host.ex#L127

arbulu89: Indeed, it will cause conflicts

arbulu89 (approved):

PR #1530, opened 2023-06-15
Update health summary endpoint

arbulu89:
# Description

Update the `/sap_systems/health` endpoint view.
Now, we need to show the database cluster and the application cluster in the home view, in individual columns, so the endpoint now sends them as `application_cluster_id/application_cluster_health` and `database_cluster_id/database_cluster_health`.

The old `clusters_health` and `cluster_id` get deprecated (better to deprecate rather than creating a new version, wdyt?).
This is how it looks like in the swagger view:
image

## How was this tested?

Unit tested

### DIFF
--- a/lib/trento/application/usecases/sap_systems/health_summary_service.ex
+++ b/lib/trento/application/usecases/sap_systems/health_summary_service.ex
@@ -19,8 +19,6 @@ defmodule Trento.SapSystems.HealthSummaryService do
 
   alias Trento.Repo
 
-  @type instance_list :: [DatabaseInstanceReadModel.t() | ApplicationInstanceReadModel.t()]
-
   @spec get_health_summary :: [map()]
   def get_health_summary do
     SapSystemReadModel
@@ -62,7 +60,10 @@ defmodule Trento.SapSystems.HealthSummaryService do
     |> HealthService.compute_aggregated_health()
   end
 
-  @spec compute_cluster_health(instance_list) :: Health.t()
+  @spec compute_cluster_health(
+          [DatabaseInstanceReadModel.t()]
+          | [ApplicationInstanceReadModel.t()]
+        ) :: Health.t()
   defp compute_cluster_health(instances) do
     cluster_id =
       Enum.find_value(instances, nil, fn
@@ -77,7 +78,8 @@ defmodule Trento.SapSystems.HealthSummaryService do
     end
   end
 
-  @spec compute_hosts_health(instance_list) :: Health.t()
+  @spec compute_hosts_health([DatabaseInstanceReadModel.t() | ApplicationInstanceReadModel.t()]) ::
+          Health.t()
   defp compute_hosts_health(instances) do
     instances
     |> Enum.filter(fn %{host: host} -> host end)
--- a/lib/trento/application/usecases/sap_systems/health_summary_service.ex
+++ b/lib/trento/application/usecases/sap_systems/health_summary_service.ex
@@ -5,7 +5,7 @@ defmodule Trento.SapSystems.HealthSummaryService do
 
   import Ecto.Query
 
-  require Trento.Domain.Enums.ClusterType, as: ClusterType
+  require Trento.Domain.Enums.Health, as: HealthEnum
 
   alias Trento.{
     ApplicationInstanceReadModel,
@@ -47,10 +47,8 @@ defmodule Trento.SapSystems.HealthSummaryService do
       sid: sid,
       sapsystem_health: health,
       database_health: compute_database_health(database_instances),
-      application_cluster_health:
-        compute_cluster_health(application_instances, [ClusterType.ascs_ers()]),
-      database_cluster_health:
-        compute_cluster_health(database_instances, [ClusterType.hana_scale_up()]),
+      application_cluster_health: compute_cluster_health(application_instances),
+      database_cluster_health: compute_cluster_health(database_instances),
       hosts_health: compute_hosts_health(all_instances),
       application_instances: application_instances,
       database_instances: database_instances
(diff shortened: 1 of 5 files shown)


lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -60,16 +64,24 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
 
-  @spec compute_clusters_health(instance_list) :: Health.t()
-  defp compute_clusters_health(instances) do
+  @spec compute_cluster_health(instance_list, [ClusterType.t()]) :: Health.t()
+  defp compute_cluster_health(instances, cluster_types) do
     instances
     |> reject_unclustered_instances()
     |> clusters_from_instance()
-    |> keep_only_hana_scale_up_clusters()
+    |> filter_by_cluster_type(cluster_types)
```
arbulu89, line 72: If it was done from scratch, I would left the simple `reject/filter` enum functions here, without hiding them in functions, but just followed what we had.
If you want I can put them here directly. Wdyt?

lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -60,16 +64,24 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
     |> health_from_cluster()
     |> HealthService.compute_aggregated_health()
   end
 
+  @spec reject_unclustered_instances(instance_list) :: instance_list
```
arbulu89, line 77: Just changed some functions order...
They were not following the order of appearance

lib/trento_web/openapi/v1/schema/sap_system.ex
```diff
@@ -93,10 +93,35 @@ defmodule TrentoWeb.OpenApi.V1.Schema.SAPSystem do
...
+          format: :uuid
+        },
+        database_id: %Schema{type: :string, description: "Database ID", format: :uuid},
         sapsystem_health: ResourceHealth,
         database_health: ResourceHealth,
         hosts_health: ResourceHealth,
-        clusters_health: ResourceHealth
+        clusters_heatlh: %Schema{
+          allOf: [
+            ResourceHealth,
```
arbulu89, line 118: From the `open_api_spex` author:
https://github.com/open-api-spex/open_api_spex/issues/388#issuecomment-951455225

lib/trento/application/usecases/sap_systems/health_summary_service.ex
arbulu89: Nevermind, i have changed the code XD

lib/trento_web/views/v1/health_overview_view.ex
```diff
@@ -37,12 +48,21 @@ defmodule TrentoWeb.V1.HealthOverviewView do
   defp extract_database_id([%DatabaseInstanceReadModel{sap_system_id: sap_system_id} | _]),
     do: sap_system_id
 
-  @spec extract_cluster_id([DatabaseInstanceReadModel.t()]) :: String.t()
+  @spec extract_cluster_id([ApplicationInstanceReadModel.t() | DatabaseInstanceReadModel.t()]) ::
```
jamie-suse, line 51: Isn't the spec `@spec extract_cluster_id([ApplicationInstanceReadModel.t() | DatabaseInstanceReadModel.t()]) :: String.t() | nil`?

test/trento/application/usecases/health_summary_service_test.exs
```diff
@@ -48,49 +49,139 @@ defmodule Trento.HealthSummaryServiceTest do
...
       insert(:sap_system, deregistered_at: DateTime.utc_now())
 
-      database_instance =
+      database_instances = [
         insert(
-          :database_instance_without_host,
+          :database_instance,
           sap_system_id: sap_system_id,
+          instance_number: "00",
           sid: "HDD",
```
jamie-suse, line 82: Why are these `sid` fields String literals? In our factories file we seem to use `Faker.StarWars.planet()` or `Faker.UUID.v4()` for `sid`s 🤔

lib/trento/application/usecases/sap_systems/health_summary_service.ex
```diff
@@ -60,44 +62,19 @@ defmodule Trento.SapSystems.HealthSummaryService do
...
-  @spec keep_only_hana_scale_up_clusters([ClusterReadModel.t()]) :: [ClusterReadModel.t()]
-  defp keep_only_hana_scale_up_clusters(clusters) do
-    Enum.filter(clusters, fn
-      %ClusterReadModel{type: ClusterType.hana_scale_up()} -> true
-      _ -> false
-    end)
+  @spec compute_cluster_health(instance_list) :: Health.t()
+  defp compute_cluster_health(instances) do
+    cluster_id =
+      Enum.find_value(instances, nil, fn
```
nelsonkopliku, line 68: If I am getting this right, it means that among a list of either application or database instances, we might or might not find a cluster id meaning the instance is part of a cluster (hana/netweaver) or not.

lib/trento_web/views/v1/health_overview_view.ex
```diff
@@ -37,12 +48,21 @@ defmodule TrentoWeb.V1.HealthOverviewView do
...
+          String.t()
   defp extract_cluster_id([]), do: nil
 
   defp extract_cluster_id([%DatabaseInstanceReadModel{host: %{cluster_id: cluster_id}} | _]),
     do: cluster_id
 
+  defp extract_cluster_id(application_instances) do
+    Enum.find_value(application_instances, nil, fn
```
nelsonkopliku, line 59: Now that I see this I am wondering whether the previous still makes sense :thinking: 
```elixir
defp extract_cluster_id([%DatabaseInstanceReadModel{host: %{cluster_id: cluster_id}} | _]),
    do: cluster_id
```
We are basically getting whatever the cluster id is on the first instance of the list, while in the similar function in the service we are looking in all the instances to find a cluster id, if any.

I am bit confused :sweat_smile:

nelsonkopliku (commented):
Nice job!
I just commented on a couple of type specs and raised a doubt about the extraction of the cluster id from a list of instances.

lib/trento_web/views/v1/health_overview_view.ex
arbulu89: Yes!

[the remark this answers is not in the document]
arbulu89: Again right.
PD: Well, we are pattern matching against an empty list in the first clause, but for populated lists in the nexts

lib/trento/application/usecases/sap_systems/health_summary_service.ex
arbulu89: Yes!
This is to cover unclustered/clustered scenarios.
An unclustered app/db has `unknown` as health

lib/trento_web/views/v1/health_overview_view.ex
nelsonkopliku: Alright, thanks for the refresher :smile:

test/trento/application/usecases/health_summary_service_test.exs
arbulu89: I'm removing those values from the factory usage. Actually, we don't really care what value they get from the factory.
Anyway, the creation of the `sid` in the factories needs some refactor. We already knew that and I think we might already have some tech-debt ticket to track it

nelsonkopliku (approved):
Thanks for the adjustments and for the answers! LGTM

PR #1537, opened 2023-06-16
Add Deregistration modal

jamie-suse:
# Description

Adds the Deregistration Modal that will appear to confirm the user's intention to deregister a host.

## How was this tested?

Added Jest test for component, also added Storybook for component

### DIFF
--- a/assets/js/components/Button/Button.jsx
+++ b/assets/js/components/Button/Button.jsx
@@ -12,12 +12,16 @@ const getSizeClasses = (size) => {
 
 const getButtonClasses = (type) => {
   switch (type) {
+    case 'primary-white-fit':
+      return 'bg-white hover:opacity-75 focus:outline-none text-jungle-green-500 w-fit transition ease-in duration-200 text-center font-semibold rounded shadow';
     case 'primary-white':
       return 'bg-white hover:opacity-75 focus:outline-none text-jungle-green-500 w-full transition ease-in duration-200 text-center font-semibold rounded shadow';
     case 'transparent':
       return 'bg-transparent hover:opacity-75 focus:outline-none w-full transition ease-in duration-200 font-semibold';
     case 'secondary':
-      return 'bg-persimmon hover:opacity-75 focus:outline-none  text-gray-800 w-full transition ease-in duration-200 text-center font-semibold rounded shadow';
+      return 'bg-persimmon hover:opacity-75 focus:outline-none text-gray-800 w-full transition ease-in duration-200 text-center font-semibold rounded shadow';
+    case 'default-fit':
+      return 'bg-jungle-green-500 hover:opacity-75 focus:outline-none text-white w-fit transition ease-in duration-200 text-center font-semibold rounded shadow';
     default:
       return 'bg-jungle-green-500 hover:opacity-75 focus:outline-none text-white w-full transition ease-in duration-200 text-center font-semibold rounded shadow';
   }
--- a/assets/js/components/Button/Button.jsx
+++ b/assets/js/components/Button/Button.jsx
@@ -12,10 +12,10 @@ const getSizeClasses = (size) => {
 
 const getButtonClasses = (type) => {
   switch (type) {
-    case 'primary-white-fit':
-      return 'bg-white hover:opacity-75 focus:outline-none text-jungle-green-500 w-fit transition ease-in duration-200 text-center font-semibold rounded shadow';
     case 'primary-white':
       return 'bg-white hover:opacity-75 focus:outline-none text-jungle-green-500 w-full transition ease-in duration-200 text-center font-semibold rounded shadow';
+    case 'primary-white-fit':
+      return 'bg-white hover:opacity-75 focus:outline-none text-jungle-green-500 w-fit transition ease-in duration-200 text-center font-semibold rounded shadow';
     case 'transparent':
       return 'bg-transparent hover:opacity-75 focus:outline-none w-full transition ease-in duration-200 font-semibold';
     case 'secondary':
--- a/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
+++ b/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
@@ -6,14 +6,14 @@ import Modal from '@components/Modal';
 import Button from '@components/Button';
 
 function DeregistrationModal({
-  hostName,
+  hostname,
   isOpen = false,
   onCleanUp,
   onCancel,
 }) {
   return (
     <Modal
-      title={`Clean up data discovered by agent on host ${hostName}`}
+      title={`Clean up data discovered by agent on host ${hostname}`}
       open={isOpen}
       onClose={onCancel}
     >
--- a/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
+++ b/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
@@ -18,8 +18,9 @@ function DeregistrationModal({
       onClose={onCancel}
     >
       <div className="text-gray-500">
-        This will ignore all the past events collected by the agent instance
-        until this point. This the action is not reversible.
+        This action will cause Trento to stop tracking all the components
+        discovered by the agent in this host, including the host itself and any
+        other component depending on it.
       </div>
       <div className="flex justify-start gap-2 mt-4">
         <Button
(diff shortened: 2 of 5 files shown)


assets/js/components/Button/Button.jsx
```diff
@@ -12,12 +12,16 @@ const getSizeClasses = (size) => {
 
 const getButtonClasses = (type) => {
   switch (type) {
+    case 'primary-white-fit':
```
jamie-suse, line 15: I created this style and `default-fit` as the current button styles use Tailwind class `w-full` which stretches the buttons to take the entire width of their container. While from following the design, I need `w-fit`, so that the width of the button is only as big as it's contents.

There may be a nicer way to achieve this. Thoughts?

jamie-suse: Similarly in the design docs, the `Modal` component's width is also `w-fit` instead of `w-full`, so that it is not stretched to the entire screen width. But currently the `Modal`'s `w-full` class is overriding any classes set from above. Thoughts about this @jagabomb?

image

image

arbulu89, line 15: I would put this after `primary-white`, to keep the "most-used" one first

assets/js/components/DeregistrationModal/DeregistrationModal.jsx
```diff
@@ -0,0 +1,47 @@
...
+  onClose = () => {},
+}) {
+  return (
+    <Modal
+      title={`Clean up data discovered by agent on host ${hostName}`}
+      open={isOpen}
+      onClose={onClose}
+    >
+      <div className="text-gray-500">
+        This will ignore all the past events collected by the agent instance
```
arbulu89, line 21: The text looks wrong.
Can you update it with the latest text proposed by @stefanotorresi ?

assets/js/components/DeregistrationModal/DeregistrationModal.stories.jsx
```diff
@@ -0,0 +1,73 @@
...
+    },
+    onClose: {
+      description: 'Callback function to run when "Cancel" button is clicked',
+      action: 'Cancel',
+      control: false,
+    },
+  },
+};
+
+export function Default({ hostName, onCleanUp, onClose }) {
```
arbulu89, line 39: Let's use the new format for declaring stories:
https://github.com/trento-project/web/blob/main/assets/js/components/DottedPagination/DottedPagination.stories.jsx#L28

https://storybook.js.org/docs/react/writing-stories/stories-for-multiple-components

Otherwise we cannot manage the args

assets/js/components/DeregistrationModal/DeregistrationModal.stories.jsx
```diff
@@ -0,0 +1,73 @@
...
+        {deregistered
+          ? `Host ${hostName} deregistered`
+          : 'Click me to open modal'}
+      </Button>
+
+      <DeregistrationModal
+        hostName={hostName}
+        isOpen={open}
+        onCleanUp={() => {
+          onCleanUp();
```
arbulu89, line 62: What are this specific `onCleanUp` and `onClose` used for?
Shouldn't we simply use the `setDeregistered` and `setOpen`?
Anyway, this will change once you adapt to the new format

assets/js/components/DeregistrationModal/DeregistrationModal.test.jsx
```diff
@@ -0,0 +1,26 @@
...
+    render(
+      <DeregistrationModal
+        hostName={hostName}
+        isOpen
+        onCleanUp={() => {}}
+        onClose={() => {}}
+      />
+    );
+
+    expect(await screen.findByText(new RegExp(hostName))).toBeTruthy();
```
arbulu89, line 20: This `RegExp` looks out of place. Cannot you simply find the hostname?
There must be tons of examples in our codebase for that

arbulu89 (changes requested):
Good job!
Let's change some few little things first

assets/js/components/DeregistrationModal/DeregistrationModal.jsx
```diff
@@ -0,0 +1,47 @@
...
+import { EOS_CLEANING_SERVICES } from 'eos-icons-react';
+
+import Modal from '@components/Modal';
+import Button from '@components/Button';
+
+function DeregistrationModal({
+  hostName,
+  isOpen = false,
+  onCleanUp,
+  onClose = () => {},
```
CDimonaco, line 12: I will suggest to not use any default for the onClose, we don't have any default for onCleanUp, I don't see the reason for having the onclose defined with default, this could led to an unexpected flow

CDimonaco: I will also suggest to use onCancel intestead of onClose, we have onCleanUp because the cta is cleanup, we have cancel as cta button, and onClose, move to onCancel for consistency

assets/js/components/DeregistrationModal/DeregistrationModal.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+
+import { EOS_CLEANING_SERVICES } from 'eos-icons-react';
+
+import Modal from '@components/Modal';
+import Button from '@components/Button';
+
+function DeregistrationModal({
+  hostName,
```
CDimonaco, line 9: Let's use hostname as props name 👀

CDimonaco (changes requested):
Just some comments, I agree with @arbulu89 review

assets/js/components/Button/Button.jsx
jagabomb: > Similarly in the design docs, the `Modal` component's width is also `w-fit` instead of `w-full`, so that it is not stretched to the entire screen width. But currently the `Modal`'s `w-full` class is overriding any classes set from above. Thoughts about this @jagabomb?
Hi @jamie-suse, I do prefer the modal that is smaller in width as it wraps the content more tightly.

rtorrero (approved):
Nothing to add to what's already been said. Thanks @jamie-suse !

assets/js/components/DeregistrationModal/DeregistrationModal.test.jsx
jamie-suse: This is just some information for anyone wondering:

I was using the `RegExp` to be able to do a _partial_ match on the Host name, as it is part of a bigger string. If you just provide the string like that, then `testing-library` will only look for strings that do an _exact_ match, which doesn't work in this case.

However I found that with the `TextMatch` functions, you can provide an extra argument: `{ exact: false }` to be able to do the partial text match without a `RegExp`.

https://testing-library.com/docs/queries/about/#textmatch

We only do a similar thing in [StatusPill.test.jsx](https://github.com/trento-project/web/blob/main/assets/js/components/HostDetails/StatusPill.test.jsx#L11)

dottorblaster (approved):
Green light for now but maybe the button needs more flexibility

assets/js/components/Button/Button.jsx
```diff
@@ -14,10 +14,14 @@ const getButtonClasses = (type) => {
   switch (type) {
     case 'primary-white':
       return 'bg-white hover:opacity-75 focus:outline-none text-jungle-green-500 w-full transition ease-in duration-200 text-center font-semibold rounded shadow';
+    case 'primary-white-fit':
+      return 'bg-white hover:opacity-75 focus:outline-none text-jungle-green-500 w-fit transition ease-in duration-200 text-center font-semibold rounded shadow';
```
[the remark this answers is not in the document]
jamie-suse: I think we could create a tech-debt ticket for the `Button` component.

But for the `Modal` component, I need it to be fixed now in order to adhere to the design. Should I do this refactor in a separate PR first?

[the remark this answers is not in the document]
dottorblaster: Absolutely, track that and we can tackle that when we merge the deregistration branch :+1:

assets/js/components/DeregistrationModal/DeregistrationModal.test.jsx
```diff
@@ -0,0 +1,26 @@
...
+    render(
+      <DeregistrationModal
+        hostname={hostname}
+        isOpen
+        onCleanUp={() => {}}
+        onCancel={() => {}}
+      />
+    );
+
+    expect(await screen.findByText(hostname, { exact: false })).toBeTruthy();
```
arbulu89, line 20: Curious that we need async waits here

arbulu89 (approved):
Thank you!

assets/js/components/DeregistrationModal/DeregistrationModal.test.jsx
jamie-suse: Ah that is because the internal `Modal` component uses a transition, so we have to use the async queries to wait for the animation to finish and the elements to appear.

CDimonaco (approved):

PR #1544, opened 2023-06-19
Add select host checks endpoint

nelsonkopliku:
# Description

Exposes a `POST /api/v1/hosts/:id/checks` for host checks selection.

## How was this tested?
Automated test.

### DIFF
--- a/lib/trento_web/controllers/v1/host_controller.ex
+++ b/lib/trento_web/controllers/v1/host_controller.ex
@@ -2,6 +2,8 @@ defmodule TrentoWeb.V1.HostController do
   use TrentoWeb, :controller
   use OpenApiSpex.ControllerSpecs
 
+  alias TrentoWeb.OpenApi.Schema
+
   alias Trento.{
     Heartbeats,
     Hosts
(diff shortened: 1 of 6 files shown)


lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -51,4 +52,33 @@ defmodule TrentoWeb.V1.HostController do
...
+        in: :path,
+        required: true,
+        type: %OpenApiSpex.Schema{type: :string, format: :uuid}
+      ]
+    ],
+    request_body: {"Checks Selection", "application/json", Schema.Checks.ChecksSelectionRequest},
+    responses: [
+      accepted: "The Selection has been successfully collected",
+      not_found: Schema.NotFound.response(),
+      bad_request: Schema.BadRequest.response(),
```
nelsonkopliku, line 71: I have a doubt on this. I am tempted to remove it. I couldn't make it respond a 400.

arbulu89 (approved):
Nice!
PD: We will have fun with the conflicts at some point hehe

lib/trento_web/controllers/v1/host_controller.ex
nelsonkopliku: I took advantage by cleaning up also other controllers defining an unnecessary bad_request :see_no_evil: 

> I guess the command dispatch command returns {:error, :host_not_registered}, right?
>
yes if the host is not registered yet. It might also return `{:error, :host_rolling_up}` if the aggregate is in the rollup process.

test/trento/application/usecases/hosts_test.exs
```diff
@@ -25,4 +29,42 @@ defmodule Trento.HostsTest do
       assert 6 = Hosts.get_all_sles_subscriptions()
     end
   end
+
+  describe "Check Selection" do
+    test "should dispatch command on Check Selection" do
+      host_id = Faker.UUID.v4()
+      selected_checks = Enum.map(0..4, fn _ -> Faker.Cat.name() end)
```
dottorblaster, line 36: Maybe we want to use UUIDs here (file under the usual topic: not mandatorily a uuid, but still for the sake of uniqueness)

dottorblaster (approved):
Just a nitpick but good job overall :+1:

test/trento/application/usecases/hosts_test.exs
nelsonkopliku: Uniqueness here is not strictly necessary, however I changed in favor of an uuid.
Hope it does not become _uuid all the things_ :sweat_smile:

dottorblaster: Oh well for tests I tend to prefer them because there's really no risk of collisions :sweat_smile:

PR #1549, opened 2023-06-20
Enforce expectation names uniqueness in CheckResultOutline test

nelsonkopliku:
# Description

This should remove flakiness for the specific test changed.

### DIFF
--- a/assets/js/components/ExecutionResults/CheckResultOutline.test.jsx
+++ b/assets/js/components/ExecutionResults/CheckResultOutline.test.jsx
@@ -43,11 +43,12 @@ describe('CheckResultOutline Component', () => {
     const checkID = faker.datatype.uuid();
     const clusterName = faker.lorem.word();
 
-    const expectationName1 = faker.company.name();
-    const expectationName2 = faker.color.human();
-    const expectationName3 = faker.color.human();
-    const expectSameExpectationName1 = faker.lorem.word();
-    const expectSameExpectationName2 = faker.lorem.word();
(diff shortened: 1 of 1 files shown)


EMaksy (approved):
Yeah it's a lot more consistent this way :D
LGTM :D

dottorblaster (approved):
LGTM but since those are strings, could we just use uuids?

assets/js/components/ExecutionResults/CheckResultOutline.test.jsx
```diff
@@ -43,11 +43,11 @@ describe('CheckResultOutline Component', () => {
     const checkID = faker.datatype.uuid();
     const clusterName = faker.lorem.word();
 
-    const expectationName1 = faker.company.name();
-    const expectationName2 = faker.color.human();
-    const expectationName3 = faker.color.human();
-    const expectSameExpectationName1 = faker.lorem.word();
-    const expectSameExpectationName2 = faker.lorem.word();
+    const expectationName1 = faker.datatype.uuid();
```
nelsonkopliku, line 46: @dottorblaster I applied the suggested change about uuids.
 
I am afraid, though, that the _should expectation names be  uuids?_ doubt might rise, which is not the case. 

As long as it clear that we just want strings that are unique in a restricted context and not necessarily uuids, we should be good.

dottorblaster: Yeah I agree, for me it's clear, I don't know if you want to leave a comment or something :sweat_smile:

PR #1551, opened 2023-06-20
Allow rollup on deregistered aggregates

CDimonaco:
# Description

This pr fixes the bug we currently have in aggregates, we reject commands when they are deregistered, but we need to allow the rollup command when deregistered to tombstone the aggregates.

This is caused by #1513 

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -278,14 +278,6 @@ defmodule Trento.Domain.SapSystem do
   end
 
   def execute(
-        %SapSystem{deregistered_at: deregistered_at},
-        _
-      )
-      when not is_nil(deregistered_at) do
-    {:error, :sap_system_not_registered}
-  end
-
-  def execute(
         %SapSystem{sap_system_id: sap_system_id} = snapshot,
         %RollUpSapSystem{}
       ) do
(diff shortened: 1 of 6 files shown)


lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -295,6 +287,14 @@ defmodule Trento.Domain.SapSystem do
     }
   end
 
+  def execute(
```
arbulu89, line 290: So this does nothing after all :laughing: 
We don't have any command afterward.
Anyway, leave it, maybe we need it in the future

test/trento/domain/cluster/cluster_test.exs
```diff
@@ -923,6 +922,15 @@ defmodule Trento.ClusterTest do
         assert match?({:error, :cluster_not_registered}, aggregate_run(initial_events, command)),
                "Command #{inspect(command)} should be rejected by the aggregate"
       end
+
+      commands_to_accept = [
+        %RollUpCluster{}
```
arbulu89, line 927: Here we could include the `RegisterClusterHost` command as well, for completeness

test/trento/domain/host/host_test.exs
```diff
@@ -753,6 +752,15 @@ defmodule Trento.HostTest do
       for command <- commands_to_reject do
         assert_error(initial_events, command, {:error, :host_not_registered})
       end
+
+      commands_to_accept = [
+        %RollUpHost{}
```
arbulu89, line 757: The same. We could include the `RegisterHost`

arbulu89 (approved):
Nice issue hehe
Rejecting the most important command during a deregistration XD
Didn't think about it neither

rtorrero:
Thanks for the quick fix @CDimonaco . I gave it a manual try and the issue seems gone now. +1 on what xabi said about adding the `Register*` commands to the allowed commands lists, after that, good to merge :+1:

nelsonkopliku (approved):
LGTM. Nothing to add.

dottorblaster (approved):
LGTM

PR #1553, opened 2023-06-20
Enforce fact and value name uniqueness

nelsonkopliku:
# Description

Another hopefully useful improvement against flaky tests.
I could observe that some cases are due to duplicated fact or value names coming from the factories.

### DIFF
--- a/assets/js/lib/test-utils/factories/executions.js
+++ b/assets/js/lib/test-utils/factories/executions.js
@@ -12,8 +12,8 @@ const resultEnum = () =>
 const expectationReturnTypeEnum = () =>
   faker.helpers.arrayElement(['expect', 'expect_same']);
 
-export const executionValueFactory = Factory.define(({ sequence }) => ({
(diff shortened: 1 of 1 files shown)


arbulu89 (commented):
Hey!
Do you have there some of the failed tests?
I don't really understand how we can have equal values using the `sequence` option + a randomly generated word

arbulu89 (approved):
@nelsonkopliku wow, it must have been a life time coincidence hehe
The `lorem.word` has [183 options](https://github.com/faker-js/faker/blob/651d1a8cd11b2d562c7a6473b8a544b81d5fbb95/src/locales/en/lorem/words.ts), and just to match even with the sequence. This for 2 different factories.
What the hell.
It is not related to the `rewind` as the value comes from different factories as far as I see.

Up to you, if it was me, I would choose other faker option or add some hardcode word like `fact` or `value` and continue using the `sequence`, it gives it some meaningful information. Even reading the factory.

But I don't want to be picky hehe
Feel free to merge as it is

PR #1560, opened 2023-06-21
Handle linking to unregistered hosts

rtorrero:
Captura desde 2023-06-21 09-50-03

# Description

Check for unregistered hosts to avoid linking and show a warning

### DIFF
--- a/assets/js/components/ClusterDetails/AscsErsClusterDetails.test.jsx
+++ b/assets/js/components/ClusterDetails/AscsErsClusterDetails.test.jsx
@@ -200,4 +200,31 @@ describe('ClusterDetails AscsErsClusterDetails component', () => {
     expect(sidContainer).toHaveTextContent(sid);
     expect(sidContainer.querySelector('a')).toBeNull();
   });
+
+  it('should display a host link for registered hosts', () => {
+    const {
+      name,
+      cib_last_written: cibLastWritten,
+      provider,
+      details,
+    } = clusterFactory.build({ type: 'ascs_ers' });
+
+    const hosts = buildHostsFromAscsErsClusterDetails(details);
+    renderWithRouter(
+      <AscsErsClusterDetails
+        clusterName={name}
+        hosts={hosts}
+        cibLastWritten={cibLastWritten}
+        provider={provider}
+        sapSystems={[]}
+        details={details}
+      />
+    );
+    const registeredHostContainer = screen.getByText(hosts[0].hostname);
+
+    expect(registeredHostContainer).toHaveAttribute(
+      'href',
+      `/hosts/${hosts[0].id}`
+    );
+  });
 });
--- a/assets/js/components/ClusterDetails/AscsErsClusterDetails.test.jsx
+++ b/assets/js/components/ClusterDetails/AscsErsClusterDetails.test.jsx
@@ -104,13 +104,7 @@ describe('ClusterDetails AscsErsClusterDetails component', () => {
 
     nodes.forEach(
       async (
-        {
-          id: hostId,
-          name: nodeName,
-          role,
-          virtual_ip: virtualIp,
-          filesysten,
-        },
+        { id: hostId, name: nodeName, role, virtual_ip: virtualIp, filesysten },
         index
       ) => {
         await waitFor(() => {
@@ -210,7 +204,7 @@ describe('ClusterDetails AscsErsClusterDetails component', () => {
     } = clusterFactory.build({ type: 'ascs_ers' });
 
     const hosts = buildHostsFromAscsErsClusterDetails(details);
-    const unregisteredHost = hosts.pop()
+    const unregisteredHost = hosts.pop();
 
     renderWithRouter(
       <AscsErsClusterDetails
@@ -222,10 +216,10 @@ describe('ClusterDetails AscsErsClusterDetails component', () => {
         details={details}
       />
     );
-    const unregisteredHostContainer = screen.getByText(unregisteredHost.hostname);
-
-    expect(unregisteredHostContainer).not.toHaveAttribute(
-      'href'
+    const unregisteredHostContainer = screen.getByText(
+      unregisteredHost.hostname
     );
+
+    expect(unregisteredHostContainer).not.toHaveAttribute('href');
   });
 });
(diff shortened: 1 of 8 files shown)


arbulu89 (commented):
For the record, knowing that the PR is in draft.
This change will cause massive conflicts on main, as this file is even deleted.
I encourage to base this work on `main` branch.

arbulu89 (commented):
@rtorrero I think we could put this logic in the already existing `HostLink` component, or create a new one, as we need to have this logic in the ascs/ers clusters as well.
What do you think?

rtorrero:
> @rtorrero I think we could put this logic in the already existing `HostLink` component, or create a new one, as we need to have this logic in the ascs/ers clusters as well. What do you think?

I had this conversation a few weeks with @dottorblaster and we went for the current approach to avoid adding artifacts to places where `HostLink` might have been used in a different context.

But your point avoid avoiding duplicating the logic in the ascs/ers clusters is a valid concern. As you suggested in Slack, creating a separate `ClusterNodeLink` component seems the way to go.

assets/js/components/ClusterDetails/AscsErsClusterDetails.test.jsx
```diff
@@ -200,4 +200,31 @@ describe('ClusterDetails AscsErsClusterDetails component', () => {
     expect(sidContainer).toHaveTextContent(sid);
     expect(sidContainer.querySelector('a')).toBeNull();
   });
+
+  it('should display a host link for registered hosts', () => {
```
arbulu89, line 204: This is already tested in `should show nodes information`.
It has a typo though.
In line 108 we should jave `id: hostID` instead of `id: clusterID`.
If you want to have an alternative test, we should test that when the host is not there, we don't have a link

assets/js/components/ClusterDetails/HanaClusterDetails.stories.jsx
```diff
@@ -80,6 +80,13 @@ export const Hana = {
   },
 };
 
+export const WithUnregisteredHost = {
+  args: {
+    ...Hana.args,
+    hosts: [hosts.pop()],
```
arbulu89, line 86: I would use the `slice` even here

assets/js/components/ClusterDetails/HanaClusterDetails.test.jsx
```diff
@@ -181,4 +182,51 @@ describe('HanaClusterDetails component', () => {
     expect(sidContainer).toHaveTextContent(sid);
     expect(sidContainer.querySelector('a')).toBeNull();
   });
+
+  it('should display a host link in the site details if the host is registered', () => {
+    const registeredClusterNode = hanaClusterDetailsNodesFactory.build({
+      name: 'registeredhost',
```
arbulu89, line 188: Do we need this fixes name?
The factory will already create a name. `registeredClusterNode.name`

assets/js/components/ClusterDetails/ClusterNodeLink.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { EOS_WARNING_OUTLINED } from 'eos-icons-react';
+
+import Tooltip from '@components/Tooltip';
+import HostLink from '@components/HostLink';
+
+function ClusterNodeLink({ name, hostId }) {
```
arbulu89, line 7: I would use the same strategy as `HostLink` and pass the name as `children`,
so you can use `<ClusterNodeLink hostId={id}>name</ClusterNodeLink>`

assets/js/components/ClusterDetails/ClusterNodeLink.test.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { screen, render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { renderWithRouterMatch } from '@lib/test-utils';
+import ClusterNodeLink from '@components/ClusterDetails/ClusterNodeLink';
+
+describe('ClusterNodeLink', () => {
+  it('renders HostLink when hostId is provided', () => {
+    const hostId = '12345678-abcd-1234-abcd-1234567890ab';
```
arbulu89, line 9: Let's use `HostFactory`

assets/js/components/ClusterDetails/ClusterNodeLink.test.jsx
```diff
@@ -0,0 +1,32 @@
...
+    });
+
+    const hostLinkElement = screen.getByRole('link', { name });
+
+    expect(hostLinkElement).toBeInTheDocument();
+    expect(hostLinkElement).toHaveAttribute('href', `/hosts/${hostId}`);
+  });
+
+  it('renders warning span when hostId is not provided', () => {
+    const name = 'host02';
```
arbulu89, line 25: Factory

assets/js/components/ClusterDetails/ClusterNodeLink.test.jsx
```diff
@@ -0,0 +1,32 @@
...
+import ClusterNodeLink from '@components/ClusterDetails/ClusterNodeLink';
+
+describe('ClusterNodeLink', () => {
+  it('renders HostLink when hostId is provided', () => {
+    const hostId = '12345678-abcd-1234-abcd-1234567890ab';
+    const name = 'host01';
+    const path = '/hosts/:hostId';
+
+    renderWithRouterMatch(<ClusterNodeLink hostId={hostId} name={name} />, {
+      path,
```
arbulu89, line 14: Do you need this `path` and `route` thing?
You are not clicking the url.
I think `renderWithRouter` would be enough

arbulu89 (changes requested):
Looking good!
Some minor changes and we are ready to merge
@jagabomb could you have a look on how it looks?
Changes available in PR storybook. Stories `AscsErsClusterDetails` and `HanaClusterDetails`

arbulu89 (approved):
Much better :smiley: 
Feel free to merge once you have @jagabomb 's approval

CDimonaco (approved):
Good job LGTM when @jagabomb approves

jagabomb (approved):
LGTM. I would like to also double check this with @abravosuse  when he is back. But so far this solution makes sense to me.

nelsonkopliku (approved):

abravosuse (approved):
I have reviewed AscsErsClusterDetails With Unregistered Host and HanaClusterDetails With Unregistered Host in the Storybook and they both look good to me.

PR #1561, opened 2023-06-21
Add ensa version to sap systems overview entries

arbulu89:
# Description

Last episode in the `ensa version` series!
At the field to the SAP systems overview page.

image
**The screenshot is not taken directly using `main` branch. It has the `deregistration` branch rebased.**

I didn't add storybook stories this time, as the components have multiple selectors usage in different levels.
It could be fixed, but I thought that this was not the correct time.
I have played a bit using a mocked redux store in storybook, but even though it works, it is pretty horrible, as it takes away all of the options to customize the incoming arguments

## How was this tested?

I have added some tests that are not related to this addition, as the view was almost untested.

### DIFF
--- a/assets/js/components/SapSystemsOverview/SapSystemsOverview.test.jsx
+++ b/assets/js/components/SapSystemsOverview/SapSystemsOverview.test.jsx
@@ -2,28 +2,202 @@ import React from 'react';
 import { screen, waitFor } from '@testing-library/react';
 import 'intersection-observer';
 import '@testing-library/jest-dom';
-import { sapSystemFactory } from '@lib/test-utils/factories';
+import {
+  clusterFactory,
+  hostFactory,
+  sapSystemFactory,
+} from '@lib/test-utils/factories';
 import { renderWithRouter, withState } from '@lib/test-utils';
 import { filterTable, clearFilter } from '@components/Table/Table.test';
 
 import SapSystemsOverview from './SapSystemsOverview';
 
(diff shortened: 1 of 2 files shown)


assets/js/components/SapSystemsOverview/SapSystemsOverview.test.jsx
```diff
@@ -2,28 +2,202 @@ import React from 'react';
...
+          applicationInstances,
+          databaseInstances,
+        },
+      };
+
+      const [StatefulSapSystemList] = withState(<SapSystemsOverview />, state);
+
+      renderWithRouter(StatefulSapSystemList);
+
+      const rows = screen.getByRole('table').querySelectorAll('tbody > tr');
```
dottorblaster, line 76: This approach is fine but you could also just do a query for 'ENSA1', 'ENSA2' and count the number of nodes that appear

dottorblaster (approved):
LGTM, just one nitpick

assets/js/components/SapSystemsOverview/SapSystemsOverview.test.jsx
arbulu89: Yes, as well.
I wanted to have a more detailed test, where I check the cell position within the table together with the content

nelsonkopliku (approved):

EMaksy (approved):
Thank you! --> LGTM :D

PR #1567, opened 2023-06-23
Update contracts usage

nelsonkopliku:
# Description

Bumped contracts reference and updated deprecated usages of `new(!)` functions.
Relates to https://github.com/trento-project/contracts/pull/47


CDimonaco (approved):
LGTM

lib/trento/application/integration/checks/checks.ex
```diff
@@ -17,16 +17,15 @@ defmodule Trento.Integration.Checks do
...
-      ExecutionRequested.new!(
-        execution_id: execution_id,
-        group_id: cluster_id,
-        targets:
-          Enum.map(hosts, fn %{host_id: host_id} ->
-            Target.new!(agent_id: host_id, checks: selected_checks)
-          end),
-        env: build_env(env)
-      )
+    execution_requested = %ExecutionRequested{
```
arbulu89, line 20: The same as the other contract PR. Shouldn't we go with `struct` function to have a 1:1 replacement, and most probably apply validation.

mix.exs
```diff
@@ -98,7 +98,7 @@ defmodule Trento.MixProject do
       {:timex, "~> 3.7.7"},
       {:trento_contracts,
        github: "trento-project/contracts",
-       ref: "b8e1036d0177c029ac40d34f237c366a7fb08bec",
+       ref: "53bce57d28b18fd4d73cab30a6e0793b8204139b",
```
arbulu89, line 101: This must be updated once the contracts project PR is merged I guess

arbulu89 (approved):
All clear after off-line talk.
Just use the latest "after-merge" commit number

mix.exs
arbulu89: @nelsonkopliku , at the end you didn't use the merge commit, you used an intermediate commit that it is not in `main` :see_no_evil: 
I think we should've put this instead: https://github.com/trento-project/contracts/commit/cd285f49e0cbdf168ae9aa3d86d92252b3754139

nelsonkopliku: right, my bad :sweat_smile: opening a pr to pin to that commit. pardon eheh

nelsonkopliku: @arbulu89 here you go https://github.com/trento-project/web/pull/1580

PR #1568, opened 2023-06-23
Add missing target_type query param to catalog when loading execution results page

nelsonkopliku:
# Description

Adds a missing `target_type=cluster` when loading the execution results page.


EMaksy (approved):
LGTM! :D

PR #1577, opened 2023-06-29
Add ChecksSelection component

dottorblaster:
Just externalized the checks selection into a component of its own. Stories available :metal:

### DIFF
--- a/assets/js/components/ChecksSelection/ChecksSelection.stories.jsx
+++ b/assets/js/components/ChecksSelection/ChecksSelection.stories.jsx
@@ -12,9 +12,18 @@ const catalog = [
   catalogCheckFactory.build({ group: 'Corosync' }),
   catalogCheckFactory.build({ group: 'SBD' }),
   catalogCheckFactory.build({ group: 'SBD' }),
+  catalogCheckFactory.build({ group: 'Miscellaneous' }),
+  catalogCheckFactory.build({ group: 'Miscellaneous' }),
 ];
 
-const resourceID = faker.datatype.uuid();
+const selectedChecks = [
+  catalog[0].id,
+  catalog[1].id,
+  catalog[5].id,
+  catalog[6].id,
+];
+
+const targetID = faker.datatype.uuid();
 
 export default {
   title: 'ChecksSelection',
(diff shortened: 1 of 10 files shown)


dottorblaster:
@nelsonkopliku @EMaksy I will be away for a couple weeks, so I'll leave some notes on this to make the merge and the iteration (if any) easiest as it can be.

- The logic that displays the selected checks has been simplified a bit
- You can use the onClear handler to clear any redux state that holds any saga result. It gets called every time a check gets selected, and on mount
- The component is completely drop-in: remaining stuff is using this to build the actual hosts check result page and dropping this into the old clusters checks selection page, making the same component declared locally into that directory (`ClusterDetails`) useless (and the directory itself slimmer).

Right now I don't have any other remark in my mind but feel free to ask!

assets/js/components/ChecksSelection/ChecksSelection.stories.jsx
```diff
@@ -0,0 +1,115 @@
...
+    onSave: { action: 'Save' },
+    onClear: {
+      action: 'Clear',
+      description:
+        'Gets called on mount and when checks are selected. It can be used to clear any external state.',
+    },
+  },
+};
+
+export const Default = {
```
arbulu89, line 86: We could maybe have an entry with already selected checks

assets/js/components/ChecksSelection/ChecksSelection.stories.jsx
```diff
@@ -0,0 +1,115 @@
...
+  args: {
+    catalog,
+    resourceID,
+  },
+};
+
+export const Loading = {
+  args: {
+    catalog,
```
arbulu89, line 95: You could have to use the default values:
```
args: {
   ...Default,
   loading: true,
}
```

arbulu89 (approved):
Fantastic job man!
Eager to see how it will be applied in the cluster and host views :slightly_smiling_face:

nelsonkopliku:
@arbulu89 addressed the feedbacks and took the opportunity to align naming from `resource` to `target` in 27e715ae5a9fcb39b5c08d204a5ba56d97cd91af

assets/js/components/ChecksSelection/ChecksSelection.stories.jsx
arbulu89: In fact, i liked the default empty. I would've added a new story :see_no_evil:

arbulu89 (approved):
Thank you @nelsonkopliku 
:heart: 
PD: Still added a comment XD, but no need of a new review, the code looks great

PR #1570, opened 2023-06-26
Project broadcast application instance moved

fabriziosestito:
Add projection and channel broadcast of the `ApplicationInstanceMoved` event.

PR #1579, opened 2023-06-30
ApplicationInstanceMoved FE side-effects

fabriziosestito:
Continuation of: #1570 

Adds FE side effects triggered by the instance moved event.

### DIFF
--- a/assets/js/state/sagas/sapSystems.js
+++ b/assets/js/state/sagas/sapSystems.js
@@ -3,6 +3,7 @@ import {
   SAP_SYSTEM_REGISTERED,
   SAP_SYSTEM_HEALTH_CHANGED,
   APPLICATION_INSTANCE_REGISTERED,
+  APPLICATION_INSTANCE_MOVED,
   APPLICATION_INSTANCE_HEALTH_CHANGED,
   APPLICATION_INSTANCE_DEREGISTERED,
   SAP_SYSTEM_DEREGISTERED,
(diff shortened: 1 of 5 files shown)


assets/js/state/sagas/sapSystems.js
```diff
@@ -64,6 +66,16 @@ function* applicationInstanceRegistered({ payload }) {
   );
 }
 
+export function* applicationInstanceMoved({ payload }) {
+  yield put(updateApplicationInstanceHost(payload));
+  yield put(
+    notify({
+      text: `The application instance ${payload.instance_number} has been moved.`,
```
arbulu89, line 73: Could we have `The application instance ${instance_number} in ${sid} has been moved`

assets/js/state/sapSystems.js
```diff
@@ -84,6 +84,20 @@ export const sapSystemsListSlice = createSlice({
         return sapSystem;
       });
     },
+    updateApplicationInstanceHost: (state, action) => {
```
arbulu89, line 87: I think we can destruct here: `(state, { payload: {sap_system__id, old_host_id, _instance_number }})`.
We do it in some places and not in others though :sweat_smile:

arbulu89 (approved):
Great!
Just some nitpicks from my side. Ready to merge

assets/js/state/sapSystems.js
fabriziosestito: I think i will destructure only the payload since this way is cleared that we are comparing payload with an instance.

arbulu89 (approved):

CDimonaco (approved):

PR #1581, opened 2023-07-03
Add error message to Deregistration Modal

jamie-suse:
# Description

Adds an error message to Deregistration Modal. This will be displayed to the user when an error occurs after requesting to deregister a Host by clicking the "Clean up" button.

## How was this tested?

Added unit tests, updated Storybook.

### DIFF
--- a/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
+++ b/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
@@ -8,6 +8,7 @@ import Button from '@components/Button';
 function DeregistrationModal({
   hostname,
   isOpen = false,
+  isError = false,
   onCleanUp,
   onCancel,
 }) {
@@ -22,6 +23,14 @@ function DeregistrationModal({
         discovered by the agent in this host, including the host itself and any
         other component depending on it.
       </div>
+      {isError && (
+        <div
+          className="bg-red-100 border border-red-400 text-red-700 px-4 py-3 mt-3 rounded relative"
+          role="alert"
+        >
+          Error occurred when requesting deregistration of host
+        </div>
+      )}
       <div className="flex justify-start gap-2 mt-4">
         <Button
           type="default-fit"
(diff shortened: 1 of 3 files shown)


assets/js/components/DeregistrationModal/DeregistrationModal.jsx
```diff
@@ -22,6 +23,14 @@ function DeregistrationModal({
         discovered by the agent in this host, including the host itself and any
         other component depending on it.
       </div>
+      {isError && (
+        <div
+          className="bg-red-100 border border-red-400 text-red-700 px-4 py-3 mt-3 rounded relative"
+          role="alert"
+        >
+          Error occurred when requesting deregistration of host
```
CDimonaco, line 31: Please extract this message to a constant ad use as default message for another prop called errorMessage, so we could have this granularity, and if we have to provide a more appropriate message for a specific case we could do it.

CDimonaco (changes requested):
Looks great!

Only some nitpicks you can find in the comments.

tldr; add another optional props errorMessage with what tmessage as default so we can have error prop set and show a error message and we can override it.

Could you add some screenshots/gifs?

arbulu89 (commented):
I thought we were going to use the toast :sweat_smile: 
@jagabomb At the end we show the error in the modal itself?
Because in this case, what happens on an error message? The `clean up` and `cancel` are still there? The user can click `clean up` many times as he wants on errors?
Maybe we would need a spinner or something, because you need to have the modal there until you receive the answer...
There are many things

CDimonaco:
@arbulu89 I suggested to add the alert inside the modal, so we can have the error inside the modal component and handle it.

This is a controlled component, so when the error is set the user cannot click cleanup anymore can only close the modal.

As a controlled component is responsible to the outside component controlling the lifecycle, for example setting the error prop to null or false when the modal is reopened, because that modal will be only one instance for the pages as you can see in other pr.

For me is a compromise, and with the cleanup button disabled when error is present we can avoid multiple trigger scenario, we can define a more fine behaviour in the full pr for frontend deregistration, leaving the modal as dummy as possible, Wdyt?

CDimonaco:
You're right, let's go to draft untile we clarify the flow

PR #1586, opened 2023-07-04
Fix ApplicationInstanceMoved notifications

fabriziosestito:
Fixes the `undefined` notification text by adding the sid to the ApplicationInstanceMoved broadcast.

### DIFF
--- a/lib/trento/application/projectors/sap_system_projector.ex
+++ b/lib/trento/application/projectors/sap_system_projector.ex
@@ -267,7 +267,9 @@ defmodule Trento.SapSystemProjector do
           new_host_id: new_host_id
         },
         _,
-        _
+        %{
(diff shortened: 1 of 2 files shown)


lib/trento/application/projectors/sap_system_projector.ex
```diff
@@ -267,7 +267,9 @@ defmodule Trento.SapSystemProjector do
           new_host_id: new_host_id
         },
         _,
-        _
+        %{
+          application_instance: %ApplicationInstanceReadModel{sid: sid}
```
arbulu89, line 271: Nice! without a db call, as it should be hehe

arbulu89 (approved):
Fantastic!
Thank you mate

PR #1587, opened 2023-07-04
Add ability to mark Hosts as `deregisterable`

jamie-suse:
# Description

Adds ability to mark Hosts as `deregisterable`. This will be used to display a button in the _Hosts List_ and _Host Details_ pages to allow the user to deregister a Host.

For this PR, the `deregisterable` field is updated on a Host during the initial data fetch and when the `Heartbeat Succeeded` or `Heartbeat Failed` events are received for a given Host.

## How was this tested?

**Unit tests still need to be added!!**


assets/js/state/hosts.js
```diff
@@ -71,6 +87,8 @@ export const hostsListSlice = createSlice({
   },
 });
 
+export const DEREGISTER_HOST = 'DEREGISTER_HOST';
```
arbulu89, line 90: We need a new name for this, the most difficult thing as always XD hehe
`CHECK_HOST_IS_DEREGISTARABLE` maybe?
And the next would be:
`CANCEL_CHECK_HOST_IS_DEREGISTARABLE_ACTION` or similar.

Suggest alternatives please XD

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,54 @@
...
+      .map((host) =>
+        put({
+          type: DEREGISTER_HOST,
+          payload: host,
+        })
+      )
+  );
+}
+
+function* deregisterHostDetail(payload) {
```
arbulu89, line 32: `hostDeregisterable` maybe better?

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,54 @@
...
+  // eslint-disable-next-line no-undef
+  yield delay(config.deregistrationDebounce ?? 0);
```
arbulu89, line 34: Let's pass this `config` value as argument. The purest, the better.
We could get the value in the main sagas index file.
We will need to pass it through multiple functions, but it would be much better, to make things testable either

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,54 @@
...
+  yield race({
+    response: call(deregisterHostDetail, payload),
+    cancel: take(
+      ({ type, payload: cancelPayload }) =>
+        type === CANCEL_DEREGISTER_HOST && cancelPayload.id === payload.id
+    ),
+  });
+}
+
+export function* watchDeregisterHost() {
```
arbulu89, line 48: `watchHostDeregisterable`

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,54 @@
...
+  yield delay(config.deregistrationDebounce ?? 0);
+  yield put(setHostDeregisterable(payload));
+}
+
+function* deregisterHost({ payload }) {
+  yield race({
+    response: call(deregisterHostDetail, payload),
+    cancel: take(
+      ({ type, payload: cancelPayload }) =>
+        type === CANCEL_DEREGISTER_HOST && cancelPayload.id === payload.id
```
arbulu89, line 43: If we extract this to an actual function, we could test it, and skip the `race` testing as it is difficult. But testing this `matching` function looks worthy

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,54 @@
...
+  );
+}
+
+function* deregisterHostDetail(payload) {
+  // eslint-disable-next-line no-undef
+  yield delay(config.deregistrationDebounce ?? 0);
+  yield put(setHostDeregisterable(payload));
+}
+
+function* deregisterHost({ payload }) {
```
arbulu89, line 38: `checkHostIsDeregisterable` maybe?

assets/js/state/sagas/index.js
```diff
@@ -183,6 +192,8 @@ function* heartbeatSucceded({ payload }) {
       icon: '❤️',
     })
   );
+  yield put(setHostNotDeregisterable(payload));
```
arbulu89, line 195: I would leave the `notify` action last

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,54 @@
...
+  setHostDeregisterable,
+} from '@state/hosts';
 import { notify } from '@state/actions/notifications';
 
+export function* markDeregisterableHosts(hosts) {
+  yield all(
+    hosts
+      .filter(({ heartbeat }) => heartbeat !== 'passing')
+      .map((host) =>
+        put({
```
arbulu89, line 24: Let's put this as a function, so we don't need to copy it.
Something like:
```
export const checkHostIsDeregistarable = (host) => ({
          type: CHECK_HOST_IS_DEREGISTARABLE,
          payload: host,
})
```
In the reducer file. Until now, we have some of these in the actions files, but we agreed on not using them anymore.

arbulu89 (commented):
@jamie-suse Looking good.
2 major things:
- Naming XD. I have proposed some ideas. At the end, we are checking if the hosts are deregisterable or not, so let's try to name accordingly
- Let's use `action` functions instead of plain `type, payload` objects. We started doing so in the `actions` folder, bu now we will put these functions in the reducers themselves

assets/js/state/hosts.test.js
```diff
@@ -1,7 +1,39 @@
...
+
+    const expectedState = {
+      hosts: [{ ...host1, deregisterable: true }, host2],
+    };
+
+    expect(hostsReducer(initialState, action)).toEqual(expectedState);
+  });
+
+  it('should mark host as not deregisterable', () => {
+    const [host1, host2] = hostFactory.buildList(2);
```
arbulu89, line 23: Maybe this looks better:
```
const host1 = hostFactory.build({ deregisterable: true })
const host2 = hostFactory.build()
const initialState = { hosts: [host1, host2] };
```

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,56 @@
...
+    type === CANCEL_CHECK_HOST_IS_DEREGISTERABLE && hostId === payload.id;
+
+function* checkHostIsDeregisterable({ payload }) {
+  yield race({
+    response: call(hostDeregisterable, payload),
+    cancel: take(matchHost(payload.id)),
+  });
+}
+
+export function* watchHostDeregisterable() {
```
arbulu89, line 50: Let's put all the `watch*` function at the end all together

arbulu89 (commented):
Looking better @jamie-suse 
Some additional nitpicks

CDimonaco:
Could we have the actions that accept an array of hosts, instead of a single one? This way we avoid spreading a lot of actions with subsequent rerender, in the normal case we have to check multiple host so make sense to accept and array, it's a simple change

For the rest looks very good!

assets/js/state/hosts.test.js
CDimonaco: +1

assets/js/state/sagas/hosts.js
CDimonaco: +1

CDimonaco (commented):
Totally agree with @arbulu89 review, I suggest only the multiple host action version, so we could avoid multiple actions and refresh and also the ui is updated one shot, all the hosts, without having one host updating one after another and the ui can get quite strange to see

Anyway good job, more comments when it's ready to review

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,38 @@
...
+  HOST_DEREGISTERED,
+  removeHost,
+  setHostsDeregisterable,
+} from '@state/hosts';
 import { notify } from '@state/actions/notifications';
 
+export function* markDeregisterableHosts(hosts) {
+  yield put(
+    setHostsDeregisterable(
+      hosts.filter(({ heartbeat }) => heartbeat !== 'passing')
```
jamie-suse, line 14: I am unsure if this logic should be the responsibility of the Sagas, or the Reducer. Thoughts?

assets/js/state/sagas/hosts.test.js
```diff
@@ -1,9 +1,90 @@
...
+      setHostsDeregisterable([criticalHost, unknownHost])
+    );
+  });
+
+  it('should mark a host as deregisterable', async () => {
+    const host = hostFactory.build();
+
+    const dispatched = [];
+
+    await runSaga(
```
jamie-suse, line 39: I've had to use `redux-saga`'s `runSaga()` function in this file because our `recordSaga()` utility function does not allow for multiple parameters to be passed to functions; which is required in order to test `hostDeregisterable()` and `checkHostDeregisterable()`. Thoughts on this?

arbulu89 (commented):
Hey @jamie-suse ,
The prod code looks good to me.
I have some doubts on passing the `debounce` as argument, but i have to confess that i don't have any other alternative now.

I would revisit a bit the tests so they look simpler. Adding loops look an overkill, as they make the test difficult to understand.

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,38 @@
...
+    )
+  );
+}
+
+export function* hostDeregisterable(debounce, host) {
```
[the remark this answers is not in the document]
jamie-suse: Agreed! I was going to propose this in my comment, but I forgot to write it 🤦‍♂️

assets/js/state/sagas/hosts.js
```diff
@@ -1,7 +1,38 @@
...
 
+export function* markDeregisterableHosts(hosts) {
+  yield put(
+    setHostListDeregisterable(
+      hosts.filter(({ heartbeat }) => heartbeat !== 'passing')
+    )
+  );
+}
+
+function* hostDeregisterable({ debounce, host }) {
```
jamie-suse, line 19: I have excluded testing for `hostDeregisterable()` as it is the same as
the test for `checkHostDeregisterable()`, which simply calls it with a
cancellable condition. Thoughts on this?

assets/js/state/sagas/hosts.js
```diff
@@ -12,6 +54,20 @@ export function* hostDeregistered({ payload }) {
   );
 }
 
+function* watchHostDeregisterableDetail() {
+  while (true) {
```
jamie-suse, line 58: I have to use this pattern instead of [takeEvery()](https://redux-saga.js.org/docs/api/#takeeverypattern-saga-args) in order to be able to pass `debounce` into the action's `payload`.

https://redux-saga.js.org/docs/api/#notes-3

arbulu89 (approved):
@jamie-suse Nice!
All good from my side!

assets/js/state/hosts.js
```diff
@@ -59,6 +59,25 @@ export const hostsListSlice = createSlice({
         return host;
       });
     },
+    setHostListDeregisterable: (state, { payload }) => {
+      const ids = payload.map((host) => host.id);
+
+      state.hosts = state.hosts.map((host) => {
+        if (ids.includes(host.id)) {
+          host.deregisterable = true;
```
CDimonaco, line 67: Just a very nitpick comment, it's true that redux toolkit abstract the immutability for us, so we can assign to state properties without worring about immutability (using immer under the hood https://immerjs.github.io/immer/), but, we are mapping on an array, and it's good practice to not reassign the paramters like here, if you want you can change this condition to `return {...host, deregisterable: true}` and the full `{...host}` after.

CDimonaco (approved):
lgtm just the comment if you want take a look :D good job

CDimonaco (approved):

PR #1592, opened 2023-07-05
Remove host aggregate changeset

arbulu89:
# Description

Remove host aggregate `changeset` function. 
Initially, it was added to handle the polymorphic embed, but now this is handled by the `deftype` already.
Having this was causing an issue in the Host rollup, as the snapshot has this `Host` type, and the changeset function was not casting most of the values.


jamie-suse (approved):
+0, -6 😍 nice

CDimonaco (approved):

nelsonkopliku (approved):
:eyes:

PR #1600, opened 2023-07-06
Load selectable checks for a host

nelsonkopliku:
# Description

First of a series of PRs that will bring checks to hosts.

The main thing that this change allows to do is to load the selectable checks for a specific host

image

You can browse to `/hosts/:hostID/settings` and see it loads the catalog for the host.

If running it locally make sure to have at least one check in wanda with `env.target_type == "host"` in the `when` condition, otherwise since there is no host checks currently available the page would be empty (story added)

image

We need to iterate multiple times to get to the final state of the design, so what we're focusing on now is making it work, then make it beautiful. Otherwise the PRs would be huge.

Next: add state/saga for the host check selection and trigger the API call when clicking on Save Check Selection.

## How was this tested?

Automated tests.

### DIFF
--- a/assets/js/components/HostDetails/HostChecksSelection.stories.jsx
+++ b/assets/js/components/HostDetails/HostChecksSelection.stories.jsx
@@ -1,0 +1,80 @@
+import React from 'react';
+import { faker } from '@faker-js/faker';
+import { MemoryRouter } from 'react-router-dom';
+
+import { catalogCheckFactory, hostFactory } from '@lib/test-utils/factories';
+import HostChecksSelection from './HostChecksSelection';
+
+const catalog = [
+  ...catalogCheckFactory.buildList(3, { group: faker.animal.cat() }),
+  ...catalogCheckFactory.buildList(6, { group: faker.animal.dog() }),
+  ...catalogCheckFactory.buildList(2, { group: faker.lorem.word() }),
+];
+
+const selectedChecks = [
+  catalog[0].id,
+  catalog[1].id,
+  catalog[2].id,
+  catalog[5].id,
+  catalog[6].id,
+];
+
+const host = hostFactory.build({ provider: 'azure' });
+
+export default {
+  title: 'HostChecksSelection',
+  component: HostChecksSelection,
+  decorators: [
+    (Story) => (
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
+    ),
+  ],
+  argTypes: {
+    host: {
+      control: 'object',
+      description: 'The host for which to select checks',
+      table: {
+        type: { summary: 'object' },
+      },
+    },
+    catalog: {
+      control: 'object',
+      description: 'Catalog data',
+      table: {
+        type: { summary: 'object' },
+      },
+    },
+    catalogError: {
+      control: 'text',
+      description: 'Error occurred while loading che relevant checks catalog',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: null },
+      },
+    },
+    catalogLoading: {
+      control: { type: 'boolean' },
+      description: 'Whether the catalog is loading',
+      table: {
+        type: { summary: 'string' },
+        defaultValue: { summary: false },
+      },
+    },
+    onUpdateCatalog: {
+      action: 'Update catalog',
+      description: 'Called on mount to load the catalog for the host.',
+    },
+  },
+};
+
+export const Default = {
+  args: {
+    host: { ...host, selected_checks: selectedChecks },
+    catalog,
+    catalogError: null,
+    catalogLoading: false,
+    onUpdateCatalog: () => {},
+  },
+};
--- a/assets/js/components/HostDetails/HostChecksSelection.stories.jsx
+++ b/assets/js/components/HostDetails/HostChecksSelection.stories.jsx
@@ -19,7 +19,10 @@ const selectedChecks = [
   catalog[6].id,
 ];
 
-const host = hostFactory.build({ provider: 'azure' });
+const host = hostFactory.build({
+  provider: 'azure',
+  selected_checks: selectedChecks,
+});
 
 export default {
   title: 'HostChecksSelection',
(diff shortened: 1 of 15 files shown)


assets/js/components/HostDetails/HostInfoBox.jsx
```diff
@@ -0,0 +1,28 @@
...
+        orientation="vertical"
+        data={[
+          {
+            title: 'Provider',
+            content: provider,
+            render: (content) => <ProviderLabel provider={content} />,
+          },
+          { title: 'Agent version', content: agentVersion },
+          {
+            title: '',
```
EMaksy, line 19: Hey any reason to keep empty title and content besides css ? 
Removing it seems not to harm :D

nelsonkopliku: little hack to have it as Jurgen designed it :see_no_evil:

assets/js/components/HostSettings/HostSettings.jsx
```diff
@@ -0,0 +1,64 @@
...
+    loading: catalogLoading,
+  } = useSelector(getCatalog());
+
+  const catalogEnv = {
+    provider: host?.provider,
+    target_type: 'host',
+  };
+
+  if (!host) {
+    return <div>Loading...</div>;
```
EMaksy, line 33: What about using the Loading Spinner component here?

assets/js/components/HostSettings/HostSettings.test.jsx
```diff
@@ -0,0 +1,54 @@
...
+    });
+
+    expect(screen.getByText('Provider')).toBeVisible();
+    expect(screen.getByText('Azure')).toBeVisible();
+    expect(screen.getByText('Agent version')).toBeVisible();
+    expect(screen.getByText(agentVersion)).toBeVisible();
+    expect(screen.getByText(group0)).toBeVisible();
+    expect(screen.getByText(group1)).toBeVisible();
+    expect(screen.getByText(group2)).toBeVisible();
+
```
EMaksy, line 49: What about making the test a bit stronger and adding a click event to check if the created groups also have the checks in them ? :D

nelsonkopliku: I see, however that's already tested in the `assets/js/components/ChecksSelection/ChecksSelectionGroup.test.jsx`.
That should be enough. What do you think?

assets/js/components/HostDetails/HostInfoBox.jsx
arbulu89: @jagabomb Could you have a look into this design?
It doesn't follow the same spacing than the Cluster check selection, where we as well have 2 elements.
I personally wouldn't add this space, and be consistent in what we have in the cluster details view. We could even consider putting the provider in 2nd position

assets/js/components/HostSettings/HostSettings.jsx
```diff
@@ -0,0 +1,65 @@
...
+import LoadingBox from '@components/LoadingBox';
+import { HostInfoBox } from '@components/HostDetails';
+
+import ChecksSelection from '@components/ChecksSelection/ChecksSelection';
+
+import { updateCatalog } from '@state/actions/catalog';
+import { getCatalog } from '@state/selectors/catalog';
+import { getHost } from '@state/selectors';
+
+function HostSettings() {
```
arbulu89, line 16: I would be nice to have a storybook story for this. I think we should aim for that.

arbulu89 (commented):
It looks good in general.
I would ask @jagabomb to visit the design of the check selection view.

Two side notes:
- I would personally like to have a storybook story for the host settings, as it looks pretty easy to have
- Having the `HostSettings` inside `HostDetails` would be more consistent, as we do this way in the clusters details view. Anyway, I don't know if some work is in progress for this, as there are many leftovers from previous `CheckSelection` refactor

assets/js/components/HostSettings/HostSettings.test.jsx
EMaksy: Aye this works!

assets/js/components/HostSettings/HostSettings.jsx
nelsonkopliku: I didn't add it because I believed that the stories we have for the major composing parts could be enough and because we consciously decided not to hook redux with storybook for such _pages_ top level containers. 
In fact, none of the top level pages have stories (except for AboutPage which is different).

Anyhow, if the existing stories are not enough and we want a comprehensive story about the composed page we could add an intermediate component (maybe `HostChecksSelection`? suggestions welcome) that simply gets the host and catalog from the page component, and renders everything, so we could story that one. 
How would that work?

arbulu89: Well, i think many of the top level views have storybook stories, at least almost all of the new ones.
Using the `...Page` component thing. Like the `ExecutionResults` and `ExecutionResultsPage` combo (used really really often in the new components). We don't need to hook redux this way.

In fact, in my personal implementations it even helped to develop faster/better, as i'm not forced to have updated redux states while i'm just changing simple styles, etc

nelsonkopliku:
Thanks @arbulu89 for the review.

> Two side notes:
> - I would personally like to have a storybook story for the host settings, as it looks pretty easy to have
>
answered inline in the comment
> - Having the `HostSettings` inside `HostDetails` would be more consistent, as we do this way in the clusters details view. 
>
I see where this comes from. I did that in light of our ADR about the frontend, however I reckon we're still not in the position to adhere that one. I can move back the `HostSettings` in `HostDetails` as we do for the cluster, however sooner or later that might be refactored.

> Anyway, I don't know if some work is in progress for this, as there are many leftovers from previous `CheckSelection` refactor
>

So, well yes its a WIP and Jurgen would need to review more thoroughly as we get closer to his design, which we'd need to iterate over to get closer :smile:

assets/js/components/HostDetails/HostInfoBox.jsx
arbulu89: Could you at least put a comment saying that this is a temporary hack?

assets/js/components/HostDetails/HostChecksSelection.stories.jsx
```diff
@@ -0,0 +1,80 @@
...
+    catalog: {
+      control: 'object',
+      description: 'Catalog data',
+      table: {
+        type: { summary: 'object' },
+      },
+    },
+    catalogError: {
+      control: 'text',
+      description: 'Error occurred while loading che relevant checks catalog',
```
arbulu89, line 51: Typo, `che` -> `the`

arbulu89 (approved):
Thank you @nelsonkopliku ,
Some nitpicks. Feel free to ignore them XD
Ready to merge in any case

PR #1599, opened 2023-07-06
Add cleanup button component

arbulu89:
# Description

Add `CleanUpButton` component. It will be used in the `HostList` and `HostDetail` views.
It uses a Spinner when the button is in "cleaning" state.
Check on storybook (that width will adjust to the parent component size of course).

PR #1601, opened 2023-07-06
Deregister hosts list frontend

arbulu89:
# Description
Add deregistration button and modal.
It includes a simple saga to handle this.
We need to add the same button in the host details view, with the unique different that when the call is successful, we should navigate to the host list.

deregister

## Blocked by
https://github.com/trento-project/web/pull/1599
https://github.com/trento-project/web/pull/1587

## How was this tested?

Tested

### DIFF
--- a/assets/js/components/HostsList.jsx
+++ b/assets/js/components/HostsList.jsx
@@ -1,10 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import { useSearchParams } from 'react-router-dom';
 import { useSelector, useDispatch } from 'react-redux';
 import { EOS_WARNING_OUTLINED } from 'eos-icons-react';
 
 import Table from '@components/Table';
+import DeregistrationModal from '@components/DeregistrationModal';
 import HealthIcon from '@components/Health/HealthIcon';
 import Tags from '@components/Tags';
 import HostLink from '@components/HostLink';
@@ -16,8 +17,10 @@ import HealthSummary from '@components/HealthSummary/HealthSummary';
 import { getCounters } from '@components/HealthSummary/summarySelection';
 import ProviderLabel from '@components/ProviderLabel';
 import Tooltip from '@components/Tooltip';
+import CleanUpButton from '@components/CleanUpButton';
 
-import { addTagToHost, removeTagFromHost } from '@state/hosts';
+import { addTagToHost, removeTagFromHost, deregisterHost } from '@state/hosts';
+
 import { post, del } from '@lib/network';
 import { agentVersionWarning } from '@lib/agent';
 
@@ -50,6 +53,8 @@ function HostsList() {
   );
 
   const [searchParams, setSearchParams] = useSearchParams();
+  const [cleanUpModalOpen, setCleanUpModalOpen] = useState(false);
+  const [selectedHost, setSelectedHost] = useState(undefined);
 
   const dispatch = useDispatch();
 
@@ -177,6 +182,21 @@ function HostsList() {
           />
         ),
       },
+      {
+        title: '',
+        key: 'deregisterable',
+        className: 'w-48',
+        render: (content, item) =>
+          content && (
+            <CleanUpButton
+              cleaning={item.deregistering}
+              onClick={() => {
+                setSelectedHost(item);
+                setCleanUpModalOpen(true);
+              }}
+            />
+          ),
+      },
     ],
   };
 
@@ -199,6 +219,8 @@ function HostsList() {
       id: host.id,
       tags: (host.tags && host.tags.map((tag) => tag.value)) || [],
       sap_systems: sapSystemList,
+      deregisterable: host.deregisterable,
+      deregistering: host.deregistering,
     };
   });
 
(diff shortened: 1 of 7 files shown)


assets/js/components/HostsList.jsx
```diff
@@ -177,6 +182,21 @@ function HostsList() {
           />
         ),
       },
+      {
+        title: '',
+        key: 'deregisterable',
+        className: 'w-48',
+        render: (content, item) =>
+          content ? (
```
jamie-suse, line 190: What do you think about conditional rendering using `content && <CleanUpButton ... />` here?

assets/js/state/hosts.js
```diff
@@ -59,6 +59,22 @@ export const hostsListSlice = createSlice({
         return host;
       });
     },
+    setHostDeregistering: (state, action) => {
+      state.hosts = state.hosts.map((host) => {
+        if (host.id === action.payload.id) {
+          host.deregistering = true;
```
jamie-suse, line 65: @CDimonaco left a [good comment on my PR](https://github.com/trento-project/web/pull/1587#discussion_r1257914093) about making this reducer function more pure that I think applies here also

assets/js/components/HostsList.test.jsx
```diff
@@ -99,6 +99,97 @@ describe('HostsLists component', () => {
...
+        name: 'Clean up',
+      })[1];
+
+      await user.click(cleanUpModalButton);
+
+      const actions = store.getActions();
+      const expectedActions = [
+        {
+          type: 'DEREGISTER_HOST',
+          payload: expect.objectContaining({
```
jamie-suse, line 183: I guess this is more of a smell than anything, it's a result of the `data` variable in `HostsList.jsx` not simply sending all data from `hosts`. Is there a reason why we do that?

arbulu89: I'm not really sure if I'm following you.
This is a jest option to partially check objects, without the need of comparing all the fields.
In this case, the unique field that I'm interested in the test is the `id`. If I just leave the object itself, the comparison fails, because the passed value has much more fields, that are irrelevant for this usage.
If you want to be purist, i would need to change:
```
        onCleanUp={() => {
          setCleanUpModalOpen(false);
          dispatch(deregisterHost({id: selectedHost.id}));
        }}
```
and only send that value in the payload. But I'm not sure how better this would be honestly, as we are all the time passing the whole host object to the reducers.

jamie-suse: Yeah I guess there is the purist way above, I just mean to ask why we filter out some of the host data into `data` [here](https://github.com/trento-project/web/pull/1601/files#diff-ece88c9cf0057cb7af611ba3ec1cb6e74bd25f22b05668431918c7fb46fd3379L183)?
(As opposed to simply using the raw `hosts` list, instead)

arbulu89: well, the table is not a 1:1 relation with the factory, or the redux state (it doesn't need to be).
That's why there are some divergencies.

assets/js/state/hosts.js
```diff
@@ -79,6 +79,22 @@ export const hostsListSlice = createSlice({
...
     },
+    setHostDeregistering: (state, action) => {
+      state.hosts = state.hosts.map((host) => {
+        if (host.id === action.payload.id) {
+          return { ...host, deregistering: true };
+        }
+        return host;
+      });
+    },
+    setHostNotDeregistering: (state, action) => {
```
nelsonkopliku, line 90: I have a doubt about naming. I would have thought of something that talks about _starting_ a deregistration and _completing_ a deregistration.
No hard feelings tho :smile:

nelsonkopliku (approved):
Cool! LGTM

CDimonaco (approved):
Lgtm, I left some comments, nitpicks but we could work on it

assets/js/state/sagas/hosts.test.js
```diff
@@ -1,18 +1,39 @@
...
+import { networkClient } from '@lib/network';
+import { notify } from '@state/actions/notifications';
 import { hostFactory } from '@lib/test-utils/factories';
 
+const axiosMock = new MockAdapter(networkClient);
+
 describe('Hosts sagas', () => {
+  beforeEach(() => {
+    axiosMock.reset();
+    jest.spyOn(console, 'error').mockImplementation(() => null);
```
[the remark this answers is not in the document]
arbulu89: To avoid error logs in the test output.
We are doing this in other places as well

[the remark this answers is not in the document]
CDimonaco: perfect!

assets/js/components/HostsList.jsx
```diff
@@ -177,6 +182,21 @@ function HostsList() {
...
       },
+      {
+        title: '',
+        key: 'deregisterable',
+        className: 'w-48',
+        render: (content, item) =>
+          content && (
+            <CleanUpButton
+              cleaning={item.deregistering}
+              onClick={() => {
```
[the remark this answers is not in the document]
arbulu89: @CDimonaco This is how it looks like now: https://github.com/trento-project/web/pull/1601/commits/3614c2eca667ae9d058ee6f194943b99c077dae5

Personally, I find it uglier. What do you think?

CDimonaco (approved):

CDimonaco:
lgtm

PR #1602, opened 2023-07-06
Broadcast tags and value objects in restore events

arbulu89:
# Description

Broadcast preloaded objects in restoration events. Otherwise, objects such as `tags` and `sles_subscriptions` are not shown on restoration, and the user is forced to refresh the page to get them.

## How was this tested?

Tested on projections

### DIFF
--- a/lib/trento/application/projectors/host_projector.ex
+++ b/lib/trento/application/projectors/host_projector.ex
@@ -224,12 +224,15 @@ defmodule Trento.HostProjector do
         _,
         _
       ) do
-    host = Repo.get!(HostReadModel, id)
+    host =
+      HostReadModel
+      |> Repo.get!(id)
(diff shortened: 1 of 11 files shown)


CDimonaco (approved):

lib/trento/application/projectors/host_projector.ex
```diff
@@ -224,12 +224,15 @@ defmodule Trento.HostProjector do
...
       ) do
-    host = Repo.get!(HostReadModel, id)
+    host =
+      HostReadModel
+      |> Repo.get!(id)
+      |> Repo.preload([:sles_subscriptions, :tags])
 
     TrentoWeb.Endpoint.broadcast(
       "monitoring:hosts",
       "host_registered",
```
jamie-suse, line 234: Should this event now be `host_restored`?

jamie-suse (approved):

PR #1611, opened 2023-07-10
Only broadcast `HostRemovedFromCluster` when a host is part of a cluster

jamie-suse:
Description
===========

Relates to https://github.com/trento-project/web/pull/1474

> This PR fixes possible concurrency issues when a host is removed from a cluster (delta-deregistration flow).
> The issue is caused by two aggregates acting concurrently, so the HostAddedToCluster event and HostRemovedFromCluster event ordering is not guaranteed.
>
> The solution is to change the projector logic to skip the event if the host is not part of the cluster anymore.
>
> **We still need to conditionally broadcast to the FE the event, but this will come once the FE PR is merged.**

This change will broadcast `HostRemovedFromCluster` _only_ when a given host is part of a cluster.

How was this tested?
--------------------

Added conditions in unit tests to assert for the existence/nonexistence of the broadcast messages when `HostRemovedFromCluster` is called

### DIFF
--- a/lib/trento/application/projectors/host_projector.ex
+++ b/lib/trento/application/projectors/host_projector.ex
@@ -281,7 +281,8 @@ defmodule Trento.HostProjector do
         %{host: %Trento.HostReadModel{cluster_id: nil}}
       ) do
     TrentoWeb.Endpoint.broadcast("monitoring:hosts", "host_details_updated", %{
-      id: host_id
+      id: host_id,
+      cluster_id: nil
     })
   end
 
--- a/lib/trento/application/projectors/host_projector.ex
+++ b/lib/trento/application/projectors/host_projector.ex
@@ -276,6 +276,16 @@ defmodule Trento.HostProjector do
   end
 
   def after_update(
+        %HostRemovedFromCluster{host_id: host_id},
+        _,
+        %{host: %Trento.HostReadModel{cluster_id: nil}}
+      ) do
+    TrentoWeb.Endpoint.broadcast("monitoring:hosts", "host_details_updated", %{
+      id: host_id
+    })
+  end
+
+  def after_update(
         %HostDetailsUpdated{},
         _,
         %{host: host}
--- a/test/trento/application/projectors/host_projector_test.exs
+++ b/test/trento/application/projectors/host_projector_test.exs
@@ -148,13 +148,12 @@ defmodule Trento.HostProjectorTest do
     }
 
     ProjectorTestHelper.project(HostProjector, event, "host_projector")
+    projection = Repo.get!(HostReadModel, host_id)
+    assert nil == projection.cluster_id
 
     assert_broadcast "host_details_updated",
                      %{id: ^host_id},
                      1000
-
-    projection = Repo.get!(HostReadModel, host_id)
-    assert nil == projection.cluster_id
   end
 
   test "should not set the cluster_id to nil if a HostRemovedFromCluster event is received and the host is not part of the cluster anymore" do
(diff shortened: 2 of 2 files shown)


arbulu89 (changes requested):

lib/trento/application/projectors/host_projector.ex
```diff
@@ -275,6 +275,16 @@ defmodule Trento.HostProjector do
     end
   end
 
+  def after_update(
+        %HostRemovedFromCluster{host_id: host_id},
+        _,
+        %{host: %Trento.HostReadModel{cluster_id: nil}}
+      ) do
+    TrentoWeb.Endpoint.broadcast("monitoring:hosts", "host_details_updated", %{
+      id: host_id
```
jamie-suse, line 284: Is this the correct payload to broadcast? `cluster_id` will always be `nil` because of the pattern-matching, is it useful to transmit this?

test/trento/application/projectors/host_projector_test.exs
```diff
@@ -148,8 +148,12 @@ defmodule Trento.HostProjectorTest do
     }
 
     ProjectorTestHelper.project(HostProjector, event, "host_projector")
-    projection = Repo.get!(HostReadModel, host_id)
 
+    assert_broadcast "host_details_updated",
```
arbulu89, line 152: Why did you change the order between the `assert_broadcast` and `Repo.get`?
Every other test has the `assert_broadcast` as last thing in the test

test/trento/application/projectors/host_projector_test.exs
```diff
@@ -172,6 +176,10 @@ defmodule Trento.HostProjectorTest do
     projection = Repo.get!(HostReadModel, host_id)
 
     assert cluster_id == projection.cluster_id
+
+    refute_broadcast "host_details_updated",
+                     %{id: ^host_id},
+                     1000
```
arbulu89, line 182: I'm just a bit worried about this `1000` value.
In the `refute` this blocks the test until the timeout is expired, so it will be waiting for 1 second to have a green test.
Maybe we should use the default `100` value?
@CDimonaco @fabriziosestito Any suggestion on this?

arbulu89 (approved):
All good from my side!
Only commented the `refute_broadcast` timeout value, which slows down or tests

test/trento/application/projectors/host_projector_test.exs
CDimonaco: I think the 1000 value was set by me when I started doing broadcast test, this was set by testing manually different values, if the default timeout is sufficient we can switch to that

CDimonaco (approved):

test/trento/application/projectors/host_projector_test.exs
arbulu89: I mean, in `assert_broadcast` is fine, because it will just wait until the message is sent. And it will be a long test only on failure.
The `refute` will always wait for the time, even in passing tests.
That's the difference

jamie-suse: I reduced this timeout to 100 mS 👍

arbulu89: Just a least request XD
Could you test if putting this same `refute_broadcast` with 100ms in the test above fails? Of course, just temporarily, in your local env.
It should fail. I want to be sure that the 100ms makes sense

jamie-suse: Just did on my local machine and `refute_broadcast` can catch messages using a `timeout` as low as 1 mS 😵 I guess we may see different performance in the CI environment though.

arbulu89:
@jamie-suse Green light from my side!

PR #1616, opened 2023-07-11
Fix application instance deletion after_update

arbulu89:
# Description

Fix the `application_instance` deletion.
With the old code, there was a risk that the `after_update` could try to get a non existing `SapSystemReadModel` (SAP system not registered yet because some App instance was missing). This causes a breaking failure.
Now, the instance deletion data is retrieved before the deletion and passed to the `after_update` in the 3rd argument.


fabriziosestito (approved):

rtorrero (approved):
Nice catch @arbulu89, LGTM

nelsonkopliku (approved):

PR #1618, opened 2023-07-11
Host Checks selection saga

nelsonkopliku:
# Description

This PR adds a saga to enable saving of host checks selection.

Note that:
- the newly added `checksSelection` slice is intentionally similar to the existing `clusterChecksSelection`. 
  As a followup of this work I'd like to converge to just one slice for check selection based on targetType/targetID for both host check selection and cluster check selection
- UX/UI intentionally resembles the current cluster check selection appareance. 
We'll follow up by aligning both host and cluster check selection to proposed design and get rid of some legacy 
- discovered a bug in host projector that emits a host detail with too many null fields, polluting the state when read from websocket (item in the backlog created)

Successful saving
successful-saving-host-checks-selection

Failed saving
failing-saving-host-checks-selection gif

I didn't want to add the bigger refactoring of checks selection saga/state right now. It could have been too much.

## How was this tested?
Automated tests added.

### DIFF
--- a/assets/js/components/ClusterDetails/ChecksSelection.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelection.jsx
@@ -8,7 +8,7 @@ import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
 import { remove, uniq, toggle, groupBy } from '@lib/lists';
 import { getCatalog } from '@state/selectors/catalog';
 import { updateCatalog } from '@state/actions/catalog';
-import { checksSelected } from '@state/actions/cluster';
+import { checksSelected } from '@state/clusters';
 import { executionRequested } from '@state/actions/lastExecutions';
 
 import CatalogContainer from '@components/ChecksCatalog/CatalogContainer';
@@ -149,7 +149,12 @@ function ChecksSelection({ clusterId, cluster }) {
             <button
               className="flex justify-center items-center bg-jungle-green-500 hover:opacity-75 text-white font-bold py-2 px-4 rounded"
               onClick={() =>
-                dispatch(checksSelected(selectedChecks, clusterId))
+                dispatch(
+                  checksSelected({
+                    checks: selectedChecks,
+                    clusterID: clusterId,
+                  })
+                )
               }
               type="button"
               data-testid="save-selection-button"
--- a/assets/js/components/ClusterDetails/ChecksSelection.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelection.jsx
@@ -101,6 +101,14 @@ function ChecksSelection({ clusterId, cluster }) {
     }
   }, [loading]);
 
+  const saveSelection = () =>
+    dispatch(
+      checksSelected({
+        checks: selectedChecks,
+        clusterID: clusterId,
+      })
+    );
+
   return (
     <div className="bg-white rounded p-3">
       <CatalogContainer
@@ -148,14 +156,7 @@ function ChecksSelection({ clusterId, cluster }) {
           <div className="place-items-end flex">
             <button
               className="flex justify-center items-center bg-jungle-green-500 hover:opacity-75 text-white font-bold py-2 px-4 rounded"
-              onClick={() =>
-                dispatch(
-                  checksSelected({
-                    checks: selectedChecks,
-                    clusterID: clusterId,
-                  })
-                )
-              }
+              onClick={saveSelection}
               type="button"
               data-testid="save-selection-button"
             >
--- a/assets/js/components/ClusterDetails/ChecksSelection.jsx
+++ b/assets/js/components/ClusterDetails/ChecksSelection.jsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/no-array-index-key */
 
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 
 import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
@@ -101,13 +101,14 @@ function ChecksSelection({ clusterId, cluster }) {
     }
   }, [loading]);
 
-  const saveSelection = () =>
+  const saveSelection = useCallback(() =>
     dispatch(
       checksSelected({
         checks: selectedChecks,
         clusterID: clusterId,
       })
-    );
+    )
+  );
 
   return (
     <div className="bg-white rounded p-3">
--- a/assets/js/components/HostDetails/HostChecksSelection.jsx
+++ b/assets/js/components/HostDetails/HostChecksSelection.jsx
@@ -32,8 +32,8 @@ function HostChecksSelection({
         catalogError={catalogError}
         loading={catalogLoading}
         selected={selectedChecks}
-        onSave={(selection, targetID, targetName) =>
-          onSaveSelection(selection, targetID, targetName)
+        onSave={(newSelectedChecks, targetID, targetName) =>
+          onSaveSelection(newSelectedChecks, targetID, targetName)
         }
         onUpdateCatalog={() => onUpdateCatalog()}
         onClear={() => {
--- a/assets/js/components/HostDetails/HostChecksSelection.jsx
+++ b/assets/js/components/HostDetails/HostChecksSelection.jsx
@@ -15,6 +15,8 @@ function HostChecksSelection({
   catalogError,
   catalogLoading,
   onUpdateCatalog,
+  onSaveSelection,
+  isSavingSelection,
 }) {
   return (
     <div className="w-full px-2 sm:px-0">
@@ -29,16 +31,12 @@ function HostChecksSelection({
         catalogError={catalogError}
         loading={catalogLoading}
         selected={selectedChecks}
-        onSave={(_selectedChecks, _hostID) => {
-          // TODO: dispatch check selection for a host
-        }}
+        onSave={(selection, targetID) => onSaveSelection(selection, targetID)}
         onUpdateCatalog={() => onUpdateCatalog()}
         onClear={() => {
           // TODO
         }}
-        saving={false}
-        error={null}
-        success={false}
+        saving={isSavingSelection}
       />
     </div>
   );
--- a/assets/js/components/HostDetails/HostChecksSelection.jsx
+++ b/assets/js/components/HostDetails/HostChecksSelection.jsx
@@ -27,11 +27,14 @@ function HostChecksSelection({
       <HostInfoBox provider={provider} agentVersion={agentVersion} />
       <ChecksSelection
         targetID={hostID}
+        targetName={hostName}
         catalog={catalog}
         catalogError={catalogError}
         loading={catalogLoading}
         selected={selectedChecks}
-        onSave={(selection, targetID) => onSaveSelection(selection, targetID)}
+        onSave={(selection, targetID, targetName) =>
+          onSaveSelection(selection, targetID, targetName)
+        }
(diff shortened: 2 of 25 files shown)


assets/js/components/HostDetails/HostSettingsPage.jsx
```diff
@@ -40,7 +47,7 @@ function HostSettingsPage() {
       selectedChecks={selectedChecks}
       catalog={catalog}
       catalogError={catalogError}
-      catalogLoading={catalogLoading}
+      loading={catalogLoading}
```
arbulu89, line 50: I know that you do this to take advantage of the `...rest` thing, but it loses meaning in my opinion

assets/js/components/HostDetails/HostSettingsPage.jsx
```diff
@@ -49,6 +56,19 @@ function HostSettingsPage() {
...
       }
+      saving={saving}
+      error={savingError}
+      success={savingSuccess}
+      onSave={(selection, targetID) =>
+        dispatch(
+          checksSelected({
+            targetID,
+            targetType: 'host',
+            targetName: host.hostname,
```
arbulu89, line 67: Getting this `host` value from outside the function itself looks a bit risky, doesn't it?

assets/js/state/sagas/hosts.js
```diff
@@ -64,6 +73,34 @@ export function* deregisterHost({ payload }) {
   }
 }
 
+export function* checksSelected({ payload }) {
+  const { targetID, targetName } = payload;
```
arbulu89, line 77: why `targetID` in a host related saga, and not `hostID`?
The same for the `targetName`

assets/js/state/hosts.js
```diff
@@ -43,6 +43,14 @@ export const hostsListSlice = createSlice({
         return host;
       });
     },
+    updateSelectedChecks: (state, action) => {
+      state.hosts = state.hosts.map((host) => {
+        if (host.id === action.payload.targetID) {
```
arbulu89, line 48: why `targetID` and not `hostID`?

assets/js/state/hosts.test.js
```diff
@@ -83,4 +85,27 @@ describe('Hosts reducer', () => {
...
+  it('should update the check selection for a host', () => {
+    const initialCheckSelection = [
+      faker.datatype.uuid(),
+      faker.datatype.uuid(),
+    ];
+    const host1 = hostFactory.build({ selected_checks: initialCheckSelection });
+    const host2 = hostFactory.build();
+    const initialState = { hosts: [host1, host2] };
+
+    const newChecksSelection = ['ABABAB', 'CDCDCD'];
```
arbulu89, line 98: Maybe we should use `uuid` here, to match the format with the initial check selection

assets/js/state/sagas/clusters.js
```diff
@@ -1,6 +1,44 @@
-import { put, takeEvery } from 'redux-saga/effects';
-import { CLUSTER_DEREGISTERED, removeCluster } from '@state/clusters';
+import { call, put, select, takeEvery } from 'redux-saga/effects';
+import { post } from '@lib/network';
+
 import { notify } from '@state/actions/notifications';
+import {
+  CLUSTER_DEREGISTERED,
```
arbulu89, line 6: Maybe better to put the constants and function grouped together, in order

assets/js/state/sagas/clusters.js
```diff
@@ -1,6 +1,44 @@
...
+} from '@state/clusters';
+import {
+  setClusterChecksSelectionSavingError,
+  setClusterChecksSelectionSavingSuccess,
+  startSavingClusterChecksSelection,
+  stopSavingClusterChecksSelection,
+} from '@state/clusterChecksSelection';
+import { getClusterName } from '@state/selectors/cluster';
+
+function* checksSelected({ payload }) {
```
arbulu89, line 19: No test for this? I know that it is a migration, but it looks a good opportunity to add tests

arbulu89 (commented):
It looks good in general. I just dropped some comments.
I have some doubts how this will evolve though. What are the "follow up" tasks that you mention in the PR description?

PD: 
Don't take the next comment as a blocker to merge the PR.
I personally find difficult to review such a dense PR (that includes changes in some many places). I might have overlooked some changes. I think we should drive to have more atomic things. For example, a PR with the redux state changes, that in this case, would already unify the check selection for clusters/hosts first. Instead of duplicating now the code, and refactoring later (or followup work as you describe). The need for refactoring new code looks like a smell to me.

Because now, we have some redux state as `checkSelection` which is used only by hosts, but according the name it looks a generic thing.

jagabomb (approved):
LGTM

nelsonkopliku:
Thanks @arbulu89 for the feedback!

> What are the "follow up" tasks that you mention in the PR description?
>

Getting to the state you're mentioning :smile: 
- a revisited redux state and unified checks selection for host and cluster
- adding the navigation from host detail to checks selection
- bringing checks selection and host selection UX/UI to jurgen's proposal
- cleanup the legacy

For the rest I understand it all, it's another way to get to the same place imho.

All in all I am aware of the smells, I just got to make a choice and accept some tradeoffs.
I got to pay a bit the handover when picking up this job and I preferred having a working feature as promised.
This luckily also allows to be more confident in the upcoming required actions.

I didn't think this PR was exceeding the acceptable size, tho :sweat_smile: 
I will take the feedback about more atomic PRs for the future.

assets/js/components/HostDetails/HostSettingsPage.jsx
nelsonkopliku: No strong preference, I thought it could make sense. I can revert that.

assets/js/components/HostDetails/HostSettingsPage.jsx
nelsonkopliku: What do you mean by risky? The fact that a host might not be loaded yet?
If that is the case we should be fine with the early return of a loading box if we don't have a host on line 30

assets/js/state/sagas/clusters.js
nelsonkopliku: My idea was to shortly converge to just one `assets/js/sagas/checksSelection.js` for both clusters and hosts, and basically move there what we have in `assets/js/state/sagas/hosts.js` about `checksSelected` and related tests.

And remove this one

assets/js/state/sagas/hosts.js
nelsonkopliku: because the idea is to extract this as `checksSelection` saga, where givent a targetType, targetName and targetID we can orchestrate checks selection feature.

If it helps I can make this totally host specific for now.

assets/js/state/hosts.js
nelsonkopliku: Because that is what we get from the action payload as it currently is. 

Can change as per this [comment](https://github.com/trento-project/web/pull/1618/files#r1261176463)

assets/js/components/HostDetails/HostSettingsPage.jsx
nelsonkopliku: Correct that is the intended evolution providing targetID, targetName and targetType.

assets/js/components/HostDetails/HostSettingsPage.jsx
nelsonkopliku: removed the ...rest usage

assets/js/components/ClusterDetails/ChecksSelection.jsx
```diff
@@ -101,6 +101,14 @@ function ChecksSelection({ clusterId, cluster }) {
     }
   }, [loading]);
 
+  const saveSelection = () =>
+    dispatch(
+      checksSelected({
```
arbulu89, line 106: Wouldn't be better to pass `selectedChecks` and `clusterId` as the function arguments?
Using global args looks a bit unsafe.
But talking from my ignorance... XD

assets/js/components/HostDetails/HostChecksSelection.jsx
```diff
@@ -29,16 +31,12 @@ function HostChecksSelection({
         catalogError={catalogError}
         loading={catalogLoading}
         selected={selectedChecks}
-        onSave={(_selectedChecks, _hostID) => {
-          // TODO: dispatch check selection for a host
-        }}
+        onSave={(selection, targetID) => onSaveSelection(selection, targetID)}
```
arbulu89, line 34: I would personally continue using the `selectedChecks` term.
Adding `selection` adds yet another term for the same concept.
Or use `selection` generally

assets/js/components/HostDetails/HostSettingsPage.jsx
```diff
@@ -21,16 +23,36 @@ function HostSettingsPage() {
...
 
+  const refreshCatalog = () =>
+    dispatch(
+      updateCatalog({
+        provider: host.provider,
+        target_type: 'host',
+      })
+    );
+
+  const saveSelection = (selection, targetID) =>
```
arbulu89, line 47: I already mentioned, but I think `hostName` should come as arg.
Actually you have it in `HostCheckSelection`, so it would be free.

assets/js/state/index.js
```diff
@@ -23,6 +24,7 @@ export const store = configureStore({
     hostsList: hostsListReducer,
     clustersList: clustersListReducer,
     clusterChecksSelection: clusterChecksSelectionReducer,
+    checksSelection: hostChecksSelectionReducer,
```
arbulu89, line 27: Maybe `hostChecksSelection` better?

arbulu89 (approved):
Ok to merge from my side.
I still have my concerns on how this will evolve and how the different target types checks selection logic will be, but I guess we will see it soon XD

assets/js/components/ClusterDetails/ChecksSelection.jsx
nelsonkopliku: We would need to have different names tho, otherwise it will clash.
And at the end we would be still passing the same variables, with a different name.

I am open to suggestions.

assets/js/components/HostDetails/HostChecksSelection.jsx
nelsonkopliku: again name clashing here.
`selection` here is the possibly different selection emitted when a user clicks on the save selection button, not the `selectedChecks` used to initially feed the component.

Maybe rename it to `newSelectedChecks`/`newChecksSelection`?

assets/js/components/HostDetails/HostSettingsPage.jsx
nelsonkopliku: d89913dfd016ccfd45ce79493562594576c85bd4

assets/js/components/ClusterDetails/ChecksSelection.jsx
arbulu89: I don't really have a suggestion...
Anyway, now that i checked a bit, we should maybe wrap this functions in `useCallback`, otherwise the function is redefined in every render. Maybe a cheap operation, but with the `useCallback` it is avoided.
Maybe @dottorblaster has stronger opinions on this

EMaksy (commented):
Hey man nice job! 
As I'm a bit unfamiliar with Sagas, so  I only left a comment --> The rest LGTM :D
Well done.

assets/js/components/HostDetails/HostChecksSelection.jsx
nelsonkopliku: renamed the function argument to `newSelectedChecks`

PR #1623, opened 2023-07-12
Add 'Clean up' button to Host Details page

jamie-suse:
[CleanUpHostDetails.webm](https://github.com/trento-project/web/assets/113615552/17bd7b50-2159-47d2-8053-c9696e088a51)

# Description

Adds the '_Clean up_' button to the Host Details page. Allows the user to deregister a host.

## How was this tested?

Added frontend unit tests to test the behaviour of the button.

### DIFF
--- a/assets/js/components/HostDetails/HostDetails.jsx
+++ b/assets/js/components/HostDetails/HostDetails.jsx
@@ -43,10 +43,8 @@ function HostDetails() {
   const dispatch = useDispatch();
   const navigate = useNavigate();
 
-  const cleanUpHost = (selectedHost) => {
+  const cleanUpHost = ({ id, hostname }) => {
     setCleanUpModalOpen(false);
-
-    const { id, hostname } = selectedHost;
     dispatch(deregisterHost({ id, hostname, navigate }));
   };
 
--- a/assets/js/components/HostsList.jsx
+++ b/assets/js/components/HostsList.jsx
@@ -64,10 +64,8 @@ function HostsList() {
     setCleanUpModalOpen(true);
   };
 
-  const cleanUpHost = (host) => {
+  const cleanUpHost = ({ id, hostname }) => {
     setCleanUpModalOpen(false);
-
-    const { id, hostname } = host;
     dispatch(deregisterHost({ id, hostname, navigate }));
   };
 


assets/js/components/HostDetails/HostDetails.jsx
```diff
@@ -36,6 +38,18 @@ function HostDetails() {
   const { grafanaPublicUrl } = config;
 
   const [exportersStatus, setExportersStatus] = useState([]);
+  const [cleanUpModalOpen, setCleanUpModalOpen] = useState(false);
+
+  const dispatch = useDispatch();
+
+  const openDeregistrationModal = () => {
```
arbulu89, line 45: I think we can use directly the `setCleanUpModalOpen(true)` , right?
I know that you did to be consistent with the other.
Anyway, up to you, not a big deal hehe

assets/js/components/HostDetails/HostDetails.test.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { screen } from '@testing-library/react';
+import { screen, fireEvent } from '@testing-library/react';
```
arbulu89, line 2: Use `userEvent` instead of `fireEvent`. It is the recommended library.
You will need to setup the user, etc
Check other test files with this approach

assets/js/components/HostDetails/HostDetails.test.jsx
```diff
@@ -66,4 +67,115 @@ describe('HostDetails component', () => {
       )
     ).toBeInTheDocument();
   });
+
+  it("should not display 'Clean up' button when host is not deregisterable", () => {
```
arbulu89, line 71: double quotes? I guess the ci will complain about this hehe
I would remove the single quotes in any case from `Clean up` and put them in the test string

arbulu89 (changes requested):
Looks great!
Only 2 things.
- Could you add a gif on how it works?
- Change `fireEvent` by `userEvent`
We can merge after that XD

assets/js/components/HostDetails/HostDetails.test.jsx
arbulu89, line 71: We could maybe add a `deregistration` describe to group the tests about this topic.
Again, not strong opinions

CDimonaco: +1

assets/js/components/HostDetails/HostDetails.jsx
```diff
@@ -36,6 +38,14 @@ function HostDetails() {
   const { grafanaPublicUrl } = config;
 
   const [exportersStatus, setExportersStatus] = useState([]);
+  const [cleanUpModalOpen, setCleanUpModalOpen] = useState(false);
+
+  const dispatch = useDispatch();
+
+  const cleanUpHost = () => {
+    setCleanUpModalOpen(false);
+    dispatch(deregisterHost(host));
```
arbulu89, line 47: We should get the `host` as argument in the function.
If we use the global variable, we have the risk of having strange things.
For example, if the function was used when `host` is not loaded, things wouldn't work.

arbulu89 (changes requested):
Okey!
Once we have the `navigation` sorted out the code looks good

assets/js/components/HostDetails/HostDetails.jsx
```diff
@@ -36,6 +38,15 @@ function HostDetails() {
...
 
   const [exportersStatus, setExportersStatus] = useState([]);
+  const [cleanUpModalOpen, setCleanUpModalOpen] = useState(false);
+
+  const dispatch = useDispatch();
+  const navigate = useNavigate();
+
+  const cleanUpHost = (selectedHost) => {
+    setCleanUpModalOpen(false);
+    dispatch(deregisterHost({ host: selectedHost, navigate }));
```
arbulu89, line 48: We careful with this!
If you change the payload format for the function, you need to change it every place that it is used, the host list view in this case.
I would suggest to go with `deregisterHost({ ...selectedHost, navigate })`, but if you prefer this way don't forget to change the host list view call as well

arbulu89 (changes requested):

assets/js/components/HostDetails/HostDetails.jsx
arbulu89: Most probably not...

assets/js/state/sagas/hosts.js
```diff
@@ -52,6 +52,7 @@ export function* deregisterHost({ payload }) {
   yield put(setHostDeregistering(payload));
   try {
     yield call(del, `/hosts/${payload.id}`);
+    payload.navigate?.('/hosts');
```
arbulu89, line 55: Never seen this format before.

arbulu89 (approved):

assets/js/state/sagas/hosts.js
dottorblaster, line 55:
```suggestion
    const { navigate } = payload;
    navigate('/hosts');
```

dottorblaster (changes requested):

assets/js/state/sagas/hosts.js
jamie-suse: But if `payload.navigate` doesn't exist and then we try to call `navigate()` we will get `Uncaught TypeError: navigate is not a function`.

With optional chaining if `navigate` doesn't exist, the call to `navigate()` just returns `undefined`, essentially a NOOP.

We need to do this because we only want to navigate back to `HostsList` from `HostDetails`. It doesn't really make sense to do it from `HostsList` because we are already there and it is updated by websocket events. So as a result [we don't pass in `navigate`](https://github.com/trento-project/web/blob/c52a178fe0df9c39910cf31ced48c4141427544d/assets/js/components/HostsList.jsx#L68) to the payload to make this a NOOP.

jamie-suse: Fancy [Optional chaining with function calls](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#optional_chaining_with_function_calls).

Basically, if `payload.navigate` doesn't exist, then when we try to call `payload.navigate()`, we instead just returned `undefined`, acting like a NOOP.

If we didn't do this, then when calling `payload.navigate()`, we would get the error thrown: `Uncaught TypeError: payload.navigate is not a function`

jamie-suse: We need to do this because we only want to navigate back to `HostsList` from `HostDetails`. It doesn't really make sense to do it from `HostsList` because we are already there and it is updated by websocket events. So as a result [we don't pass in `navigate`](https://github.com/trento-project/web/blob/c52a178fe0df9c39910cf31ced48c4141427544d/assets/js/components/HostsList.jsx#L68) to the payload to make this a NOOP.

CDimonaco, line 55: I agree to remove the optional chaining and unify the flows, we need to redirect always even on the host list, it's a no-op for the router, and we avoid to pass or not pass parameters, we don't have types and this cause only confusion imho.

Unify the things, pass the navigate always and remove the optional chaining

CDimonaco (changes requested):

assets/js/state/sagas/hosts.test.js
```diff
@@ -84,15 +84,15 @@ describe('Hosts sagas', () => {
   it('should send host deregister request', async () => {
     const host = hostFactory.build();
 
+    const mockNavigate = (route) => route;
```
jamie-suse, line 87: As we have chosen to always provide a function `navigate()` to the action payload for the `deregisterHost()` Saga, now in the test we need to provide a mocked `navigate` function. Is this function fine?

dottorblaster: You could have just a () => {} but this is fine as well

dottorblaster (approved):

assets/js/components/HostsList.jsx
```diff
@@ -66,7 +66,9 @@ function HostsList() {
 
   const cleanUpHost = (host) => {
     setCleanUpModalOpen(false);
-    dispatch(deregisterHost({ ...host, navigate }));
+
+    const { id, hostname } = host;
```
CDimonaco, line 70: Pls destructure on the function argument level 👁️

CDimonaco (approved):
LGTM

CDimonaco (approved):

PR #1624, opened 2023-07-12
Projectors refactor

CDimonaco:
# Description

This pr refactor the usage of projectors, using the changesets with the current entity as the base entity for the changes.

Removes the queries from `after_update` callbacks, only the necessary association preloads for broadcasts are present.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/application/projectors/cluster_projector.ex
+++ b/lib/trento/application/projectors/cluster_projector.ex
@@ -63,8 +63,10 @@ defmodule Trento.ClusterProjector do
       deregistered_at: deregistered_at
     },
     fn multi ->
+      cluster = Repo.get!(ClusterReadModel, cluster_id)
+
       changeset =
-        ClusterReadModel.changeset(%ClusterReadModel{id: cluster_id}, %{
+        ClusterReadModel.changeset(cluster, %{
           deregistered_at: deregistered_at
         })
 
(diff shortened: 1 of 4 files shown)


lib/trento/application/projectors/cluster_projector.ex
```diff
@@ -63,8 +63,10 @@ defmodule Trento.ClusterProjector do
       deregistered_at: deregistered_at
     },
     fn multi ->
+      cluster = Repo.get!(ClusterReadModel, cluster_id)
```
arbulu89, line 66: Just as a comment, in other `project` functions, we have this pipelined:
```
changeset =
    ClusterReadModel
    |> Repo.get!(cluster_id)
    |> ClusterReadModel.changeset(%{
         deregistered_at: deregistered_at
    })
```
Just in case if you prefer it :stuck_out_tongue_closed_eyes:

nelsonkopliku (approved):
LGTM

arbulu89 (changes requested):
@CDimonaco I still see some `Repo.queries` in the `after_updates` that shouldn't be there:
- Cluster projection, `ClusterDeregistered` event. **This is the unique real change i ask you XD**
- Host projection, `HostAddedToCluster`, you could add the `returning: true` value to the `insert`, so you get the data from the database, and you don't need to query in the `after_update`. Anyway `insert` first `query` later is accepted as a good solution as well.
[Ref](https://hexdocs.pm/ecto/Ecto.Repo.html#c:insert/2-upserts)

arbulu89:
@CDimonaco Besides the review, can you make me a small favor? XD
Could you change the line 286 in `HostProjector` and update `Trento.HostReadModel` by simply `HostReadModel`?

arbulu89 (approved):
:100:

lib/trento/application/projectors/cluster_projector.ex
CDimonaco: You triggered my ocd 👁️ 

Everything refactored to this style

CDimonaco:
Everything done!

arbulu89 (approved):
:100:  :1st_place_medal:

PR #1629, opened 2023-07-13
Add test to check for Clean up button visibility

rtorrero:
# Description
This PR simply adds:
 - e2e test to check if the "Clean up" button in the hosts overview is only displayed after a failed heartbeat
 - e2e test to check if the "Clean up" button displays a confirmation and removes the host after confirmation

### DIFF
--- a/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
+++ b/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
@@ -24,18 +24,14 @@ function DeregistrationModal({
       </div>
       <div className="flex justify-start gap-2 mt-4">
         <Button
+          data-testid="cleanup-confirm"
           type="default-fit"
           className="inline-block mx-0.5 border-green-500 border w-fit"
           size="small"
           onClick={onCleanUp}
         >
           <EOS_CLEANING_SERVICES size="base" className="fill-white inline" />
-          <span
-            id="cleanup-confirm"
-            className="text-white text-sm font-bold pl-1.5"
-          >
-            Clean up
-          </span>
+          <span className="text-white text-sm font-bold pl-1.5">Clean up</span>
         </Button>
         <Button
           type="primary-white-fit"
--- a/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
+++ b/assets/js/components/DeregistrationModal/DeregistrationModal.jsx
@@ -30,7 +30,12 @@ function DeregistrationModal({
           onClick={onCleanUp}
         >
           <EOS_CLEANING_SERVICES size="base" className="fill-white inline" />
-          <span className="text-white text-sm font-bold pl-1.5">Clean up</span>
+          <span
+            id="cleanup-confirm"
+            className="text-white text-sm font-bold pl-1.5"
+          >
+            Clean up
+          </span>
         </Button>
         <Button
           type="primary-white-fit"
--- a/test/e2e/cypress/e2e/hosts_overview.cy.js
+++ b/test/e2e/cypress/e2e/hosts_overview.cy.js
@@ -172,7 +172,7 @@ context('Hosts Overview', () => {
   });
 
   describe('Deregistration', () => {
-    const host = {
+    const hostToDeregister = {
       name: 'vmdrbddev01',
       id: '240f96b1-8d26-53b7-9e99-ffb0f2e735bf',
     };
@@ -181,11 +181,11 @@ context('Hosts Overview', () => {
       before(() => {
         cy.visit('/hosts');
         cy.url().should('include', '/hosts');
-        cy.task('startAgentHeartbeat', [host.id]);
+        cy.task('startAgentHeartbeat', [hostToDeregister.id]);
       });
 
-      it(`should not display a clean-up button for host ${host.name}`, () => {
-        cy.contains(host.name).within(() => {
+      it(`should not display a clean-up button for host ${hostToDeregister.name}`, () => {
+        cy.contains(hostToDeregister.name).within(() => {
           cy.get('td:nth-child(9)').should('not.exist');
         });
       });
@@ -198,7 +198,7 @@ context('Hosts Overview', () => {
         }
       });
 
-      it(`should display the cleanup button for host ${host.name} once heartbeat is lost`, () => {
+      it(`should display the cleanup button for host ${hostToDeregister.name} once heartbeat is lost`, () => {
         cy.task('stopAgentsHeartbeat');
 
         cy.get('tr:nth-child(1) > .w-48 > [data-testid="cleanup-button"]', {
@@ -226,7 +226,7 @@ context('Hosts Overview', () => {
 
         cy.get('[data-testid="cleanup-confirm"]').click();
 
-        cy.contains(host.name).should('not.exist');
+        cy.contains(hostToDeregister.name).should('not.exist');
       });
     });
   });
--- a/test/e2e/cypress/e2e/hosts_overview.cy.js
+++ b/test/e2e/cypress/e2e/hosts_overview.cy.js
@@ -171,7 +171,7 @@ context('Hosts Overview', () => {
     });
   });
 
-  describe('Clean-up', () => {
+  describe('Deregistration', () => {
     describe('Clean-up buttons should be visible only when needed', () => {
       const heartbeatTargets = agents();
       const failingHost = heartbeatTargets.shift();
@@ -180,9 +180,10 @@ context('Hosts Overview', () => {
       });
 
       it('should hide all Clean-up buttons when heartbeat starts except the first one', () => {
-        cy.get(
-          ':nth-child(1) > .w-48 > .bg-white > .text-jungle-green-500'
-        ).should('contain.text', 'Clean up');
+        cy.get('tr:nth-child(1) > td:nth-child(9)').should(
+          'contain.text',
+          'Clean up'
+        );
 
         for (let i = 2; i < 11; i++) {
           cy.get(
@@ -191,7 +192,7 @@ context('Hosts Overview', () => {
         }
       });
 
-      it('should display a confirmation to remove the host after clicking on Clean up', () => {
+      it('should allow to deregister a host after clean up confirmation', () => {
         cy.get(
           ':nth-child(1) > .w-48 > .bg-white > .text-jungle-green-500'
         ).click();
@@ -200,10 +201,8 @@ context('Hosts Overview', () => {
           'contain.text',
           'This action will cause Trento to stop tracking all the components discovered by the agent in this host, including the host itself and any other component depending on it.'
         );
-      });
 
-      it('should remove a host after confirming host cleanup', () => {
-        cy.get('.bg-jungle-green-500').click();
+        cy.get('#cleanup-confirm').click();
         cy.get(`#host-${failingHost}`).should('not.exist');
       });
     });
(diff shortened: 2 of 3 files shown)


test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,42 @@ context('Hosts Overview', () => {
       });
     });
   });
+
+  describe('Clean-up', () => {
```
arbulu89, line 174: Maybe `Deregistration`?

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,42 @@ context('Hosts Overview', () => {
...
+    describe('Clean-up buttons should be visible only when needed', () => {
+      const heartbeatTargets = agents();
+      const failingHost = heartbeatTargets.shift();
+      before(() => {
+        cy.task('startAgentHeartbeat', heartbeatTargets);
+      });
+
+      it('should hide all Clean-up buttons when heartbeat starts except the first one', () => {
+        cy.get(
+          ':nth-child(1) > .w-48 > .bg-white > .text-jungle-green-500'
```
arbulu89, line 184: What about this: `tr:nth-child(1) > td:nth-child(9)` where the 1st `1` is the row number.
The other query selector looks totally confusing

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,42 @@ context('Hosts Overview', () => {
...
+        ).click();
+
+        cy.get('.min-h-screen > .w-full').should(
+          'contain.text',
+          'This action will cause Trento to stop tracking all the components discovered by the agent in this host, including the host itself and any other component depending on it.'
+        );
+      });
+
+      it('should remove a host after confirming host cleanup', () => {
+        cy.get('.bg-jungle-green-500').click();
```
arbulu89, line 206: This test depends in the previous one. The test should be executable by itself, so it should set the host in deregisterable state, open the modal, and everything else.

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,42 @@ context('Hosts Overview', () => {
...
+          ':nth-child(1) > .w-48 > .bg-white > .text-jungle-green-500'
+        ).click();
+
+        cy.get('.min-h-screen > .w-full').should(
+          'contain.text',
+          'This action will cause Trento to stop tracking all the components discovered by the agent in this host, including the host itself and any other component depending on it.'
+        );
+      });
+
+      it('should remove a host after confirming host cleanup', () => {
```
arbulu89, line 205: Cannot we get the button by name of something? The class usage looks weak and really not meaningful

arbulu89 (commented):
@rtorrero I would go with a new file for the deregistration (`deregistration.cy.js`). This topic is so big, and involves more things that a unique file would be better.

Besides this, I miss the fact that we test that the button is there when the heartbeat fails. 
I would prefer to have something like
- Heartbeat sending - button is nothere
- Heartbeat stopped - button shows up
- Heartbeat sending again - button goes away

rtorrero:
Hey @arbulu89, thanks for the quick review :-)

> @rtorrero I would go with a new file for the deregistration (`deregistration.cy.js`). This topic is so big, and involves more things that a unique file would be better.

 I'm not sure about this: up until now, we have organized all the tests by section / view:
 - about
 - checks_catalog
 - clusters_overview
 - ... etc
 
I'd find it confusing to mix these and now introduce splitting by "feature"

Also, organizing the test files by section / view reduces the total run time of the e2e test suite as it minimizes transitioning from one part of trento to a different one (the deregistration / cleanup feature spans through multiple views). We would check the cluster details view for the cluster details specific tests and then we'd need to navigate again to it while running the deregistration tests)

I'd instead use http calls to our backend in the views that are affected by the deregistration (e.g. cluster details) to deregister a single host and, while I'd use the button in those views that have it.

> Besides this, I miss the fact that we test that the button is there when the heartbeat fails. I would prefer to have something like
> 
>     * Heartbeat sending - button is nothere
> 
>     * Heartbeat stopped - button shows up
> 
>     * Heartbeat sending again - button goes away

I was indeed doing this but as you said, depending on the previous test. I'll improve this and to it explicitly as you suggest, being able to run tests independently is indeed important.

assets/js/components/DeregistrationModal/DeregistrationModal.jsx
```diff
@@ -30,7 +30,12 @@ function DeregistrationModal({
           onClick={onCleanUp}
         >
           <EOS_CLEANING_SERVICES size="base" className="fill-white inline" />
-          <span className="text-white text-sm font-bold pl-1.5">Clean up</span>
+          <span
+            id="cleanup-confirm"
```
arbulu89, line 34: This is just for testing?
You should maybe use `data-testid` i guess

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,41 @@ context('Hosts Overview', () => {
...
+
+  describe('Deregistration', () => {
+    describe('Clean-up buttons should be visible only when needed', () => {
+      const heartbeatTargets = agents();
+      const failingHost = heartbeatTargets.shift();
+      before(() => {
+        cy.task('startAgentHeartbeat', heartbeatTargets);
+      });
+
+      it('should hide all Clean-up buttons when heartbeat starts except the first one', () => {
```
arbulu89, line 182: I personally still don't like this testing approach.
I don't really think we should use so many agents. I would prefer to use one/two agent, and test thing more thorougly.
As commented, a thing that I would do is:
1. Send heartbeat for this agent. Check button is not there
2. Stop sending hearbeat. Check the button shows up. This is the most important test. I know that it takes time (10 seconds) to reach the heartbeat failed scenario, but otherwise we are not testing this explicitily
3. Send hearbeat again and check button goes away

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,41 @@ context('Hosts Overview', () => {
       });
     });
   });
+
+  describe('Deregistration', () => {
+    describe('Clean-up buttons should be visible only when needed', () => {
+      const heartbeatTargets = agents();
+      const failingHost = heartbeatTargets.shift();
+      before(() => {
+        cy.task('startAgentHeartbeat', heartbeatTargets);
```
arbulu89, line 179: Shouldn't we navigate to the hosts view here?
I know that we do in the main `beforeAll`, but if we don't do it here, we are already basing our tests in the order of things. For example, before this describe we have the `healthDetection` tests, and these tests would already be affected by those.
What do you think?

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,41 @@ context('Hosts Overview', () => {
...
+        );
+
+        for (let i = 2; i < 11; i++) {
+          cy.get(
+            `:nth-child(${i}) > .w-48 > .bg-white > .text-jungle-green-500`
+          ).should('not.exist');
+        }
+      });
+
+      it('should allow to deregister a host after clean up confirmation', () => {
```
arbulu89, line 195: Should this test belong to the current `describe`. It is not about the button visualization. This already runs a deregistration, and most probably we want to have more things in this, like restoration and other things

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,41 @@ context('Hosts Overview', () => {
...
+
+      it('should hide all Clean-up buttons when heartbeat starts except the first one', () => {
+        cy.get('tr:nth-child(1) > td:nth-child(9)').should(
+          'contain.text',
+          'Clean up'
+        );
+
+        for (let i = 2; i < 11; i++) {
+          cy.get(
+            `:nth-child(${i}) > .w-48 > .bg-white > .text-jungle-green-500`
```
arbulu89, line 190: Still old query string (the same in the next lines)

arbulu89 (commented):
Some additional comments from my side

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,41 @@ context('Hosts Overview', () => {
       });
     });
   });
+
+  describe('Deregistration', () => {
+    describe('Clean-up buttons should be visible only when needed', () => {
+      const heartbeatTargets = agents();
+      const failingHost = heartbeatTargets.shift();
```
jamie-suse, line 177: `shift()` mutates the `heartbeatTargets` array, is this safe for the tests?

assets/js/components/DeregistrationModal/DeregistrationModal.jsx
jamie-suse: I know the `click` event will bubble up to the `button` element, so it still works, but isn't it more proper to place this ID on the `button`?

In any case, the ID for testing might not even be required, in previous tests we selected the _confirm clean up button_ like this:

```javascript
screen.getAllByRole('button', { name: 'Clean up' })[1]
```

https://github.com/trento-project/web/blob/73f91dca23b50266544b02bced0a3299dfcdfde8/assets/js/components/HostsList.test.jsx#L173

test/e2e/cypress/e2e/hosts_overview.cy.js
rtorrero: That's exactly the intention, to leave that host out of the array so that it doesn't get a heartbeat

assets/js/components/DeregistrationModal/DeregistrationModal.jsx
rtorrero: @jamie-suse That seems from react's testing library, not sure if I can use it in cypress. In any case, based on the official docs https://docs.cypress.io/guides/references/best-practices, `data-cy` seems to be the preferred way.

rtorrero: (or data-test or data-testid)

test/e2e/cypress/e2e/hosts_overview.cy.js
rtorrero: Good point @arbulu89 I guess that if any of the previous tests would do a `cy.visit(...)` it could break. I'll add a 
```
cy.visit('/hosts');
cy.url().should('include', '/hosts');
```
Maybe for a future PR we could change the top `before()` for a `beforeEach()` so that this URL visit/check is performed for all tests without having to add it explicitly

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -170,4 +170,64 @@ context('Hosts Overview', () => {
       });
     });
   });
+
+  describe('Deregistration', () => {
+    const host = {
```
arbulu89, line 175: Maybe we could call this `hostToDeregister`. But not a big deal

.photofinish.toml
```diff
@@ -202,3 +202,13 @@ files = [
 files = [
     "./test/fixtures/scenarios/host-details/9cd46919-5f19-59aa-993e-cf3736c71053_cloud_discovery_unknown.json",
 ]
+
+[vmdrbddev01]
```
arbulu89, line 206: This is not used anywhere.
I would avoid adding things that you only use locally for your tests.
If you need in reallity, just add in the corresponding PR

arbulu89 (approved):
Looks good @rtorrero 
I wouldn't add the photofinish new feature if it is not used in the tests.
I understand that you added because you were using it locally for your development

PR #1635, opened 2023-07-17
Refactor check selection saga

nelsonkopliku:
# Description

This PR adds a centralized saga/state/selectors for `checkSelection` aiming to support hosts and clusters now and other target types later. 

Following up, this will be hooked in host checks selection and cluster checks selection.

## How was this tested?

Automated tests.


dottorblaster (changes requested):
A couple things then we can merge

dottorblaster (approved):
LGTM

PR #1639, opened 2023-07-18
Fix database sid display

arbulu89:
# Description
Fix database sid visualization in HostList view and HanaClusterDetails view.
Before this change, the database sid was only available when the SAP system was completely registered. If the database was only there, the sid was not being displayed.
This happened because the database instance list was being obtained from the SAP system list redux state instead of its own database state

image

## How was this tested?
Tests updated


assets/js/lib/test-utils/data/databases.js
```diff
@@ -0,0 +1,116 @@
+export default [
```
arbulu89, line 1: The databases slice default state added, as this is how it is loaded in the initial fetch (and next updates)

assets/js/state/selectors/cluster.js
```diff
@@ -21,13 +21,17 @@ export const getClusterName =
 export const getClusterSapSystems = (clusterID) => (state) => {
   const clusterHostIDs = getClusterHostIDs(clusterID)(state);
   const {
-    sapSystemsList: { sapSystems, applicationInstances, databaseInstances },
+    sapSystemsList: { sapSystems, applicationInstances },
   } = state;
+  const {
+    databasesList: { databases, databaseInstances },
+  } = state;
```
dottorblaster, line 28:
```suggestion
    sapSystemsList: { sapSystems, applicationInstances },
    databasesList: { databases, databaseInstances },
  } = state;
```

dottorblaster (changes requested):
Just a tiny nitpick

nelsonkopliku (approved):
Cool! Nice catch!

dottorblaster (approved):
If the CI is happy, I am

CDimonaco (approved):

PR #1641, opened 2023-07-18
Use centralized checks selection for both hosts and clusters

nelsonkopliku:
# Description

This PR:
- adds the cluster checks selection saga in the centralized checks selection saga
- makes the new centralized checks selection used both in host and cluster checks selection pages.
- cleans up legacy

Note that the changes in the legacy `assets/js/components/ClusterDetails/ChecksSelection.jsx` and test are basically transient as that component will be removed soon in favor of the generalized one in `assets/js/components/ChecksSelection/ChecksSelection.jsx`

## How was this tested?

Automated tests.


assets/js/components/ClusterDetails/ChecksSelection.jsx
```diff
@@ -40,9 +46,12 @@ const getGroupSelectedState = (checks, selectedChecks) => {
...
 
-  const { saving, savingError, savingSuccess } = useSelector(
-    (state) => state.clusterChecksSelection
-  );
+  const { status } = useSelector(getClusterCheckSelection(clusterId));
+  const { saving, savingError, savingSuccess } = {
+    saving: isSaving(status),
+    savingSuccess: isSuccessfullySaved(status),
+    savingError: isSavingFailed(status),
+  };
```
dottorblaster, line 54:
```suggestion
const saving = isSaving(status);
const savingSuccess = isSuccessfullySaved(status);
const savingError = isSavingFailed(status);
```

assets/js/state/checksSelection.js
```diff
@@ -5,6 +5,10 @@ const SAVING = 'SAVING';
 const SUCCESSFULLY_SAVED = 'SUCCESSFULLY_SAVED';
 const SAVING_FAILED = 'SAVING_FAILED';
 
+export const isSaving = (status) => SAVING === status;
+export const isSuccessfullySaved = (status) => SUCCESSFULLY_SAVED === status;
+export const isSavingFailed = (status) => SAVING_FAILED === status;
+
```
dottorblaster, line 11: The idiomatic way to do this is creating selectors that return boolean values. `reselect` would help, but we still didn't introduce that.

dottorblaster (changes requested):
A couple changes and then we can merge

assets/js/state/checksSelection.js
nelsonkopliku: Cool! Would you mind tracking that so we can plan for a refactor?

dottorblaster: Yes absolutely! Tracking that on Jira right now, could you still move this to selectors and base the functions on `resourceID`?

nelsonkopliku: sure. You meant `targetID` I guess :smile:

dottorblaster: `whateverID`, I just get lost into these names :joy:

PR #1664, opened 2023-07-24
Switch to UUIDs inside catalog Jest testing

dottorblaster:
# Description

Sometimes we get failures in our CI due to this test using animal names as group IDs and causing a clash when rendering the elements, because the group IDs are used as keys for the shadow DOM.

Switching to UUIDs so we don't get clashes and the test doesn't get crazy. 

## How was this tested?

Jest


nelsonkopliku (approved):
Flakiness: terminated :smile:

EMaksy (approved):
Hey LGTM :D

PR #1665, opened 2023-07-24
E2e restoration

rtorrero:
# Description

This PR adds e2e tests for the restoration of hosts after they have been deregistered. It checks that the right details are restored and visible in:
 - hosts overview
 - sap systems overview
 - clusters overview
 - dbs overview
 - cluster details
 - sap system details
 - hana db details
 
Notice that to do this, we have added additional photofinish scenarios that reuse the existing discovery json messages.

### DIFF
--- a/.photofinish.toml
+++ b/.photofinish.toml
@@ -203,14 +203,14 @@ files = [
     "./test/fixtures/scenarios/host-details/9cd46919-5f19-59aa-993e-cf3736c71053_cloud_discovery_unknown.json",
 ]
 
-[host-vmdrbddev01-restore]
+[host-vmhdbdev01-restore]
 
 files = [
-    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/240f96b1-8d26-53b7-9e99-ffb0f2e735bf_host_discovery.json",
-    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/240f96b1-8d26-53b7-9e99-ffb0f2e735bf_subscription_discovery.json",
-    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/240f96b1-8d26-53b7-9e99-ffb0f2e735bf_cloud_discovery.json",
-    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/240f96b1-8d26-53b7-9e99-ffb0f2e735bf_ha_cluster_discovery.json",
-    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/240f96b1-8d26-53b7-9e99-ffb0f2e735bf_sap_system_discovery.json"
+    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/13e8c25c-3180-5a9a-95c8-51ec38e50cfc_host_discovery.json",
+    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/13e8c25c-3180-5a9a-95c8-51ec38e50cfc_subscription_discovery.json",
+    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/13e8c25c-3180-5a9a-95c8-51ec38e50cfc_cloud_discovery.json",
+    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/13e8c25c-3180-5a9a-95c8-51ec38e50cfc_ha_cluster_discovery.json",
+    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/13e8c25c-3180-5a9a-95c8-51ec38e50cfc_sap_system_discovery.json"
 ]
 
 [host-vmhdbprd02-restore]
--- a/.photofinish.toml
+++ b/.photofinish.toml
@@ -212,3 +212,28 @@ files = [
     "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/240f96b1-8d26-53b7-9e99-ffb0f2e735bf_ha_cluster_discovery.json",
     "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/240f96b1-8d26-53b7-9e99-ffb0f2e735bf_sap_system_discovery.json"
 ]
+
+[sapsystem-NWD-restore]
+
+files = [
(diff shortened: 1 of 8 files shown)


jamie-suse (commented):
This might be overkill, but what do you think about first asserting that the host/database/cluster/SAP system does *not* exist, then running the scenario and asserting that it has been restored?
I guess these assertions for nonexistence will take extra time to run though. Just a thought.

.photofinish.toml
```diff
@@ -202,3 +202,93 @@ files = [
 files = [
     "./test/fixtures/scenarios/host-details/9cd46919-5f19-59aa-993e-cf3736c71053_cloud_discovery_unknown.json",
 ]
+
+[host-vmdrbddev01-restore]
```
arbulu89, line 206: I don't know if we should use any `drbd` thing for anything hehe
As they are candidate number one to be removed from our photofinish setup.
Maybe it will happen, but hey, just pointing it out

.photofinish.toml
```diff
@@ -202,3 +202,93 @@ files = [
...
+
+files = [
+    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/240f96b1-8d26-53b7-9e99-ffb0f2e735bf_host_discovery.json",
```
arbulu89, line 209: We could maybe open an improvement ticket to photifinish, to accept wildcards so we don't need to repeat so many similar files

test/e2e/cypress/e2e/clusters_overview.cy.js
```diff
@@ -165,4 +165,19 @@ context('Clusters Overview', () => {
       cy.contains(hanaCluster1.name).should('not.exist');
     });
   });
+
+  describe('Restoration', () => {
```
arbulu89, line 169: The tests is totally dependant on the orther of tests. Here you assume that the previous test was already executed and worked.
I think we should have a `before` clause where we check if the host is there, and if it is, deregister it.

Other option that I would accept is to move this test to the `Deregistration` describe, as in essence, the restoration is part of the deregistration.

test/e2e/cypress/e2e/databases_overview.cy.js
```diff
@@ -18,4 +18,19 @@ context('Databases Overview', () => {
       cy.contains(hdqDatabase.sid).should('not.exist');
     });
   });
+
+  describe('Restoration', () => {
+    const databaseToRestore = {
```
arbulu89, line 23: We have seen some recent failure on Databases and SAP systems restoration. The instances belonging to them were not recovered properly after a browser refresh.
Maybe we should consider those events.

test/e2e/cypress/e2e/clusters_overview.cy.js
```diff
@@ -165,4 +165,19 @@ context('Clusters Overview', () => {
...
+    const hanaClusterToRestore = {
+      name: 'hana_cluster_1',
+      hosts: [
+        '13e8c25c-3180-5a9a-95c8-51ec38e50cfc',
+        '0a055c90-4cb6-54ce-ac9c-ae3fedaf40d4',
+      ],
+    };
+
+    it(`should show cluster ${hanaClusterToRestore.name}`, () => {
```
arbulu89, line 178: I find this test too basic.
I think we agreed that we need to test if the previously added tags are restored properly

arbulu89 (changes requested):

test/e2e/cypress/e2e/databases_overview.cy.js
rtorrero: I've added the test and indeed it fails. Oddly enough I can't consistently reproduce the bug manually, but cypress catches it.

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -232,6 +232,11 @@ context('Hosts Overview', () => {
 
         cy.get(`#host-${hostToDeregister.id}`).should('not.exist');
       });
+
+      it(`should show host ${hostToDeregister.name} registered again`, () => {
```
arbulu89, line 236: Would be a lot hassle to add the same `tag` testing here?

arbulu89, line 236: Maybe we could use the `restore` word in the test description

arbulu89 (commented):
@rtorrero much better!
Some last nitpicks.
If you want, we can leave the "remaining instances are there after restoration" e2e test for the PR which includes the fix

test/e2e/cypress/e2e/databases_overview.cy.js
```diff
@@ -7,15 +7,37 @@ context('Databases Overview', () => {
...
+
+    it(`should include both instances in DB ${hdqDatabase.sid} after restoring the primary instance`, () => {
+      cy.contains('tr', hdqDatabase.sid).should('exist').click();
+      cy.contains('div', hdqDatabase.instances[0].name).should('exist');
+      cy.contains('div', hdqDatabase.instances[1].name).should('exist');
+    });
+
+    it('should show the ACTIVE pill in the right host', () => {
+      cy.contains('div', hdqDatabase.instances[1].name).within(() => {
+        cy.contains('ACTIVE');
```
arbulu89, line 39: Could we put this `Active` in the `hdbDatabase.instances.state` or something like that.
Besides that, we should check that the 1st row doesn't include it, as this is the current bug

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -173,8 +173,10 @@ context('Hosts Overview', () => {
 
   describe('Deregistration', () => {
     const hostToDeregister = {
-      name: 'vmdrbddev01',
-      id: '240f96b1-8d26-53b7-9e99-ffb0f2e735bf',
+      name: 'vmhdbdev01',
+      id: '13e8c25c-3180-5a9a-95c8-51ec38e50cfc',
+      tablePos: 7,
```
arbulu89, line 178: This looks so ugly :sweat_smile: 
I would rename it to `tableRow` anyway

rtorrero: I improved it a bit by removing the dependency on the table position, should look better

PR #1682, opened 2023-07-28
Make checks selection great again

dottorblaster:
# Description

AW YEAH

## How was this tested?
Existing tests updated, some rearchitected

### DIFF
--- a/assets/js/components/ChecksSelection/ChecksSelection.jsx
+++ b/assets/js/components/ChecksSelection/ChecksSelection.jsx
@@ -1,9 +1,7 @@
-import React, { useState, useEffect } from 'react';
+import React, { useEffect } from 'react';
 import classNames from 'classnames';
 
 import { remove, uniq, toggle, groupBy } from '@lib/lists';
-
-import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
 
 import CatalogContainer from '@components/ChecksCatalog/CatalogContainer';
 import ChecksSelectionGroup, {
@@ -13,8 +11,6 @@ import ChecksSelectionGroup, {
   allSelected,
 } from './ChecksSelectionGroup';
 import ChecksSelectionItem from './ChecksSelectionItem';
-import FailAlert from './FailAlert';
-import ExecutionSuggestion from './ExecutionSuggestion';
 
 const isSelected = (selectedChecks, checkID) =>
   selectedChecks ? selectedChecks.includes(checkID) : false;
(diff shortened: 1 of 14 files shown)


nelsonkopliku (commented):
Great job so far!

I tried the branch locally and noticed a couple of things:
- I cannot select checks in the host checks selection (clicking the switch does not seem to trigger a state change)
- (ie in clusters) it is possible to trigger an execution with an empty checks selection, ending up in an infinite _starting_ state cause wanda would never start an execution for an empty selection

The second item relates to the question: _when should we be able to start an execution?_ which imho should be when we have a saved non-empty checks selection.

assets/js/components/ChecksSelection/ChecksSelection.jsx
```diff
@@ -108,45 +92,13 @@ function ChecksSelection({
...
-                      setSelectedChecks(toggle(check.id, selectedChecks));
-                      onClear();
+                      onChange(toggle(check.id, selectedChecks));
                     }}
                   />
                 ))}
               </ChecksSelectionGroup>
             ))}
           </div>
-          <div className="place-items-end flex">
```
nelsonkopliku, line 119: What was the reason to move the buttons outside?

assets/js/components/ClusterSettingsPage/ClusterSettingsPage.jsx
```diff
@@ -49,23 +76,41 @@ function ClusterSettingsPage() {
...
+      <div className="flex flex-wrap">
+        <div className="flex w-1/2 h-auto overflow-hidden overflow-ellipsis break-words">
+          <PageHeader>
+            Cluster Settings for{' '}
+            <span className="font-bold">{clusterName}</span>
+          </PageHeader>
+        </div>
+        <div className="flex w-1/2 justify-end">
+          <div className="flex w-fit whitespace-nowrap">
+            <Button type="primary" className="mx-1" onClick={saveSelection}>
```
nelsonkopliku, line 98: For a consistent ux we could disable also this button while saving the selection.

assets/js/components/ChecksSelection/ChecksSelection.jsx
dottorblaster: Just positioning as per mockups, of course moving those to a completely different place I had to move the state declarations as well

nelsonkopliku (commented):
I think that once we adjusted the logic that determines whether an execution can be started or not and aligned the behavior in host checks selection, we're good to go.

Given the duplication of the action buttons, we just need to keep remembering to align behaviors in the two selections :smile: 
Maybe we can refactor to a container component later.

assets/js/components/ClusterSettingsPage/ClusterSettingsPage.jsx
```diff
@@ -49,23 +82,46 @@ function ClusterSettingsPage() {
...
+              onClick={saveSelection}
+              disabled={saving}
+            >
+              Save Checks Selection
+            </Button>
+            <Button
+              type="primary"
+              className="mx-1"
+              onClick={requestExecution}
+              disabled={canStartExecution(selection, savingSuccessful)}
```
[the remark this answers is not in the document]
dottorblaster: LOL you're right about the typo

nelsonkopliku (approved):
From a behavior perspective I think it is good to go.
Not sure about tiny details like the color of the disabled start execution, but that I'd leave to @jagabomb 

LGTM

dottorblaster:
Thanks! Anything concerning the `disabled` style I think we can leave it to a new PR, I just added the attribute otherwise we would end up not having that one

PR #1686, opened 2023-08-01
Skip tests that are still missing bugfixes

rtorrero:
# Description

This PR adds .skips for tests that still require bugfixes to pass.

### DIFF
--- a/test/e2e/cypress/e2e/databases_overview.cy.js
+++ b/test/e2e/cypress/e2e/databases_overview.cy.js
@@ -31,12 +31,12 @@ context('Databases Overview', () => {
       cy.contains('tr', hdqDatabase.sid).should('exist').click();
     });
 
-    it(`should include both instances in DB ${hdqDatabase.sid} after restoring the primary instance`, () => {
+    it.skip(`should include both instances in DB ${hdqDatabase.sid} after restoring the primary instance`, () => {
       cy.contains('div', hdqDatabase.instances[0].name).should('exist');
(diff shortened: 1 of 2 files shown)


jamie-suse (approved):

arbulu89 (commented):
@jamie-suse My PR would remove the 2nd skip.
Is your PR removing the other 2?

jamie-suse:
> @jamie-suse My PR would remove the 2nd skip. Is your PR removing the other 2?

@arbulu89 @rtorrero Yeah, I can remove them in my PR :+1:

arbulu89:
> > @jamie-suse My PR would remove the 2nd skip. Is your PR removing the other 2?
> 
> @arbulu89 @rtorrero Yeah, I can remove them in my PR +1

@jamie-suse but are both passing or the tests make sense?
I want to understand if we are making anything logic with this PR

jamie-suse:
> @jamie-suse but are both passing or the tests make sense?
> I want to understand if we are making anything logic with this PR

Both tests are passing in my PR. They make sense to me as one is testing that the database restoration works, and the other is testing that when a SAP system is deregistered, the SID should be removed from Hosts (this test is in the Hosts List).

arbulu89:
> > @jamie-suse but are both passing or the tests make sense?
> > I want to understand if we are making anything logic with this PR
> 
> Both tests are passing in my PR. They make sense to me as one is testing that the database restoration works, and the other is testing that when a SAP system is deregistered, the SID should be removed from Hosts (this test is in the Hosts List).

Okey them. Thank you for clarifying.

PR #1687, opened 2023-08-01
Fix SR status broadcast on database instance registration

arbulu89:
# Description

Fix system replication status broadcast on database instance registration.
Basically, we need all database instances to get the value for each type (primary/secondary).

I want to test this with the whole "restoration" fix for the instances (https://github.com/trento-project/web/pull/1673)

PD: I do agree that this got complex, so querying with a api call on this events as refactor would be a good idea.
PD2: In fact, it has "forced" me to change some domain events order to make it work without horrible "data enrichment" messages to the frontend...

## How was this tested?

Everything tested in the projection. I didn't add any test on the view, as the test would be the same.

### DIFF
--- a/lib/trento_web/views/v1/sap_system_view.ex
+++ b/lib/trento_web/views/v1/sap_system_view.ex
@@ -28,6 +28,15 @@ defmodule TrentoWeb.V1.SapSystemView do
     |> Map.delete(:host)
   end
 
+  def render("database_instance_with_sr_status.json", %{
+        instance: instance,
+        database_instances: database_instances
+      }) do
+    "database_instance.json"
+    |> render(%{instance: instance})
+    |> add_system_replication_status(database_instances)
+  end
+
   def render("database_registered.json", %{database: database}) do
     database
     |> Map.from_struct()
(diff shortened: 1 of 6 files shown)


rtorrero (approved):
@arbulu89 looks good to me. Just for a better understanding: what made you have to change the order of the restoration event? (not questioning it, I think it's intuitively makes more sense to me this way)

arbulu89:
> @arbulu89 looks good to me. Just for a better understanding: what made you have to change the order of the restoration event? (not questioning it, I think it's intuitively makes more sense to me this way)

Already commented in our internal Slack channel, but here we go again :laughing: 

In order to calculate the `Active` pill, we need to have all the database instances (specially the primary) present in the database.
During the restoration, we can have a corner case (or normal case, depending what you do...), where you deregister the database simply removing the Primary instance, and letting the Secondary there, without removing, but it is not showing up in the console, so it stays as "zombie".

When you restore the Primary instance, the Secondary was already present there, so it is sent to the Frontend together with the DatabaseRestore event, but as the primary was not there, the `Active` pill couldn't be calculated, so it is not set.
To calculate it we need the Primary.

Changing the order, maker the Primary instance read model being present in the database for the moment we do the query to do the calculation

I'm saying this is a corner case, because in a normal flow, you couldn't never have a Secondary there without previously having a Primary, but as it stayed as zombie...

lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -152,32 +161,43 @@ defmodule TrentoWeb.V1.SapSystemView do
...
+    end)
   end
+
+  defp map_system_replication_status_to_secondary(
+         %{system_replication: "Secondary"} = instance,
+         system_replication_status
+       ),
+       do: %{instance | system_replication_status: system_replication_status}
+
+  defp map_system_replication_status_to_secondary(%{system_replication: "Primary"} = instance, _),
```
jamie-suse, line 199: Just a question: why do primary instances get set to `""` here? To me the name `map_system_replication_status_to_secondary()` makes me think that primary instances won't be affected by this function

arbulu89: The code is exactly as it was before.
The secondary state pill is only sent to `Secondary` instances, that's why it is set to an empty string (as it was before).
About the naming, well, this is pattern matching technique, so the 1st and 2nd functions must have the same name, and at the end, the functions sets the status only to secondary instances

CDimonaco (approved):

jamie-suse (approved):

PR #1689, opened 2023-08-02
Reduce flakyness

rtorrero:
# Description
I've been tasking a look at the past month's CI failures, and in an attempt to reduce the red there:

 - remove the deprecated `headless: true`. for the cypress action. It's default since 8.0
 - switch `exist` with `be.visible` which should be more specific. This should not affect flakyness but should be more robust (We could actually take a look at other tests and see where we can also update this)
 - on the test that I saw most of the flakyness at (checks_catalog), `cy.get('div.check-row').should('have.length', 4);` has been added to block the execution until the rows about to be clicked are visible.
 - a `contains` that looks for the heartbeat toast before trying to click the button has been added in the host_details which was another test that showed failures from time to time.
 - chromatic deployment action showed failures when being run from the repository_dispatch that gets triggered from wanda repo, so it now checks if `action_name` is `repository_dispatch` to avoid running.

### DIFF
--- a/test/e2e/cypress/e2e/checks_catalog.cy.js
+++ b/test/e2e/cypress/e2e/checks_catalog.cy.js
@@ -68,8 +68,9 @@ context('Checks catalog', () => {
   describe('Individual checks data is expanded', () => {
     it('should expand check data when clicked', () => {
       cy.get('.check-panel').should('not.exist');
+      cy.get('div.check-row').should('have.length', 4);
       cy.get('div.check-row').first().click();
-      cy.get(`.check-panel`).should('exist');
+      cy.get(`.check-panel`).should('be.visible');
     });
   });
 
--- a/test/e2e/cypress/e2e/host_details.cy.js
+++ b/test/e2e/cypress/e2e/host_details.cy.js
@@ -359,12 +359,14 @@ context('Host Details', () => {
   describe('Deregistration', () => {
     describe('"Clean up" button should be visible only for an unhealthy host', () => {
       it('should not display the "Clean up" button for healthy host', () => {
-        cy.task('startAgentHeartbeat', [selectedHost.id]);
         cy.contains('button', 'Clean up').should('not.exist');
       });
 
       it('should show the "Clean up" button once heartbeat is lost and debounce period has elapsed', () => {
         cy.task('stopAgentsHeartbeat');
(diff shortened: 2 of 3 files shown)


test/e2e/cypress/e2e/checks_catalog.cy.js
```diff
@@ -68,8 +68,9 @@ context('Checks catalog', () => {
   describe('Individual checks data is expanded', () => {
     it('should expand check data when clicked', () => {
       cy.get('.check-panel').should('not.exist');
+      cy.get('div.check-row').should('have.length', 4);
```
arbulu89, line 71: Having this `4` hardcoded would make the test fail if at some point we add a new checks group.
Could we check maybe if the groups are there some other way?
Maybe `.should('have.length.least', 1)`

test/e2e/cypress/e2e/host_details.cy.js
```diff
@@ -359,12 +359,14 @@ context('Host Details', () => {
   describe('Deregistration', () => {
     describe('"Clean up" button should be visible only for an unhealthy host', () => {
       it('should not display the "Clean up" button for healthy host', () => {
-        cy.task('startAgentHeartbeat', [selectedHost.id]);
         cy.contains('button', 'Clean up').should('not.exist');
       });
 
       it('should show the "Clean up" button once heartbeat is lost and debounce period has elapsed', () => {
         cy.task('stopAgentsHeartbeat');
+        cy.contains(`The host ${selectedHost.hostName} heartbeat is failing.`, {
```
arbulu89, line 367: Just curiosity. Why do we need this check?

arbulu89 (approved):
It looks good!
Just a minor suggestion

nelsonkopliku (approved):
LGTM. Let's see how it goes :smile:

test/e2e/cypress/e2e/host_details.cy.js
rtorrero: Watching the screenshot of the failing tests, it looked like the heartbeat toast was showing up still when the "Clean Up" button failed to be clicked. So basically this adds another point to sync with the state before clicking. I know it shouldn't be needed, but it does look like it helped after many tests. The long term solution is a bit of a refactor of the e2e tests that I've been already giving some thought.

PR #1690, opened 2023-08-03
Add API backward compatibility check in pipeline

nelsonkopliku:
# Description

This PR introduces a backward compatibility check in the pipeline for our APIs, for both V1 and V2.
Relies on https://github.com/OpenAPITools/openapi-diff

- [Backward incompatible change makes stops the pipeline](https://github.com/trento-project/web/actions/runs/5752421716/job/15593405750#step:11:67)
- [Backward compatible change lets the pipeline continue](https://github.com/trento-project/web/actions/runs/5752530118/job/15593744241#step:11:66)
- [No changes pass](https://github.com/trento-project/web/actions/runs/5752614002/job/15594004127#step:11:50)

---

I tried this action at first https://github.com/marketplace/actions/openapi-diff-action but it was failing with a weird
> Error: The template is not valid. swimmwatch/openapi-diff-action/v1.0.1/action.yml (Line: 77, Col: 12): The maximum allowed memory size was exceeded while evaluating the following expression: format('compatible_val="{0}"
>

[see this run](https://github.com/trento-project/web/actions/runs/5751458228/job/15590268392#step:12:111)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -148,7 +148,7 @@ jobs:
 
   static-code-analysis:
     name: Static Code Analysis
-    needs: [elixir-deps, npm-deps]
+    needs: [elixir-deps, npm-deps, api-bc-check]
     runs-on: ubuntu-20.04
     steps:
       - name: Cancel Previous Runs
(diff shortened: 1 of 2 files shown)


dottorblaster:
Nice! Maybe we could have this action instead of the plain `docker run`?

https://github.com/addnab/docker-run-action

Too bad the official action doesn't work :cry:

nelsonkopliku:
> Nice! Maybe we could have this action instead of the plain `docker run`?
> 
> https://github.com/addnab/docker-run-action
> 
> Too bad the official action doesn't work cry

@dottorblaster already tried and does not work well with how the openapi diff image works

dottorblaster (approved):
Then, LGTM

PR #1704, opened 2023-08-09
Pull request environment

CDimonaco:
# Description

This pr create the github workflows for creating and updating an ephimeral pull request environments.

### Prerequisites

- The pr should be authored by one of the web repository mantainers.
- The pr should have the env label

### Execution

The environment is created/updated when the prerequisities are met, the author of pr can suspend the environment update/creation simply removing the env label from pull request. When the label is attached again the environment gets synced.

The environment has the `web` project code from the pr branch, and as dependency `wanda`, with the rolling release container.

The web project is seeded with `photofinish` using the `healthy-27-node-SAP-cluster` scenario.

The databases, users and rabbit configuration are created and deleted on-demand.

When the pr is closed all the environment included his dependencies are deleted.

When the environment is deployed, is accessible on `https://<prnumber>.prenv.trento.suse.com` 

## How was this tested?

Act locally and manual triggering

### DIFF
--- a/.github/workflows/pr_env.yaml
+++ b/.github/workflows/pr_env.yaml
@@ -1,0 +1,150 @@
+name:  Pull request environment
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - labeled
+
+env:
+  MANTAINERS: "[\"cdimonaco\", \"dottorblaster\", \"fabriziosestito\", \"rtorrero\", \"nelsonkopliku\", \"arbulu89\",\"jagabomb\",\"emaksy\",\"jamie-suse\"]"
+  PR_ENV_LABEL: env
+  PR_NUMBER: "${{ github.event.pull_request.number }}"
+
+jobs:
+  check_env_creation_privilege:
+    name: Check if the environment creation criteria are met, store in the job output
+    runs-on: ubuntu-latest
+    outputs:
+      create_env: ${{ steps.check.outputs.create_env }}
+    steps:
+      - id: check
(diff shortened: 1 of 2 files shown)


CDimonaco:
You can try the pr environment with this pr here: https://1704.prenv.trento.suse.com

abravosuse (approved):
Code apart, it looks good to me. I just would like to understand, when time comes, is how to install Trento Server based on a particular PR.

nelsonkopliku (commented):
This is super! Thanks for the effort.
Just have a couple of things I need to better understand.

All the envs would be released on the same ec2, right? Not an ec2 for each PR.

Since agents are missing also the PR env is lacking in a couple of features  - heartbeating and facts gathering.

No heartbeat means at most having some icons red/gray, but no facts gathering means always getting a timed out checks execution, limiting a bit the value of an env where we might also be interested in other checks execution scenarios other than timeout.

What about launching web and wanda containers in demo mode? It'd give us a more comprehensive experience.

At least until we are able to have also fake agents.

CDimonaco:
I prefer to not fake anything in this type of environment, this is not a demo type of env but more like a "development" environment online, so I want to directly use photofinish or attach real agent if i need one, or in case of fake one, this should be a direct action without automating it.

It's our development environment but online, so each action is performed by us without any automatic thingy

cc @dottorblaster wdyt?

dottorblaster:
@nelsonkopliku  fair point! I'd like to have a stab at this as-is, I need it to test some reselect goodies. We can set MIX_ENV=demo for wanda in a subsequent iteration.

dottorblaster (approved):

PR #1706, opened 2023-08-10
Restore a SAP system only when the database is present

CDimonaco:
# Description

This pr fix a SAP system restore bug, we want to restore the SAP system only if the database is present.

## How was this tested?

Automated tests


jamie-suse (approved):

abravosuse (approved):
Code apart, I've tested thjs in my rolling environment with the web image from http://ghcr.io/trento-project/trento-web:1706-env and it works.

PR #1710, opened 2023-08-11
Fix host checks selection ui

EMaksy:
# Description
This PR addresses an issue where the selected "checks" were not being updated correctly when saved in the host checks selection view. I've added a useEffect to ensure proper state updates, see demo.

Furthermore, I've refactored the code by moving the state to the container component HostSettingsPage. 
Also added a helper redux selector function to get all hosts from a check.

## DEMO before changes
host_check_selection_broadcast_before

## DEMO after changes
host_check_selection_broadcast

## How was this tested?
Updated existing test
Added a test for new getHostSelectedChecks function

### DIFF
--- a/assets/js/components/HostDetails/HostChecksSelection.jsx
+++ b/assets/js/components/HostDetails/HostChecksSelection.jsx
@@ -21,7 +21,7 @@ function HostChecksSelection({
   isSavingSelection,
   onSaveSelection,
   onSelectedChecksChange,
-  enableHostChecksExecution,
+  hostChecksExecutionEnabled,
 }) {
   return (
     <div className="w-full px-2 sm:px-0">
@@ -47,7 +47,7 @@ function HostChecksSelection({
               type="primary"
               className="mx-1"
               onClick={() => {}}
-              disabled={enableHostChecksExecution}
+              disabled={hostChecksExecutionEnabled}
             >
               <EOS_PLAY_CIRCLE className="fill-white inline-block align-sub" />{' '}
               Start Execution
--- a/assets/js/components/HostDetails/HostChecksSelection.jsx
+++ b/assets/js/components/HostDetails/HostChecksSelection.jsx
@@ -4,9 +4,7 @@ import { EOS_PLAY_CIRCLE } from 'eos-icons-react';
 import PageHeader from '@components/PageHeader';
 import BackButton from '@components/BackButton';
 import Button from '@components/Button';
-import ChecksSelection, {
-  canStartExecution,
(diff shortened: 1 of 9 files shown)


assets/js/state/selectors/index.js
```diff
@@ -10,6 +13,11 @@ export const isIdByKey =
 export const getHost = (id) => (state) =>
   state.hostsList.hosts.find((host) => host.id === id);
 
+export const getHostSelectedChecks = (hostID) => (state) => {
+  const host = getHost(hostID)(state);
+  return get(host, 'selected_checks', defaultEmptyArray);
+};
+
```
dottorblaster, line 20: Could you create a `host.js` file and move this inside? This way we could start moving all the host-related selectors into their own yard

assets/js/components/HostDetails/HostChecksSelection.jsx
```diff
@@ -49,7 +49,9 @@ function HostChecksSelection({
               type="primary"
               className="mx-1"
               onClick={() => {}}
-              disabled={!canStartExecution(selectedChecks, isSavingSelection)}
+              disabled={
+                !canStartExecution(hostSelectedChecks, isSavingSelection)
+              }
```
dottorblaster, line 54: I think you can move this check outside and avoid passing the `hostSelectedChecks` prop

dottorblaster (changes requested):
Brilliant! A couple things to adjust then we can merge :+1:

assets/js/state/selectors/index.js
EMaksy: Yeah good idea! i moved only host related functions to `host.js`

assets/js/components/HostDetails/HostChecksSelection.jsx
EMaksy: Lets try :+1:

assets/js/components/HostDetails/HostChecksSelection.jsx
```diff
@@ -16,14 +14,14 @@ function HostChecksSelection({
...
   selectedChecks,
-  hostSelectedChecks,
   catalog,
   catalogError,
   catalogLoading,
   onUpdateCatalog,
   isSavingSelection,
   onSaveSelection,
   onSelectedChecksChange,
+  enableHostChecksExecution,
```
dottorblaster, line 24: `hostChecksExecutionEnabled` I think it would be better

dottorblaster (approved):
Just a nitpick and we can merge

PR #1715, opened 2023-08-16
Sid linking from clusters

rtorrero:
# Description

This PR adds a tooltip for the SapSystemLink component so that when the linked SapSystem is not registered, a warning is shown.

It also adjusts the ClusterList to use this new component.

## How was this tested?

Automated tests added

### DIFF
--- a/assets/js/components/ClustersList.jsx
+++ b/assets/js/components/ClustersList.jsx
@@ -5,6 +5,7 @@ import Tags from '@components/Tags';
 import PageHeader from '@components/PageHeader';
 import { addTagToCluster, removeTagFromCluster } from '@state/clusters';
 import ClusterLink from '@components/ClusterLink';
+import SapSystemLink from '@components/SapSystemLink';
 import { ExecutionIcon } from '@components/ClusterDetails';
 import { post, del } from '@lib/network';
 import { useSearchParams } from 'react-router-dom';
@@ -24,6 +25,26 @@ const getClusterTypeLabel = (type) => {
   }
 };
 
+const getSapSystemLinkBySID = (
+  applicationInstances,
+  databaseInstances,
+  sid
+) => {
+  const foundInstance = applicationInstances
+    .map((instance) => ({ ...instance, type: 'sap_systems' }))
+    .concat(
+      databaseInstances.map((instance) => ({
+        ...instance,
+        type: 'databases',
+      }))
+    )
+    .find((instance) => instance.sid === sid);
+
+  return foundInstance
+    ? { sap_system_id: foundInstance.sap_system_id, type: foundInstance.type }
+    : null;
+};
+
 const addTag = (tag, clusterId) => {
   post(`/clusters/${clusterId}/tags`, {
     value: tag,
@@ -36,6 +57,9 @@ const removeTag = (tag, clusterId) => {
 
 function ClustersList() {
   const clusters = useSelector((state) => state.clustersList.clusters);
+  const { applicationInstances, databaseInstances } = useSelector(
+    (state) => state.sapSystemsList
+  );
   const dispatch = useDispatch();
   const [searchParams, setSearchParams] = useSearchParams();
 
@@ -71,7 +95,23 @@ function ClustersList() {
         filterFromParams: true,
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
-        render: (_, { sid }) => sid.join(', '),
+        render: (_, { sid }) =>
+          sid.map((singleSid) => {
+            const linkData = getSapSystemLinkBySID(
+              applicationInstances,
+              databaseInstances,
+              singleSid
+            );
+
+            return (
+              <SapSystemLink
+                systemType={linkData ? linkData.type : null}
+                sapSystemId={linkData ? linkData.sap_system_id : null}
+              >
+                {singleSid}
+              </SapSystemLink>
+            );
+          }),
       },
       {
         title: 'Hosts',
--- a/assets/js/components/ClustersList.jsx
+++ b/assets/js/components/ClustersList.jsx
@@ -94,21 +94,25 @@ function ClustersList() {
         filterFromParams: true,
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
-        render: (_, { sid }) =>
-          sid.map((singleSid) => {
+        render: (_, { sid, id }) =>
+          sid.map((singleSid, index) => {
             const linkData = getSapSystemLinkBySID(
               applicationInstances,
               databaseInstances,
               singleSid
             );
             return (
-              <SapSystemLink
-                key={sid}
-                systemType={get(linkData, 'type', null)}
-                sapSystemId={get(linkData, 'sap_system_id', null)}
-              >
-                {singleSid}
-              </SapSystemLink>
+              <React.Fragment key={singleSid}>
+                <SapSystemLink
+                  key={id + singleSid}
+                  systemType={get(linkData, 'type', null)}
+                  sapSystemId={get(linkData, 'sap_system_id', null)}
+                >
+                  {singleSid}
+                </SapSystemLink>
+
+                {index !== sid.length - 1 && ', '}
+              </React.Fragment>
             );
           }),
       },
--- a/assets/js/components/ClustersList.jsx
+++ b/assets/js/components/ClustersList.jsx
@@ -102,7 +102,7 @@ function ClustersList() {
               singleSid
             );
             return (
-              <React.Fragment key={singleSid}>
+              <Fragment key={singleSid}>
                 <SapSystemLink
                   key={id + singleSid}
                   systemType={get(linkData, 'type', null)}
@@ -112,7 +112,7 @@ function ClustersList() {
                 </SapSystemLink>
 
                 {index !== sid.length - 1 && ', '}
-              </React.Fragment>
+              </Fragment>
             );
           }),
       },
--- a/assets/js/components/ClustersList.jsx
+++ b/assets/js/components/ClustersList.jsx
@@ -46,9 +46,6 @@ const removeTag = (tag, clusterId) => {
 
 function ClustersList() {
   const clusters = useSelector((state) => state.clustersList.clusters);
-  const { applicationInstances, databaseInstances } = useSelector(
-    (state) => state.sapSystemsList
-  );
   const concatenatedInstances = useSelector(getConcatenatedInstances());
   const dispatch = useDispatch();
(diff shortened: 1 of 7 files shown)


assets/js/components/ClustersList.jsx
```diff
@@ -71,7 +95,23 @@ function ClustersList() {
...
+            const linkData = getSapSystemLinkBySID(
+              applicationInstances,
+              databaseInstances,
+              singleSid
+            );
+
+            return (
+              <SapSystemLink
+                systemType={linkData ? linkData.type : null}
+                sapSystemId={linkData ? linkData.sap_system_id : null}
```
dottorblaster, line 109: You can use lodash's `get` function to get `type` and `sap_system_id` out of `linkData`

assets/js/components/ClustersList.jsx
```diff
@@ -71,7 +95,23 @@ function ClustersList() {
...
+        render: (_, { sid }) =>
+          sid.map((singleSid) => {
+            const linkData = getSapSystemLinkBySID(
+              applicationInstances,
+              databaseInstances,
+              singleSid
+            );
+
+            return (
+              <SapSystemLink
```
dottorblaster, line 107: `map`ping and returning JSX elements should require a key prop to be passed, can you do that?

assets/js/components/ClustersList.test.jsx
```diff
@@ -143,7 +143,8 @@ describe('ClustersList component', () => {
 
       renderWithRouter(StatefulClustersList);
 
-      expect(screen.getByText('HA1, HA2')).toBeVisible();
+      expect(screen.getByText('HA1')).toBeVisible();
+      expect(screen.getByText('HA2')).toBeVisible();
```
dottorblaster, line 147: Commas should be still there

dottorblaster (changes requested):

assets/js/components/ClustersList.jsx
```diff
@@ -24,6 +25,26 @@ const getClusterTypeLabel = (type) => {
...
+      databaseInstances.map((instance) => ({
+        ...instance,
+        type: 'databases',
+      }))
+    )
+    .find((instance) => instance.sid === sid);
+
+  return foundInstance
+    ? { sap_system_id: foundInstance.sap_system_id, type: foundInstance.type }
+    : null;
```
dottorblaster, line 45: You can directly return `foundInstance` without any fuss

dottorblaster (changes requested):

assets/js/components/ClustersList.jsx
```diff
@@ -24,6 +25,26 @@ const getClusterTypeLabel = (type) => {
   }
 };
 
+const getSapSystemLinkBySID = (
+  applicationInstances,
+  databaseInstances,
+  sid
+) => {
+  const foundInstance = applicationInstances
```
arbulu89, line 33: I see this pattern of concatenating the app and db instances repeated.
Maybe we could create a specific selector for it.
It would be more reusable and cleaner I would say

assets/js/components/SapSystemLink.jsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
+import { EOS_WARNING_OUTLINED } from 'eos-icons-react';
```
arbulu89, line 2: Are we sure we want to change this component?
In the past, for this exact feature (adding a not registered tooltip to the link) we have created a new component, to not change the behaviour of this one (`ClusterNodeLink`).
Maybe we should follow the same pattern

assets/js/components/ClustersList.jsx
```diff
@@ -71,7 +95,23 @@ function ClustersList() {
...
+              databaseInstances,
+              singleSid
+            );
+
+            return (
+              <SapSystemLink
+                systemType={linkData ? linkData.type : null}
+                sapSystemId={linkData ? linkData.sap_system_id : null}
+              >
+                {singleSid}
```
arbulu89, line 111: What happens on multiple sids here?
Are they separated by spaces?

arbulu89 (changes requested):
Hey @rtorrero ,
Some comments.
Besides that, I'm a bit surprised with the implementation, because as far as I see, it doesn't follow what our backlog ticket asks:
```
- Add link to SID colum in Cluster overview
- When SAP system or HANA Database is not registered, remove link from SID column in Cluster overview 
```
(Anyway, I personally like more the current implementation behaviour, being honest)

assets/js/components/ClustersList.jsx
rtorrero: I've used the tests to render how it would look like:

Captura desde 2023-08-21 13-40-38

Captura desde 2023-08-21 13-50-23

assets/js/components/ClustersList.test.jsx
rtorrero: @dottorblaster see (first screenshot)->
https://github.com/trento-project/web/pull/1715/files#r1300067494

I guess that because of the svg element inbetween, the `screen.getByText` is not catching it. Is it ok to leave the test like this?

dottorblaster: Thanks @rtorrero, it's ok.

assets/js/components/SapSystemLink.jsx
rtorrero: As I was looking how to add this I noticed that we have already a bunch of of *Link components that to basically very similar things. I'm leaning more towards keeping less components and add more when we need them. Wdyt?

assets/js/components/ClustersList.jsx
rtorrero: After speaking with @dottorblaster we decided it might look better if we add the coma back. @jagabomb wdyt?
Captura desde 2023-08-21 15-37-22

assets/js/components/ClustersList.jsx
```diff
@@ -24,6 +28,12 @@ const getClusterTypeLabel = (type) => {
   }
 };
 
+const getSapSystemLinkBySID = (instances, sid) => {
```
arbulu89, line 31:
```
const getSapSystemLinkBySID = (instances, sid) =>
    instances.find((instance) => instance.sid === sid);
```

And I would rename to `getSapSysmtemBySID`

assets/js/components/ClustersList.jsx
```diff
@@ -71,7 +82,26 @@ function ClustersList() {
         filterFromParams: true,
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
-        render: (_, { sid }) => sid.join(', '),
+        render: (_, { sid, id }) =>
+          sid.map((singleSid, index) => {
```
arbulu89, line 86: I personally prefer the way it is done in the `HostsList.jsx` component, in the `SID` column, but well, they both do the same at the end.

arbulu89 (commented):
Some additional suggestions

assets/js/components/ClustersList.jsx
rtorrero: No strong opinion here but consistency is nice :-)

assets/js/components/ClustersList.jsx
```diff
@@ -71,7 +79,28 @@ function ClustersList() {
...
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
-        render: (_, { sid }) => sid.join(', '),
+        render: (_, { sid, id }) => {
+          const sidsArray = sid.map((singleSid, index) => [
+            index > 0 && ', ',
+            <SapSystemLink
+              key={`${id}-${singleSid}`}
+              systemType={get(
+                getSapSystemBySID(allInstances, singleSid),
```
arbulu89, line 88: We shouldn't use `getSapSystemBySID` twice, as it gets the same data.

assets/js/components/ClustersList.jsx
```diff
@@ -71,7 +79,28 @@ function ClustersList() {
         filterFromParams: true,
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
-        render: (_, { sid }) => sid.join(', '),
+        render: (_, { sid, id }) => {
+          const sidsArray = sid.map((singleSid, index) => [
+            index > 0 && ', ',
+            <SapSystemLink
+              key={`${id}-${singleSid}`}
+              systemType={get(
```
arbulu89, line 87: Is not the lodash `get` an overkill?
`system?.type` would do the same.
Anyway, as this comes from other discussion, feel free to use it

assets/js/components/HostsList.jsx
```diff
@@ -48,8 +41,7 @@ const removeTag = (tag, hostId) => {
 function HostsList() {
   const hosts = useSelector((state) => state.hostsList.hosts);
   const clusters = useSelector((state) => state.clustersList.clusters);
-  const { applicationInstances } = useSelector((state) => state.sapSystemsList);
-  const { databaseInstances } = useSelector((state) => state.databasesList);
+  const concatenatedInstances = useSelector(getAllInstances());
```
arbulu89, line 44: Maybe better `allInstances`

assets/js/components/SapSystemLink.test.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { screen, render, fireEvent as userEvent } from '@testing-library/react';
```
arbulu89, line 2: Man! What is this? hehe
https://testing-library.com/docs/ecosystem-user-event/

assets/js/state/selectors/index.js
```diff
@@ -91,3 +91,16 @@ export const getInstancesOnHost = (hostId) => (state) => {
 
   return [...foundApplicationInstances, ...foundDatabaseInstances];
 };
+
+export const getAllInstances = () => (state) => {
```
arbulu89, line 95: I know that I'm coming again with this, but maybe `getAllSAPInstances` would be more meaningful.
Wdyt?

arbulu89 (changes requested):
Just putting read for the `userEvent` thing XD

assets/js/components/SapSystemLink.test.jsx
rtorrero: oh god, vscode trying to be smart, sorry about that

rtorrero: (and the person driving it :sweat_smile: )

arbulu89 (approved):
Thanks @rtorrero 
All good from my side!
PD: I guess this needs to be extrapolated to the HANA and ASCS/ERS cluster views

dottorblaster (approved):
Go go go!

PR #1717, opened 2023-08-17
Upgrade to Storybook 7.3.1

dottorblaster:
# Description

Just upgrading as a whole to Storybook 7.3.1 because of failing CI on main


EMaksy (approved):
Thanks for taking care of the upgrade, LGTM :D

PR #1731, opened 2023-08-22
Add ability to mark instances as absent

jamie-suse:
# Description

When an instance is no longer responding, the corresponding `MarkDatabaseInstanceAbsent`/`MarkApplicationInstanceAbsent` command is dispatched (instead of `Deregister*Instance`). This even then applies an `absent` marker on the corresponding read/write models.

When an instance is registered, the `absent` field should be cleared.

## How was this tested?

Updated unit tests to suit behaviour.


lib/trento/domain/sap_system/commands/mark_application_instance_absent.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Trento.Domain.Commands.MarkApplicationInstanceAbsent do
+  @moduledoc """
+  Mark an application instance as absent from the monitoring system.
```
arbulu89, line 3: I don't really like this `from the monitoring system`. It is kind of implicit, isn't it?

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -234,6 +240,9 @@ defmodule Trento.Domain.SapSystem do
     |> Multi.execute(fn _ ->
       maybe_emit_database_instance_health_changed_event(instance, command)
     end)
+    |> Multi.execute(fn _ ->
```
arbulu89, line 243: I think this function should come after the `maybe_emit_database_instance_registered_event`.
It looks more logical, and we spare an unnecessary event emission (this one, as the next event is putting this value to false in this case in the 1st registration, right)

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -244,20 +253,28 @@ defmodule Trento.Domain.SapSystem do
   # Restore sap system
   # Same registration rules
   def execute(
-        %SapSystem{deregistered_at: deregistered_at} = sap_system,
-        %RegisterApplicationInstance{} = instance
+        %SapSystem{
+          application: %Application{instances: instances},
+          deregistered_at: deregistered_at
+        } = sap_system,
+        %RegisterApplicationInstance{host_id: host_id, instance_number: instance_number} = command
```
arbulu89, line 260: I wouldn't rename this to `command`. It was ok as `instance`

jamie-suse: I thought to rename it as I need to use `instance = get_instance(...)` further down

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -244,20 +253,28 @@ defmodule Trento.Domain.SapSystem do
...
       )
       when not is_nil(deregistered_at) do
+    instance = get_instance(instances, host_id, instance_number)
+
     sap_system
     |> Multi.new()
+    |> Multi.execute(fn _ ->
+      maybe_emit_application_instance_marked_present_event(instance, command)
```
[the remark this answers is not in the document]
jamie-suse: I was trying to make this in-line with the existing logic in  `execute(%SapSystem{}, %RegisterDatabaseInstance{})`. But I've changed it :+1:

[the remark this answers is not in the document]
jamie-suse: Referring to the logic seen here: https://github.com/trento-project/web/blob/bac7caa68fd67b6d7d4576b3f8306531be21afbd/lib/trento/domain/sap_system/sap_system.ex#L223

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -300,6 +315,54 @@ defmodule Trento.Domain.SapSystem do
...
+        %SapSystem{sap_system_id: sap_system_id, database: %Database{instances: instances}},
+        %MarkDatabaseInstanceAbsent{
+          instance_number: instance_number,
+          host_id: host_id,
+          absent_at: absent_at
+        }
+      ) do
+    %Instance{absent_at: instance_absent_at} = get_instance(instances, host_id, instance_number)
+
+    case instance_absent_at do
```
arbulu89, line 328:
```
case get_instance(instances, host_id, instance_number) do
    %Instance{absent_at: nil} ->
        %DatabaseInstanceMarkedAbsent{
          instance_number: instance_number,
          host_id: host_id,
          sap_system_id: sap_system_id,
          absent_at: absent_at
        }

_ ->
    nil
end
```

jamie-suse:
FYI I had to rebase, so I squashed most commits down

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4208,361 @@ defmodule Trento.SapSystemTest do
     end
   end
 
+  describe "delta deregistration" do
```
arbulu89, line 4211: I wouldn't call this `delta deregistration`.
Maybe `instance marked absent/present` or similar?

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4208,361 @@ defmodule Trento.SapSystemTest do
...
+    test "newly registered instances should be marked as present" do
```
arbulu89, line 4212: Do we need this 1st test?
We have plenty of tests where we have a regular registration. We could add the `absent_at` pattern matching in any of them.

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4208,361 @@ defmodule Trento.SapSystemTest do
...
+          sap_system_id: sap_system_id,
+          sid: sid
+        ),
+        build(:application_instance_registered_event,
+          sap_system_id: sap_system_id,
+          sid: sid,
+          features: "MESSAGESERVER",
+          host_id: host_id
+        ),
+        build(
```
arbulu89, line 4235: This is not correct. We shouldn't have a `sap_system_registered` event unless we have an application server as well

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4208,361 @@ defmodule Trento.SapSystemTest do
...
+                         absent_at: nil
+                       }
+                     ]
+                   }
+                 } = state
+        end
+      )
+    end
+
+    test "absent instances receiving 'mark as absent' command should mark as absent" do
```
arbulu89, line 4415: I would put this test before the others.
The tests can tell the story of the flow as well, so we would set to absent before setting it back to present, right?

arbulu89, line 4415: I would rename this to `should mark as absent a previously registered instance`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4208,361 @@ defmodule Trento.SapSystemTest do
...
+      ensa_version = EnsaVersion.ensa1()
+      host_id = Faker.UUID.v4()
+      absent_db_instance_number = "01"
+      present_db_instance_number = "02"
+      absent_app_instance_number = "03"
+      present_app_instance_number = "04"
+      absent_db_absent_at = DateTime.utc_now()
+      absent_app_absent_at = DateTime.utc_now()
+
+      initial_events = [
```
arbulu89, line 4427: Aren't we mixing to many things here?
Wouldn't be better to have one tests for the databases and other for the applications?
After all they are totally different code and flows?

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -1199,4 +1385,17 @@ defmodule Trento.Domain.SapSystem do
         false
     end)
   end
+
+  defp update_instance(instances, instance_number, host_id, key, value) do
+    Enum.map(
+      instances,
+      fn instance ->
+        if instance.instance_number == instance_number and instance.host_id == host_id do
```
arbulu89, line 1393: Can we pattern match instead of this `if` ?

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -1199,4 +1385,17 @@ defmodule Trento.Domain.SapSystem do
         false
     end)
   end
+
+  defp update_instance(instances, instance_number, host_id, key, value) do
```
arbulu89, line 1389: I don't really like this `key/value` thing.
Could we send the whole new update and use maybe `Map.merge`?
https://hexdocs.pm/elixir/1.12/Map.html#merge/2

test/trento/domain/sap_system/sap_system_test.exs
arbulu89: If you still want to keep it, i think we should still split the database and application parts.
But the test must be fixed in any case.
The `database_instance_registered` must not go in the initial events, as this is the event which triggers the change, so I would suggest to use the command in the 2nd param of `assert_events_and_state`

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -95,15 +101,15 @@ defmodule Trento.Domain.SapSystem do
...
-    field :sap_system_id, Ecto.UUID
-    field :sid, :string, default: nil
-    field :health, Ecto.Enum, values: Health.values()
-    field :ensa_version, Ecto.Enum, values: EnsaVersion.values(), default: EnsaVersion.no_ensa()
-    field :rolling_up, :boolean, default: false
-    field :deregistered_at, :utc_datetime_usec, default: nil
-
-    embeds_one :database, Database
-    embeds_one :application, Application
+    field(:sap_system_id, Ecto.UUID)
```
arbulu89, line 104: Formatting error. Remove parenthesis please

lib/trento/domain/sap_system/sap_system.ex
```diff
@@ -1199,4 +1385,17 @@ defmodule Trento.Domain.SapSystem do
         false
     end)
   end
+
+  defp update_instance(instances, instance_number, host_id, fields) do
+    Enum.map(
+      instances,
+      fn
+        %Instance{instance_number: ^instance_number, host_id: ^host_id} = instance ->
+          struct(instance, fields)
```
arbulu89, line 1394: Maybe I would put `Kernel.struct` to exactly know the module. Up to you

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4210,365 @@ defmodule Trento.SapSystemTest do
...
+          sap_system_id: sap_system_id,
+          sid: sid
+        ),
+        build(
+          :database_instance_registered_event,
+          sap_system_id: sap_system_id,
+          sid: sid,
+          host_id: host_id,
+          instance_number: absent_db_instance_number,
+          system_replication: nil,
```
arbulu89, line 4234: Do we need this `nil`s in the `system_replication` fields?
The factory should populate things you don't really need, right?

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4210,365 @@ defmodule Trento.SapSystemTest do
     end
   end
 
+  describe "instance marked absent/present" do
+    test "should mark as absent a previously registered database instance" do
```
arbulu89, line 4214: I'm not sure if mixing 2 different flows is a good idea, as it "hides" them.
I'm talking about the scenario where you test here that an already marked instance is not re-marked.
No need for a change, the test is totally correct.
For the future

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4210,365 @@ defmodule Trento.SapSystemTest do
...
+          host_id: host_id,
+          instance_number: "01",
+          system_replication: nil,
+          system_replication_status: nil
+        ),
+        build(:application_instance_registered_event,
+          sap_system_id: sap_system_id,
+          sid: sid,
+          host_id: host_id,
+          instance_number: absent_app_instance_number,
```
arbulu89, line 4328: This is not an `app` instance in reallity, as it is a `MESSAGESERVER`

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4210,365 @@ defmodule Trento.SapSystemTest do
...
+                         absent_at: ^absent_app_absent_at
+                       }
+                     ]
+                   }
+                 } = state
+        end
+      )
+    end
+
+    test "receiving a 'register instance' command for an already-registered database instance sets instance as present if absent" do
```
arbulu89, line 4393: I would rephrase this to:
```
should mark as present and already registered absent database instance
```
We use the `should` wording at the beginning of each test by default

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4210,365 @@ defmodule Trento.SapSystemTest do
...
+          %DatabaseInstanceMarkedAbsent{
+            instance_number: present_db_instance_number,
+            host_id: host_id,
+            sap_system_id: sap_system_id,
+            absent_at: absent_at
+          }
+        ],
+        fn state ->
+          assert %SapSystem{
+                   database: %SapSystem.Database{
```
arbulu89, line 4283: `SapSystem.Database` to `Database` (already aliased in top).
The same for `Instance` and `Application` in all tests

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -4202,6 +4210,365 @@ defmodule Trento.SapSystemTest do
...
+          host_id: host_id,
+          instance_number: absent_app_instance_number,
+          features: "ABAP"
+        ),
+        build(
+          :application_instance_marked_absent_event,
+          sap_system_id: sap_system_id,
+          host_id: host_id,
+          instance_number: absent_app_instance_number,
+          absent_at: DateTime.utc_now()
```
arbulu89, line 4509: Shouldn't this `absent_at` be created by the factory?

arbulu89 (changes requested):
Some tiny things.
Ready to merge after them

arbulu89 (approved):
Alright!
Green light from my side.
@rtorrero would you like to give a second look if I missed something?

rtorrero (approved):

PR #1733, opened 2023-08-22
Forbid not registered instances deregistration

arbulu89:
# Description
Now that SAP instances deregistration will be triggered by external users, i want to enforce some better rules.
The application and database instances cannot be deregistered if they are not registered already.

PD: the new `application_instance_not_registered` and `database_instance_not_registered` atoms will be added to the fallback controller in next PRs, when they are really used

## How was this tested?

Tested

### DIFF
--- a/lib/trento/domain/sap_system/sap_system.ex
+++ b/lib/trento/domain/sap_system/sap_system.ex
@@ -1020,41 +1020,6 @@ defmodule Trento.Domain.SapSystem do
     end
   end
 
-  defp maybe_emit_application_instance_deregistered_event(
-         %SapSystem{application: nil},
-         %DeregisterApplicationInstance{}
-       ),
-       do: {:error, :application_instance_not_registered}
-
-  defp maybe_emit_application_instance_deregistered_event(
-         %SapSystem{application: %Application{instances: []}},
-         %DeregisterApplicationInstance{}
-       ),
-       do: {:error, :application_instance_not_registered}
-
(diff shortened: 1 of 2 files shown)


CDimonaco (approved):
LGTM for me, just a nitpick we could order the clauses `maybe_emit_application_instance_deregistered_event` `maybe_emit_database_instance_deregistered_event`, to match the actual usage in the aggregate (invert the current order)

But it's really up to you, no strong opinions

arbulu89:
> LGTM for me, just a nitpick we could order the clauses `maybe_emit_application_instance_deregistered_event` `maybe_emit_database_instance_deregistered_event`, to match the actual usage in the aggregate (invert the current order)
> 
> But it's really up to you, no strong opinions

Oh, yes, you are right! Good catch

CDimonaco:
> > LGTM for me, just a nitpick we could order the clauses `maybe_emit_application_instance_deregistered_event` `maybe_emit_database_instance_deregistered_event`, to match the actual usage in the aggregate (invert the current order)
> > But it's really up to you, no strong opinions
> 
> Oh, yes, you are right! Good catch. This should be a strong opinion indeed hehe As we should enforce the order with precedence of occurrence

You're right, I was not strong on that because in the end they are two different functions and the clauses between same functions are already ordered, anyway lgtm, good job!

rtorrero:
Thanks @arbulu89, looks good to me :+1:

PR #1737, opened 2023-08-23
Allow additional classes to be specified on the clean up button

rtorrero:
# Description
This PR:
 - Allows the CleanUpButton component to be provided with additional classes
 - Changes the HostList to remove the outlining of the Clean Up button by providing additional classes

This is also needed for the upcoming PR to include the Clean Up button in the sap system instances
Captura desde 2023-08-23 10-12-57

### DIFF
--- a/assets/js/components/Button/Button.jsx
+++ b/assets/js/components/Button/Button.jsx
@@ -5,6 +5,8 @@ const getSizeClasses = (size) => {
   switch (size) {
     case 'small':
       return 'py-1 px-2 text-sm';
+    case 'fit':
+      return 'text-sm';
     default:
       return 'py-2 px-4 text-base';
   }
--- a/assets/js/components/CleanUpButton/CleanUpButton.jsx
+++ b/assets/js/components/CleanUpButton/CleanUpButton.jsx
@@ -1,17 +1,21 @@
 import React from 'react';
 
 import { EOS_CLEANING_SERVICES } from 'eos-icons-react';
+import classNames from 'classnames';
 
 import Button from '@components/Button';
 import Spinner from '@components/Spinner';
 
-function CleanUpButton({ cleaning, onClick, className = '' }) {
-  const baseClasses = 'inline-block mx-0.5 border-green-500 border w-fit';
+function CleanUpButton({ cleaning, onClick, className }) {
+  const buttonClasses = classNames(
+    'inline-block mx-0.5 border-green-500 border w-fit',
+    className
+  );
 
   return (
     <Button
       type="primary-white"
-      className={`${baseClasses} ${className}`}
+      className={buttonClasses}
       size="small"
       disabled={cleaning}
       onClick={() => onClick()}
--- a/assets/js/components/CleanUpButton/CleanUpButton.jsx
+++ b/assets/js/components/CleanUpButton/CleanUpButton.jsx
@@ -5,11 +5,13 @@ import { EOS_CLEANING_SERVICES } from 'eos-icons-react';
 import Button from '@components/Button';
 import Spinner from '@components/Spinner';
 
-function CleanUpButton({ cleaning, onClick }) {
+function CleanUpButton({ cleaning, onClick, className = '' }) {
+  const baseClasses = 'inline-block mx-0.5 border-green-500 border w-fit';
+
   return (
     <Button
       type="primary-white"
-      className="inline-block mx-0.5 border-green-500 border w-fit"
+      className={`${baseClasses} ${className}`}
       size="small"
       disabled={cleaning}
       onClick={() => onClick()}
--- a/assets/js/components/CleanUpButton/CleanUpButton.jsx
+++ b/assets/js/components/CleanUpButton/CleanUpButton.jsx
@@ -6,7 +6,7 @@ import classNames from 'classnames';
 import Button from '@components/Button';
 import Spinner from '@components/Spinner';
 
-function CleanUpButton({ cleaning, onClick, className }) {
+function CleanUpButton({ className, size = 'small', cleaning, onClick }) {
   const buttonClasses = classNames(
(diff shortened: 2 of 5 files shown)


assets/js/components/CleanUpButton/CleanUpButton.jsx
```diff
@@ -5,11 +5,13 @@ import { EOS_CLEANING_SERVICES } from 'eos-icons-react';
...
 
-function CleanUpButton({ cleaning, onClick }) {
+function CleanUpButton({ cleaning, onClick, className = '' }) {
+  const baseClasses = 'inline-block mx-0.5 border-green-500 border w-fit';
+
   return (
     <Button
       type="primary-white"
-      className="inline-block mx-0.5 border-green-500 border w-fit"
+      className={`${baseClasses} ${className}`}
```
arbulu89, line 14: Can we use `classNames`?
`import classNames from 'classnames';`

assets/js/components/CleanUpButton/CleanUpButton.jsx
```diff
@@ -5,11 +5,13 @@ import { EOS_CLEANING_SERVICES } from 'eos-icons-react';
 import Button from '@components/Button';
 import Spinner from '@components/Spinner';
 
-function CleanUpButton({ cleaning, onClick }) {
+function CleanUpButton({ cleaning, onClick, className = '' }) {
```
arbulu89, line 8: You don't need to default to empty string I think.
I would work without default as well

arbulu89 (commented):
Could you add a new story book entry with the outlined example?

arbulu89 (approved):
Thank you!
I don't know if @jagabomb wants to have a look before merging

assets/js/components/CleanUpButton/CleanUpButton.jsx
```diff
@@ -1,15 +1,21 @@
...
+  );
+
   return (
     <Button
       type="primary-white"
-      className="inline-block mx-0.5 border-green-500 border w-fit"
+      className={buttonClasses}
       size="small"
       disabled={cleaning}
       onClick={() => onClick()}
```
CDimonaco, line 21: Just a comment, when we have this "dummy" components, without any particular logic on the event handlers, for example take a modal which has on the `onClose` prop the id of something passed like `onClose(id)`, that is a complex handler.

In that button we have a direct onclick handler, without any logic and it's a common pratice, but totally optional, to pass directly the handler, like `onClick={onClick}`, so the entire event chain of the dom event is passed up to the parent. The parent can listen if want to the event as first argument of the onClick prop callback.

This is completely optional in that case, there are different opinions about that, for sure it's true in pure design systems component, take bootstrap for example, you want to give the callee the direct facade to the html event.

If you want change it, otherwise leave as is and we could have this comment as a future discussion.

CDimonaco (approved):
Good job, LGTM

The comment is completely up to you, we can leave it for future reference and discussion, feel free to merge :D

jagabomb:
> @rtorrero @jagabomb
> I have to confess that for me this implentation is a step back.
> I think that we could achieve the same keeping the button using some styles.
> If using the link is mandatory, i would create other component. This `if` thing looks a bad implementation.
> It doesn't have the spinning state neither

@arbulu89 I think you might be right, maybe using the Button component is overkill and maybe we need a simpler TextLink component which does the same thing as a the Button component but removes the additional padding? Just a thought from my side.

CDimonaco:
@arbulu89 I'm with you, I approved before but the new implementation with if, looks a little bit clumsy

assets/js/components/CleanUpButton/CleanUpButton.stories.jsx
```diff
@@ -26,3 +33,10 @@ export const Cleaning = {
     cleaning: true,
   },
 };
+
+export const NoOutline = {
+  args: {
+    ...Default.args,
+    className: 'border-none shadow-none',
```
arbulu89, line 40: Could we put `size: 'fit'` as the example

assets/js/components/CleanUpButton/CleanUpButton.stories.jsx
```diff
@@ -13,6 +13,13 @@ export default {
       },
     },
     onClick: { action: 'Click button' },
+    className: {
```
arbulu89, line 16: Add `size` to the options

jagabomb:
> @jagabomb i've updated the component to add a variant with a size setting that allows to fit the button to the available size. If this still continues to be an issue, do you mind if we solve it in a separate PR to avoid stalling this? I need this component to submit the next PR and I think there we will be able to show how it looks and if the issue is a problem. WDYT?
> 
> PS: I can later add an additional component, something like but I'd need a bit of time to do a better implementation

@rtorrero I am happy for us to proceed. My feedback is all proactive so it can always be addressed later.

assets/js/components/CleanUpButton/CleanUpButton.jsx
arbulu89: Actually, now that I read this, I'm wondering how it works (JS mysteries to me XD).
Because in the component, we have `() => onClick()`, but in the call to the component we have:
```
onClick={() => {
    openDeregistrationModal(item);
    openDeregistrationModal(item);
 }}
```
so it is kind of done twice.
Sorry for my JS dummy questions hehe
Maybe it is totally normal

jagabomb (approved):
LGTM, feedback will be addressed later.

assets/js/components/Button/Button.jsx
```diff
@@ -5,6 +5,8 @@ const getSizeClasses = (size) => {
   switch (size) {
     case 'small':
       return 'py-1 px-2 text-sm';
+    case 'fit':
```
arbulu89, line 8: We could even add a story for this

arbulu89 (approved):
Good to go from my side

PR #1738, opened 2023-08-23
Use ProviderLabel on sap and db detail views

arbulu89:
# Description
Simple fix to show the provider with the icon. If the provider value is not given a blank value is used (as in the hosts list view).
Have a look on storybook

image


nelsonkopliku (approved):
Noice!

dottorblaster (approved):
LGTM

PR #1740, opened 2023-08-23
Host checks execution api

nelsonkopliku:
# Description

Disclaimer: don't be scared, mainly tests

This PR exposes a new endpoint to request checks execution for a host.
- an execution can be started only for registered hosts
- an execution can be started only for non empty selections (cluster endpoint needs to be fixed about this)
- target type is provided to the ExecutionRequested message
- target type is handled on ExecutionStarted and ExecutionCompleted
- unsupported target types are handled

## How was this tested?

Automated tests ajdusted/added plus manual test in integration with wanda.

### DIFF
--- a/lib/trento/application/integration/checks/checks.ex
+++ b/lib/trento/application/integration/checks/checks.ex
@@ -1,8 +1,10 @@
 defmodule Trento.Integration.Checks do
   @moduledoc """
-  Checks runner service integration
+  Checks Engine service integration
   """
 
+  alias Trento.Domain.Commands.CompleteChecksExecution
+  alias Trento.Domain.Enums.Health
   alias Trento.Infrastructure.Messaging
 
   alias Trento.Checks.V1.{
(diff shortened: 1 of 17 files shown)


lib/trento/application/integration/checks/checks.ex
```diff
@@ -38,10 +62,39 @@ defmodule Trento.Integration.Checks do
...
+      correlation_id: execution_id
+    )
+  end
+
+  def complete_execution(_execution_id, _target_id, _health, :host) do
+    # TODO dispatch host execution completed command
+    :ok
+  end
+
+  def complete_execution(_, _, _, _), do: {:error, :target_not_supported}
```
EMaksy, line 83: Also here the same idea like above :D

dottorblaster (approved):
LGTM

EMaksy (approved):
Hey man great stuff, a big thanks!
Just some small questions as comments, but besides this nothing to add!
LGTM! :+1:

PR #1745, opened 2023-08-24
Refactor sap system and database reducers

arbulu89:
# Description
A small refactor of the SAP system and Database reducers code.
I have introduced a new `updateInstance` function to reduce the amount of duplicated code.
I have come up with this as I need some new functions for new stories implementation, and the duplicated code was getting big.

Besides that, some reordering of functions, to have 1st the main SAP system and database function, and later on their instances functions.

## How was this tested?

Tests added to the modified functions

### DIFF
--- a/assets/js/state/databases.js
+++ b/assets/js/state/databases.js
@@ -1,9 +1,5 @@
 import { createSlice } from '@reduxjs/toolkit';
-import {
-  upsertInstances,
-  payloadMatchesInstance,
-  updateInstance,
-} from './instances';
+import { instancesMatch, upsertInstances, updateInstance } from './instances';
 
 const initialState = {
   loading: false,
@@ -73,8 +69,7 @@ export const databasesListSlice = createSlice({
     },
     removeDatabaseInstance: (state, { payload: instance }) => {
       state.databaseInstances = state.databaseInstances.filter(
-        (databaseInstance) =>
-          !payloadMatchesInstance(databaseInstance, instance)
+        (databaseInstance) => !instancesMatch(databaseInstance, instance)
       );
     },
     updateDatabaseInstanceHealth: (state, { payload: instance }) => {
--- a/assets/js/state/databases.js
+++ b/assets/js/state/databases.js
@@ -1,5 +1,9 @@
 import { createSlice } from '@reduxjs/toolkit';
-import { upsertInstances, maybeUpdateInstanceHealth } from './instances';
+import {
+  upsertInstances,
+  payloadMatchesInstance,
+  updateInstance,
+} from './instances';
 
 const initialState = {
   loading: false,
@@ -27,12 +31,6 @@ export const databasesListSlice = createSlice({
     appendDatabase: (state, action) => {
       state.databases = [...state.databases, action.payload];
     },
-    upsertDatabaseInstances: (state, action) => {
-      state.databaseInstances = upsertInstances(
-        state.databaseInstances,
-        action.payload
-      );
-    },
     removeDatabase: (state, { payload: { id } }) => {
       state.databases = state.databases.filter(
         (database) => database.id !== id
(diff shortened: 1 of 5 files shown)


dottorblaster (approved):
Maybe in the future we could consider moving `instances.js` in a `@lib` package on its own, what do you think?

assets/js/state/databases.js
```diff
@@ -99,6 +65,34 @@ export const databasesListSlice = createSlice({
...
+        state.databaseInstances,
+        action.payload
+      );
+    },
+    removeDatabaseInstance: (state, { payload }) => {
+      state.databaseInstances = state.databaseInstances.filter(
+        (databaseInstance) => !payloadMatchesInstance(databaseInstance, payload)
+      );
+    },
+    updateDatabaseInstanceHealth: (state, { payload }) => {
```
CDimonaco, line 79: Could we rename the parameter from `payload` to `instance` is more understandable, instead of passing pieces of the payload parameter and the parameter itself.

The explanation on the signature of upsert/update instance is clear, but when we use it we could be more explicit

assets/js/state/databases.js
```diff
@@ -99,6 +65,34 @@ export const databasesListSlice = createSlice({
...
+      state.databaseInstances = updateInstance(
+        state.databaseInstances,
+        payload,
+        { health: payload.health }
+      );
+    },
+    updateDatabaseInstanceSystemReplication: (state, { payload }) => {
+      state.databaseInstances = updateInstance(
```
CDimonaco, line 87: Same as other

assets/js/state/databases.js
```diff
@@ -99,6 +65,34 @@ export const databasesListSlice = createSlice({
...
     },
+    upsertDatabaseInstances: (state, action) => {
+      state.databaseInstances = upsertInstances(
+        state.databaseInstances,
+        action.payload
+      );
+    },
+    removeDatabaseInstance: (state, { payload }) => {
+      state.databaseInstances = state.databaseInstances.filter(
+        (databaseInstance) => !payloadMatchesInstance(databaseInstance, payload)
```
CDimonaco, line 76: Same comment

assets/js/state/databases.test.js
```diff
@@ -71,4 +73,57 @@ describe('Databases reducer', () => {
...
+
+  it('should update the health of a database instance', () => {
+    const instance = databaseInstanceFactory.build();
+    const newHealth = 'newHealth';
+
+    const initialState = {
+      databaseInstances: [instance],
+    };
+
+    const payload = {
```
CDimonaco, line 85: Even in the tests

assets/js/state/sapSystems.js
```diff
@@ -146,24 +86,70 @@ export const sapSystemsListSlice = createSlice({
         return sapSystem;
       });
     },
-    removeSAPSystem: (state, { payload: { id } }) => {
-      state.sapSystems = state.sapSystems.filter(
-        (sapSystem) => sapSystem.id !== id
+    // When a new ApplicationInstanceRegistered comes in,
+    // it need to be appended to the list of the application instances of the relative sap system
+    upsertApplicationInstances: (state, action) => {
+      state.applicationInstances = upsertInstances(
```
CDimonaco, line 92: Same

assets/js/state/sapSystems.js
```diff
@@ -146,24 +86,70 @@ export const sapSystemsListSlice = createSlice({
...
+        state.applicationInstances,
+        action.payload
       );
+    },
+    // When a new DatabaseInstanceRegistered comes in,
+    // it need to be appended to the list of the database instances of the relative sap system
+    upsertDatabaseInstancesToSapSystem: (state, action) => {
+      state.databaseInstances = upsertInstances(
```
CDimonaco, line 100: Same

assets/js/state/sapSystems.js
```diff
@@ -146,24 +86,70 @@ export const sapSystemsListSlice = createSlice({
...
+        state.databaseInstances,
+        action.payload
+      );
+    },
+    removeApplicationInstance: (state, { payload }) => {
       state.applicationInstances = state.applicationInstances.filter(
-        (applicationInstance) => applicationInstance.sap_system_id !== id
+        (applicationInstance) =>
+          !payloadMatchesInstance(applicationInstance, payload)
```
CDimonaco, line 108: Same

assets/js/state/sapSystems.js
```diff
@@ -146,24 +86,70 @@ export const sapSystemsListSlice = createSlice({
...
     },
-    updateSAPSystem: (state, { payload }) => {
-      state.sapSystems = state.sapSystems.map((sapSystem) => {
-        if (sapSystem.id === payload.id) {
-          sapSystem = { ...sapSystem, ...payload };
+    updateApplicationInstanceHost: (
+      state,
+      { payload: { sap_system_id, old_host_id, new_host_id, instance_number } }
+    ) => {
+      state.applicationInstances = updateInstance(
```
CDimonaco, line 120: Same

CDimonaco (changes requested):
Very good refactoring thanks man!

Just the renaming of payload parameter everywhere in the reducer, I spent some time to understand what the function was doing, the function signature is clear but I think we can be more clear even when we use it :D 

Other than this, lgtm for me

arbulu89:
> Maybe in the future we could consider moving `instances.js` in a `@lib` package on its own, what do you think?

The functions are strictly attached on how the redux state is designed. I think it could be risky, as it would make it more generic. I think it should belong to the `state` folder, but as a refactor even for this is/was planned, I don't know how it would look like

assets/js/state/databases.js
arbulu89: You mean to have `updateDatabaseInstanceHealth: (state, { payload: instance }) => {`, right?

arbulu89: @CDimonaco Done!
I have taken the chance to rename the `payloadMatchesInstance` function to `instancesMatch`, as it is more meaningful

CDimonaco (approved):
Thanks for addressing the feedbacks! LGTM

PR #1746, opened 2023-08-24
Instance deregistration sagas

arbulu89:
# Description

Add instances deregistration sagas.

It skips by now the navigation part. I guess we should navigate from the SAP system/Database details view when all the instances are deregistered. But it is not trivial, so I think it will be better to work on that once we have the whole flow working, as i'm not sure how it will behave.

## How was this tested?

Tests added

### DIFF
--- a/assets/js/state/databases.js
+++ b/assets/js/state/databases.js
@@ -98,7 +98,7 @@ export const databasesListSlice = createSlice({
         }
       );
     },
-    setDatabaseInstanceNotDeregistering: (state, { payload: instance }) => {
+    unsetDatabaseInstanceDeregistering: (state, { payload: instance }) => {
       state.databaseInstances = updateInstance(
         state.databaseInstances,
         instance,
@@ -140,7 +140,7 @@ export const {
   addTagToDatabase,
   removeTagFromDatabase,
   setDatabaseInstanceDeregistering,
-  setDatabaseInstanceNotDeregistering,
+  unsetDatabaseInstanceDeregistering,
 } = databasesListSlice.actions;
 
 export default databasesListSlice.reducer;
--- a/assets/js/state/databases.js
+++ b/assets/js/state/databases.js
@@ -1,4 +1,4 @@
-import { createSlice } from '@reduxjs/toolkit';
+import { createAction, createSlice } from '@reduxjs/toolkit';
 import { instancesMatch, upsertInstances, updateInstance } from './instances';
 
 const initialState = {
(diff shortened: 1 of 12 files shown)


assets/js/state/databases.js
```diff
@@ -116,6 +139,8 @@ export const {
   updateDatabaseInstanceSystemReplication,
   addTagToDatabase,
   removeTagFromDatabase,
+  setDatabaseInstanceDeregistering,
+  setDatabaseInstanceNotDeregistering,
```
rtorrero, line 143: Not 100% sure that I understand the reasoning behind the naming properly, but wouldn't something like `unsetDatabaseInstanceDeregistering` be more accurate?

assets/js/state/sagas/databases.test.js
```diff
@@ -82,4 +102,52 @@ describe('SAP Systems sagas', () => {
...
+
+    expect(dispatched).toEqual([
+      setDatabaseInstanceDeregistering(instance),
+      setDatabaseInstanceDeregisteringToSAPSystem(instance),
+      setDatabaseInstanceNotDeregistering(instance),
+      setDatabaseInstanceNotDeregisteringToSAPSystem(instance),
+    ]);
+  });
+
+  it('should notify error on database instance deregistration request', async () => {
```
rtorrero, line 128: I think I'm missing something in the title that tells me why it should notify an error, something like:
`should notify error if 404 is received during database instance deregistration request`

wdyt?

assets/js/state/databases.js
rtorrero, line 143: Not 100% sure that I understand the reasoning behind the naming properly, but wouldn't something like unsetDatabaseInstanceDeregistering be more accurate and easier to read?

assets/js/state/sagas/sapSystems.test.js
```diff
@@ -92,4 +110,48 @@ describe('SAP Systems sagas', () => {
...
+      payload: instance,
+    });
+
+    expect(dispatched).toEqual([
+      setApplicationInstanceDeregistering(instance),
+      setApplicationInstanceNotDeregistering(instance),
+    ]);
+  });
+
+  it('should notify error on application instance deregistration request', async () => {
```
rtorrero, line 134: Same as commented before

rtorrero (approved):
Looks good to me, only minor comments on naming / tests description

jamie-suse (approved):

assets/js/state/sagas/databases.test.js
arbulu89: What do you think of?
`should notify an error on database instance deregistration request failure`.
Being 404 is just a test coincidence. It would notify the same error in other cases (422 for example)

assets/js/state/databases.js
arbulu89: I just replicated the naming we use in the host reducer.
In case we want to change, I think we should change all of them.
I'm not against changing anyway. Would you do it in the PR, or in a new one where I could replace all the cases?

arbulu89: I just changed all of them, as it is not a big deal. It was only this way in the host reducer files

assets/js/state/sagas/databases.test.js
rtorrero: Perfect :+1:

PR #1747, opened 2023-08-24
Implement instance style changes on absent case

rtorrero:
# Description

This PR just includes the UI changes to the instances list. Websocket handling of the event will come in a separate PR.

See following screenshots for an idea on how this impacts the sap systems overview and the databases overview:

sapsystems

hanadbs

### DIFF
--- a/assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
+++ b/assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
@@ -1,4 +1,3 @@
-import classNames from 'classnames';
 import { EOS_DATABASE_OUTLINED } from 'eos-icons-react';
 import React from 'react';
 import InstanceOverview from '@components/InstanceOverview';
@@ -15,22 +14,18 @@ const databaseInstanceColumns = [
   { name: 'System Replication' },
   { name: 'Cluster' },
   { name: 'Host' },
+  {},
 ];
 
 function PlainDatabaseItemOverview({ instances, asDatabaseLayer = false }) {
   return (
-    <div
-      className={classNames('flex bg-white dark:bg-gray-800 shadow mb-2', {
-        'rounded-lg': !asDatabaseLayer,
-        'rounded-b-lg': asDatabaseLayer,
-      })}
-    >
-      <div className="flex-auto p-6">
-        <div className="w-full text-gray-800 dark:text-white flex items-center transition-colors duration-200 justify-start pb-4">
+    <div className="flex bg-white dark:bg-gray-800 shadow mb-2 rounded-lg">
+      <div className="flex-auto">
+        <div className="w-full text-gray-800 dark:text-white flex items-center transition-colors duration-200 justify-start p-4">
           <span className="text-left">
-            <EOS_DATABASE_OUTLINED size={25} className="fill-blue-500" />
+            <EOS_DATABASE_OUTLINED size={25} className="fill-black-500" />
           </span>
-          <h2 className="mx-2">
+          <h2 className="mx-2 font-bold">
             {asDatabaseLayer ? 'Database Layer' : 'Database Instances'}
           </h2>
         </div>
--- a/assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
+++ b/assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
@@ -31,7 +31,7 @@ function PlainDatabaseItemOverview({ instances, asDatabaseLayer = false }) {
         </div>
         <div className="table w-full">
           <div className="table-header-group bg-grey bg-gray-100">
-            <div className="table-row">
+            <div className="table-row border-b">
               {databaseInstanceColumns.map(({ key, name, cssClass }) => (
                 <div
                   key={key}
--- a/assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
+++ b/assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
@@ -14,7 +14,7 @@ const databaseInstanceColumns = [
   { key: 'systemReplication', name: 'System Replication' },
   { key: 'cluster', name: 'Cluster' },
   { key: 'hostname', name: 'Host' },
-  { key: 'cleanupButton', cssClass: 'w-24' },
+  { key: 'cleanupButton', cssClass: 'w-48' },
 ];
 
 function PlainDatabaseItemOverview({ instances, asDatabaseLayer = false }) {
--- a/assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
+++ b/assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
@@ -8,13 +8,13 @@ export function DatabaseInstance({ instance }) {
 }
 
 const databaseInstanceColumns = [
-  { name: 'Health', cssClass: 'w-20' },
-  { name: 'Instance Nr', cssClass: 'w-24' },
-  { name: 'Features' },
-  { name: 'System Replication' },
-  { name: 'Cluster' },
-  { name: 'Host' },
-  {},
+  { key: 'health', name: 'Health', cssClass: 'w-20' },
+  { key: 'instanceNr', name: 'Instance Nr', cssClass: 'w-24' },
+  { key: 'features', name: 'Features' },
+  { key: 'systemReplication', name: 'System Replication' },
+  { key: 'cluster', name: 'Cluster' },
+  { key: 'hostname', name: 'Host' },
+  { key: 'cleanupButton', cssClass: 'w-24' },
 ];
 
 function PlainDatabaseItemOverview({ instances, asDatabaseLayer = false }) {
@@ -32,9 +32,9 @@ function PlainDatabaseItemOverview({ instances, asDatabaseLayer = false }) {
         <div className="table w-full">
           <div className="table-header-group bg-grey bg-gray-100">
             <div className="table-row">
-              {databaseInstanceColumns.map(({ name, cssClass }) => (
+              {databaseInstanceColumns.map(({ key, name, cssClass }) => (
                 <div
-                  key={name}
+                  key={key}
                   className={`table-cell p-2 text-left text-xs font-medium text-gray-500 uppercase ${cssClass}`}
                 >
                   {name}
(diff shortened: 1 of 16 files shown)


assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
```diff
@@ -15,22 +15,18 @@ const databaseInstanceColumns = [
   { name: 'System Replication' },
   { name: 'Cluster' },
   { name: 'Host' },
+  { name: ' ' },
```
arbulu89, line 18: Do we need this empty `name`. You could go with `{}` simply, or even better `{ cssClass: 'w-48' }` to have some fixed width (it can be any other of course)

assets/js/components/InstanceOverview/InstanceOverview.jsx
```diff
@@ -18,17 +20,24 @@ function InstanceOverview({
...
+    'table-row border-b'
+  );
 
   return (
-    <div className="table-row border-b">
-      <div className="table-cell p-2">
-        <HealthIcon health={health} />
+    <div className={rowClasses}>
+      <div className="table-cell p-2 px-5">
+        <HealthIcon health={absentAt ? 'absent' : health} />
```
arbulu89, line 40: We wanted to have a tooltip with some text when the instance was absent, right?

assets/js/components/InstanceOverview/InstanceOverview.jsx
```diff
@@ -52,6 +61,14 @@ function InstanceOverview({
       <div className="table-cell p-2">
         <HostLink hostId={hostId}>{host && host.hostname}</HostLink>
       </div>
+      {absentAt && (
+        <div className="table-cell p-2">
+          <CleanUpButton
+            size="fit"
+            className="bg-gray-100 border-none shadow-none"
+          ></CleanUpButton>
```
arbulu89, line 69: We can simply close the tag with ` />`

assets/js/components/InstanceOverview/InstanceOverview.jsx
```diff
@@ -18,17 +20,24 @@ function InstanceOverview({
...
     features,
     host_id: hostId,
+    absent_at: absentAt,
   },
 }) {
   const isDatabase = DATABASE_TYPE === instanceType;
-
   const host = useSelector(getHost(hostId));
   const cluster = useSelector(getCluster(host?.cluster_id));
+  if (instanceNumber == '01') {
```
arbulu89, line 29: As a reminder that this needs to go away

arbulu89 (commented):
An initial review.
In the screenshot I see that the database layer doesn't have space for the clean up button.
This might be due the fact that you don't put a fixed width as I suggest in one comment

PD:
I think this could be a good moment to maybe refactor this view to make it storybook friendly...
I find now the lack of it a big drawback.
It wouldn't be a big deal, as this views are pretty simple. It would be the same we do with other views adding the `...Page` suffix kind of component file

assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
rtorrero: I'd rather use your first suggestion of leaving it as `{}` as I think that having a fixed might use more space in the cell than what we actually need.

rtorrero:
Something that I still didn't manage is to vertically center the health icon in the instances list. It was like this before, but according to @jagabomb figma links, it should be centered

assets/js/components/InstanceOverview/InstanceOverview.jsx
```diff
@@ -17,14 +19,30 @@ function InstanceOverview({
...
+    'table-row border-b'
+  );
   return (
-    <div className="table-row border-b">
-      <div className="table-cell p-2">
-        <HealthIcon health={health} />
+    <div className={rowClasses}>
+      <div className="table-cell p-2 px-5">
+        {absentAt ? (
+          <Tooltip content="Instance currently not registered." place="bottom">
```
arbulu89, line 34: I would change text to `Instance current absent` or `Instance not found` (as in the mockup), because it is registered indeed

assets/js/components/InstanceOverview/InstanceOverview.jsx
```diff
@@ -17,14 +19,30 @@ function InstanceOverview({
...
+            <span
```
arbulu89, line 35: Why do we need this double span here? Cannot we simply put the icon component, you can put the `data-testid` there if you want I guess

assets/js/components/InstanceOverview/InstanceOverview.test.jsx
```diff
@@ -0,0 +1,90 @@
...
+import userEvent from '@testing-library/user-event';
+import '@testing-library/jest-dom';
+import { databaseInstanceFactory } from '@lib/test-utils/factories';
+import { renderWithRouter } from '@lib/test-utils';
+import { DATABASE_TYPE, APPLICATION_TYPE } from '@lib/model';
+import { faker } from '@faker-js/faker';
+import InstanceOverview from './InstanceOverview';
+
+describe('InstanceOverview', () => {
+  it('renders HealthIcon for a registered instance', () => {
```
arbulu89, line 12: funny test. Was it necessary? XD
Is there any scenario where we shouldn't have an icon?

assets/js/components/InstanceOverview/InstanceOverview.test.jsx
```diff
@@ -0,0 +1,90 @@
...
+    const srInstance = databaseInstanceFactory.build({
+      system_replication: 'Secondary',
+      system_replication_status: 'ACTIVE',
+    });
+
+    renderWithRouter(
+      <InstanceOverview instanceType={DATABASE_TYPE} instance={srInstance} />
+    );
+
+    const replicationText = screen.getByText('HANA Secondary');
```
arbulu89, line 35: We could check that `ACTIVE` is present as well

assets/js/components/InstanceOverview/InstanceOverview.test.jsx
arbulu89, line 12: we start all the tests with a `should` statement by convention (i didn't find any that we don't)

assets/js/components/InstanceOverview/InstanceOverview.test.jsx
```diff
@@ -0,0 +1,90 @@
...
+      <InstanceOverview
+        instanceType={APPLICATION_TYPE}
+        instance={absentInstance}
+      />
+    );
+
+    expect(screen.queryByRole('button', { name: 'Clean up' })).toBeVisible();
+  });
+
+  it('does not render a clean up button for registered instances', () => {
```
arbulu89, line 76: for `not absent` or `present instances` maybe?

assets/js/components/SapSystemsOverview/SapSystemItemOverview.jsx
```diff
@@ -16,20 +16,21 @@ const applicationInstanceColumns = [
   { name: 'Features' },
   { name: 'Cluster' },
   { name: 'Host' },
+  {},
```
arbulu89, line 19: The same to avoid the `key` issue

arbulu89 (changes requested):
Some comments.

- I'm already regretting the button styling using our Button component. It is making things over complicated
- In the mock up, I see that the texts (like instance number) are using a different gray color. I guess this affects to system replication texts I guess
- Can we add an absent row story for the `SapSystemsOverview` view?

assets/js/components/InstanceOverview/InstanceOverview.jsx
rtorrero: I basically followed how it was done elsewhere, I guess the reasoning is that our Tooltip component doesn't allow `data-testid`. If we add it, wouldn't that override the `data-testid` that EOS already adds by default? (`data-testid="eos-svg-component"`)

assets/js/components/InstanceOverview/InstanceOverview.test.jsx
rtorrero: Hope it makes more sense now

assets/js/components/DatabasesOverview/DatabaseItemOverview.jsx
```diff
@@ -15,22 +14,18 @@ const databaseInstanceColumns = [
   { name: 'System Replication' },
   { name: 'Cluster' },
   { name: 'Host' },
+  {},
```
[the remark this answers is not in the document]
rtorrero: you are right, I missed this

assets/js/components/InstanceOverview/InstanceOverview.jsx
arbulu89: I think the `Tooltip` usage you based this has other purposes, like using the span to put a text.
There are other examples where we don't have this `span`s.
Here you have an empty span which doesn't do anything.
You can perfectly pass the `data-testid` to `Tooltip` as it receives the all the props as `rest`

assets/js/components/SapSystemsOverview/SapSystemsOverview.stories.jsx
```diff
@@ -39,6 +40,44 @@ const enrichedDatabaseInstances = sapSystems[0].database_instances
     };
   });
 
+const sapSystemsWithAbsentInstances = sapSystemFactory.buildList(2);
+
+const enrichedAbsentApplicationInstances =
+  sapSystemsWithAbsentInstances[0].application_instances
+    .concat(sapSystemsWithAbsentInstances[1].application_instances)
+    .map((instance) => {
```
arbulu89, line 48: Now that this have this thing duplicated 4 times, time for a function?

arbulu89 (commented):
More things

assets/js/components/Table/CollapsibleTableRow.jsx
```diff
@@ -27,7 +27,7 @@ function CollapsibleTableRow({
       </tr>
       {collapsibleDetailRenderer && (
         <tr
-          className={`overflow-y-hidden ${
+          className={`bg-gray-100 overflow-y-hidden ${
```
arbulu89, line 30: @jagabomb A comment on this. 
Changing this will change our collapsible table rows, as this would be a default color now.
I'm not sure we want this...
You can check how it looks in `DatabasesOverview` in storybook for example.
Just for your information, as I'm a bit afraid of doing this "generic" changes for specific things

dottorblaster (commented):
Just some comments, I leave the final notes to @arbulu89, just wanted to interject on some topics

arbulu89 (approved):
Perfect!

jagabomb (approved):
LGTM, @rtorrero have discussed some small UI requests for further PRs.

PR #1758, opened 2023-08-28
Refactor sap system selectors

arbulu89:
# Description

Refactor sap systems/databases selectors to use `createSelector`. This will enable to update SapSystemsOverview and DatabasesOverview views to make storybook addition easy.

Besides that, I have moved the functions to a new `sapSystems` file, used lodash in some operations and removed the camelCase transformation that is only used in these selectors.

## How was this tested?

Tests added after refactoring


assets/js/state/selectors/cluster.js
```diff
@@ -24,6 +24,11 @@ export const getClusterName = (clusterID) => (state) => {
   return get(cluster, 'name', '');
 };
 
+export const getClusterByHost = (state, hostID) => {
```
arbulu89, line 27: Moved from `index`. Now it uses lodash find function

assets/js/state/selectors/sapSystems.js
```diff
@@ -0,0 +1,118 @@
...
+  state.sapSystemsList.sapSystems.find((sapSystem) => id === sapSystem.id);
+
+export const getDatabase = (id) => (state) =>
+  state.databasesList.databases.find((database) => id === database.id);
+
+export const getEnrichedSapSystemDetails = createSelector(
+  [
+    (id, state) => getSapSystem(id)(state),
+    (id, state) =>
+      enrichInstances(
```
arbulu89, line 47: I'm not sure if adding this operation here is a good idea, or it should better go in the body of the function

assets/js/state/selectors/sapSystem.js
```diff
@@ -0,0 +1,118 @@
...
+  (enrichedInstances) => enrichedInstances
+);
+
+export const getSapSystem = (id) => (state) =>
+  state.sapSystemsList.sapSystems.find((sapSystem) => id === sapSystem.id);
+
+export const getDatabase = (id) => (state) =>
+  state.databasesList.databases.find((database) => id === database.id);
+
+export const getEnrichedSapSystemDetails = createSelector(
```
arbulu89, line 43: In essence, the next `getEnrichedSapSystemDetails` and `getEnrichedDatabaseDetails` functions have been changes in the whole PR.

assets/js/state/selectors/sapSystem.js
```diff
@@ -0,0 +1,118 @@
...
+  [
+    (id, state) => getSapSystem(id)(state),
+    (id, state) =>
```
dottorblaster, line 46: Can you switch this to `(state, id)` form? Usually `state` comes as the first parameter when you use `createSelector`

dottorblaster (changes requested):
Just a thing then we can merge. Awesome work, thank you so much!

assets/js/state/selectors/cluster.js
dottorblaster: YEEEEEEEAH

assets/js/state/selectors/sapSystem.js
arbulu89: I can and I will!
Thank you for the tip

arbulu89: Done

dottorblaster (approved):
LGTM

rtorrero (approved):
Thanks @arbulu89, this looks much clearer now

PR #1775, opened 2023-08-31
Add `absent_at` field to projectors and broadcast events

jamie-suse:
# Description

This change makes the SAP system and database projectors listen to the `*MarkedAsAbsent`/`*MarkedAsPresent` events, whereupon they update the read models and broadcast this change through the websocket channel as events named `application_instance_absent_at_changed`/`database_instance_absent_at_changed`

## How was this tested?

Added unit tests for the new behaviour in the projectors

### DIFF
--- a/lib/trento_web/views/v1/sap_system_view.ex
+++ b/lib/trento_web/views/v1/sap_system_view.ex
@@ -97,6 +97,21 @@ defmodule TrentoWeb.V1.SapSystemView do
 
   def render("application_instance_health_changed.json", %{health: health}), do: health
 
+  def render("instance_absent_at_changed.json", %{
+        instance_number: instance_number,
+        host_id: host_id,
+        sap_system_id: sap_system_id,
+        sid: sid,
+        absent_at: absent_at
+      }),
+      do: %{
+        instance_number: instance_number,
+        host_id: host_id,
+        sap_system_id: sap_system_id,
+        sid: sid,
+        absent_at: absent_at
+      }
+
   def render("sap_systems.json", %{sap_systems: sap_systems}) do
     render_many(sap_systems, __MODULE__, "sap_system.json")
   end
--- a/test/trento/application/projectors/database_projector_test.exs
+++ b/test/trento/application/projectors/database_projector_test.exs
@@ -18,6 +18,8 @@ defmodule Trento.DatabaseProjectorTest do
     DatabaseHealthChanged,
     DatabaseInstanceDeregistered,
     DatabaseInstanceHealthChanged,
+    DatabaseInstanceMarkedAbsent,
+    DatabaseInstanceMarkedPresent,
     DatabaseInstanceSystemReplicationChanged,
     DatabaseRestored
   }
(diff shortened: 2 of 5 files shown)


lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -97,6 +97,21 @@ defmodule TrentoWeb.V1.SapSystemView do
 
   def render("application_instance_health_changed.json", %{health: health}), do: health
 
+  def render("instance_absent_at_changed.json", %{
```
arbulu89, line 100: I would follow the same pattern we use in others and wrap the data in a `instance` field, like what we do in `database_instance_health_changed.json` and `database_instance_system_replication_changed.json"`

arbulu89, line 100: Let's put this function at the end, just after `def render("instance_deregistered.json", %{`

test/trento/application/projectors/database_projector_test.exs
```diff
@@ -295,6 +297,67 @@ defmodule Trento.DatabaseProjectorTest do
                      1000
   end
 
+  test "should broadcast database_instance_absent_at_changed when DatabaseInstanceMarkedAbsent event is received" do
```
arbulu89, line 300: The main things under test is the projection, not the broadcasting.
Let's change the test text to:
`should update the absent_at field on DatabaseInstanceMarkedAbsent event is received`.

The same for others

test/trento/application/projectors/database_projector_test.exs
```diff
@@ -295,6 +297,67 @@ defmodule Trento.DatabaseProjectorTest do
...
+    )
+  end
+
+  test "should broadcast database_instance_absent_at_changed when DatabaseInstanceMarkedPresent event is received" do
+    %{
+      sap_system_id: sap_system_id,
+      instance_number: instance_number,
+      host_id: host_id,
+      sid: sid
+    } = insert(:database_instance_without_host)
```
arbulu89, line 338: Let's put some value to `absent_at` first:
```
insert(:database_instance_without_host, absent_at: DateTime.utc_now())
```

test/trento/application/projectors/sap_system_projector_test.exs
```diff
@@ -213,6 +215,85 @@ defmodule Trento.SapSystemProjectorTest do
     )
   end
 
+  test "should broadcast application_instance_absent_at_changed when ApplicationInstanceMarkedAbsent event is received" do
+    insert(:sap_system, id: sap_system_id = Faker.UUID.v4())
```
arbulu89, line 219:
```
%{id: sap_system_id} = insert(:sap_system)
```

test/trento/application/projectors/sap_system_projector_test.exs
```diff
@@ -213,6 +215,85 @@ defmodule Trento.SapSystemProjectorTest do
...
+    event = build(:application_instance_registered_event, sap_system_id: sap_system_id)
+
+    ProjectorTestHelper.project(SapSystemProjector, event, "sap_system_projector")
+
+    %{
```
arbulu89, line 224: This doesn't make sense. You already get this information from the factory, you don't need to re-query.
You can "copy/paste" what you did in the database projection tests

test/trento/application/projectors/sap_system_projector_test.exs
```diff
@@ -213,6 +215,85 @@ defmodule Trento.SapSystemProjectorTest do
     )
   end
 
+  test "should broadcast application_instance_absent_at_changed when ApplicationInstanceMarkedAbsent event is received" do
+    insert(:sap_system, id: sap_system_id = Faker.UUID.v4())
+    event = build(:application_instance_registered_event, sap_system_id: sap_system_id)
```
arbulu89, line 220: You don't need this event, right?

arbulu89 (changes requested):
Some tiny requests.
We are fine after those are fixed

test/trento/application/projectors/database_projector_test.exs
```diff
@@ -295,6 +297,67 @@ defmodule Trento.DatabaseProjectorTest do
                      1000
   end
 
+  test "should update the absent_at field on DatabaseInstanceMarkedAbsent event is received" do
```
arbulu89, line 300: it looks like we should replace `on` by `when`, right?

PR #1780, opened 2023-09-01
Host checks execution scheduler

EMaksy:
# Description
This Pr will execute host checks execution every 5 minutes similar to the cluster checks execution scheduler

## How was this tested?
Added automated test

### DIFF
--- a/test/trento/application/usecases/clusters_test.exs
+++ b/test/trento/application/usecases/clusters_test.exs
@@ -7,10 +7,15 @@ defmodule Trento.ClustersTest do
 
   import Trento.Factory
 
-  alias Trento.Clusters
+  alias Trento.{ClusterEnrichmentData, ClusterReadModel, Clusters}
 
-  alias Trento.ClusterEnrichmentData
-  alias Trento.ClusterReadModel
+  alias Trento.Checks.V1.{
+    ExecutionRequested,
+    Target
+  }
+
+  require Trento.Domain.Enums.ClusterType
+  require Logger
 
   setup [:set_mox_from_context, :verify_on_exit!]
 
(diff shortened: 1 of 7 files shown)


test/trento/application/usecases/clusters_test.exs
```diff
@@ -64,6 +70,39 @@ defmodule Trento.ClustersTest do
...
     end
+
+    test "should request cluster checks execution for all clusters" do
+      checks = [Faker.UUID.v4(), Faker.UUID.v4()]
+      %{id: cluster_id} = insert(:cluster, id: Faker.UUID.v4(), selected_checks: checks)
+      %{id: host_id1} = insert(:host, id: Faker.UUID.v4(), cluster_id: cluster_id)
+      %{id: host_id2} = insert(:host, id: Faker.UUID.v4(), cluster_id: cluster_id)
+
+      expect(Trento.Infrastructure.Messaging.Adapter.Mock, :publish, fn "executions", event ->
+        assert %ExecutionRequested{
```
dottorblaster, line 81: If you move this match directly inside the function signature I think it'll work the same

dottorblaster (changes requested):
Just one thing

test/trento/application/usecases/clusters_test.exs
EMaksy: Ah thanks, nice catch :D

lib/trento/application/usecases/clusters/clusters.ex
```diff
@@ -129,7 +129,7 @@ defmodule Trento.Clusters do
     |> select_merge([c, e], %{cib_last_written: e.cib_last_written})
   end
 
-  defp maybe_request_checks_execution(%{selected_checks: []}), do: :ok
+  defp maybe_request_checks_execution(%{selected_checks: []}), do: {:error, :no_checks_selected}
```
nelsonkopliku, line 132: Now, on an empty check selection the `ClusterController.request_checks_execution` operation returns 422 instead of 202, which is correct, desirable and still covered by the endpoint's openapi spec (falling in the 422 case)

I cannot but think, though, that the very same endpoint might start suddenly returning 422s where it before returned 202s and as a client being integrated it might be annoying.

What about tracking and deferring this? 
The `ClusterController.request_checks_execution` operation needs anyway an adjustment (for instance, it would never return a bad_request as defined in the spec)

We could introduce this change and bump the version of the endpoint, if needed.

wdyt?

test/trento/application/usecases/clusters_test.exs
```diff
@@ -64,6 +70,44 @@ defmodule Trento.ClustersTest do
 
       assert {:error, :amqp_error} = Clusters.request_checks_execution(cluster_id)
     end
+
+    test "should request cluster checks execution when checks are selected" do
+      checks = [Faker.UUID.v4(), Faker.UUID.v4()]
+      %{id: cluster_id} = insert(:cluster, id: Faker.UUID.v4(), selected_checks: checks)
```
nelsonkopliku, line 76: what about covering also the case where there is a deregistered cluster?

test/trento/application/usecases/hosts_test.exs
```diff
@@ -171,5 +173,35 @@ defmodule Trento.HostsTest do
 
       assert {:error, :amqp_error} = Hosts.request_checks_execution(host_id)
     end
+
+    test "should request host checks execution for hosts when checks are selected" do
+      checks = [Faker.UUID.v4(), Faker.UUID.v4()]
+      %{id: id} = insert(:host, id: Faker.UUID.v4(), selected_checks: checks)
```
nelsonkopliku, line 179: What about covering also the case where there is a deregistered host?

nelsonkopliku (changes requested):
Great job!
Just a couple of things.

lib/trento/application/usecases/clusters/clusters.ex
EMaksy: Sure let's tackle it at another time, we may refactor it. I was just thinking about keeping the host and cluster checks execution similar.
I will create a new ticket for this!

nelsonkopliku (commented):
Super! LGTM

nelsonkopliku (approved):

dottorblaster (approved):
Feel free to merge when the CI is green

PR #1781, opened 2023-09-01
rowKey function for table rows

dottorblaster:
# Description

We need to have a function inside the table component to determine the key for a row in a flexible way

## How was this tested?

Jest tests


arbulu89 (approved):
Thank you!
Most probably we want to use this function now in all of our tables.
Right?
To avoid the bad practice of the index usage.
By now, I'll at least use it in the place where it creates undesared behaviours.

nelsonkopliku (approved):
LGTM

PR #1782, opened 2023-09-04
Use rowKey in sap system and database overview tables

arbulu89:
# Description
Use the new `rowKey` parameter for the SAP system and database tables.
These tables have the delete function while they have react state, so unless we use a unique key, the table rows get incorrect state once the list is changed.


dottorblaster (approved):
If the CI is happy I am

PR #1790, opened 2023-09-05
Application Instance Move only if clustered

CDimonaco:
# Description

The application instances, will be moved when the host id changes AND the application instance is clustered.

This new behavior, aims to fix the 2379 bug,

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/application/integration/discovery/policies/sap_system_policy.ex
+++ b/lib/trento/application/integration/discovery/policies/sap_system_policy.ex
@@ -121,6 +121,8 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
          },
          host_id
        ) do
+    cluster_id = Trento.Clusters.get_cluster_id_by_host_id(host_id)
+
     Enum.map(instances, fn instance ->
       RegisterApplicationInstance.new(%{
         sid: sid,
@@ -134,7 +136,8 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
         start_priority: parse_start_priority(instance),
         host_id: host_id,
         health: parse_dispstatus(instance),
-        ensa_version: parse_ensa_version(instance)
+        ensa_version: parse_ensa_version(instance),
+        cluster_id: cluster_id
       })
     end)
   end
--- a/lib/trento/application/integration/discovery/policies/sap_system_policy.ex
+++ b/lib/trento/application/integration/discovery/policies/sap_system_policy.ex
@@ -32,7 +32,11 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
   @application_type 2
   @diagnostics_type 3
 
-  @spec handle(map, [ApplicationInstanceReadModel.t() | DatabaseInstanceReadModel.t()]) ::
+  @spec handle(
+          map,
+          [ApplicationInstanceReadModel.t() | DatabaseInstanceReadModel.t()],
+          String.t()
+        ) ::
           {:ok,
            [
              DeregisterApplicationInstance.t()
@@ -47,13 +51,14 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
           "agent_id" => agent_id,
           "payload" => payload
         },
-        current_instances
+        current_instances,
+        cluster_id
       ) do
     with {:ok, sap_systems} <- SapSystemDiscoveryPayload.new(payload),
          {:ok, register_instance_commands} <-
            sap_systems
            |> Enum.flat_map(fn sap_system ->
-             build_register_instances_commands(sap_system, agent_id)
+             build_register_instances_commands(sap_system, agent_id, cluster_id)
            end)
            |> Enum.reduce_while(
              {:ok, []},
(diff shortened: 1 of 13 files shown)


lib/trento/application/integration/discovery/policies/sap_system_policy.ex
```diff
@@ -121,6 +121,8 @@ defmodule Trento.Integration.Discovery.SapSystemPolicy do
          },
          host_id
        ) do
+    cluster_id = Trento.Clusters.get_cluster_id_by_host_id(host_id)
```
arbulu89, line 124: Would it be better to run this in the `discovery.ex` file?
We do the same to get the current instances, so the policy stays 100% pure, without external side-effects

lib/trento/domain/sap_system/commands/register_application_instance.ex
```diff
@@ -38,6 +38,7 @@ defmodule Trento.Domain.Commands.RegisterApplicationInstance do
     field :db_host, :string
     field :tenant, :string
     field :host_id, Ecto.UUID
+    field :cluster_id, Ecto.UUID
```
arbulu89, line 41: Could we add a little doc lines in top?
Something like:

>`cluster_id` value is used to know if the application instance is clustered or not. This information is required in order to decide whether the instance was moved by the cluster in a failover scenario or not

test/trento/application/integration/discovery/policies/sap_system_policy_test.exs
```diff
@@ -96,6 +121,30 @@ defmodule Trento.Integration.Discovery.SapSystemPolicyTest do
              |> SapSystemPolicy.handle([])
   end
 
+  test "should return the expected commands when a sap_system payload of type application and diagnostics is handled and the host is part of a cluster" do
```
arbulu89, line 124: Do we need this test?
It doesn't test any new code as far as I see

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -537,6 +539,91 @@ defmodule Trento.SapSystemTest do
       )
     end
 
+    test "should not move an application instance if the host_id changed and the application instance is not clustered" do
```
arbulu89, line 542: I would redo and rephrase this test. At the end, we want to check if already registered instance numbers in new not clustered hosts are registered as new instances:

`should register a new not clustered application instance if it has an already existing instance number`

arbulu89 (commented):
Some suggestions.
To have some e2e tests (other PR) we could do:
1s test:
- Register clustered ASCS. Send the same payload from other host. This should emit a movement event
- Register a not clustered APP with an already existing instance number in other host. This should register a new instance

lib/trento/application/integration/discovery/policies/sap_system_policy.ex
CDimonaco: Done

test/e2e/cypress/e2e/sap_systems_overview.cy.js
```diff
@@ -344,4 +344,54 @@ context('SAP Systems Overview', () => {
...
+
+      cy.get('table.table-fixed > tbody > tr')
+        .eq(1)
+        .find('div.table-row-group')
+        .eq(0)
+        .find('div.table-row')
+        .its('length')
+        .should('eq', 4);
+
+        cy.contains('vmnwdev01').should('not.exist');
```
arbulu89, line 380: I would put this `vmnwdev01` in the `nwdSystem` const.

test/trento/application/integration/discovery/policies/sap_system_policy_test.exs
arbulu89: we are still in the same. I would remove this test, as it doesn't test anything new. But well, it is not harmful neither

test/trento/domain/sap_system/sap_system_test.exs
```diff
@@ -537,6 +539,189 @@ defmodule Trento.SapSystemTest do
...
+          tenant: tenant,
+          instance_number: instance_number,
+          instance_hostname: instance_hostname,
+          features: "ABAP",
+          http_port: http_port,
+          https_port: https_port,
+          start_priority: start_priority,
+          host_id: new_host_id,
+          health: :passing,
+          ensa_version: ensa_version
```
arbulu89, line 607: Maybe we could put `cluster_id: nil`, to make it more obvious

arbulu89 (approved):
@CDimonaco Alright!
All good in my eyes.
Simple commented some few cosmetic things

test/trento/domain/sap_system/sap_system_test.exs
CDimonaco: done

test/e2e/cypress/e2e/sap_systems_overview.cy.js
CDimonaco: done

arbulu89 (approved):

PR #1793, opened 2023-09-06
Host details storybook

arbulu89:
# Description

Make `HostDetails` view "storybookable".
As with many other views, create a `*Page` container and put there all the redux interactions.

## How was this tested?

Tests updated. We still have e2e tests to test the whole usage of the view

### DIFF
--- a/assets/js/components/HostDetails/HostDetailsPage.jsx
+++ b/assets/js/components/HostDetails/HostDetailsPage.jsx
@@ -1,0 +1,83 @@
+import React, { useState, useEffect } from 'react';
+import { useParams, useNavigate } from 'react-router-dom';
+import { useSelector, useDispatch } from 'react-redux';
+
+import { networkClient } from '@lib/network';
+
+import { TARGET_HOST } from '@lib/model';
+
+import { getClusterByHost } from '@state/selectors/cluster';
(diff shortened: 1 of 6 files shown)


assets/js/components/HostDetails/HostDetailsPage.jsx
```diff
@@ -0,0 +1,85 @@
...
+      savingChecks={saving}
+      selectedChecks={hostSelectedChecks}
+      slesSubscriptions={host.sles_subscriptions}
+      cleanUpHost={() => {
+        dispatch(
+          deregisterHost({ id: hostID, hostname: host.hostname, navigate })
+        );
+      }}
+      requestHostChecksExecution={() => {
+        dispatch(hostExecutionRequested(host, hostSelectedChecks, navigate));
```
arbulu89, line 78: Some doubts on this. If we should get the `host` and `hostSelectedChecks` as function arguments, instead of "global" variables from the file.
This would force me to pass the whole `host` object to the details view as this is needed by the state function

nelsonkopliku (approved):
LGTM

dottorblaster (approved):
I think everything is legit, LGTM

Thanks for this! Deeply missed :)

EMaksy (approved):
Wonderful! -> LGTM

PR #1797, opened 2023-09-07
Split FE and BE tests in the CI

rtorrero:
# Description

This PR :
- splits the FE tests from the elixir tests. Since they don't depend on each other this might gain a bit of speed as it means they can run in parallel.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -199,9 +199,36 @@ jobs:
       - name: Check JS Code Format
         run: cd assets && npm run format:check
 
+  test-fe:
+    name: Run FE tests
+    needs: npm-deps
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Cancel Previous Runs
+        uses: styfle/cancel-workflow-action@0.11.0
+        with:
+          access_token: ${{ github.token }}
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


nelsonkopliku:
The frontend test step is now a dead end, no one is depending on it. I am afraid we are missing something.
If everything suceceds, except for frontend tests, we'd be committing to OBS something potentially broken.

dottorblaster:
I didn't propose a retry, I just said that splitting FE tests from BE tests (that is needed) doesn't prevent us from having to retry. The goal for me is having a CI that works, not a CI that automatically retries.

rtorrero:
> I didn't propose a retry, I just said that splitting FE tests from BE tests (that is needed) doesn't prevent us from having to retry. The goal for me is having a CI that works, not a CI that automatically retries.

my bad, I re-read your message on slack and it makes more sense now :sweat_smile: . I'll drop the retry.

nelsonkopliku:
I am afraid we still could be committing broken stuff to obs 

pipeline

rtorrero:
> I am afraid we still could be committing broken stuff to obs
> 
> pipeline

I was just doing the push to address that (in fact I think I beat you :smile: for a few seconds)

nelsonkopliku (approved):
LGTM

dottorblaster (approved):
LGTM

PR #1800, opened 2023-09-11
Saptune domain

arbulu89:
# Description

Saptune domain code. Really simple: it gets a command with the status and emit the event if there is any change.
PD: The `SaptuneStatus` module is a minimal implementation of what it is coming in the payload PR, just to proceed with this implementation

## How was this tested?
Tested with UT


CDimonaco (approved):

EMaksy (approved):
Hey man nice job --> LGTM :D

PR #1802, opened 2023-09-11
Add tooltip above `Start Execution` button

jamie-suse:
# Description

This PR present 3 main visual changes:

## Checks Selection tooltip

Adds a tooltip above the `Start Execution` button in the _Cluster Settings_ and _Host Checks Selection_ pages that says:

>Click Start Execution or wait for Trento to periodically run checks.

If no checks or no _new_ checks are saved on the page, this tooltip does not appear.

## _Host Details_ page tooltip

This PR also adds the tooltip for

>Select some Checks first!

that appears below the `Start Execution` button in the _Host Details_ page; in order ot align the behaviour as seen in the _Cluster Details_ page.

## Align _Cluster Details_ page

This PR also aligns the styling of the `Start Execution` button in the _Cluster Details_ page with the rest of the app.

## How was this tested?

Added assertions in React component tests to test appearance/disappearance of tooltip.

Updated Storybooks.

### DIFF
--- a/assets/js/components/HostDetails/HostChecksSelection.jsx
+++ b/assets/js/components/HostDetails/HostChecksSelection.jsx
@@ -5,6 +5,7 @@ import PageHeader from '@components/PageHeader';
 import BackButton from '@components/BackButton';
 import Button from '@components/Button';
 import ChecksSelection from '@components/ChecksSelection';
+import Tooltip from '@components/Tooltip';
 
 import HostInfoBox from './HostInfoBox';
 
@@ -23,6 +24,7 @@ function HostChecksSelection({
   onSelectedChecksChange,
   hostChecksExecutionEnabled,
   onStartExecution = () => {},
+  checksExecutionTooltipVisible,
 }) {
   return (
     <div className="w-full px-2 sm:px-0">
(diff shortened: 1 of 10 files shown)


jamie-suse:
One thing to note is that if you save some new checks in _Host Checks Selection_, then click the _Start Execution_ button, the tooltip disappears as intended. After if you then navigate back to _Host Details_ page, the tooltip will appear again - instead of being hidden because the user has already executed the checks.

Do you think this is a big issue?

nelsonkopliku (changes requested):
Just a quick recap of the offline discussion:
- host check selection and cluster check selection should have the same ux/ui regarding the tooltip
- host detail and cluster detail should have the same ux/ui (ie Select some checks first tooltip and should the `Click Start Execution or wait for ...` tooltip be there in that page?)
- enabled/disabled button ui consitency between cluster/host detail/selection

assets/js/components/HostDetails/HostChecksSelection.jsx
```diff
@@ -23,6 +24,7 @@ function HostChecksSelection({
   onSelectedChecksChange,
   hostChecksExecutionEnabled,
   onStartExecution = () => {},
+  checksExecutionTooltipVisible,
```
dottorblaster, line 27: Could you rename this to `isExecutionTooltipVisible`?

assets/js/components/HostDetails/HostSettingsPage.jsx
```diff
@@ -29,6 +29,8 @@ function HostSettingsPage() {
       setSelection(hostSelectedChecks);
     }
   }, [hostSelectedChecks]);
+  const [checksExecutionTooltipVisible, setChecksExecutionTooltipVisible] =
```
dottorblaster, line 32: `hostSelectedChecks` should update itself reactively if I remember correctly, did you check if you actually need a state here?

dottorblaster (changes requested):
Just a couple things then we can merge

nelsonkopliku (approved):

assets/js/components/HostDetails/HostSettingsPage.jsx
jamie-suse: Good point :+1: removed this

dottorblaster (approved):
LGTM

PR #1810, opened 2023-09-12
Fix instance movement logic

arbulu89:
# Description
Fix instance movement logic.
- If the instance with same number, host and features exist -> nop
- The instance finding matching updated to check the `features` as well

## How was this tested?

UT test added together with manual testing


arbulu89:
Tested with @abravosuse in a real env and it worked as expected

CDimonaco (approved):
LGTM

PR #1821, opened 2023-09-13
Saptune status projection

arbulu89:
# Description

Saptune status projection.

The read model and the view follows a flat hierarchy adding `saptune_status` directly. In the future we could have other entries like `saptune_notes` or `saptune_solutions`.
We could have other "tree" kind of representations as well. Check the open api schema to see how it looks like

## How was this tested?

Tested with UT


CDimonaco (approved):

EMaksy (approved):
Hey nice, everything looks good!

PR #1825, opened 2023-09-14
Enable navigation from host details to check results

dottorblaster:
# Description
Just removing the `disabled` on the "show results" button

## How was this tested?
Storybook

### DIFF
--- a/assets/js/components/HostDetails/HostDetails.jsx
+++ b/assets/js/components/HostDetails/HostDetails.jsx
@@ -115,15 +115,12 @@ function HostDetails({
 
               <Button
                 type="primary-white"
-                className="mx-0.5 border-green-500 border disabled:bg-slate-50 disabled:text-slate-500 disabled:border-gray-400"
+                className="mx-0.5 border-green-500 border"
                 size="small"
                 onClick={() => navigate(`/hosts/${hostID}/executions/last`)}
               >
                 <EOS_CLEAR_ALL
-                  className={classNames('inline-block ', {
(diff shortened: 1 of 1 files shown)


assets/js/components/HostDetails/HostDetails.jsx
```diff
@@ -118,15 +118,14 @@ function HostDetails({
                 className="mx-0.5 border-green-500 border disabled:bg-slate-50 disabled:text-slate-500 disabled:border-gray-400"
                 size="small"
                 onClick={() => navigate(`/hosts/${hostID}/executions/last`)}
-                disabled
               >
                 <EOS_CLEAR_ALL
```
nelsonkopliku, line 122: Just the icon color needs adjusting
```js
<EOS_CLEAR_ALL className="inline-block fill-jungle-green-500" />
```

assets/js/components/HostDetails/HostDetails.jsx
```diff
@@ -118,15 +118,14 @@ function HostDetails({
                 className="mx-0.5 border-green-500 border disabled:bg-slate-50 disabled:text-slate-500 disabled:border-gray-400"
```
nelsonkopliku, line 118:
```suggestion
                className="mx-0.5 border-green-500 border"
```
I guess that we can get rid of the disabled styles, this button never goes disabled.

dottorblaster: Yeah you're right

PR #1834, opened 2023-09-18
Align `primary` button style `disabled` state with UI proposal

jamie-suse:
# Description

Currently the styling of the `primary`/default button in the `disabled` state does not reflect the UI proposal. This PR aligns the implementation with the proposal.

## How was this tested?

As this is purely a CSS styling change, the behaviour of the UI does not change, and therefore no tests were added.


nelsonkopliku (approved):
LGTM!

nelsonkopliku (approved):

nelsonkopliku (approved):

PR #1843, opened 2023-09-19
Port check results detail page selectors to reselect

dottorblaster:
# Description

Porting check results detail page selectors to reselect

## How was this tested?

Automatic tests added


nelsonkopliku (approved):
Sweet, thanks!

PR #1858, opened 2023-09-22
Exclude checks result from host's health when empty selection is saved

nelsonkopliku:
# Description

This PR makes sure to ignore the checks result from the host's aggregated health computation when an empty selection is saved.

Imagine a host whose heartbeat is successful, but the latest checks execution is warning: the resulting host health would be warning.
If we now deselect all the checks for the host and save the empty selection, no other execution would run and the checks_health would remain in warning, keeping also the host's aggregated health in warning state, which would not make sense if there are no selected checks.

## How was this tested?

Automated test added

### DIFF
--- a/lib/trento/domain/host/host.ex
+++ b/lib/trento/domain/host/host.ex
@@ -389,17 +389,20 @@ defmodule Trento.Domain.Host do
   end
 
   def execute(
-        %Host{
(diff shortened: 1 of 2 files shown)


dottorblaster (approved):
LGTM

PR #1871, opened 2023-09-26
Add health to host read model

nelsonkopliku:
# Description

This PR adds the aggregated health of a host to the host read model

- migration added
- host read model has also `health` defaulting to `:unknown`
- on a  `HostHealthChanged` event the health is projected to the read model and broadcasted via websocket
- `HostView` takes `health` into account where necessary (ie ignored in `host_details_updated`)
- `HostRegistered` is upcasted to contain also host's aggregated `health`

Frontend will follow up.

## How was this tested?

Automated tests updated/added


dottorblaster (approved):
Aw yiss

PR #1878, opened 2023-09-27
Handle fact value with null value

arbulu89:
# Description

Handle `null` value in the `ObjectTree`. Without this fix the `Object.keys(data)` above panics.
Related to: https://github.com/trento-project/wanda/pull/297

image

PD: As the objectree treats arrays and objects as the same, I think having the `{}` is good enough.

## How was this tested?

UT added

### DIFF
--- a/assets/js/components/ExecutionResults/CheckResultDetail/FactValue.jsx
+++ b/assets/js/components/ExecutionResults/CheckResultDetail/FactValue.jsx
@@ -3,11 +3,13 @@ import React from 'react';
 import ObjectTree from '@components/ObjectTree';
 
 function FactValue({ className, data }) {
-  return typeof data === 'object' ? (
-    <ObjectTree className={className} data={data} />
-  ) : (
-    <span className={className}>{`${data}`}</span>
-  );
+  if (data === null) {
+    return <span className={className}>null</span>;
+  }
+  if (typeof data === 'object') {
+    return <ObjectTree className={className} data={data} />;
+  }
+  return <span className={className}>{`${data}`}</span>;
 }
 
 export default FactValue;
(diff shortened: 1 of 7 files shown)


assets/js/components/ObjectTree/tree.js
```diff
@@ -46,6 +46,10 @@ export const treeify = (name, data) => ({
     const { [key]: element } = data;
     const dataType = typeof element;
 
+    if (element == null) {
```
arbulu89, line 49: This is the piece of code which fixes the actual error

assets/js/components/ExecutionResults/CheckResultDetail/FactValue.jsx
```diff
@@ -3,11 +3,13 @@ import React from 'react';
 import ObjectTree from '@components/ObjectTree';
 
 function FactValue({ className, data }) {
-  return typeof data === 'object' ? (
-    <ObjectTree className={className} data={data} />
-  ) : (
-    <span className={className}>{`${data}`}</span>
-  );
+  if (data === null) {
```
nelsonkopliku, line 6: Just curious, this is needed because `typeof null` is `object`, right? :smile: 

Besides this I am wondering whether it would make sense having this check in `ObjectTree` itself. wdyt?

nelsonkopliku (approved):
LGTM. Just a tiny comment.

assets/js/components/ExecutionResults/CheckResultDetail/FactValue.jsx
arbulu89: > Just curious, this is needed because typeof null is object, right? 😄
Yes, it is a bug in js that it won't be fixed in 10000 years.

> Besides this I am wondering whether it would make sense having this check in ObjectTree itself. wdyt?
We have the same fix in the ObjectTree. We need to have in both places because we can get a plain fact value, so we don't always use ObjectTree

nelsonkopliku: fine with that :wink:

EMaksy (approved):
Hey man thanks codewise LGTM!

PR #1887, opened 2023-09-29
Update storybook control type from string to text

EMaksy:
# Description
The "string" control type does not enable us to change the value of the story anymore. 
This is just a small patch to update it to the newer control type "text" from the string datatype.

See storybook docs: 
https://storybook.js.org/docs/react/essentials/controls#annotation


nelsonkopliku (approved):
LGTM from a train 🚆

dottorblaster (approved):
Yes

arbulu89 (approved):

PR #1877, opened 2023-09-27
Use aggregated host health in the UI

nelsonkopliku:
# Description

Please include a summary of the changes and the related issue, if it exists. 
Also, include relevant motivation and context for this PR. List any
dependencies that are required for this change.

Fixes # (issue)

## Did you add the right label?

Remember to add the right labels to this PR.

## How was this tested?

Describe what kind of tests for this functionality have been added.

PR #1891, opened 2023-10-02
Fix test after behaviour change

arbulu89:
# Description

Fix test after https://github.com/trento-project/web/pull/1877
It fixes `Error: Unable to fire a "click" event - please provide a DOM element.` issue

Otherwise the filtering fails as it is not dependant on the `heartbeat` anymore.


dottorblaster (approved):
LGTM

nelsonkopliku (approved):
Thanks. Guess we missed it. Whoops

PR #1892, opened 2023-10-02
Upgrade react-markdown to 9.0.0 and remark-gfm to 4.0.0

dottorblaster:
# Description
Special care needed for this upgrade since new versions were released at the same time for both packages and they needed to be upgraded in combo

## How was this tested?

Jest (mocks added), manually


jamie-suse (approved):

nelsonkopliku (approved):
Thanks! LGTM

PR #1900, opened 2023-10-04
Send email notification on host health going critical

nelsonkopliku:
# Description

This PR makes sure an email notification is sent when a `HostHealthChanged` event is emitted and the health is `critical` instead of on a `HeartbeatFailed`.

## How was this tested?

Automated tests updated.

image

### DIFF
--- a/guides/alerting/alerting.md
+++ b/guides/alerting/alerting.md
@@ -3,7 +3,7 @@
 Alerting feature notifies the SAP Administrator about important updated in the Landscape being monitored/observed by Trento.
 
 Some of the notified events:
-- **Host heartbeat failed**
+- **Host Health detected critical**
 - **Cluster Health detected critical**
 - **Database Health detected critical**
 - **SAP System Health detected critical**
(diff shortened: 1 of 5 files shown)


jamie-suse (approved):
looks good :eyes:

abravosuse (commented):
Just one suggestion: replace
`Check your Trento installation for remediation suggestions`
with
`Check the corresponding Details View in the Trento Console for remediation suggestions`

nelsonkopliku:
> Just one suggestion: replace `Check your Trento installation for remediation suggestions` with `Check the corresponding Details View in the Trento Console for remediation suggestions`

@abravosuse That is gonna change for _all_ the sent emails. Ok with that?

abravosuse:
Let's think about this. What components can have their health state changed to critical? Hosts, SAP systems, HANA databases and clusters, right? And they all have their own details view, correct? So I think it's safe to change that message for all alert emails. Right?

abravosuse (approved):
Awesome!

PR #1906, opened 2023-10-05
Saptune services status icons

EMaksy:
# Description

Adding Saptune Service Status  icons

## Demo
image

## How was this tested?
Added automated test

### DIFF
--- a/assets/js/components/SaptuneDetails/SaptuneDetails.jsx
+++ b/assets/js/components/SaptuneDetails/SaptuneDetails.jsx
@@ -16,7 +16,11 @@ const renderService = (serviceName, services) => {
   const { enabled, active } = service;
 
   return (
-    <SaptuneServiceStatus service={service} enabled={enabled} active={active} />
+    <SaptuneServiceStatus
+      serviceName={serviceName}
+      enabled={enabled}
+      active={active}
+    />
   );
 };
 
--- a/assets/js/components/SaptuneDetails/SaptuneDetails.jsx
+++ b/assets/js/components/SaptuneDetails/SaptuneDetails.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { useParams } from 'react-router-dom';
 
-import { map } from 'lodash';
+import { find, map } from 'lodash';
 
 import BackButton from '@components/BackButton';
 import ListView from '@components/ListView';
@@ -10,6 +10,15 @@ import PageHeader from '@components/PageHeader';
 import SaptuneTuningState from './SaptuneTuningState';
 import SaptuneVersion from './SaptuneVersion';
 import SaptuneServiceStatus from './SaptuneServiceStatus';
+
+const renderService = (serviceName, services) => {
+  const service = find(services, { name: serviceName });
+  const { enabled, active } = service;
+
+  return (
+    <SaptuneServiceStatus service={service} enabled={enabled} active={active} />
+  );
+};
 
 const renderNote = (noteID) => (
   <a
@@ -22,6 +31,7 @@ const renderNote = (noteID) => (
     {noteID}
   </a>
 );
+
 const renderNotes = (notes) => {
   if (notes.length === 0) {
     return <span>-</span>;
@@ -51,6 +61,7 @@ function SaptuneDetails({
 }) {
   const { hostID: paramHostID } = useParams();
   const resolvedHostID = hostID || paramHostID;
+
   return (
     <div>
       <BackButton url={`/hosts/${resolvedHostID}`}>
@@ -95,27 +106,15 @@ function SaptuneDetails({
           data={[
             {
               title: 'saptune.service',
-              content: (
-                <SaptuneServiceStatus
-                  serviceName="saptune"
-                  services={services}
-                />
-              ),
+              content: renderService('saptune', services),
             },
             {
               title: 'sapconf.service',
-              content: (
-                <SaptuneServiceStatus
-                  serviceName="sapconf"
-                  services={services}
-                />
-              ),
+              content: renderService('sapconf', services),
             },
             {
               title: 'tuned.service',
-              content: (
-                <SaptuneServiceStatus serviceName="tuned" services={services} />
-              ),
+              content: renderService('tuned', services),
             },
           ]}
         />
--- a/assets/js/components/SaptuneDetails/SaptuneDetails.jsx
+++ b/assets/js/components/SaptuneDetails/SaptuneDetails.jsx
@@ -1,27 +1,53 @@
 import React from 'react';
 import { useParams } from 'react-router-dom';
 
-import { find, map } from 'lodash';
+import { find, map, get } from 'lodash';
 
 import BackButton from '@components/BackButton';
+import HealthIcon from '@components/Health/HealthIcon';
 import ListView from '@components/ListView';
-
 import PageHeader from '@components/PageHeader';
 
 import SaptuneTuningState from './SaptuneTuningState';
 import SaptuneVersion from './SaptuneVersion';
 
+const servicesIcons = {
+  saptune: {
+    'enabled/active': <HealthIcon health="passing" />,
+    'enabled/inactive': <HealthIcon health="warning" />,
+    'disabled/active': <HealthIcon health="warning" />,
+    'disabled/inactive': <HealthIcon health="critical" />,
+  },
+  sapconf: {
+    'enabled/active': <HealthIcon health="critical" />,
+    'enabled/inactive': <HealthIcon health="warning" />,
+    'disabled/active': <HealthIcon health="critical" />,
+    'disabled/inactive': <HealthIcon health="passing" />,
+  },
+  tuned: {
+    'enabled/active': <HealthIcon health="warning" />,
+    'enabled/inactive': <HealthIcon health="warning" />,
+    'disabled/active': <HealthIcon health="warning" />,
+    'disabled/inactive': <HealthIcon health="passing" />,
+  },
+};
+
 const renderService = (serviceName, services) => {
   const currentService = find(services, { name: serviceName });
+  const { enabled, active } = currentService;
 
-  if (!currentService.enabled) {
+  if (!enabled) {
     return <span>-</span>;
   }
 
+  const text = `${enabled}/${active}`;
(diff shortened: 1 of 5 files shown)


assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
```diff
@@ -14,42 +14,43 @@ import {
...
+    name: 'tuned',
+  };
+
+  const services = [
+    customSaptuneService,
+    customSapconfService,
+    customTunedService,
+  ];
 
+  const solution = saptuneSolutionFactory.build({ partial: false });
```
arbulu89, line 41: Why did you move these constants out?
I think they belong to the previous test.

For the 2nd test, you could use factory values:
```
const { configured_version: configuredVersion, enabled_solution: enabledSolution, applied_solution: appliedSolution } =
      saptuneStatusFactory.build();
```

I don't see having global constants as good idea here

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
```diff
@@ -102,4 +105,35 @@ describe('SaptuneDetails', () => {
       staging.solutions_ids
     );
   });
+
+  it('should render all the icons properly', () => {
```
arbulu89, line 109: I see this test super weak. It should check all the combinations.
I would extract the service logic to `SaptuneService.jsx` component and test them all with a `it.each` kind of test.

arbulu89 (changes requested):
I think we need to adjust the tests

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
EMaksy: fair point i wanted to reuse the same data for the tests, but actually this was the moment where i should have just created a new component :D

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
EMaksy: Good idea, it was challenging to test it, i will move it :D

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
```diff
@@ -22,7 +22,7 @@ describe('SaptuneDetails', () => {
     };
 
     const customSapconfService = {
-      active: 'inactive',
+      active: 'active',
```
arbulu89, line 25: why changes in this file?

assets/js/components/SaptuneDetails/SaptuneServiceStatus.jsx
```diff
@@ -0,0 +1,47 @@
...
+
+  return (
+    <div className="flex">
+      {icon}
+      <span className="ml-1">{text}</span>
+    </div>
+  );
+};
+
+function SaptuneServiceStatus({ serviceName, services }) {
```
arbulu89, line 44: I think this should be `{ serviceName, enabled, active}` and the next piece of code should remain in the `SaptuneDetails` file:
```
const currentService = find(services, { name: serviceName });
const { enabled, active } = currentService;
```
Passing the whole list of services looks really a smell (just seeing how strange the tests become)

assets/js/components/SaptuneDetails/SaptuneServiceStatus.test.jsx
```diff
@@ -0,0 +1,103 @@
...
+        name: service,
+      };
+
+      const services = [customService];
+
+      renderWithRouter(
+        <SaptuneServiceStatus services={services} serviceName={service} />
+      );
+
+      const serviceStatusIcons = screen.getAllByTestId('eos-svg-component');
```
arbulu89, line 98: `getByTestId`, we should only have one icon

assets/js/components/SaptuneDetails/SaptuneServiceStatus.test.jsx
```diff
@@ -0,0 +1,103 @@
...
+      active: 'inactive',
+      icon: 'fill-jungle-green-500',
+    },
+  ])(
+    'should render the service and icon properly for service $service',
+    ({ service, enabled, active, icon }) => {
+      const customService = {
+        active,
+        enabled,
+        name: service,
```
arbulu89, line 89: You can name `name` above if this is the term we should in the factory

assets/js/components/SaptuneDetails/SaptuneDetails.jsx
```diff
@@ -1,29 +1,15 @@
...
 
 import BackButton from '@components/BackButton';
 import ListView from '@components/ListView';
-
 import PageHeader from '@components/PageHeader';
 
 import SaptuneTuningState from './SaptuneTuningState';
 import SaptuneVersion from './SaptuneVersion';
-
-const renderService = (serviceName, services) => {
```
arbulu89, line 14: I think we could keep this as:
```
const renderService = (serviceName, services) => {
  const { enabled, active } = find(services, { name: serviceName });
  return <SaptuneServiceStatus serviceName={serviceName} enabled={enabled} active={active} />;
}
```

arbulu89 (changes requested):
Some suggestions @EMaksy

assets/js/components/SaptuneDetails/SaptuneDetails.jsx
EMaksy: Yeah sure i moved :D

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
EMaksy: it was a leftover from the previous test, so more  varied icons can be rendered.
I will revert it

assets/js/components/SaptuneDetails/SaptuneServiceStatus.jsx
EMaksy: Yeah sure this is cleaner :D

assets/js/components/SaptuneDetails/SaptuneServiceStatus.jsx
```diff
@@ -0,0 +1,41 @@
...
+  },
+};
+
+function SaptuneServiceStatus({ service, enabled, active }) {
+  if (!enabled) {
+    return <span>-</span>;
+  }
+
+  const status = `${enabled}/${active}`;
+  const icon = get(servicesIcons, [service.name, status], null);
```
arbulu89, line 32: `service.name` ?, why don't we pass the `serviceName` directly here?

assets/js/components/SaptuneDetails/SaptuneServiceStatus.test.jsx
```diff
@@ -0,0 +1,117 @@
...
+import '@testing-library/jest-dom';
+
+import { renderWithRouter } from '@lib/test-utils';
+
+import SaptuneServiceStatus from './SaptuneServiceStatus';
+
+describe('SaptuneServiceStatus', () => {
+  it.each([
+    {
+      service: { name: 'saptune', enabled: 'enabled', active: 'active' },
```
arbulu89, line 12: I don't think we need this `service` subtree.
Just `serviceName`, `enabled` and `active` should be alright.
Once you adjust the code to get the `serviceName` it will be ok

assets/js/components/SaptuneDetails/SaptuneDetails.stories.jsx
```diff
@@ -119,3 +119,37 @@ export const StagingDisabled = {
...
+  },
+};
+export const SaptuneServiceStatusFailing = {
+  args: {
+    ...Default.args,
+
+    services: [
+      { name: 'saptune', enabled: 'disabled', active: 'inactive' },
+      { name: 'sapconf', enabled: 'disabled', active: 'active' },
+      { name: 'tuned', enabled: null, active: null },
```
arbulu89, line 140: Is `null/null` failing?

arbulu89 (changes requested):
More things

assets/js/components/SaptuneDetails/SaptuneServiceStatus.jsx
EMaksy: yeah much better!

assets/js/components/SaptuneDetails/SaptuneDetails.stories.jsx
EMaksy: No there is no failing state for tuned, i thought better nothing then something in the failing case :D 

`  tuned: {
    'enabled/active': <HealthIcon health="warning" />,
    'enabled/inactive': <HealthIcon health="warning" />,
    'disabled/active': <HealthIcon health="warning" />,
    'disabled/inactive': <HealthIcon health="passing" />,
  },`
  
  Or what do you think? i also can delete it :D

arbulu89: Okey, it is fine them

assets/js/components/SaptuneDetails/SaptuneServiceStatus.test.jsx
```diff
@@ -0,0 +1,99 @@
...
+      icon: 'fill-yellow-500',
+    },
+    {
+      serviceName: 'tuned',
+      enabled: 'disabled',
+      active: 'inactive',
+      icon: 'fill-jungle-green-500',
+    },
+  ])(
+    'should render the service and icon properly for service $service.serviceName',
```
arbulu89, line 84: This looks wrong. It should be `$serviceName`

assets/js/components/SaptuneDetails/SaptuneServiceStatus.test.jsx
```diff
@@ -0,0 +1,99 @@
...
+    ({ serviceName, enabled, active, icon }) => {
+      renderWithRouter(
+        <SaptuneServiceStatus
+          serviceName={serviceName}
+          enabled={enabled}
+          active={active}
+        />
+      );
+
+      const serviceStatusIcons = screen.getByTestId('eos-svg-component');
```
arbulu89, line 94: `serviceStatusIcon` in singular

arbulu89 (approved):
Okey!
All good now.
Some minimal changes, but you can merge without other review.

PR #1908, opened 2023-10-05
Notify user when cluster health changes

jamie-suse:
# Description

This change notifies the user when a cluster's health changes.

Changes made:
- Created a new Phoenix view `cluster_health_changed.json` to use when broadcasting the `cluster_health_changed` websocket event.
- This view now also contains the cluster name as well as the previous cluster ID and health. The cluster name is required when displaying the notification to the user
- The Saga watching for the `cluster_health_changed` websocket event now displays a notification to the user stating that the cluster's health has changed

## How was this tested?

- Added test for new Phoenix view `cluster_health_changed.json`
- Updated projector test
- Added test for Saga


nelsonkopliku (approved):

dottorblaster (approved):
Thanks!

PR #1909, opened 2023-10-06
Improve renderSolution function if solution is empty

EMaksy:
# Description

A quick fix to ensure that SaptuneDetails is rendered properly if solutions are empty
If solution is empty then applied and enabled solution will render '-'

## Demo
image
image

## How was this tested? 
Added automated test

### DIFF
--- a/assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
+++ b/assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
@@ -110,4 +110,36 @@ describe('SaptuneDetails', () => {
       staging.solutions_ids
     );
   });
+
+  it('should display dashes for enabled and applied solutions when solution is null', () => {
+    const solution = null;
+
+    const {
+      configured_version: configuredVersion,
+      package_version: packageVersion,
+      services,
+      staging,
+      tuning_state: tuningState,
+    } = saptuneStatusFactory.build();
+    const { hostname, id: hostID } = hostFactory.build();
+    renderWithRouter(
+      <SaptuneDetails
+        appliedSolution={solution}
+        enabledSolution={solution}
+        hostname={hostname}
+        hostID={hostID}
+        configuredVersion={configuredVersion}
+        packageVersion={packageVersion}
+        services={services}
+        staging={staging}
+        tuningState={tuningState}
+      />
+    );
+    expect(screen.getByText('Enabled Solution').nextSibling).toHaveTextContent(
+      `-`
+    );
+    expect(screen.getByText('Applied Solution').nextSibling).toHaveTextContent(
+      `-`
+    );
+  });
 });
--- a/assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
+++ b/assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
@@ -111,53 +111,44 @@ describe('SaptuneDetails', () => {
     );
   });
 
-  it.each([
-    { solution: null },
-    { solution: { notes: [] } },
-    { solution: undefined },
-  ])(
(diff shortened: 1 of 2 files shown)


arbulu89 (commented):
The solution is good.
Let's add a test and go for it

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
```diff
@@ -102,4 +102,36 @@ describe('SaptuneDetails', () => {
...
+      configured_version: configuredVersion,
+      package_version: packageVersion,
+      services,
+      staging,
+      tuning_state: tuningState,
+    } = saptuneStatusFactory.build();
+    const { hostname, id: hostID } = hostFactory.build();
+    renderWithRouter(
+      <SaptuneDetails
+        appliedSolution={solution}
```
arbulu89, line 119: I would put `null` directly here. It is more visual

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
```diff
@@ -102,4 +102,36 @@ describe('SaptuneDetails', () => {
...
+        enabledSolution={solution}
```
arbulu89, line 120: Let's add `enabledNotes` and `appliedNotes` with empty lists, so we check a dash is rendered.
And a "disabled" staging (empty solution and notes)

arbulu89 (commented):
Looking good.
I would create a more complete test to check all the "empty" value visualizations.
And update the test name accordingly

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
```diff
@@ -110,4 +110,54 @@ describe('SaptuneDetails', () => {
       staging.solutions_ids
     );
   });
+
+  it.each([
+    { solution: null },
```
arbulu89, line 115: why all of this strange things?
Except the `null` option the other 2 are not real cases. They won't happen, so we don't need to test them.

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
```diff
@@ -110,4 +110,54 @@ describe('SaptuneDetails', () => {
...
+    { solution: { notes: [] } },
+    { solution: undefined },
+  ])(
+    'should display dashes for "Enabled Solution," "Applied Solution," "Staged Notes," and "Staged Solutions" when the solution is null, and staging is disabled',
```
arbulu89, line 119: Too long I would say. `should display empty values when no data is coming` would suffice in my opinion

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
```diff
@@ -110,4 +110,54 @@ describe('SaptuneDetails', () => {
...
+        tuning_state: tuningState,
+      } = saptuneStatusFactory.build();
+      const { hostname, id: hostID } = hostFactory.build();
+      renderWithRouter(
+        <SaptuneDetails
+          appliedSolution={solution}
+          enabledSolution={solution}
+          hostname={hostname}
+          hostID={hostID}
+          configuredVersion={configuredVersion}
```
arbulu89, line 138: Send `enabledNotes` and `appliedNotes` as empty lists and check we display them as "empty"

arbulu89 (changes requested):

assets/js/components/SaptuneDetails/SaptuneDetails.test.jsx
EMaksy: Fair enough, the other cases are irrelevant, i just thought some extra error testing won't harm :D

arbulu89 (approved):
Noice!

PR #1913, opened 2023-10-09
Fix "no tuning" value by "not tuned"

arbulu89:
# Description

By mistake I used the `no tuning` backend value when it should be `not tuned`.
I just used the frontend value :facepalm: 

## How was this tested?

Old tests are working


EMaksy (approved):
Ah damn, also slipped trough my eyes :D 
LGTM, let's merge and i push the other fix based on this fix!

PR #1935, opened 2023-10-18
Replace "SAP INSTANCES" title in Trento dashboard/home page

EMaksy:
# Description
Almost what the title says, just a replace of the title from "SAP INSTANCES" to "APPLICATION INSTANCES" .
### Before
Screenshot from 2023-10-18 13-16-00

### After
Screenshot from 2023-10-18 13-15-34

### DIFF
--- a/assets/js/components/HealthSummary/HomeHealthSummary.jsx
+++ b/assets/js/components/HealthSummary/HomeHealthSummary.jsx
@@ -24,7 +24,7 @@ const healthSummaryTableConfig = {
       ),
     },
     {
-      title: 'SAP Instances',
+      title: 'Application instances',
       key: 'sapsystemHealth',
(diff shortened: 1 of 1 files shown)


abravosuse (approved):
LGTM

arbulu89:
> LGTM

@abravosuse @jagabomb Just we aware that this forces to header to have 2 lines instead of one. If we want to adjust a bit the styles

abravosuse:
I think it depends on the width of the screen. Right now, if the screen is not wide enough, you also get a two-line headers, which is OK for me:
image

arbulu89 (approved):
If @abravosuse is happy, I'm happy

jagabomb (approved):
LGTM!

PR #1950, opened 2023-10-23
Normalize redux state content as snake_case

arbulu89:
# Description

Remove the last existence of camelCased redux state content and normalize its usage in snake_case. This way, the component is responsible of transforming the data in the best way. Now, all the factories produce the data as it is supplied by the backend.
Now, the need to have something like `keysToCamel` goes away

## How was this tested?

Tests updated


EMaksy (approved):
Hey man great job! 
LGTM :D

PR #1953, opened 2023-10-24
Checks catalog stories

arbulu89:
# Description

Make `ChecsCatalog` component storybookable and add main stories.

## How was this tested?

Tests updated

### DIFF
--- a/assets/js/components/ChecksCatalog/ChecksCatalog.stories.jsx
+++ b/assets/js/components/ChecksCatalog/ChecksCatalog.stories.jsx
@@ -11,7 +11,7 @@ const groupName3 = faker.string.uuid();
 const group1 = catalogCheckFactory.buildList(5, { group: groupName1 });
 const group2 = catalogCheckFactory.buildList(5, { group: groupName2 });
 const group3 = catalogCheckFactory.buildList(5, { group: groupName3 });
-const catalogData = group1.concat(group2).concat(group3);
+const catalogData = group1.concat(group2, group3);
 
 function ContainerWrapper({ children }) {
   return (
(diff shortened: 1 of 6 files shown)


assets/js/components/ChecksCatalog/ChecksCatalog.stories.jsx
```diff
@@ -0,0 +1,82 @@
...
+import { catalogCheckFactory } from '@lib/test-utils/factories';
+import ChecksCatalog from './ChecksCatalog';
+
+const groupName1 = faker.string.uuid();
+const groupName2 = faker.string.uuid();
+const groupName3 = faker.string.uuid();
+const group1 = catalogCheckFactory.buildList(5, { group: groupName1 });
+const group2 = catalogCheckFactory.buildList(5, { group: groupName2 });
+const group3 = catalogCheckFactory.buildList(5, { group: groupName3 });
+const catalogData = group1.concat(group2).concat(group3);
```
EMaksy, line 14: What do you think about using the spread operator as an alternative ? 
`const catalogData1 = [...group1, ...group2, ...group3];`
Just wondering which we should prefer :)

arbulu89: I would say in this case it is pretty irrelevant for us, the array elements number is small.
By default, I think that using `concat` is more efficient and meaningful than spread operator for arrays.
PD: Indeed, i have done a [research](https://stackoverflow.com/questions/48865710/spread-operator-vs-array-concat), and it looks like `concat` is faster for elements that we know are arrays.

EMaksy (approved):
Hey man well done :+1:

assets/js/components/ChecksCatalog/ChecksCatalog.stories.jsx
EMaksy: Interesting thanks!

dottorblaster (approved):
Thanks!

PR #1954, opened 2023-10-24
Fix redirection when a query string is involved

nelsonkopliku:
# Description

Fixes redirection when also a query string is involved in the initial request.

## How was this tested?

Automated test added.


dottorblaster (approved):
LGTM

PR #1958, opened 2023-10-26
Checks catalog only first group expanded

arbulu89:
# Description

Update the checks catalog view to only have the 1st group expanded initially.

## How was this tested?

Tests updated. Have a look in storybook!


abravosuse (approved):
Based on the storybook, it looks good to me.

EMaksy (approved):
Nice, looks great!

PR #1959, opened 2023-10-26
Handle when no checks available in the checks results overview

jamie-suse:
# Description

This change displays a message to the user that "Checks are coming soon" in the _Checks Results Overview_ when no checks are available in the Checks Catalog.

image

This change also displays an error message when there is a catalog error, and the loading spinner when the Checks Catalog hasn't finished loading.

## How was this tested?

Updated unit tests and end-to-end tests

### DIFF
--- a/assets/js/components/CheckResultsOverview/CheckResultsOverview.jsx
+++ b/assets/js/components/CheckResultsOverview/CheckResultsOverview.jsx
@@ -14,14 +14,12 @@ const pendingStates = [RUNNING_EXECUTION_STATE, REQUESTED_EXECUTION_STATE];
 
 function CheckResultsOverview({
   catalogData,
-  catalogLoading = false,
-  catalogError = null,
   data,
   error = null,
   loading = false,
   onCheckClick,
 }) {
-  if (catalogLoading || loading || pendingStates.includes(data?.status)) {
+  if (loading || pendingStates.includes(data?.status)) {
     return (
       <div className="flex flex-col items-center px-4">
         <h1 className="text-center text-2xl font-bold">Check Summary</h1>
@@ -33,10 +31,10 @@ function CheckResultsOverview({
     );
   }
 
-  if (error || catalogError) {
+  if (error) {
     return (
       <div className="flex flex-col items-center mt-2 px-4">
-        <div className="text-center text-xs">{error || catalogError}</div>
+        <div className="text-center text-xs">{error}</div>
       </div>
     );
   }
--- a/assets/js/components/CheckResultsOverview/CheckResultsOverview.jsx
+++ b/assets/js/components/CheckResultsOverview/CheckResultsOverview.jsx
@@ -14,7 +14,7 @@ const pendingStates = [RUNNING_EXECUTION_STATE, REQUESTED_EXECUTION_STATE];
 
 function CheckResultsOverview({
   data,
-  catalogDataEmpty,
+  catalogDataEmpty = false,
   error = null,
   loading = false,
   onCheckClick,
--- a/assets/js/components/CheckResultsOverview/CheckResultsOverview.jsx
+++ b/assets/js/components/CheckResultsOverview/CheckResultsOverview.jsx
@@ -7,17 +7,21 @@ import {
   RUNNING_EXECUTION_STATE,
 } from '@state/lastExecutions';
 
+import ChecksComingSoon from '@static/checks-coming-soon.svg';
 import CheckResultCount from './CheckResultCount';
 
 const pendingStates = [RUNNING_EXECUTION_STATE, REQUESTED_EXECUTION_STATE];
 
 function CheckResultsOverview({
+  catalogData,
+  catalogLoading = false,
+  catalogError = null,
   data,
   error = null,
   loading = false,
   onCheckClick,
 }) {
-  if (loading || pendingStates.includes(data?.status)) {
+  if (catalogLoading || loading || pendingStates.includes(data?.status)) {
     return (
       <div className="flex flex-col items-center px-4">
         <h1 className="text-center text-2xl font-bold">Check Summary</h1>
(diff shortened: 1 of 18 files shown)


jamie-suse:
Note: Currently updating Storybook, but thought it would be good to get reviews on the application code in the meantime

assets/js/components/CheckResultsOverview/CheckResultsOverview.jsx
```diff
@@ -29,10 +33,24 @@ function CheckResultsOverview({
...
+        <div className="text-center text-xs">{error || catalogError}</div>
+      </div>
+    );
+  }
+
+  if (!catalogData?.length) {
+    return (
+      <div className="flex flex-col items-center h-full">
+        <h1 className="text-center text-2xl font-bold">Check Results</h1>
+        <h6 className="opacity-60 text-xs">Checks coming soon</h6>
```
jamie-suse, line 48: Bringing focus to the wording of the message here, thoughts?

assets/js/components/HostDetails/HostDetails.jsx
```diff
@@ -72,6 +82,26 @@ function HostDetails({
...
+
+  const refreshCatalog = () =>
+    dispatch(
+      updateCatalog({
+        provider,
+        target_type: TARGET_HOST,
+      })
+    );
+
+  useEffect(() => {
```
jamie-suse, line 95: This is what I saw elsewhere; is there a more efficient way to call the Catalog here?

dottorblaster: Don't overcomplicate this, the browser is already architected to be efficient on the underlying level :+1:

dottorblaster (changes requested):
Overall it looks very good, would you avoid passing all these props inside the component and just manage the computation on the outer layer? I mean, you could use `loading` and just use `loading={loading || catalogLoading}` from outside.

This would let you use just one prop, namely `comingSoon`, to handle the conditional rendering inside the `ChecksResultsOverview`.

assets/js/components/CheckResultsOverview/CheckResultsOverview.jsx
nelsonkopliku: Currently in the cluster detail, and specifically for ASC/ERS clusters we have `Coming soon for ASCS/ERS`.

To me the thing is: 
we either do an effort to detect a meaningful user-friendly name of the target 
- Coming soon for Hosts
- Coming soon for Clusters
- Coming soon for ASCS/ERS clusters

Or we use a generic message.

The generic message, imho, would be good enough to start with,

dottorblaster (approved):
LGTM

assets/js/components/ClusterDetails/HanaClusterDetails.jsx
```diff
@@ -87,12 +93,36 @@ function HanaClusterDetails({
...
-    RUNNING_STATES.includes(lastExecution?.data?.status);
+    RUNNING_STATES.includes(executionData?.status);
+
+  const dispatch = useDispatch();
+
+  const refreshCatalog = () =>
+    dispatch(
+      updateCatalog({
+        provider,
+        target_type: TARGET_HOST,
```
nelsonkopliku, line 113:
```suggestion
        target_type: TARGET_CLUSTER,
```

assets/js/components/ClusterDetails/HanaClusterDetails.jsx
```diff
@@ -1,4 +1,11 @@
-import React from 'react';
+import React, { useEffect } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+
+import { updateCatalog } from '@state/actions/catalog';
+import { getCatalog } from '@state/selectors/catalog';
+import { RUNNING_STATES } from '@state/lastExecutions';
+
+import { TARGET_HOST } from '@lib/model';
```
nelsonkopliku, line 8:
```suggestion
import { TARGET_CLUSTER } from '@lib/model';
```

nelsonkopliku (changes requested):
Besides the target type in the hana cluster detail there are two other things:
- in the host detail the last execution call is not being triggered
- also in an ASC/ERS cluster details we should be using the overview that you've been working on

assets/js/components/HostDetails/HostDetails.jsx
```diff
@@ -71,6 +76,32 @@ function HostDetails({
...
+    data: catalogData,
+    loading: catalogLoading,
+    error: catalogError,
+  } = useSelector(getCatalog());
+
+  const {
+    data: lastExecutionData,
+    loading: lastExecutionLoading,
+    error: lastExecutionError,
+  } = lastExecution || {};
```
dottorblaster, line 103: Could you move all of this to HostDetailsPage?

dottorblaster (changes requested):
Some last changes then I think we are ok

assets/js/components/ClusterDetails/ClusterDetailsPage.jsx
```diff
@@ -1,32 +1,50 @@
...
-  const dispatch = useDispatch();
+  const provider = get(cluster, 'provider');
+  const type = get(cluster, 'type');
+
+  const catalog = useSelector(getCatalog());
+
   const lastExecution = useSelector(getLastExecution(clusterID));
+
   useEffect(() => {
+    dispatch(
```
nelsonkopliku, line 39: On page load (ie refresh when in a cluster detail) there are two pairs of requests made against the catalog and the last execution.

First we have a GET /api/v1/checks/catalog?target_type=cluster and then a complete GET /api/v1/checks/catalog?provider=azure&target_type=cluster&cluster_type=hana_scale_up

I guess we should check whether we have everything we need in order to dispatch an update of the catalog, and then dispatch it.

We could have the `cluster` variable as a dependency of this useEffect and if it is not null/undefined, then it means we have everything in place and then trigger the requests. What do you think?

dottorblaster: I told him to write that `useEffect` that way :)

thing is, putting an object in useEffect dependencies can trigger unwated side effects because every time the object is replaced it gets a new reference and can lead to additional calls made. This way otherwise we make the right call when things change punctually. I personally thought we can live with an extra API call :+1:

nelsonkopliku: I guess there can be a middle ground where we do not use the object (cluster), but we check on what we are passing (provider and type) and make the call only of we have both of them.

Would that work?

nelsonkopliku (approved):
I guess we're good! Thanks!

We could also include the `dispatch(updateLastExecution(clusterID));` in the `if (provider && type) {}` :see_no_evil: 

However, great job.

dottorblaster (approved):
LGTM

PR #1973, opened 2023-11-02
Remove Check-related action buttons from Host Details view

jamie-suse:
# Description

After the implementation of PR [#1959](https://github.com/trento-project/web/pull/1959), it was decided that as no Host Checks are currently available, the presence of the Check-related action buttons would be confusing to the user.

For example, if no Checks are available, the user can still click the _Check Selection_ button to be taken to the _Check Settings_ page. However as there are no Checks available, the UI displays a message to the user that the Checks Catalog is empty and a button to try again. This can be a confusing UX flow.

image

This change removes the Check-related action buttons from the UI in the _Host Details_ page to make the UI easier to understand.

image

## How was this tested?

Removed the tests checking for the presence of the checks-actions buttons

### DIFF
--- a/assets/js/components/HostDetails/HostDetails.jsx
+++ b/assets/js/components/HostDetails/HostDetails.jsx
@@ -1,18 +1,14 @@
 import React, { useState } from 'react';
 import { get } from 'lodash';
-import { EOS_CLEAR_ALL, EOS_PLAY_CIRCLE, EOS_SETTINGS } from 'eos-icons-react';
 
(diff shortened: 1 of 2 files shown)


dottorblaster (approved):
LGTM

arbulu89 (commented):
@jamie-suse Wouldn't be better to add some `if` condition to remove the buttons but keep the code? 
I mean, at some point, we want this logic present, and now that you removed, it will be more work to put it back hehe

PR #1980, opened 2023-11-07
Use databasesList slice properly to get db instances

arbulu89:
# Description

Fixes 2 selectors which uses the `sapSystems` slice incorrectly to get data of `databasesList` slice.
This creates the situation that if the SAP system using the database is deregistered, the database instances are not returned anymore, leaving "empty" HANA database (without instances).
I have checked, and as far as I could test, it doesn't have any "not wanted" side-effect

## How was this tested?

Tests updated. I did some manual testing as well.


nelsonkopliku (approved):
LGTM :eyes:

dottorblaster (approved):

PR #2002, opened 2023-11-15
Tags context

arbulu89:
# Description

Move the `tags` code to a proper phoenix content. Luckily enough, the `trento_web` part already follows correct conventions.
Based on: https://github.com/trento-project/docs/blob/main/adr/0010-web-dashboard-directory-structure-and-contexts.md

PD:
More references:
https://hexdocs.pm/phoenix/directory_structure.html#the-lib-hello-directory
https://hexdocs.pm/phoenix/contexts.html
https://hexdocs.pm/phoenix/testing_contexts.html

## How was this tested?

I have included some missing tests, mostly to give an example on how and where they should go.

### DIFF
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -71,9 +71,10 @@ defmodule Trento.Factory do
     HostReadModel,
     HostTelemetryReadModel,
     SapSystemReadModel,
-    SlesSubscriptionReadModel,
-    Tag
+    SlesSubscriptionReadModel
   }
+
+  alias Trento.Tags.Tag
 
   alias Trento.Integration.Discovery.{
     DiscardedDiscoveryEvent,
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -578,7 +578,7 @@ defmodule Trento.Factory do
 
   def tag_factory do
     %Tag{
-      value: sequence(:value, &"#{Faker.Beer.hop()}0#{&1}"),
+      value: sequence(:value, &"#{Faker.Color.name()}0#{&1}"),
       resource_id: Faker.UUID.v4(),
       resource_type: :host
     }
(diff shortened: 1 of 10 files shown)


nelsonkopliku (approved):

jamie-suse (approved):

EMaksy (commented):
LGTM!

test/support/factory.ex
```diff
@@ -577,7 +578,7 @@ defmodule Trento.Factory do
 
   def tag_factory do
     %Tag{
-      value: sequence(:value, &"#{Faker.Beer.hop()}0#{&1}"),
+      value: sequence(:value, &"#{Faker.Color.name()}0#{&1}"),
```
rtorrero, line 581: As a home brewer (and starwars fan) i'm very offended by this change (it's fine, just kidding :rofl:)

test/trento/tags_test.exs
```diff
@@ -0,0 +1,49 @@
...
+      %Tag{value: value, resource_id: resource_id} = insert(:tag)
+      :ok = Tags.delete_tag(value, resource_id)
+      assert nil == Repo.get_by(Tag, resource_id: resource_id)
+    end
+
+    test "delete_tag/2 returns a not found error if the resource does not exist" do
+      %Tag{value: value} = insert(:tag)
+      {:error, :not_found} = Tags.delete_tag(value, Faker.UUID.v4())
+    end
+  end
```
rtorrero, line 48: Would we want to test that nothing gets inserted on duplicate or is that overkill?

rtorrero (approved):
LGTM!

test/support/factory.ex
arbulu89: Beers fans aside XD
This was needed, as some beer names have spaces, and it was breaking validation rules

test/trento/tags_test.exs
arbulu89: Ok, a bit of an overkill to me, but it is not harmful neither

PR #2003, opened 2023-11-15
 ASCS/ERS checks

dottorblaster:
# Description
Enabling checks selection and execution for ASCS/ERS clusters

## How was this tested?

Unit tests added, additional stories not needed

### DIFF
--- a/assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
+++ b/assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
@@ -1,6 +1,10 @@
 import React, { useState, useEffect } from 'react';
 import { get } from 'lodash';
+import { EOS_SETTINGS, EOS_CLEAR_ALL, EOS_PLAY_CIRCLE } from 'eos-icons-react';
 
+import { RUNNING_STATES } from '@state/lastExecutions';
+
+import Button from '@components/Button';
 import PageHeader from '@components/PageHeader';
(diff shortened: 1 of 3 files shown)


assets/js/components/ClusterDetails/AscsErsClusterDetails.jsx
```diff
@@ -65,13 +70,19 @@ const nodeDetailsConfig = {
...
+  selectedChecks,
+  hasSelectedChecks,
   cibLastWritten,
   provider,
   hosts,
   sapSystems,
   details,
   catalog,
+  lastExecution,
+  onStartExecution,
```
nelsonkopliku, line 84: Should we provide a default noop function to `onStartExecution` and `navigate`?
```js
onStartExecution = () => {},
navigate = () => {},
```
We do that in the HanaClusterDetails component.
What's the way we prefer?

nelsonkopliku (approved):
Super! I love when things click quite seamlessly! :rocket: 

Just some notes:
- there is a bit of duplication between `HanaClusterDetails`, `AscsErsClusterDetails` and `HostDetails` when it comes to the check related action buttons and the logic they carry. Let's just be aware and ready to tackle that.
- we need to also send **ensa version** to the catalog api call when talking about ascs/ers cluster (at least in the checks selection). A followup PR would be totally ok.
- the catalog api call needs to start using the v2 exposed by wanda. I will take care of this as part of the work related to the catalog I am addressing.

PR #2004, opened 2023-11-15
Increase timeout for Rabbit consumer tests

dottorblaster:
# Description

Increased the timeout for RabbitMQ consumer tests because apparently with this larger suite a bigger timeout is needed.


jamie-suse (approved):

nelsonkopliku (approved):
:clock1:

PR #2005, opened 2023-11-15
Improve Error notification box usage

nelsonkopliku:
# Description

A tiny UI improvement in the usage of the notification box (OCD alert :smile: )

From 

image

image

to

image

image

### DIFF
--- a/assets/js/components/ExecutionResults/ResultsContainer.jsx
+++ b/assets/js/components/ExecutionResults/ResultsContainer.jsx
@@ -19,14 +19,12 @@ function ResultsContainer({
 }) {
   if (error) {
     return (
-      <div className="bg-white rounded p-3 shadow">
-        <NotificationBox
-          icon={<EOS_ERROR className="m-auto" color="red" size="xl" />}
-          text={errorContent}
-          buttonText="Try again"
-          buttonOnClick={onContentRefresh}
-        />
-      </div>
+      <NotificationBox
+        icon={<EOS_ERROR className="m-auto" color="red" size="xl" />}
+        text={errorContent}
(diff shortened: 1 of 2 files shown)


dottorblaster (approved):

jagabomb:
@nelsonkopliku Thanks for this. A small question from a UX point of view in the first scenario regarding the Network error for Cluster Settings, would the Start Execution button still work if the user clicked it or does it stop the user from proceeding?

nelsonkopliku:
> @nelsonkopliku Thanks for this. A small question from a UX point of view in the first scenario regarding the Network error for Cluster Settings, would the Start Execution button still work if the user clicked it or does it stop the user from proceeding?

Thanks @jagabomb, good point! 
Currently, the user would still be able to click the Start Execution button.
That button relies only on whether we have a non empty selection for the current target.

Triggering that action results in a message sent in a queue which wanda is responsible to consume, but if wanda is down it would not process anything until it boots up again.

From a user's perspective a loader with the `Checks execution starting...` message would be shown until wanda is again available and resumes processing messages (if wanda never comes up again the loader stays there indefinitely)

However, I guess I hear you: if we have a networking issue detected when loading the catalog, why should we allow the user to start an execution at all?

If it's ok with you, I would keep further changes on that out of this scope, cause it might involve further details. And we're anyway going to touch soon how that page behaves in these exceptional scenarios.

jagabomb:
@nelsonkopliku Cool, I hear you. I did not want to add an out of scope task here. I just wanted to mention it while it thinking about the UX. I will proceed to approve the PR since I am happy with the changes so far.

jagabomb (approved):

PR #2006, opened 2023-11-16
Supersede events

arbulu89:
# Description

Change to add the option to `supersede` events with new events (to rename in our case). This feature allows us to rename events to the new "phoenix context" way.

For example, from `Trento.Domain.Events.HostRegistered` to `Trento.Hosts.Events.HostRegistered`.

The usage is pretty simple:
We can include a new `superseeded_by` opt to our events that are to be changed, pointing to the new event.
For example:
```
defmodule Trento.Domain.Events.HostRegistered do
  @moduledoc """
  This event is emitted when a host is registered.
  """

  use Trento.Event

  defevent version: 3, superseded_by: Trento.Hosts.Events.HostRegistered do
```

This way, already existing legacy events are renamed for the rest of the application usage (e.g. aggregate code, event handlers, etc).
**Important**. The legacy events must continue existing in the codebase, so we will need to move them to some `legacy` folder. The previously commented change must be done in the legacy code file. Together with that, we need to create the new event (`Trento.Hosts.Events.HostRegistered`) file, which in this case will have the same content, as we are simply renaming it.

**Some background about why this way**:
Entries in the EventStore are inmutable, we cannot change them, for example, renaming the content of the entries. There are even some sql rules to forbid that.
The way to rename or update events is with the `upcasting`.
Find here some info: https://github.com/commanded/commanded/blob/master/guides/Events.md#upcasting-events

## How was this tested?

UT added. I have tested it manually as well, and it worked fine in the scenarios that I tried.

### DIFF
--- a/lib/trento/support/event.ex
+++ b/lib/trento/support/event.ex
@@ -8,14 +8,19 @@ defmodule Trento.Event do
   defmacro defevent(opts \\ [], do: block) do
     quote do
       @version Keyword.get(unquote(opts), :version, 1)
-      @superseded_by Keyword.get(unquote(opts), :superseded_by, __MODULE__)
+      @superseded_by Keyword.get(unquote(opts), :superseded_by, nil)
 
       deftype do
         field :version, :integer, default: @version
         unquote(block)
       end
 
-      def supersede, do: @superseded_by
+      def supersede do
+        case @superseded_by do
+          nil -> __MODULE__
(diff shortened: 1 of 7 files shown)


nelsonkopliku (approved):
L G T M

EMaksy (approved):
LGTM :)

jamie-suse (approved):

PR #2009, opened 2023-11-17
Extend Catalog filtering

nelsonkopliku:
# Description

This PR adds extra filters to the catalog. Besides the already present provider, the following have been added:
- target type
- cluster type
- ~ensa version~

catalog-filtering-no-ensa

Notes:
- first two commits are mainly about making some useful functions/constants/components a bit more reusable
- logic about filter dependencies (ie enable cluster type only if target type was selected) will follow up, with more extensive e2e testing

## How was this tested?

Automated tests and storybook.
https://2009.prenv.trento.suse.com/catalog

### DIFF
--- a/assets/js/components/ChecksCatalog/ChecksCatalogPage.jsx
+++ b/assets/js/components/ChecksCatalog/ChecksCatalogPage.jsx
@@ -3,22 +3,15 @@ import { useSelector, useDispatch } from 'react-redux';
 
 import { isValidProvider, isValidTargetType } from '@lib/model';
 import { isValidClusterType } from '@lib/model/clusters';
-import { isValidEnsaVersion } from '@lib/model/sapSystems';
 import { getCatalog } from '@state/selectors/catalog';
 import { updateCatalog } from '@state/actions/catalog';
 import ChecksCatalog from './ChecksCatalog';
 
-const buildUpdateCatalogAction = (
-  provider,
-  targetType,
-  clusterType,
-  ensaVersion
-) => {
+const buildUpdateCatalogAction = (provider, targetType, clusterType) => {
   const payload = {
     ...(isValidProvider(provider) ? { provider } : {}),
     ...(isValidTargetType(targetType) ? { target_type: targetType } : {}),
     ...(isValidClusterType(clusterType) ? { cluster_type: clusterType } : {}),
-    ...(isValidEnsaVersion(ensaVersion) ? { ensa_version: ensaVersion } : {}),
   };
 
   return updateCatalog(payload);
@@ -42,14 +35,12 @@ function ChecksCatalogPage() {
         selectedProvider,
         selectedTargetType,
         selectedClusterType,
-        selectedEnsaVersion,
       }) =>
         dispatch(
           buildUpdateCatalogAction(
             selectedProvider,
             selectedTargetType,
-            selectedClusterType,
-            selectedEnsaVersion
+            selectedClusterType
           )
         )
       }
--- a/assets/js/components/ChecksCatalog/ChecksCatalogPage.jsx
+++ b/assets/js/components/ChecksCatalog/ChecksCatalogPage.jsx
@@ -1,21 +1,16 @@
 import React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 
-import { isValidProvider, isValidTargetType } from '@lib/model';
-import { isValidClusterType } from '@lib/model/clusters';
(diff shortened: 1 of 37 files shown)


abravosuse (changes requested):
Feeding the ensa_version to the check from the web component won't work for multi-sid clusters. The Checks Team (Eike) is working right now on doing it from the agent instead, with the sapcontrol gatherer. Which means that, when it comes to ASCS/ERS clusters, we will no longer have ENSA2 and ENSA1 checks. We will have checks, valid for both ENSA versions, where the logic will perform one evaluation of the expectations or another -when applicable- based on the ENSA version of a managed system.  Because of this, the need for a ENSA filter in the UI decays. 

Therefore the proposed changes in this PR are OK for me, but we need to leave the ENSA filter out. Everything else is OK.

abravosuse (approved):
LGTM Nelson!

jagabomb (approved):
LGTM!

assets/js/components/TargetIcon/TargetIcon.jsx
```diff
@@ -1,27 +1,41 @@
...
+  [TARGET_CLUSTER]: 'Clusters',
+  [TARGET_HOST]: 'Hosts',
+};
+
+function TargetIcon({
+  targetType,
+  withLabel = false,
+  labelMap = targetTypeToLabel,
+  containerClassName = '',
+  iconClassName = '',
```
dottorblaster, line 21: You could just rename this as `className`

assets/js/components/TargetIcon/TargetIcon.jsx
```diff
@@ -1,27 +1,41 @@
...
+    <>
+      <span data-testid="target-icon" className={containerClassName}>
+        <span data-testid={`target-icon-${targetType}`}>
+          <IconComponent className={iconClassName} />
+        </span>
       </span>
-    </div>
+      {withLabel && (
+        <span data-testid="target-label">{labelMap[targetType]}</span>
+      )}
```
dottorblaster, line 37: You can use `children` for this and simplify the API:

```jsx
<TargetIcon targetType="cluster" className="some-class">
  Cluster
</TargetIcon>
```

assets/js/components/ChecksCatalog/ChecksCatalogPage.jsx
```diff
@@ -1,14 +1,17 @@
...
 import { updateCatalog } from '@state/actions/catalog';
 import ChecksCatalog from './ChecksCatalog';
 
-const buildUpdateCatalogAction = (provider) => {
+const buildUpdateCatalogAction = (provider, targetType, clusterType) => {
   const payload = {
-    ...(isValidProvider(provider) ? { provider, target_type: 'cluster' } : {}),
+    ...(isValidProvider(provider) ? { provider } : {}),
+    ...(isValidTargetType(targetType) ? { target_type: targetType } : {}),
+    ...(isValidClusterType(clusterType) ? { cluster_type: clusterType } : {}),
```
dottorblaster, line 14: This looked pretty clumsy previously and now it's even expanded! :smile: 

Can't we just run over the map and splice whatever `all` value? Doesn't lodash have anything inside for this?

dottorblaster (changes requested):
Some things to adjust and then we can merge :+1:

assets/js/components/ChecksCatalog/ChecksCatalogPage.jsx
nelsonkopliku: dunnow if lodash has something in that regard.

We need to either add an entry to the object if the value is not `all` or not add the the key to the object at all.

Any suggestion?

assets/js/components/TargetIcon/TargetIcon.jsx
nelsonkopliku: and keep the other as `containerClassName`? I'd like to keep both of them

assets/js/components/ChecksCatalog/ChecksCatalogPage.jsx
nelsonkopliku: Aha YES! Thanks!

nelsonkopliku:
@dottorblaster thanks! here we go 54d3e020e84eab06415fc1f7c8558af92f7ccfc9 and 9059064a464a09a2b1b5a96225ac0c16f80a1400

dottorblaster (approved):
LGTM

PR #2013, opened 2023-11-17
Generic selection component

nelsonkopliku:
# Description

Generalizes the current `ProviderSelection` to a `Select` which will be used also for other filters in the catalog.

## How was this tested?

Test and storybook


dottorblaster (changes requested):
Good job but a Select component is not only used for filters, but for selects :smile: 

Could you do the whole composition about the `all` options on the outside layer?

assets/js/components/Select/Select.jsx
```diff
@@ -5,24 +5,35 @@ import { CheckIcon, ChevronUpDownIcon } from '@heroicons/react/20/solid';
...
-    <ProviderLabel provider={provider} />
-  ) : (
-    <span>All</span>
-  );
-
-function ProviderSelection({ className, providers, selected, onChange }) {
+function Select({
+  optionsName,
+  options,
+  selected,
```
dottorblaster, line 19: This is generally called `value` :+1:

dottorblaster (changes requested):

nelsonkopliku:
@dottorblaster good point thanks! At the end I decided to let the used define whether an `All options` option is wanted b1f8931a9ff9d77272b866bb55368ddaf26c3b62

jamie-suse (approved):

assets/js/components/ChecksCatalog/ChecksCatalog.jsx
```diff
@@ -1,22 +1,20 @@
 import React, { useEffect, useState } from 'react';
 import { groupBy } from 'lodash';
 
-import { providerData } from '@components/ProviderLabel/ProviderLabel';
+import { PROVIDERS } from '@lib/model';
```
dottorblaster, line 4: Usually uppercase constants are used for primitive values, you may want to rename this :+1:

assets/js/components/ChecksCatalog/ChecksCatalog.jsx
```diff
@@ -26,10 +24,13 @@ function ChecksCatalog({ catalogData, catalogError, loading, updateCatalog }) {
...
       <div className="flex">
         <PageHeader className="font-bold">Checks catalog</PageHeader>
-        <ProviderSelection
+        <Select
+          optionsName="providers"
           className="ml-auto"
-          providers={Object.keys(updatedProvider)}
-          selected={selectedProvider}
+          options={PROVIDERS}
+          withAllOption
```
dottorblaster, line 31:
```suggestion
          options={[ALL_OPTION, ...PROVIDERS]}
```

assets/js/components/Select/Select.jsx
```diff
@@ -5,24 +5,40 @@ import { CheckIcon, ChevronUpDownIcon } from '@heroicons/react/20/solid';
...
-  ) : (
-    <span>All</span>
-  );
-
-function ProviderSelection({ className, providers, selected, onChange }) {
+function Select({
+  optionsName,
+  options,
+  value,
+  optionRenderer = defaultOptionRenderer,
```
dottorblaster, line 24: No strong opinion about this but I think `renderOption` would be more idiomatic

dottorblaster (changes requested):
Good job, just one thing to adjust. This way we could reuse this component as a generic select in the future and not only as a filter. That's why I asked you to move out the whole "all" logic outside of the component to increase the chance for composition.

assets/js/components/ChecksCatalog/ChecksCatalog.jsx
nelsonkopliku: Gonna lowercase it to just`providers` then

assets/js/components/ChecksCatalog/ChecksCatalog.jsx
nelsonkopliku: > we could reuse this component as a generic select in the future and not only as a filter.
We can already do this, even with the `withAllOption` prop, right?

nelsonkopliku:
@dottorblaster heres the last two commits https://github.com/trento-project/web/pull/2013/commits/8e02b2a2b40cb211717f72ad4cc1d510120d057d and https://github.com/trento-project/web/pull/2013/commits/40ed6f06717263ee3da3883cf290e3f590b6c3e8

dottorblaster (approved):
Great! Thanks!

nelsonkopliku:
@dottorblaster in total honesty I was confused by the **_whole_ "all" logic outside** :sweat_smile: 
I mean, we still have in the component the logic to _handle_ how to render an `all`. We just changed how we _add_ an `all` option to the options.

Anyway, thanks for the review!

PR #2017, opened 2023-11-20
Hosts context

arbulu89:
# Description

Update hosts related code to a "hosts phoenix context".
- The `hearbeats` goes into its own context.
- saptune and sles subscriptions belong to the hosts context
- the dependant value objects are moved to the hosts folder
- legacy events are moved to the a `legacy` folder
- mix typo in event -> `HeartbeatSucceded` -> `HeartbeatSucceeded`

## How was this tested?

Tests were simple renamed and moved


CDimonaco (approved):
LGTM

Good job, I'm reallu happy with that, just a sidenote comment, maybe we should put the ValueObject in the module name of value objects? 

Elixir module system gives us the ability to decouple the folder from the module name, but idk if we have to specify that in this refactor, I think yes, but no strong opinions, wdyt?

rtorrero (approved):
LGTM

PR #2024, opened 2023-11-23
Redux state directory restructure

EMaksy:
# Description

This PR aligns with ADR [0009](https://github.com/trento-project/docs/blob/main/adr/0009-frontend-directory-structure-and-architecture.md) by relocating actions within the Redux state directory. Actions have been moved to their respective slices, and those without a dedicated slice are now in sagas. These changes are inspired by the official documentation of the  [redux libary](https://redux.js.org/faq/code-structure/) 
## How was this tested?

Automated tests already cover the made changes :+1:


assets/js/state/catalog.js
```diff
@@ -1,5 +1,12 @@
 import { createSlice } from '@reduxjs/toolkit';
 
+export const UPDATE_CATALOG = 'UPDATE_CATALOG';
+
+export const updateCatalog = (env = {}) => ({
```
arbulu89, line 5: Could redo them using redux `createAction` util?
We already use it in many other places, and it is the recommended way of creating actions

assets/js/state/sagas/notifications.jsx
```diff
@@ -2,7 +2,13 @@
 import React from 'react';
 import { takeEvery } from 'redux-saga/effects';
 import { toast } from 'react-hot-toast';
-import { NOTIFICATION } from '@state/actions/notifications';
+
+export const NOTIFICATION = 'NOTIFICATION';
```
arbulu89, line 6: This looks a bit strange. In other components, this would be a slice code. Maybe we should move this piece of code to `state/notifications.js`. Otherwise, in other parts of the code we are importing actions from the saga code, which looks out of place.
What do you think?

arbulu89 (commented):
Good change in general.
2 comments:
- Could we check if we can modify those actions to use `createAction`? If it is not a big hussle at least
- I would move the `notifcations` constants and actions to a new `state/notifcations.js` slice file. It doesn't have reducer functions, but it makes more sense to have it ther

Wdyt?

nelsonkopliku (approved):
Besides what Xabi is mentioning, LGTM.

rtorrero (commented):
LGTM @EMaksy, thanks!

assets/js/state/catalog.js
EMaksy: Thanks for the suggestion, i used the createAction helper :+1:

assets/js/state/sagas/checksSelection.js
```diff
@@ -3,7 +3,7 @@ import { post } from '@lib/network';
 
 import { TARGET_HOST, TARGET_CLUSTER } from '@lib/model';
 
-import { notify } from '@state/actions/notifications';
+import { notify } from '@state/sagas/notifications';
```
arbulu89, line 6: `notify` is an actions, so it should still come from the reducer new file, don't you think?

EMaksy: Yeah!

arbulu89 (approved):
I think the `notify` should be implemented with the `createAction`, but, well not a blocker.
Green light from my side

PR #2025, opened 2023-11-23
Deploy to demo env when receives `deploy-demo` action from `wanda`

jamie-suse:
# Description

The `wanda` repo sends a _GitHub Action_ called `deploy-demo` that _should_ trigger the `web` repo to deploy to the demo environment. However currently, the `web` repo does not do the deployment upon receiving the `deploy-demo` action.

This change fixes in the logic in the CI to trigger deployments to demo.

## How was this tested?

By creating forks of `web` and `wanda` repos, and replicating + verifying the behaviour there.


nelsonkopliku (approved):
LGTM! Thanks!

PR #2031, opened 2023-11-24
Fix Snapshot sap system import in sap systems roll up events

CDimonaco:
# Description

Found out some wrong aliases in the roll up related events for sap systems.

Thx @rtorrero 

This pr aims to fix that

## How was this tested?

No tests added


rtorrero (approved):
Thanks for the fix, LGTM!

nelsonkopliku (commented):
I am aware that roll up is already tested in `roll_up_test.exs`, however it is acting on test events. 
I'd find it safer having  _actual_ rollup on aggregates tested as well. Such tests would cover this kind of currently undetected bug. wdyt?

CDimonaco:
Closing because of https://github.com/trento-project/web/pull/2030

PR #2035, opened 2023-11-27
Improved catalog filtering

nelsonkopliku:
# Description

This PR introduces the possibility to feed the `Select` component with a list of objects as options, where the `value` and `enabled` properties can be defined.

As a result disabled options can also be handled.
Currently unsupported filters have been disabled (target type host and cluster type hana scale out and ascs/ers)

## How was this tested?

Automated tests/storybook.


assets/js/components/ChecksCatalog/ChecksCatalog.jsx
```diff
@@ -7,14 +8,28 @@ import {
...
 import CheckItem from './CheckItem';
+import Pill from '../Pill/Pill';
+
+function ComingSoonPill() {
+  return (
+    <Pill size="xs" className="absolute right-2 bg-green-100 text-green-800">
+      Coming Soon
+    </Pill>
+  );
+}
```
dottorblaster, line 32: I think we can inline this or move it to a separate file

assets/js/components/Select/Select.jsx
```diff
@@ -49,15 +57,21 @@ function Select({
...
+                      {
+                        'text-gray-400': optionDisabled,
+                        'text-green-900 bg-green-100': active,
+                        'text-gray-900': !optionDisabled && !active,
+                      }
+                    )
                   }
-                  value={option}
+                  value={option.value}
+                  disabled={!option.enabled}
```
dottorblaster, line 74: Usually `disabled` things are just `disabled` because it's closer to HTML native elements.

Hence:

```suggestion
                  disabled={option.disabled}
```

assets/js/components/Select/Select.jsx
```diff
@@ -4,37 +4,45 @@ import { Listbox, Transition } from '@headlessui/react';
...
 import classNames from 'classnames';
+import { find, get } from 'lodash';
+import { createOptionRenderer } from '.';
 
 export const OPTION_ALL = 'all';
 
 const defaultOnChange = () => {};
-const defaultRenderOption = (item) => item;
+const defaultRenderOption = (optionsName) =>
+  createOptionRenderer(`All ${optionsName}`, (item) => item);
```
dottorblaster, line 14: We already discussed multiple times about not encapsulating this behavior by default into the component and I see strong pushback from you about this.

This `all` handling is a peculiarity of a single use case, and we will for sure use this Select component in other places like the tables where we currently have duplicated code or other filters like the one in the checks results page.

At this point I can just disagree and commit.

dottorblaster (changes requested):
Some tiny stuff then we can merge :+1:

assets/js/components/Select/Select.jsx
nelsonkopliku: I was doubtful myself :smile: 
I am ok with changing, it meaning that if we provide a list of options containing also the `all` option, it would be rendered just as a plain `all` text. :+1:

nelsonkopliku:
Thanks @dottorblaster. Addressed the feedbacks!

dottorblaster (approved):
LGTM

PR #2037, opened 2023-11-28
Refactor React components directory structure

EMaksy:
# Description
This PR implements directory restructuring based on [ADR 009](https://github.com/trento-project/docs/blob/main/adr/0009-frontend-directory-structure-and-architecture.md). It separates the components directory into 'common' and 'pages,' enhancing code organization. The change ensures better reuse of common components across multiple pages.

Result after the refactor:
image


assets/js/pages/ExecutionResults/TargetInfoBox.jsx
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 
 import { TARGET_CLUSTER, TARGET_HOST } from '@lib/model';
 
-import { ClusterInfoBox } from '@components/ClusterDetails';
-import HostInfoBox from '@components/HostDetails/HostInfoBox';
+import { ClusterInfoBox } from '@pages/ClusterDetails';
+import HostInfoBox from '@pages/HostSettingsPage/HostInfoBox';
```
jamie-suse, line 6: I see here we borrow `HostInfoBox` from `HostSettingsPage`; going by our new structure, shouldn't we move `HostInfoBox` into the `commmon/` directory then?

(Though this may not be for this PR)

jamie-suse: (similar for `ClusterInfoBox`)

dottorblaster: I think you're right @jamie-suse but I wouldn't follow through that in this PR :+1:

EMaksy: I see there also opportunities to refactor some places, but i would not mind to split it up in a follow up pr :)

rtorrero (commented):
Good stuff @EMaksy, I think this is a step in the right direction

assets/js/pages/ExecutionResults/TargetInfoBox.jsx
EMaksy: Moved like suggested :+1:

CDimonaco (approved):

PR #2050, opened 2023-11-30
Aggregated target health does not take into account checks selection status

jamie-suse:
# Description

This changes how the aggregated health works for hosts and clusters.

- Selected checks are no longer taken into account in the aggregated health calculation.
- If the checks' health is `:unknown`, it is not taken into account.

## How was this tested?

Added unit tests.

### DIFF
--- a/lib/trento/clusters/cluster.ex
+++ b/lib/trento/clusters/cluster.ex
@@ -294,22 +294,15 @@ defmodule Trento.Clusters.Cluster do
   def execute(
         %Cluster{
           cluster_id: cluster_id
-        } = cluster,
+        },
         %SelectChecks{
           checks: selected_checks
         }
       ) do
-    cluster
-    |> Multi.new()
-    |> Multi.execute(fn _ ->
-      [
-        %ChecksSelected{
-          cluster_id: cluster_id,
-          checks: selected_checks
-        }
-      ]
-    end)
-    |> Multi.execute(fn cluster -> maybe_emit_cluster_health_changed_event(cluster) end)
+    %ChecksSelected{
+      cluster_id: cluster_id,
+      checks: selected_checks
+    }
   end
 
   def execute(


lib/trento/clusters/cluster.ex
```diff
@@ -605,19 +605,20 @@ defmodule Trento.Clusters.Cluster do
 
   defp maybe_emit_cluster_deregistered_event(_, _), do: nil
 
-  defp maybe_add_checks_health(healths, _, []), do: healths
-  defp maybe_add_checks_health(healths, checks_health, _), do: [checks_health | healths]
+  defp maybe_add_checks_health(healths, checks_health) when checks_health != Health.unknown(),
+    do: [checks_health | healths]
+
+  defp maybe_add_checks_health(healths, _), do: healths
```
nelsonkopliku, line 611: I read this as: take into account `checks_health` when it is anything different from `unknown`, in all other cases (the `_` blackhole in the second match) only consider `healths`

The point is that _all other cases_ in the second match is really _just one case_: when it is `unknown`

What about the following?
```elixir
defp maybe_add_checks_health(healths, Health.unknown()), do: healths
defp maybe_add_checks_health(healths, checks_health), do: [checks_health | healths]
```
or, if we want to have it clearer in the first match what the second argument is
```elixir
defp maybe_add_checks_health(healths, checks_health) when checks_health == Health.unknown(), do: healths
defp maybe_add_checks_health(healths, checks_health), do: [checks_health | healths]
```

jamie-suse: ah yes, I like this way much more :+1:

test/trento/clusters/cluster_test.exs
```diff
@@ -474,6 +474,82 @@ defmodule Trento.ClusterTest do
       )
     end
 
+    test "checks health should not affect health when checks health is unknown" do
```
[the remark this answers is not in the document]
jamie-suse: Agreed, I was staring at it for too long :woozy_face:

test/trento/clusters/cluster_test.exs
```diff
@@ -474,6 +474,82 @@ defmodule Trento.ClusterTest do
...
+      cluster_id = Faker.UUID.v4()
+      selected_checks = Enum.map(0..4, fn _ -> Faker.Cat.name() end)
+
+      assert_events_and_state(
+        [
+          build(:cluster_registered_event, cluster_id: cluster_id, health: Health.passing()),
+          %ClusterChecksHealthChanged{
```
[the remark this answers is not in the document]
nelsonkopliku: I guess we might be just missing a ` default: Health.unknown()` in the cluster aggregate. Take a look at the Host's aggregate `checks_health` property
```elixir
field :checks_health, Ecto.Enum, values: Health.values(), default: Health.unknown()
```

test/trento/hosts/host_test.exs
```diff
@@ -381,7 +381,44 @@ defmodule Trento.Hosts.HostTest do
       end
     end
 
-    test "should ignore checks health when an empty checks selection is saved" do
+    test "unknown checks health should not affect aggregated cluster's health" do
```
nelsonkopliku, line 384:
```suggestion
    test "unknown checks health should not affect aggregated host's health" do
```

PR #2079, opened 2023-12-07
New empty state for catalog

dottorblaster:
# Description

Revamped the empty state for the catalog.

## How was this tested?

Unit tests added, Storybook story for an empty catalog added.


jamie-suse (approved):
Nice, super clean! :sponge:

PR #2088, opened 2023-12-12
Provide ENSA version info to Wanda

jamie-suse:
# Description

This change enables the frontend to send ENSA version information about the SAP systems belonging to a certain cluster.

## Business Logic

Note that the ENSA version value sent is an aggregated value of the ENSA versions present in the SAP systems for a given cluster.

- If a scenario where the ENSA versions are not _completely_ in agreement, then the value `'mixed_versions'` is sent.

- If the ENSA version info is not present for a SAP system, then `'mixed_versions'` is sent.

Some example scenarios are below:

| SAP Systems' ENSA versions |       Result       |
|:---------------------------|:------------------:|
| `['ensa1']`                |     `'ensa1'`      |
| `['ensa2', 'ensa2']`       |     `'ensa2'`      |
| `['ensa1', 'ensa2']`       | `'mixed_versions'` |
| `['ensa1', undefined]`     | `'mixed_versions'` |
| `[undefined]`              | `'mixed_versions'` | 

## How was this tested?

Added unit tests

### DIFF
--- a/assets/js/state/selectors/cluster.js
+++ b/assets/js/state/selectors/cluster.js
@@ -63,7 +63,7 @@ export const getClusterSapSystems = createSelector(
 );
 
 export const getEnsaVersion = createSelector(
-  [(state, clusterID) => getClusterSapSystems(state, clusterID)],
+  [getClusterSapSystems],
   (sapSystems) => {
     const ensaVersions = new Set();
     sapSystems.forEach(({ ensa_version }) => ensaVersions.add(ensa_version));
@@ -72,7 +72,7 @@ export const getEnsaVersion = createSelector(
 
     return firstEnsaVersion && ensaVersions.size === 1
       ? firstEnsaVersion
-      : 'mixed_versions';
+      : 'MIXED_VERSIONS';
   }
 );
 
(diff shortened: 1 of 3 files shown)


dottorblaster (changes requested):
Just two tiny thing then we can merge :santa:

assets/js/state/selectors/cluster.js
```diff
@@ -62,6 +62,8 @@ export const getClusterSapSystems = createSelector(
   }
 );
 
+export const MIXED_VERSIONS = 'MIXED_VERSIONS';
```
dottorblaster, line 65: Sorry! The value of this should be just `'mixed_versions'`, the contract that we have with the checks team is shaped like that :sweat_smile:

dottorblaster (changes requested):
Just one additional thing, sorry! And then we can merge :+1:

assets/js/state/selectors/cluster.js
jamie-suse: No worries, will revert :ok_hand:

dottorblaster (approved):
If the CI is happy I am as well

PR #2093, opened 2023-12-14
New catalog error state

dottorblaster:
# Description

As above: new catalog error state illustration and new error message

## How was this tested?

New jest expectation added


EMaksy (approved):
Looks great :)

PR #2096, opened 2023-12-15
Fix Start Execution button size in cluster details

dottorblaster:
# Description

Fix the size (and the margin) for the Start Execution button in the cluster details.

## How was this tested?

Existing jest tests and stories cover it


EMaksy (approved):
LGTM

PR #2105, opened 2023-12-18
Fix classnames usage in HostDetails

dottorblaster:
# Description

Just use classNames instead of template strings

## How was this tested?

Jest tests already covering this


EMaksy (approved):
Nice!

PR #2122, opened 2023-12-21
Use docker compose instead of action services in ci

CDimonaco:
# Description

As title suggest, use docker compose action instead of github action services.

This is a big improvement in our ci, we can get rid of all the limitations of github ci services and have a more robust and aligned ci setup with the development env.


dottorblaster (approved):
Fine for me, this is actually great for #1803 as well

PR #2126, opened 2023-12-22
Send filesystem type to Wanda

dottorblaster:
# Description
Sending filesystem type to Wanda when selecting checks for a cluster, and during an execution request.

## How was this tested?
Jest tests added, Elixir tests added

### DIFF
--- a/assets/js/state/selectors/cluster.test.js
+++ b/assets/js/state/selectors/cluster.test.js
@@ -394,64 +394,6 @@ describe('Cluster selector', () => {
     expect(getEnsaVersion(state, clusterID)).toEqual(MIXED_VERSIONS);
   });
 
-  it('should return filesystem type', () => {
-    const clusterID = faker.string.uuid();
-    const cluster = clusterFactory.build({ id: clusterID });
-
-    const sapSystems = sapSystemFactory.buildList(2, { ensa_version: 'ensa1' });
-    const [{ id: sapSystem1 }, { id: sapSystem2 }] = sapSystems;
-
-    const hosts = hostFactory
-      .buildList(4, { cluster_id: clusterID })
-      .concat(hostFactory.buildList(2));
-    const [{ id: host1 }, { id: host2 }, { id: host3 }, { id: host4 }] = hosts;
-
-    const applicationInstances = [
-      sapSystemApplicationInstanceFactory.build({
-        sap_system_id: sapSystem1,
-        host_id: host1,
-      }),
-      sapSystemApplicationInstanceFactory.build({
-        sap_system_id: sapSystem2,
-        host_id: host2,
-      }),
-    ];
-
-    const databases = databaseFactory.buildList(4);
-    const [{ id: database1 }, { id: database2 }] = databases;
-
-    const databaseInstances = [
-      databaseInstanceFactory.build({
-        sap_system_id: database1,
-        host_id: host3,
-      }),
-      databaseInstanceFactory.build({
-        sap_system_id: database2,
-        host_id: host4,
-      }),
-    ];
-
-    const state = {
-      hostsList: {
-        hosts,
-      },
-      databasesList: {
-        databases,
-        databaseInstances,
-      },
-      sapSystemsList: {
-        sapSystems,
-        applicationInstances,
-        databaseInstances,
-      },
-      clustersList: {
-        clusters: [cluster],
-      },
-    };
-
-    expect(getEnsaVersion(state, clusterID)).toEqual(MIXED_VERSIONS);
-  });
-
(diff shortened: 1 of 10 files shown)


nelsonkopliku (commented):
The end result could work, however my main concern is about the discrepancy between the domain and what we are modeling, connected also to the implicitness when interacting with the catalog api.

Let's get more feedback.

nelsonkopliku:
Another note: we are not sending the filesystem type (nor ensa version) to wanda during execution. Only in catalog loading

dottorblaster:
@nelsonkopliku thanks for all the comments, I'll follow up to them once we have ENSA version identification properly done and merged :+1:

assets/js/state/selectors/cluster.test.js
```diff
@@ -382,4 +390,132 @@ describe('Cluster selector', () => {
...
+        sapSystems,
+        applicationInstances,
+        databaseInstances,
+      },
+      clustersList: {
+        clusters: [cluster],
+      },
+    };
+
+    expect(getEnsaVersion(state, clusterID)).toEqual(MIXED_VERSIONS);
```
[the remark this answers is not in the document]
dottorblaster: Upsy-daisy! Leftover, thanks for reporting that!

dottorblaster:
@nelsonkopliku removed the leftover test, thanks. I'd like to address the differentiation between ascs/ers and other cluster types in the frontend in a subsequent PR, what do you think?

nelsonkopliku:
> removed the leftover test, thanks. I'd like to address the differentiation between ascs/ers and other cluster types in the frontend in a subsequent PR, what do you think?

Thank _you_! Yes, it makes sense to me as well following that up in another PR :+1: Still finishing the review tho :smile:

lib/trento/infrastructure/checks/checks.ex
```diff
@@ -93,6 +94,18 @@ defmodule Trento.Infrastructure.Checks do
     commanded().dispatch(command, correlation_id: execution_id)
   end
 
+  defp build_env(%AscsErsClusterExecutionEnv{
```
nelsonkopliku, line 97: I am not sure we need a new `AscsErsClusterExecutionEnv` struct.
We could simplify by patter matching on the cluster type of a `ClusterExecutionEnv`.

- this function would get a `ClusterExecutionEnv` and it would match only if `cluster_type` is `:ascs_ers` (or using the enum `Trento.Clusters.Enums.ClusterType.ascs_ers()`)
- the following `build_env` would need to be refactored to match only when `cluster_type` is either `:hana_scale_up` or `:hana_scale_out`
- maybe an extra `build_env` function might be needed to handle non matching cluster type, but I am not sure. Anyway if needed should not be a big deal adding it

I was also thinking about the possibility to add an extra test for the change in this module, but I see it is already covered in the tests added in `clusters_test.ex` :+1:

test/trento/clusters_test.exs
```diff
@@ -46,6 +46,104 @@ defmodule Trento.ClustersTest do
       assert :ok = Clusters.request_checks_execution(cluster_id)
     end
 
+    test "should start a checks execution on demand for ascs_ers clusters with a resource managed filesystem type" do
```
nelsonkopliku, line 49: Maybe we could group these ascs/ers related tests in a more specific `ASCS/ERS Cluster Checks Execution`.

nelsonkopliku (commented):
Thanks for addressing sending the env to wanda.

Besides some minor things my main feedback is about `AscsErsClusterExecutionEnv`. I believe we can avoid that and rely on the cluster type of the already existent `ClusterExecutionEnv` by doing some pattern matching.

lib/trento/infrastructure/checks/checks.ex
dottorblaster: I didn't like the env being created with `nil` fields due to Elixir's structs behavior but it's ok, we can keep the code simpler.

dottorblaster: (BTW no extra function needed for non matching cluster type: when a clause doesn't match, the subsequent one gets matched against the payload. In the end if nothing matches the BEAM throws the error)

lib/trento/clusters/enums/filesystem_type.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Trento.Clusters.Enums.FilesystemType do
```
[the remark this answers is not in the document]
dottorblaster: I'd like to keep it that way just for the sake of consistency with all the enums in that directory but I feel where your comment is coming from. Thanks for pointing this out!

nelsonkopliku (approved):
Super! Thanks!

PR #2134, opened 2023-12-28
Fix Jest test for ASCS details table

dottorblaster:
# Description
I just found out that the `should show nodes information` test for ASCS/ERS was running once in a while due to peculiar async/await usage. Fixed it.

## How was this tested?
It's a test itself


nelsonkopliku (approved):
LGTM :eyes:

PR #2141, opened 2024-01-04
minor: improve error message

isimluk:
The error message needs to be more generic as various time series may flow through here.


dottorblaster (approved):

dottorblaster:
The red is just related to Chromatic's token not being present in the fork, nothing alarming. We can merge.

PR #2145, opened 2024-01-08
Rename node exporter service in compose file for cross docker version compatibility

CDimonaco:
# Description

Node exporter service is renamed to node_exporter in docker compose file and the dns record is updated in prometheus configuration.

This will prevent issues between the `docker compose` utility and the `docker-compose` utility.

## How was this tested?

Manual and automatic tests

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -3,7 +3,7 @@ volumes:
   pg_data:
   prometheus_data:
 services:
-  node-exporter:
+  node_exporter:
     image: prom/node-exporter:v1.7.0
     volumes:
       - /proc:/host/proc:ro
--- a/prometheus-dev-config.yml
+++ b/prometheus-dev-config.yml
@@ -9,6 +9,6 @@ scrape_configs:
 
   - job_name: 'node'
     static_configs:
-      - targets: ['web-node-exporter-1:9100']
+      - targets: ['web_node_exporter_1:9100']
         labels:
           agentID: "240f96b1-8d26-53b7-9e99-ffb0f2e735bf"


arbulu89 (approved):
This fixes my issue definitely

docker-compose.yaml
```diff
@@ -3,7 +3,7 @@ volumes:
   pg_data:
   prometheus_data:
 services:
-  node-exporter:
+  node_exporter:
```
nelsonkopliku, line 6: What about naming this container explicitly and use that specific name in the `prometheus-dev-config.yml`? That would cover all cases, wouldn't it? :thinking:

arbulu89: Is not the project name mandatory? https://github.com/docker/compose/issues/1621
I mean, docker compose uses it to create the key name for the container

nelsonkopliku: I mean adding `container_name` to node-exporter service
```yaml
node-exporter:
    image: prom/node-exporter:v1.7.0
    container_name: acme-node-exporter
```

and then use `acme-node-exporter` in `prometheus-dev-config.yml`
```yaml
static_configs:
      - targets: ['acme-node-exporter:9100']
```

arbulu89: @nelsonkopliku That works as well for me.
I'm fine with either of the solutions, I guess yours is more robust, as it doesn't require to name the project folder as `web`

nelsonkopliku: yes, the name might conflict.
We could even use `name` entry in the root of the docker compose yaml
```yaml
version: "3"
name: trentoz
...
```
getting something like `trentoz-node-exporter-1`

Whatever works best for you guys. If I'll need to rename my local directory I will do :smile:

CDimonaco: Naming container with a static name inside a docker-compose is not a good practice, you can have conflicts as that name will be set for the container and you can have other container with the same name, that's why the docker compose project name exists.

It's a convention, if we have to start worrying about these things we go crazy in a couple of minutes, we need to rename all the services in all the compose and hope that they will not conflict with anything.

I'm with the idea of having the web prefix, we are in control of our environment, I had painful experience with container with fixed names

nelsonkopliku: I see the shortcoming of naming the container, that is fair. If I had to choose I'd rather use the project name

However, don't worry. I will get my setup aligned with the required naming :wink:

CDimonaco: perfect! You can also use direnv and set a env variable for the directory without changing anything, I use it in several places, and I have the direnv file in my global gitignore, I will merge

PR #2147, opened 2024-01-08
Enable/disable charts feature

CDimonaco:
# Description

This pr aims to enable and disable the charts feature, both in the backend/frontend.

By default the feature is enabled, can be disabled with a runtime env variable.

On the backend, a dedicated plug will return 501 for all the charts endpoint, when the feature is disabled.
On frontend, the chart components will not be displayed, instead a text with a dedicated copy is shown to the user.

Frontend result

image

## How was this tested?

Automated tests


lib/trento_web/plugs/charts_disabled_plug.ex
```diff
@@ -0,0 +1,33 @@
+defmodule TrentoWeb.Plugs.ChartsDisabledPlug do
+  @moduledoc """
+    This plug will act as a barrier for the charts endpoint, will return 501 for all the requests.
```
arbulu89, line 3: I always recommend to write docs in present term `This plug acts as a...`

lib/trento_web/plugs/unplug_charts_enabled_predicate.ex
```diff
@@ -0,0 +1,8 @@
+defmodule TrentoWeb.Plugs.UnplugChartsEnabledPredicate do
```
arbulu89, line 1: Maybe it would be better to organize these predicates in its own folder.
`/plugs/unplug/predicates/charts_enabled.ex`
`defmodule TrentoWeb.Plugs.Unplug.Predicates.ChartsEnabled`

arbulu89: Besides that, I would rename it to `ChartsDisabled` as it is true when they are disabled.
Otherwise, the usage in the router looks contradictory, as it uses the Disabled plug with the enabled predicate.

assets/js/common/ChartDisabledBox/ChartDisabledBox.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+function ChartDisabledBox() {
```
arbulu89, line 3: Maybe `ChartsDisabledBox` as it disables multiple charts (of course based on the usage) and the text says so

assets/js/common/ChartDisabledBox/ChartDisabledBox.jsx
```diff
@@ -0,0 +1,16 @@
...
+  return (
+    <div
+      className="mt-4 bg-white shadow rounded-lg py-4 px-8 mx-auto"
+      data-testid="chart-disabled-box"
+    >
+      <h2 className="font-bold text-center text-xl">
+        Charts are disabled, please check documentation for further details
```
arbulu89, line 10: Maybe `Charts are disabled. Please check the documentation for further details`.
Could we point the user to that documentation with a link?
Otherwise it doesn't look like that helpful

arbulu89 (commented):
Hey!
Good stuff.

assets/js/common/ChartDisabledBox/ChartDisabledBox.jsx
CDimonaco: Yes, but we don't have the docs yet, so atm I think we could have as is

lib/trento_web/plugs/charts_disabled_plug.ex
CDimonaco: done

lib/trento_web/plugs/unplug_charts_enabled_predicate.ex
CDimonaco: nice catch! done

assets/js/common/ChartDisabledBox/ChartDisabledBox.jsx
CDimonaco: done

assets/js/common/ChartFeatureWrapper/ChartFeatureWrapper.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import ChartsDisabledBox from '@common/ChartsDisabledBox';
+
+function ChartFeatureWrapper({ children }) {
```
arbulu89, line 4: Maybe this as `Charts..` as well :see_no_evil:

assets/js/common/ChartsDisabledBox/ChartsDisabledBox.stories.jsx
```diff
@@ -0,0 +1,10 @@
+import ChartDisabledBox from '.';
```
arbulu89, line 1: You forgot to update the names on this file

lib/trento_web/plugs/charts_disabled_plug.ex
```diff
@@ -0,0 +1,33 @@
+defmodule TrentoWeb.Plugs.ChartsDisabledPlug do
+  @moduledoc """
+    This plug acts as a barrier for the charts endpoint, will return 501 for all the requests.
```
arbulu89, line 3: Still `will` here -> `returns 501...`

arbulu89 (approved):
Great!
Commented some other small leftovers.
No need for a new review

PR #2153, opened 2024-01-09
Fix ENSa version computation run only over application instances

dottorblaster:
# Description

Fixing ENSA version computation for ASCS/ERS clusters: we only need to take into account application instances

## How was this tested?
Existing unit tests modified


nelsonkopliku (approved):
The fix looks good to me, thanks for addressing it :+1: 

I guess we can get this merged and follow up on:
- ensa version not being sent in the checks selection of an ascs/ers cluster
- ascs/ers information being sent also for hana clusters, where are not necessary

both related to what commented here https://github.com/trento-project/web/pull/2126

PR #2154, opened 2024-01-09
Remove Reset Filters button in Checks selection

nelsonkopliku:
# Description

This PR allows to decide whether the `CatalogContainer` component should also show a `Reset Filters` button when an empty catalog is detected.

Needed to be able to have that reset filter in the catalog page, but not in the checks selection.

## How was this tested?

Test added. Outdated storybooks fixed.

https://2154.prenv.trento.suse.com/clusters/0eac831a-aa66-5f45-89a4-007fbd2c5714/settings

Stories
https://643fc66e9747e47344899b8c-dbrfuwanza.chromatic.com/?path=/story/layouts-checkscatalog--empty
https://643fc66e9747e47344899b8c-dbrfuwanza.chromatic.com/?path=/story/patterns-checksselection--empty

### DIFF
--- a/assets/js/pages/ChecksCatalog/CatalogContainer.jsx
+++ b/assets/js/pages/ChecksCatalog/CatalogContainer.jsx
@@ -9,6 +9,7 @@ import LoadingBox from '@common/LoadingBox';
 function CatalogContainer({
   onClear = () => {},
   onRefresh = () => {},
+  withResetFilters = false,
   empty = false,
   catalogError = null,
   loading = false,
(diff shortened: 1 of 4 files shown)


assets/js/pages/ChecksCatalog/CatalogContainer.jsx
```diff
@@ -9,6 +9,7 @@ import LoadingBox from '@common/LoadingBox';
 function CatalogContainer({
   onClear = () => {},
   onRefresh = () => {},
+  withResetFilters = true,
```
dottorblaster, line 12: You can set `withResetFilters` to `false` by default and have the API be more idiomatic (and less verbose) whenever you want to set that prop to true

`<CatalogContainer withResetFilters />`

dottorblaster (changes requested):
Just one thing then we are ready to merge.

Thanks for keeping the stories up to date!

dottorblaster (approved):

PR #2164, opened 2024-01-11
Fix sequence usage in sles subscription factory

dottorblaster:
# Description
Sometimes we build lists of SLES subscriptions using the factory we have, that doesn't use sequences to assign different numbers to the identifier field which is actually a part of a primary key.

Given this, the problem we're incurring into is that our unit tests pipeline runs into collisions. This change should fix that.

## How was this tested?
Existing tests passing


nelsonkopliku (approved):
LGTM :eyes:

PR #2177, opened 2024-01-15
Add ENSA version info to Checks execution

jamie-suse:
# Description

Adds ENSA version info to Checks execution.

## How was this tested?

Added unit tests.

### DIFF
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -8,6 +8,7 @@ defmodule Trento.Clusters do
   require Logger
   require Trento.Clusters.Enums.ClusterType, as: ClusterType
   require Trento.Clusters.Enums.FilesystemType, as: FilesystemType
+  require Trento.Clusters.Enums.EnsaVersion, as: EnsaVersion
 
   alias Trento.Hosts.Projections.HostReadModel
 
@@ -16,6 +17,8 @@ defmodule Trento.Clusters do
   alias Trento.Clusters.ClusterEnrichmentData
 
   alias Trento.Clusters.Commands.SelectChecks
+
+  alias Trento.SapSystems.Projections.SapSystemReadModel
 
   alias Trento.Infrastructure.Checks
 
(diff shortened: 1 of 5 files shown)


nelsonkopliku:
> There is an existing enum `Trento.SapSystems.Enums.EnsaVersion`, however it does not contain a value for `:mixed_versions`. Should I create a new enum under `Trento.Clusters.Enums.EnsaVersion` that reflects this?

ensa version is a specificity of SapSystems, however  the its `mixed_versions` nuance is peculiar to a cluster.
We could indeed have a cluster specific `EnsaVersion` enum.
I am thinking of what could go wrong, but I don't see blockers.

lib/trento/clusters.ex
```diff
@@ -175,6 +198,8 @@ defmodule Trento.Clusters do
           %Checks.ClusterExecutionEnv{
             provider: provider,
             cluster_type: cluster_type,
+            ensa_version:
+              details["sap_systems"] |> Enum.map(fn s -> s["sid"] end) |> get_ensa_version,
```
nelsonkopliku, line 202: We could make `get_ensa_version` accept the cluster's `details` (or the whole cluster) and encapsulate extraction of the sids of a cluster there, so that here we'd have something like

```elixir
    ensa_version: get_ensa_version(details), # or get_ensa_version(cluster)
```

Another option might be having a private function for sids extraction from a cluster so that here we'd have something like
```elixir
    ensa_version: cluster |> get_sids |> get_ensa_version
```

Not a strong requirement, just sharing thoughts. I am also fine with current proposal. Up to you.

lib/trento/infrastructure/checks/cluster_execution_env.ex
```diff
@@ -9,10 +9,12 @@ defmodule Trento.Infrastructure.Checks.ClusterExecutionEnv do
   require Trento.Enums.Provider, as: Provider
   require Trento.Clusters.Enums.ClusterType, as: ClusterType
   require Trento.Clusters.Enums.FilesystemType, as: FilesystemType
+  require Trento.SapSystems.Enums.EnsaVersion, as: EnsaVersion
```
nelsonkopliku, line 12: The ensa version we use in the Cluster Execution Env is `Trento.Clusters.Enums.EnsaVersion`, not the SapSystem one

nelsonkopliku (changes requested):
Just a tiny thing in `cluster_execution_env.ex`: we need to use the correct `EnsaVersion` enum.

Besides that, LGTM.

nelsonkopliku (approved):
Great, thanks!

test/trento/clusters_test.exs
```diff
@@ -278,7 +286,90 @@ defmodule Trento.ClustersTest do
...
+
+        assert message.target_type == "cluster"
+
+        :ok
+      end)
+
+      assert :ok = Clusters.request_checks_execution(cluster_id)
+    end
+
+    test "should start a checks execution on demand for ascs_ers clusters with ENSA 1 version" do
```
jamie-suse, line 303: I haven't included a test(s) for the case where a cluster has only one SAP system, but that seems low-value, what do you think?

nelsonkopliku: Agreed

dottorblaster (changes requested):
One little thing then we can merge

lib/trento/clusters.ex
```diff
@@ -130,6 +133,30 @@ defmodule Trento.Clusters do
...
+
+    ensa_versions =
+      query
+      |> Repo.all()
+      |> Enum.uniq()
+
+    case length(ensa_versions) do
+      1 -> Enum.at(ensa_versions, 0)
+      _ -> EnsaVersion.mixed_versions()
+    end
```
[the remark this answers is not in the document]
jamie-suse: Thanks! I was trying to figure out how to do essentially this :laughing: looks a lot better :+1:

dottorblaster (approved):
LGTM, great job :+1:

nelsonkopliku (approved):

PR #2184, opened 2024-01-16
Docker compose utility adapt

CDimonaco:
# Description

Adapt the new `docker compose` utility dns naming, we use it only in the prometheus config.

From this PR, the `docker-compose` version 1 utility is **NOT SUPPORTED**.

## How was this tested?

Automated tests


nelsonkopliku (approved):
LGTM

arbulu89 (approved):
Allright!
I guess I need to update it :smiling_face_with_tear:

PR #2186, opened 2024-01-17
Upgrade elixir 1.15.7 otp 26

arbulu89:
# Description

Upgrade elixir to 1.15.7 and erlang to OTP26 version, in order to match with packages provided in SUSE (by now tumbleweed).
I have taken advantage of updating the dependencies as well.

Note: The `api_key` tests were updated to use new keys, as the erlang crypto module has changed the signing strategy. Even in this case, the old `api_keys` are valid to validate, so it is a non op for the user. They will see a different key in the frontend settings page, that's true

PD: To change the dockerfiles. They need some more work
PD2: For some reason the command `mix do ecto.drop, event_store.drop, ecto.create, ecto.migrate, setup` is raising an error. I needed to run `mix setup` separately afterwards to make it work.

## How was this tested?

Current tests


.github/workflows/ci.yaml
```diff
@@ -527,27 +523,27 @@ jobs:
...
-            --json /specs/changes.json \
-            --text /specs/changes.txt \
-            --html /specs/changes.html
-      - name: Upload OpenAPI diff report
-        uses: actions/upload-artifact@v4
-        if: failure()
-        with:
-          name: openapi-diff-report-${{ matrix.version }}
-          path: specs/
+      # - name: Generate main openapi.json
```
arbulu89, line 526: I'm forced to comment this stage out.
We cannot compare specs written in 2 different elixir versions.
Commenting this looks the less invasive way.
I will revert back once this PR is merged

nelsonkopliku (approved):
Thanks a lot!

Happy to see that changes  needed in the codebase were limited :clap:

.github/workflows/ci.yaml
dottorblaster: Absolutely plus one, no problem for me

dottorblaster (approved):
Great work! LGTM

EMaksy (approved):
Nice, thanks guys!

CDimonaco (approved):

PR #2187, opened 2024-01-17
Saptune summary warning icon

EMaksy:
# Description
Add a warning state icon to the saptune overview when no saptune version is provided and sap workload is happening on the host

Left on host with sap workload and right without.
Preview: 
image

## How was this tested?
Updated existing good test to check if the icon was rendered.

### DIFF
--- a/assets/js/pages/SaptuneDetails/SaptuneVersion.jsx
+++ b/assets/js/pages/SaptuneDetails/SaptuneVersion.jsx
@@ -7,11 +7,13 @@ import HealthIcon from '@common/HealthIcon';
 
 function SaptuneVersion({ isSapPresent = true, version }) {
   if (!version) {
-    return (
+    return isSapPresent ? (
       <div className="flex">
-        {isSapPresent ? <HealthIcon health="warning" /> : null}
+        <HealthIcon health="warning" />
         <span className="ml-1">Not installed</span>
       </div>
+    ) : (
+      <span>Not installed</span>
     );
   }
 
--- a/assets/js/pages/SaptuneDetails/SaptuneVersion.jsx
+++ b/assets/js/pages/SaptuneDetails/SaptuneVersion.jsx
@@ -5,9 +5,9 @@ import { isVersionSupported, SUPPORTED_VERSION } from '@lib/saptune';
 import Tooltip from '@common/Tooltip';
 import HealthIcon from '@common/HealthIcon';
 
-function SaptuneVersion({ isSapPresent = true, version }) {
+function SaptuneVersion({ sapPresent = true, version }) {
   if (!version) {
-    return isSapPresent ? (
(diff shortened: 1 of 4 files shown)


arbulu89 (commented):
Hey @EMaksy ,
Did you discuss the scenario where nodes without saptune will display this as well with @abravosuse ?
I mean, in that scenario, the host health is not changed to warning, and I think, this story comes from the need to reflect the warning point in the details view when the host list shows a warning health

abravosuse:
@EMaksy just to clarify the discussion we had on Friday:

Right now, when a version of saptune lower than 3.1.0 is detected in a host, a warning icon is displayed in the corresponding details view, regardless whether there is an SAP workload or not. But only in the case there is such workload, the warning gets added to the aggregated health of the host.

The behavior should be the same when no saptune is detected:

- a warning icon is displayed in the corresponding details view, regardless whether there is an SAP workload or not.
- But only in the case there is such workload, the warning gets added to the aggregated health of the host

EMaksy:
Thanks guys, for clarifying, I will ensure that it behaves like @abravosuse describes it :+1:

arbulu89:
@abravosuse @EMaksy if this is the case, i think the code might be ok.
Anyway, i'm surprised of this desired behaviour. Showing a warning icon if saptune is not installed in a host that doesn't need saptune doesn't look really appropriate for me. Specially because the host health is not affected by it.
My 2 cents.

arbulu89 (approved):

abravosuse:
@EMaksy apologies for the back and forth. Xabi has a point. Please scratch what I said before and go for the following behavior:

- a warning icon is displayed in the corresponding details view only if there is an SAP workload in the host
- in that case, the warning is added to the aggregated health of the host (this last part is already implemented!)

assets/js/pages/SaptuneDetails/SaptuneVersion.jsx
```diff
@@ -5,9 +5,16 @@ import { isVersionSupported, SUPPORTED_VERSION } from '@lib/saptune';
 import Tooltip from '@common/Tooltip';
 import HealthIcon from '@common/HealthIcon';
 
-function SaptuneVersion({ version }) {
+function SaptuneVersion({ isSapPresent = true, version }) {
   if (!version) {
-    return <span>Not installed</span>;
+    return isSapPresent ? (
```
arbulu89, line 10: Not the most beautiful piece of code, but I guess it works

arbulu89 (approved):
Good!
I think it makes the job

CDimonaco (approved):
LGTM for me,

As a suggestion, I reccomend to change the name of props from `isSapPresent` to `sapPresent`.

rtorrero (commented):
LGTM thanks @EMaksy

PR #2197, opened 2024-01-18
Fix broken HostDetails story

EMaksy:
# Description

This pr will add a little fix for the HostDetails story. The introduction of the new charts for the HostDetails view made a global config required in order to render the story correct.
By providing a mocked config in the decorator, the charts are disabled.

Preview: 
image

## How was this tested?
No test needed, check storybook HostDetails.stories.jsx


EMaksy:
I would create a new task for creating a new story with properly rendered charts, this pr is just a quick fix.

arbulu89 (commented):
I'm not honestly sure if this is the way to go.
I would try to already create the change including the view with and without charts, because I see that you will need to maybe undo this change.

EMaksy:
Thanks for taking care with a proper fix :)
I will close this pr!

PR #2204, opened 2024-01-19
Project host FQDN

dottorblaster:
# Description

Project the FQDN sent by the agent on the host read model.

## How was this tested?

Existing unit tests updated

### DIFF
--- a/lib/trento/discovery/payloads/host_discovery_payload.ex
+++ b/lib/trento/discovery/payloads/host_discovery_payload.ex
@@ -3,7 +3,15 @@ defmodule Trento.Discovery.Payloads.HostDiscoveryPayload do
   Host discovery integration event payload
   """
 
-  @required_fields :all
+  @required_fields [
+    :hostname,
+    :ip_addresses,
+    :agent_version,
+    :cpu_count,
+    :total_memory_mb,
+    :socket_count,
+    :os_version
+  ]
 
   use Trento.Support.Type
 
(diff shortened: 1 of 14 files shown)


jamie-suse (approved):
LGTM :rocket:

nelsonkopliku (changes requested):
I am afraid we're missing some parts of the _changing fqdn_ scenario.

**During a host's restoration**
in execute on line 173 we also need to match the fqdn between the aggregate's state and the incoming `RegisterHost` command, meaning that _the exact same host has been restored_ and so only a `HostRestored` is emitted

**During a host's details changes detection**
in executes on lines 259 and 284 fqdn is not properly taken into account.
In the first command handling we need to match also fqdn between aggregate's state and the incoming `RegisterHost` so that if they're the same no event is emitted.
In the second command handling, when the previous does not match (aka different details have been detected), we need to get the fqdn from the incoming `RegisterHost` command and emit it in `HostDetailsUpdated` (and of course update aggregate's state accordingly in the relevant apply)

Happy to answer any questions.

dottorblaster:
@nelsonkopliku thanks for catching that! About changes detection: totally right, I updated the code accordingly.

About restoration: not a big fan of matching the FQDN in that clause because I don't think it's a relevant detail to base the restoration logic upon, but I don't have a strong opinion either. Change done. :+1:

nelsonkopliku:
@dottorblaster the problem with restoration is that if all other details remain unchanged, BUT `fully_qualified_domain_name` does change, we are yes going to have a `HostRestored` event emitted, but not a `HostDetailsUpdated` resulting in a dirty state both in the aggregate and in the read models.

nelsonkopliku (approved):
Everything looks fine now, thanks!
We might want to have at least one photofinish scenario sending the fqdn, for completeness.

PR #2212, opened 2024-01-23
Add ability to encrypt and decrypt plain content

jamie-suse:
# Description

This change adds the ability to encrypt and decrypt passwords (_and any other plaintext_), utilizing the [Cloak](https://hexdocs.pm/cloak/readme.html) library for the encryption algorithms.

Note: The _Cloak_ library requires that a new secret environment variable be added. I have named it `ENCRYPTION_KEY_BASE`.

## How was this tested?

Added unit tests


jamie-suse:
Ideally we could provide `SECRET_KEY_BASE` to the Cloak Vault. However Cloak requires a key size of _exactly_ 256 bits for it's ciphers, and our `SECRET_KEY_BASE` is not.

lib/trento/vault.ex
```diff
@@ -0,0 +1,22 @@
...
+        default: {
+          Cloak.Ciphers.AES.GCM,
+          tag: "AES.GCM.V1",
+          key: Base.decode64!(Application.fetch_env!(:trento, Trento.Vault)[:encryption_key_base]),
+          iv_length: 12
+        }
+      )
+
+    {:ok, config}
+  end
```
dottorblaster, line 21: You can use just `use Cloak.Vault` avoiding this callback providing the configuration directly inside the config files.

lib/trento/crypto.ex
```diff
@@ -0,0 +1,17 @@
...
+  @spec encrypt(String.t()) :: binary()
+  def encrypt(plaintext) do
+    Vault.encrypt!(plaintext)
+  end
+
+  @spec decrypt(binary()) :: String.t()
+  def decrypt(ciphertext) do
+    Vault.decrypt!(ciphertext)
+  end
+end
```
dottorblaster, line 17: Do we really need this proxy module just to call `Vault`'s functions?

dottorblaster (changes requested):
Thanks for this! Some changes to apply, but we are definitely setting course for merging :rocket:

CDimonaco (changes requested):
Completely agree with alessio, you can avoid reimplementing the init callback with runtime configuration, and also the facade is pretty useless.

Also, having a default config in config.exs for Cloak will avoid all the repetition.

Anyway you can also avoid a new env variable, you can adapt the secret key base to be like the Cloak secret key and use that as encryption key.

nelsonkopliku:
I am ok with the changes requested, however I am not sure about adapting the secret key base.

What would happen to running installations? 
I guess they would need to be restarted in order to provide the new secret and since secret key base is used to generate api keys used by agents, I guess agents need to be restarted (and config updated very likely) with a new api key generated with the new secret.

If this is not the case, then ok with adapting already existent secret key base, otherwise I would be more cautious.

CDimonaco:
> I am ok with the changes requested, however I am not sure about adapting the secret key base.
> 
> What would happen to running installations? I guess they would need to be restarted in order to provide the new secret and since secret key base is used to generate api keys used by agents, I guess agents need to be restarted (and config updated very likely) with a new api key generated with the new secret.
> 
> If this is not the case, then ok with adapting already existent secret key base, otherwise I would be more cautious.

That's a valid point, I think we could use `secret_key_base` but truncated to 256 bits like `:binary.part(secret_key_base, {0, 32})` this way we can avoid another variable and also be predictable on the encryption key

nelsonkopliku:
@dottorblaster @CDimonaco 
- removed the facade module
- moved the config in runtime. It would be added in any environment
- leveraging existent secret key base

nelsonkopliku (approved):
:trollface:

config/runtime.exs
```diff
@@ -142,3 +142,15 @@ if config_env() in [:prod, :demo] do
   config :trento,
     api_key_authentication_enabled: System.get_env("ENABLE_API_KEY", "true") == "true"
 end
+
+secret_key_base = Application.fetch_env!(:trento, TrentoWeb.Endpoint)[:secret_key_base]
```
CDimonaco, line 146: Please use secret_key_base as variable, no need to refetch, you can move this config block to the same block as other runtime,

You need to put the same block in config.exs for the base config, in runtime we should only have runtime dependent config.

test/trento/vault_test.exs
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    defmodule RuntimeVault do
+      @moduledoc """
+      A vault used for testing purposes
+      """
+      use Cloak.Vault, otp_app: :trento
+    end
+
+    test "should not decrypt when the content has been encrypted with a different key" do
```
CDimonaco, line 45: No need to test it, Cloak is already tested, and I'm not sure anyway this is the right way to inject a genserver into a test

CDimonaco (changes requested):
Thanks for the changes, overall  a goob, we just need to fix the config, having a default in config.exs, that will apply to all the envs except for overrides, like runtime.

config/runtime.exs
```diff
@@ -142,3 +142,15 @@ if config_env() in [:prod, :demo] do
...
+
+config :trento, Trento.Vault,
+  ciphers: [
+    default: {
+      Cloak.Ciphers.AES.GCM,
+      tag: "AES.GCM.V1",
+      key: secret_key_base |> Base.decode64!() |> :binary.part({0, 32}),
+      iv_length: 12
+    }
+  ]
```
dottorblaster, line 156: Please move this into the `if` statement :+1:

test/trento/vault_test.exs
```diff
@@ -0,0 +1,77 @@
...
+        secret3: "AQpBRVMuR0NNLlYxhAx+UMg0XkqCyAXPv7UlNg7xVeig4sMLrwapysrGNvD8A9k="
+      ]
+
+      Enum.each(encrypted_secrets, fn {decrypted, encrypted} ->
+        plaintext = encrypted |> Base.decode64!() |> Vault.decrypt!()
+        assert plaintext == Atom.to_string(decrypted)
+      end)
+    end
+
+    defmodule RuntimeVault do
```
dottorblaster, line 38: This really feels like we are testing the library itself, are you sure about that?

dottorblaster (changes requested):
Really good job, just one more thing: `secret_key_base` is defined also in `dev.exs` for example so I'd try to follow the same approach we follow for the `secret_key_base` variable.

test/trento/vault_test.exs
nelsonkopliku: shame on me. removed :smile:

nelsonkopliku:
@dottorblaster @CDimonaco changes applied

CDimonaco (approved):

PR #2213, opened 2024-01-23
Input fields

nelsonkopliku:
# Description

A bit of react verbosity :smile: 
Adds Components for input fields.
Based on https://github.com/react-component/input

## How was this tested?

Automated tests, storybook.

### DIFF
--- a/assets/js/common/Input/Password.jsx
+++ b/assets/js/common/Input/Password.jsx
@@ -13,7 +13,6 @@ function Password({
   name,
   value,
   placeholder = 'Password',
-  allowToggleVisibility = false,
   disabled = false,
   onChange = () => {},
 }) {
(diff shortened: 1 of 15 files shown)


assets/js/common/Input/Password.jsx
```diff
@@ -0,0 +1,52 @@
...
+import classNames from 'classnames';
+import Input from './Input';
+
+function Password({
+  className,
+  id,
+  name,
+  value,
+  placeholder = 'Password',
+  allowToggleVisibility = false,
```
dottorblaster, line 16: I would strip off this, I think we just want visibility to always be togglable. Your call nonetheless.

assets/js/common/Input/Textarea.jsx
```diff
@@ -0,0 +1,28 @@
...
+  id,
+  name,
+  value,
+  initialValue,
+  placeholder,
+  disabled = false,
+  onChange,
+}) {
+  const [initialInputValue, setValue] = useState(initialValue);
+  const defaultOnChange = (e) => setValue(e.target.value);
```
dottorblaster, line 13: Why this? `value` wasn't enough?

dottorblaster (changes requested):
Just a couple things (mainly questions) but overall really great job, thank you so much :+1:

assets/js/common/Input/Password.jsx
nelsonkopliku: I don't mind much, but if we want one day to also use this Password component in the login page it might be useful :eyes:

assets/js/common/Input/Textarea.jsx
nelsonkopliku: As mentioned privately the reasons are two:
- I got inspired by the api that also rc-input exposes
- without `initialValue` and `defaultOnChange` we can only have value controlled from the outside

I don't mind here either, whatever you prefer.

assets/js/common/Input/Password.jsx
dottorblaster: I personally don't have a big problem with that, for example here's my Fastmail login screen that lets me toggle password visibility for the login screen :+1: it would actually be a feature to me :smile:

immagine

@jagabomb what do you think?

assets/js/common/Input/Textarea.jsx
dottorblaster: Leave it, no problem with this :+1:

assets/js/common/Input/Password.jsx
nelsonkopliku: Yes, good point. Happy to make that the default behavior in case. 👍

jagabomb: @nelsonkopliku This makes sense to me. Always have the password visibility toggle present.

nelsonkopliku:
@dottorblaster @jagabomb thanks for the feedback. Made the password togglable the default behavior.

dottorblaster (approved):

PR #2214, opened 2024-01-23
Fix required fields usage

arbulu89:
# Description
Fix `validate_required_fields` usage. We were using this function incorrectly in many places.
When we have an explicit `changeset` implemented for the `Type`, the `validate_required_fields` only works for the type itself. If there are other inline types with their own `changeset` functions, we cannot use it, as it would expand the `fields` of the main module, and not run the proper validation.
Together with this, in other to check if embedded fields are required or not, we cannot use `validate_required_fields`, we need to set the `required: true` in the `embed_cast` itself.

Using the macro `changeset` all of this is done there.

**PD: I have tried to put the required values in the places where it makes sense. Until now, any of them on the `embeds_*` was working, so it was like there were not required. Many of them actually shouldn't be required...**

As a bonus, I have removed the `sbd` from the required list (even though it was not really working) to avoid an error we were having when the cluster doesn't have sbd configured. FYI @abravosuse 

Ref: https://github.com/trento-project/web/pull/1422/files#r1196297634

## How was this tested?
Current tests passing

### DIFF
--- a/lib/trento/discovery/policies/cluster_policy.ex
+++ b/lib/trento/discovery/policies/cluster_policy.ex
@@ -278,6 +278,8 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
     end)
   end
 
+  defp parse_sbd_devices(_), do: []
+
   defp parse_node_resources(node_name, crmmon) do
     crmmon
     |> extract_cluster_resources()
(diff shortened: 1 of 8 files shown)


lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -278,6 +278,8 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
     end)
   end
 
+  defp parse_sbd_devices(_), do: []
```
arbulu89, line 281: Enable empty sbd data coming, for future non sbd scenario.
PD: without this, we were having a nasty error

CDimonaco (approved):

nelsonkopliku (approved):
:eyes:

PR #2220, opened 2024-01-25
Hana scale out discovery

arbulu89:
# Description
HANA scale out discovery.
We only need to change the policy.
In a scale out cluster how we identify different fields is different.
We need to find them in the `cluser_properties` field. Check the test fixtures to see them.
The major difficulty comes when these attributes are not always present...
This depends on if some resource agents hooks are used or not...
As far as I discovered there are 3 ways:
- No hooks. So the fields might not be there, so some fields will be populated with `unknown` value (**I have not been able to get any fixture for this, so it is just a "I hope it works" effort**)
- `SAPHanaSR.py` hook. This is old hook. In this case we don't have the `hana_<sid>_site_srHook_<site>` attributes
- `SAPHanaSrMultiTarget.py` hook, which provides the latest fields.

Now...
- The value in nodes `role` column needs to be redefined. This value is `site` based instead of host based. In scale up it doesn't really matter to show it wrong, but we need a new design most probably
- The `SAPHanaSR health state` field is as well site based. Right now, we only show the secondary value, but we could show both sites in fact
- The majority maker nodes is not visualized

@abravosuse the topics above must be discussed (we already started some days ago in fact).

In a next PR I want to add some storybook story, add some E2E test just in case and upload a real scale out fixtures set.

This is how it looks like:
scale_out

## How was this tested?

Tested in the backend

### DIFF
--- a/assets/js/pages/ClusterDetails/HanaClusterDetails.jsx
+++ b/assets/js/pages/ClusterDetails/HanaClusterDetails.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { get, groupBy } from 'lodash';
 import classNames from 'classnames';
 
+import { getClusterTypeLabel } from '@lib/model/clusters';
 import { RUNNING_STATES } from '@state/lastExecutions';
 
 import BackButton from '@common/BackButton';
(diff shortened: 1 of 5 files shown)


abravosuse (commented):
It's almost there @arbulu89 but, as discussed by Slack, maybe better to hold it until we have a final UI design. I will work on the ticket for @jagabomb tomorrow.

EMaksy (approved):
Hey man could not find anything which does not look good :+1:

rtorrero (approved):
LGTM @arbulu89, thanks!

PR #2222, opened 2024-01-25
Fix ENSA version calculation

jamie-suse:
# Description

This PR is a correction to the aggregated ENSA version calculation made in https://github.com/trento-project/web/pull/2177

The way the relevant SAP systems and resulting ENSA version are determined is by:

1. Record all hosts that have the cluster's ID in a list
2. From there, get all application instances that are hosts from the _hosts list_ **and** have an SID managed from the cluster (found in the `additional_sids` field for the cluster)
3. With the list of relevant application instances, their corresponding SAP system IDs can be found, and recorded as a list of distinct SAP system IDs.
4. Lastly, with the relevant SAP systems found, their individual ENSA versions can be recorded, and the aggregated ENSA version can be calculated as previously.

## How was this tested?

Updated unit tests to validate the new logic

### DIFF
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -179,15 +179,16 @@ defmodule Trento.Clusters do
           select: %{host_id: h.id, sap_system_id: a.sap_system_id}
       )
 
-    sap_system_ids =
+    aggregated_ensa_version =
       hosts_data
       |> Enum.map(fn %{sap_system_id: sap_system_id} -> sap_system_id end)
       |> Enum.uniq()
+      |> get_aggregated_ensa_version()
 
     env = %Checks.ClusterExecutionEnv{
       provider: provider,
       cluster_type: ClusterType.ascs_ers(),
-      ensa_version: get_ensa_version(sap_system_ids),
+      ensa_version: aggregated_ensa_version,
       filesystem_type: get_filesystem_type(cluster_id)
     }
 
@@ -231,18 +232,16 @@ defmodule Trento.Clusters do
 
   defp maybe_request_checks_execution(error), do: error
 
-  @spec get_ensa_version([String.t()]) :: EnsaVersion.t()
-  defp get_ensa_version(sap_system_ids) do
-    query =
-      from(s in SapSystemReadModel,
-        select: s.ensa_version,
-        where: s.id in ^sap_system_ids
-      )
-
+  @spec get_aggregated_ensa_version([String.t()]) :: EnsaVersion.t()
+  defp get_aggregated_ensa_version(sap_system_ids) do
     ensa_versions =
-      query
-      |> Repo.all()
-      |> Enum.uniq()
+      Repo.all(
+        from(s in SapSystemReadModel,
+          select: s.ensa_version,
+          where: s.id in ^sap_system_ids and is_nil(s.deregistered_at),
+          distinct: true
+        )
+      )
 
     case ensa_versions do
       [ensa_version] -> ensa_version
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -188,25 +188,39 @@ defmodule Trento.Clusters do
          selected_checks: selected_checks
        }) do
     hosts_data =
-      Repo.all(
-        from h in HostReadModel,
-          select: %{host_id: h.id},
-          where: h.cluster_id == ^cluster_id and is_nil(h.deregistered_at)
-      )
+      case cluster_type do
+        :ascs_ers ->
+          Repo.all(
+            from h in HostReadModel,
+              select: %{host_id: h.id},
+              join: a in ApplicationInstanceReadModel,
+              on: h.id == a.host_id,
+              where:
+                h.cluster_id == ^cluster_id and is_nil(h.deregistered_at) and
+                  a.sid in ^cluster_sids
+          )
 
-    host_ids = Enum.map(hosts_data, fn h -> h.host_id end)
-
-    sap_system_ids =
-      Repo.all(
-        from a in ApplicationInstanceReadModel,
-          select: a.sap_system_id,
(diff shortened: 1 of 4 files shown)


lib/trento/clusters.ex
```diff
@@ -181,29 +178,53 @@ defmodule Trento.Clusters do
...
+              where: h.cluster_id == ^cluster_id and is_nil(h.deregistered_at)
+          )
+      end
 
     env =
       case cluster_type do
         :ascs_ers ->
+          host_ids = Enum.map(hosts_data, fn h -> h.host_id end)
+
+          sap_system_ids =
```
arbulu89, line 216: Do you need this query?
You could get the sap system id in the query above including in in the select

arbulu89 (changes requested):
Hey @jamie-suse 
I think there is room for improvement in the code, beyond the logic now looks correct.

Anyway, out of that, and not asking to do changes about this, just food for thoughts.
I think that personally, I would design the code differently. I would have a function named `getSapSystemsByCluster(cluster_id)` or similar, which returns you all the sap systems belonging to this cluster, returning the sap system read model, and from there, you can get the needed field, in this case the ensa version.
The current code looks so tailored to the current need that those functions are not really usable anywhere else.

nelsonkopliku:
> I would design the code differently. I would have a function named `getSapSystemsByCluster(cluster_id)` or similar...

@arbulu89 this is a good point which might result in the opportunity of being able to expose the information about _which ensa versions are there in this ascs ers cluster_ directly from the backend, so we do not have the same logic applied in the frontend.

I would track this as debt though, because this _logic_ might be applied at read time by composing query results or it might be projected either in host read model or a dedicated read model if we leverage events.

jamie-suse:
Okay I will make a ticket for tracking the refactor :+1:

lib/trento/clusters.ex
```diff
@@ -225,4 +230,23 @@ defmodule Trento.Clusters do
...
+    query =
+      from(s in SapSystemReadModel,
+        select: s.ensa_version,
+        where: s.id in ^sap_system_ids
+      )
+
+    ensa_versions =
+      query
+      |> Repo.all()
+      |> Enum.uniq()
```
arbulu89, line 245: You could add `distinct` in the query above as well

lib/trento/clusters.ex
```diff
@@ -225,4 +230,23 @@ defmodule Trento.Clusters do
   end
 
   defp maybe_request_checks_execution(error), do: error
+
+  @spec get_ensa_version([String.t()]) :: EnsaVersion.t()
+  defp get_ensa_version(sap_system_ids) do
```
arbulu89, line 235: The name is a bit misleading for me, but I don't have any better suggestion, so...

lib/trento/clusters.ex
```diff
@@ -181,13 +159,52 @@ defmodule Trento.Clusters do
...
+          where:
+            h.cluster_id == ^cluster_id and is_nil(h.deregistered_at) and
+              a.sid in ^cluster_sids,
+          select: %{host_id: h.id, sap_system_id: a.sap_system_id}
+      )
+
+    sap_system_ids =
+      hosts_data
+      |> Enum.map(fn %{sap_system_id: sap_system_id} -> sap_system_id end)
+      |> Enum.uniq()
```
arbulu89, line 185: You could pipe `get_ensa_version` here and get the ensa_version directly

lib/trento/clusters.ex
```diff
@@ -225,4 +230,23 @@ defmodule Trento.Clusters do
   end
 
   defp maybe_request_checks_execution(error), do: error
+
+  @spec get_ensa_version([String.t()]) :: EnsaVersion.t()
+  defp get_ensa_version(sap_system_ids) do
+    query =
+      from(s in SapSystemReadModel,
+        select: s.ensa_version,
+        where: s.id in ^sap_system_ids
```
arbulu89, line 239: For sanity I think we should exclude deregistered SAP systems, just in case.

test/trento/clusters_test.exs
```diff
@@ -300,62 +347,62 @@ defmodule Trento.ClustersTest do
       assert :ok = Clusters.request_checks_execution(cluster_id)
     end
 
-    test "should start a checks execution on demand for ascs_ers clusters with a SAP system ENSA 1 version" do
-      sap_systems = build_list(1, :ascs_ers_cluster_sap_system, filesystem_resource_based: false)
-      [%Trento.Clusters.ValueObjects.AscsErsClusterSapSystem{sid: sid}] = sap_systems
+    test "should start a checks execution on demand for ascs_ers clusters with ENSA 1 version" do
+      %SapSystemReadModel{id: sap_system_id_1, sid: sid_1} =
+        insert(:sap_system, ensa_version: Trento.SapSystems.Enums.EnsaVersion.ensa1())
```
arbulu89, line 352: You should use the aliased short version here `EnsaVersion.ensa1()` (and in many more places)

arbulu89 (approved):
Now the logic looks good :slightly_smiling_face:

lib/trento/clusters.ex
jamie-suse: As far as I know we don't really have terminology for this. The best I can think of is `get_aggregated_ensa_version()`, what do you think?

test/trento/clusters_test.exs
jamie-suse: This is because these tests use both `Trento.SapSystems.Enums.EnsaVersion` and `Trento.Clusters.Enums.EnsaVersion`. I'm fine with renaming the cluster's ENSA version something else like `ClusterEnsaVersion` or `AggregatedEnsaVersion`, what do you think?

nelsonkopliku (approved):

arbulu89 (approved):

PR #2234, opened 2024-01-30
Add software updates settings schema

nelsonkopliku:
# Description

Initial PR supporting SUMA settings API.
Simply adding a schema where required information are going to be stored.

Consider that:
- software_update_settings is a one-row table like `settings`. Constraint added to the table
- `password` and `ca_cert` are store encrypted by leveraging https://github.com/danielberkompas/cloak_ecto

Implementation of module to get/apply suma settings will follow.

## How was this tested?

Manually by running migration and manual insertions. More thorough testing will follow when implementing the context (service)


jamie-suse (approved):

priv/repo/migrations/20240130130617_create_software_update_settings.exs
```diff
@@ -0,0 +1,26 @@
+defmodule Trento.Repo.Migrations.CreateSoftwareUpdateSettings do
+  use Ecto.Migration
+
+  def change do
+    settings_name = "software_update_settings"
+
+    create table(:software_update_settings, primary_key: false) do
+      add :name, :string, primary_key: true, default: settings_name
```
dottorblaster, line 8: Maybe this could be a UUID as we do for the settings table? Just a nitpick, I don't know how I feel about having a primary key that is a plain string :grimacing: 

Also, I don't think we need the `default` here since we have the constraint to keep this to one record only and the initial insert already populates the record with the value we care about

dottorblaster (approved):
LGTM with a minor nitpick

priv/repo/migrations/20240130130617_create_software_update_settings.exs
nelsonkopliku: ok, I will make it a uuid if that makes feel you better :smile: 

However, let's keep the default, it just comes handy when operating with that record. Is that ok?

PR #2247, opened 2024-02-01
Add SUSE Manager config component

dottorblaster:
# Description
New SUSE Manager configuration UI component

## How was this tested?
Jest tests, storybook

### DIFF
--- a/assets/js/common/SuseManagerConfig/CertifiedUploadDate.jsx
+++ b/assets/js/common/SuseManagerConfig/CertifiedUploadDate.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { format } from 'date-fns';
+import { EOS_LOCK_OUTLINED } from 'eos-icons-react';
+
+function CertificateUploadDate({ date }) {
+  if (!date) {
+    return '-';
+  }
+
+  return (
+    <div className="flex flex-row items-center">
+      <EOS_LOCK_OUTLINED className="mr-3" size="25" />
+
+      <div>
+        <div>Certificate Uploaded</div>
+        <div className="text-xs">{format(date, 'iii MMM dd, y')}</div>
+      </div>
+    </div>
+  );
+}
+
+export default CertificateUploadDate;
--- a/assets/js/common/SuseManagerConfig/SuseManagerConfig.jsx
+++ b/assets/js/common/SuseManagerConfig/SuseManagerConfig.jsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import { defaultTo, noop } from 'lodash';
+
+import Button from '@common/Button';
+
+import CertificateUploadDate from './CertifiedUploadDate';
+
+const renderPassword = (username, certUploadDate) =>
+  username && certUploadDate ? '•••••' : '.....';
+
+function SuseManagerConfig({
+  url = 'https://',
+  username,
+  certUploadDate,
+  onEditClick = noop,
+}) {
+  return (
+    <div className="container max-w-7xl mx-auto p-4 sm:p-6 lg:p-8 bg-white dark:bg-gray-800 rounded-lg">
+      <div>
+        <h2 className="text-2xl font-bold inline-block">SUSE Manager Config</h2>
+        <span className="w-32 inline-block float-right">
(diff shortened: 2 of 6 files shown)


assets/js/common/SuseManagerConfig/SuseManagerConfig.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { defaultTo, noop } from 'lodash';
+
+import Button from '@common/Button';
+
+import CertificateUploadDate from './CertifiedUploadDate';
+
+const renderPassword = (username, certUploadDate) =>
+  username && certUploadDate ? '•••••' : '.....';
```
jamie-suse, line 9: Looks great :rocket: I'm just curious why here we use `'.....'` for the empty state, instead of `'-'` like in `CertificateUploadDate`?

dottorblaster: I just copied the mockup without asking :grimacing: @jagabomb what do you think?

assets/js/common/SuseManagerConfig/SuseManagerConfig.jsx
```diff
@@ -0,0 +1,52 @@
...
+      <div>
+        <h2 className="text-2xl font-bold inline-block">SUSE Manager Config</h2>
+        <span className="w-32 inline-block float-right">
+          <Button type="primary-white" size="small" onClick={onEditClick}>
+            Edit Settings
+          </Button>
+        </span>
+      </div>
+      <p className="mt-3 mb-3 text-gray-500">
+        SUSE Manager lets you discover software vulnerabilities on nodes
```
nelsonkopliku, line 28: Slightly different from the design where we have `SUSE Manager enables software updates for hosts.`
No preferences, just aiming to get alignment with @jagabomb

nelsonkopliku (changes requested):
Looking great!

Only change needed is the filename `CertifiedUploadDate.jsx` -> `CertificateUploadDate.jsx`.

Besides that I just noticed a couple of misalignment in copy.
Jurgen might help there.

assets/js/common/SuseManagerConfig/SuseManagerConfig.jsx
jagabomb: I like it the way it is for now. In the design, I just put the ellipsis(...) for username as a way to show there needs to be a value set by the user. Unlike the certificate field which can stay empty and hence the use of the dash (–). I also considered writing Not Applicable/
Available (N/A) for the certificate field but I thought it might imply too much or lead to confusion.

assets/js/common/SuseManagerConfig/CertifiedUploadDate.jsx
```diff
@@ -0,0 +1,22 @@
...
+    return '-';
+  }
+
+  return (
+    <div className="flex flex-row items-center">
+      <EOS_LOCK_OUTLINED className="mr-3" size="25" />
+
+      <div>
+        <div>Certificate Uploaded</div>
+        <div className="text-xs">{format(date, 'iii MMM dd, y')}</div>
```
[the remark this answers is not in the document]
jagabomb: I just like a shorter date as per example but it might be too early to make a call on what is right or wrong with date structure. I think as long as there is consistency with how we display dates across Trento then we should be fine.

assets/js/common/SuseManagerConfig/SuseManagerConfig.jsx
jagabomb: I think this also works for me. The best person check this product language with would be @abravosuse.

assets/js/common/SuseManagerConfig/SuseManagerConfig.jsx
dottorblaster: Thanks both for the suggestions!

dottorblaster:
Just applied the changes, thanks everybody for the suggestions, I adopted the wording suggested in the review we had, @abravosuse if you feel like we could use something better feel free to make a suggestion!

nelsonkopliku (approved):
LGTM

jagabomb (approved):
LGTM!

PR #2248, opened 2024-02-01
Add ability to read software update settings

nelsonkopliku:
# Description

Adding a `Trento.SoftwareUpdates` context allowing, for now, to load software updates settings.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/software_updates.ex
+++ b/lib/trento/software_updates.ex
@@ -1,0 +1,45 @@
+defmodule Trento.SoftwareUpdates do
+  @moduledoc """
+  Entry point for the software updates feature.
+  """
+
+  alias Trento.Repo
+  alias Trento.SoftwareUpdates.Settings
+
+  @type software_update_settings :: %{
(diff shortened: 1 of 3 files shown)


lib/trento/software_updates.ex
```diff
@@ -0,0 +1,45 @@
...
+          ca_cert: ca_cert,
+          ca_uploaded_at: ca_uploaded_at
+        }
+      }
+    else
+      {:error, :settings_not_configured}
+    end
+  end
+
+  defp has_valid_settings?(%Settings{url: url, username: username, password: password}),
```
jamie-suse, line 43: Does this protect against empty strings?

jamie-suse (approved):

lib/trento/software_updates.ex
jamie-suse: I see now; it's to detect if any settings have been set before trying to retrieve them :+1:

PR #2250, opened 2024-02-01
Add endpoint to retrieve software update settings

jamie-suse:
# Description

This change adds a HTTP endpoint `GET /api/v1/user_settings` to allow reading of user credentials to the SAP software update system (i.e. at time of writing, SUSE Manager).

## How was this tested?

Added unit tests for new functions.

### DIFF
--- a/lib/trento_web/router.ex
+++ b/lib/trento_web/router.ex
@@ -146,7 +146,7 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
-      get "/user_settings", SoftwareUpdatesController, :get_user_settings
+      get "/software_updates/settings", SoftwareUpdatesController, :get_user_settings
 
       scope "/charts" do
         pipe_through :charts_feature
--- a/lib/trento_web/router.ex
+++ b/lib/trento_web/router.ex
@@ -146,7 +146,7 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
-      get "/software_updates/settings", SoftwareUpdatesController, :get_user_settings
+      get "/software_updates/settings", SoftwareUpdatesController, :get_software_updates_settings
 
       scope "/charts" do
         pipe_through :charts_feature
--- a/lib/trento_web/router.ex
+++ b/lib/trento_web/router.ex
@@ -146,6 +146,8 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
+      get "/user_settings", SoftwareUpdatesController, :get_user_settings
+
       scope "/charts" do
         pipe_through :charts_feature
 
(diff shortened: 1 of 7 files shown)


lib/trento_web/router.ex
```diff
@@ -146,6 +146,8 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
+      get "/user_settings", SoftwareUpdatesController, :get_user_settings
```
nelsonkopliku, line 149: What about `/software-updates/settings`?

test/trento_web/views/v1/software_updates_view_test.exs
```diff
@@ -0,0 +1,28 @@
...
+  import Phoenix.View
+  import Trento.Factory
+
+  alias Trento.SoftwareUpdates.Settings
+  alias TrentoWeb.V1.SoftwareUpdatesView
+
+  describe "renders user_settings.json" do
+    test "should render all the fields" do
+      %Settings{url: url, username: username, ca_uploaded_at: ca_uploaded_at} =
+        insert(
```
nelsonkopliku, line 13: I guess `build` would fit better than `insert`. No strict need to store anything in the database for this test? :eyes:

test/trento_web/views/v1/software_updates_view_test.exs
```diff
@@ -0,0 +1,28 @@
...
+  use TrentoWeb.ConnCase, async: true
+
+  import Phoenix.View
+  import Trento.Factory
+
+  alias Trento.SoftwareUpdates.Settings
+  alias TrentoWeb.V1.SoftwareUpdatesView
+
+  describe "renders user_settings.json" do
+    test "should render all the fields" do
```
nelsonkopliku, line 11:
```suggestion
    test "should render relevant fields" do
```
Not really rendering _all_ the fields

test/trento_web/views/v1/software_updates_view_test.exs
```diff
@@ -0,0 +1,28 @@
...
+      %Settings{url: url, username: username, ca_uploaded_at: ca_uploaded_at} =
```
nelsonkopliku, line 12:
```suggestion
      %Settings{url: url, username: username, ca_uploaded_at: ca_uploaded_at} = settings =
```
with this we can pass later in the `render(SoftwareUpdatesView, "user_settings.json", settings)` and assert that the output only contains what we expect

nelsonkopliku (changes requested):
Awesome! A few comments though.

nelsonkopliku (approved):
LGTM

nelsonkopliku (approved):

test/trento_web/controllers/v1/software_updates_controller_test.exs
```diff
@@ -0,0 +1,43 @@
+defmodule TrentoWeb.V1.SoftwareUpdatesControllerTest do
+  use TrentoWeb.ConnCase, async: true
+
+  import Mox
```
nelsonkopliku, line 4: is mox actually needed here?

jamie-suse: Nope, good catch!

PR #2252, opened 2024-02-01
Add RPM package generation

rtorrero:
# Description

This PR adds RPM package generation on OBS in addition to the existing container image building


rtorrero:
Reopened as PR#2258 to avoid CI issues with forks

PR #2263, opened 2024-02-06
Add clear settings feature to software updates

nelsonkopliku:
# Description

Enabling clear of software updates settings.

## How was this tested?

Automated test

### DIFF
--- a/test/trento/software_updates_test.exs
+++ b/test/trento/software_updates_test.exs
@@ -183,4 +183,22 @@ defmodule Trento.SoftwareUpdates.SettingsTest do
 
     assert {:error, :settings_already_configured} = SoftwareUpdates.save_settings(settings)
   end
+
+  describe "clearing software update settings" do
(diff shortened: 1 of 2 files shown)


test/trento/software_updates_test.exs
```diff
@@ -183,4 +183,37 @@ defmodule Trento.SoftwareUpdates.SettingsTest do
 
     assert {:error, :settings_already_configured} = SoftwareUpdates.save_settings(settings)
   end
+
+  describe "clearing software update settings" do
+    test "should clear settings" do
```
jamie-suse, line 188: Would you say that this test is a subset of the other test `should support idempotent sequential clear settings`?

nelsonkopliku: Yes, we could squash in the second test only. Deal?

jamie-suse: Sounds like a deal :handshake:

PR #2264, opened 2024-02-06
Rename route: `/software_updates/settings` -> `/settings/suma_credentials`

jamie-suse:
# Description

Renames the route at `/api/v1/software_updates/settings` to `/api/v1/settings/suma_credentials`.

Also updates the routing code to use the [`resources`](https://hexdocs.pm/phoenix/Phoenix.Router.html#resources/4) macro to be more in-line with typical Phoenix routing code as seen in the [Phoenix Routing Guide](https://hexdocs.pm/phoenix/routing.html#resources).

## How was this tested?

Updated existing unit tests

### DIFF
--- a/lib/trento_web/router.ex
+++ b/lib/trento_web/router.ex
@@ -146,7 +146,9 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
-      resources "/settings/suma_credentials", SoftwareUpdatesController, only: [:index]
+      resources "/settings/suma_credentials", SoftwareUpdatesController,
+        only: [:show],
+        singleton: true
 
       scope "/charts" do
         pipe_through :charts_feature
--- a/lib/trento_web/router.ex
+++ b/lib/trento_web/router.ex
@@ -146,7 +146,7 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
-      resources "/settings/suma_credentials", SoftwareUpdatesController,
+      resources "/settings/suma_credentials", SUMACredentialsController,
         only: [:show],
         singleton: true
 
--- a/lib/trento_web/router.ex
+++ b/lib/trento_web/router.ex
@@ -146,7 +146,7 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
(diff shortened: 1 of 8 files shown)


nelsonkopliku:
We need to temporarily skip the api bc check in the pipeline

nelsonkopliku (approved):
LGTM

.github/workflows/ci.yaml
```diff
@@ -529,15 +529,7 @@ jobs:
...
-        run: |
-          docker run -v "$(pwd)/specs:/specs" --rm openapitools/openapi-diff:2.0.1 \
-            /specs/main-spec.json \
-            /specs/current-spec.json \
-            --fail-on-incompatible \
-            --markdown /specs/changes.md \
-            --json /specs/changes.json \
-            --text /specs/changes.txt \
-            --html /specs/changes.html
+        run: echo "TEMPORARY SKIP"
```
dottorblaster, line 532: Either we remember to just edit this or I can merge with admin privileges if we all agree on that. I'd rather forcemerge instead of doing 2 PRs but up to you.

dottorblaster (approved):
LGTM

.github/workflows/ci.yaml
jamie-suse: That'd be awesome if so. I'll revert this change then.

lib/trento_web/router.ex
```diff
@@ -146,7 +146,7 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
-      get "/software_updates/settings", SoftwareUpdatesController, :get_software_updates_settings
+      resources "/settings/suma_credentials", SoftwareUpdatesController, only: [:index]
```
nelsonkopliku, line 149: I believe we might want to look at `singleton-resources` https://hexdocs.pm/phoenix/Phoenix.Router.html#resources/4-singleton-resources

Seems to better fit our usecase.

```suggestion
      resources "/settings/suma_credentials", SoftwareUpdatesController, only: [:show], singleton: true
```

nelsonkopliku: Let's also rename the controller and file to something like `SoftwareUpdatesCredentialsController` or even `SUMACredentialsController`.

It'd be clearer which resource the actions `show`, `update`, `delete` would be referring to

nelsonkopliku (changes requested):
A couple of extra tiny adjustments

nelsonkopliku (approved):
Thanks! LGTM (again :smile:)

CDimonaco (approved):

PR #2273, opened 2024-02-07
Add change settings feature to software updates

nelsonkopliku:
# Description

Adds the ability to change previously saved settings.

Considerations:
- a change can happen only if settings were previously saved
- partial update of settings is supported
- same validation as when first saving is triggered when attempting to apply a change

Little polishing:
- removed unncecessary on conflict clauses in tests factory `insert`
- grouped tests in `describe` sections
- tiny refactor of `save_settings` to leverage `get_settings`+`with` construct

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/software_updates.ex
+++ b/lib/trento/software_updates.ex
@@ -9,10 +9,17 @@ defmodule Trento.SoftwareUpdates do
   alias Trento.Repo
   alias Trento.SoftwareUpdates.Settings
 
-  @type software_update_settings_submission :: %{
+  @type software_update_settings_save_submission :: %{
           url: String.t(),
           username: String.t(),
           password: String.t(),
+          ca_cert: String.t() | nil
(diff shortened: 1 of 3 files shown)


dottorblaster (changes requested):
Awesome!

Can you refactor the usage of `with`? According to our coding guidelines complex else clauses should be avoided because you can't tell exactly where the failure is coming from and are difficult to read.

You can find this advice both [here](https://github.com/lucasvegi/Elixir-Code-Smells?tab=readme-ov-file#complex-else-clauses-in-with) and officially stated in the [code-related anti-patterns](https://hexdocs.pm/elixir/main/code-anti-patterns.html#complex-else-clauses-in-with) manual page. As you can see refactoring is pretty straightforward.

dottorblaster (approved):

PR #2277, opened 2024-02-08
Hana cluster find sites

arbulu89:
# Description

Backend part related to finding the site name, state and sr health state.
I have done some changes on how we put them, to have a new `sites` field in the details.
I have done this transition, because in reality, these fields are site based, and not host based.
This way, we have the same structure for scale up and scale out (in fact, considering that they are working in a new resource agent that follows this argument, it looks a good idea). And in fact, we can put the sr health state in both sites (until now we only have the secondary site health state).

So the new `sites` entry goes only in `v2` version, and I have set the other values as deprecated, even though the values is there, it is valid, and the user can use, so we don't have non backward compatibles. But marking as deprecated just shows that they will be removed in a potential `v3` version.

image

## How was this tested?

Tests added

### DIFF
--- a/lib/trento/clusters/value_objects/hana_cluster_node.ex
+++ b/lib/trento/clusters/value_objects/hana_cluster_node.ex
@@ -5,7 +5,6 @@ defmodule Trento.Clusters.ValueObjects.HanaClusterNode do
 
   @required_fields [
     :name,
-    :site,
     :hana_status,
     :attributes
   ]
@@ -17,6 +16,7 @@ defmodule Trento.Clusters.ValueObjects.HanaClusterNode do
   deftype do
     field :name, :string
     field :site, :string
+    # this attribute is deprecated, moved to the sites entry
     field :hana_status, :string
     field :attributes, {:map, :string}
     field :virtual_ip, :string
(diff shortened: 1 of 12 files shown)


rtorrero (approved):
LGTM @arbulu89, thanks!

lib/trento/clusters/value_objects/hana_cluster_node.ex
```diff
@@ -17,6 +16,7 @@ defmodule Trento.Clusters.ValueObjects.HanaClusterNode do
   deftype do
     field :name, :string
     field :site, :string
+    # this attribute is deprecated, moved to the sites entry
```
EMaksy, line 19: Like @rtorrero suggested previously, this attribute is not too concrete. Maybe just naming it would not harm. What do you think ?

EMaksy (approved):
Great Job!!!

CDimonaco (approved):

test/trento_web/views/v1/cluster_view_test.exs
```diff
@@ -6,10 +6,25 @@ defmodule TrentoWeb.V1.ClusterViewTest do
...
+      details =
+        :hana_cluster_details
+        |> build()
+        |> Map.from_struct()
+
+      cluster = build(:cluster, type: :hana_scale_up, details: details)
+
+      %{details: details} = render(ClusterView, "cluster.json", %{cluster: cluster})
+
+      refute Access.get(details, "sites")
```
CDimonaco, line 27: Love the refute used like that 😍

PR #2283, opened 2024-02-09
Add missed CI step name

rtorrero:
# Description

Add missed name in the RPM step

Edit: I'm slightly hijacking the original goal of this PR to add another one-liner: the EnvironmentFile in systemd better matches the example file name with `-` instead of `_`


arbulu89 (approved):
Thank you @rtorrero

EMaksy (approved):
Nice!

PR #2290, opened 2024-02-12
Validate ca_cert when it was provided different than null

nelsonkopliku:
# Description

A tiny improvement to the validation of `ca_cert`.
It gets validated when its provided value is different than `null`, meaning that cases like empty strings `""`, `"    "` trigger validation error both when saving or updating settings.

## How was this tested?

Automated tests


jamie-suse (approved):
LGTM :rocket:

PR #2300, opened 2024-02-13
Add SUSE Manager settings dialog component

dottorblaster:
# Description
Adding SUSE Manager settings dialog component. Examples on Storybook.

## How was this tested?
Storybook stories, Jest tests added

### DIFF
--- a/assets/js/common/SuseManagerSettingsDialog/SuseManagerSettingsModal.jsx
+++ b/assets/js/common/SuseManagerSettingsDialog/SuseManagerSettingsModal.jsx
@@ -1,0 +1,151 @@
+import React, { useState } from 'react';
+import { noop } from 'lodash';
+import { format } from 'date-fns';
+import { EOS_LOCK_OUTLINED } from 'eos-icons-react';
+
+import Button from '@common/Button';
+import Modal from '@common/Modal';
+import Input, { Password, Textarea } from '@common/Input';
+import Label from '@common/Label';
+
+import { hasError } from './errors';
+
+const defaultErrors = [];
+
+function SuseManagerSettingsModal({
+  open = false,
+  loading = false,
+  initialUsername,
+  initialUrl,
+  certUploadDate,
+  errors = defaultErrors,
+  onSave = noop,
+  onCancel = noop,
+  onClearErrors = noop,
+}) {
+  const settingsExist = Boolean(certUploadDate);
+
+  const [url, setUrl] = useState(initialUrl);
+  const [username, setUsername] = useState(initialUsername);
+  const [password, setPassword] = useState('');
+  const [certificate, setCertificate] = useState('');
+  const [editingCertificate, setEditingCertificate] = useState(!settingsExist);
+  const [editingPassword, setEditingPassword] = useState(!settingsExist);
+
+  return (
+    <Modal title="Enter SUSE Manager Settings" open={open} onClose={onCancel}>
+      <div className="grid grid-cols-6 items-center my-5 gap-6">
+        <Label className="col-span-2" required>
+          SUSE Manager URL
+        </Label>
+        <Input
+          className="col-span-4"
+          value={url}
+          placeholder="Enter a URL"
+          error={hasError('url', errors)}
+          onChange={({ target: { value } }) => {
+            setUrl(value);
+            onClearErrors();
+          }}
+        />
+        <Label
+          className="col-span-2 self-start"
+          info="Only required for self-signed certificates"
+        >
+          CA Certificate
+        </Label>
+        {editingCertificate ? (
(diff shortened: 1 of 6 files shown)


dottorblaster:
Dependsa on #2329 but still reviewable

assets/js/common/SuseManagerSettingsDialog/SuseManagerSettingsModal.jsx
```diff
@@ -0,0 +1,151 @@
...
+            placeholder="Starts with -----BEGIN CERTIFICATE-----"
+            error={hasError('ca_cert', errors)}
+            onChange={({ target: { value } }) => {
+              setCertificate(value);
+              onClearErrors();
+            }}
+          />
+        ) : (
+          <div className="col-span-4 flex flex-row items-center justify-start p-5 border border-gray-200 rounded-md">
+            <EOS_LOCK_OUTLINED className="mr-3" size="25" />
```
nelsonkopliku, line 70: I guess we could/should reuse `CertificateUploadDate` component here. Or an adjustment of it to make it reusable here.

assets/js/common/SuseManagerSettingsDialog/SuseManagerSettingsModal.jsx
```diff
@@ -0,0 +1,151 @@
...
+        </p>
+      </div>
+      <div className="flex flex-row w-80 space-x-2">
+        <Button
+          disabled={loading}
+          onClick={() => {
+            const payload = {
+              url,
+              username,
+              ...(editingCertificate && { certificate }),
```
nelsonkopliku, line 135: Let's just keep in mind that the api expects `ca_cert` in the payload: this ends up as `certificate`.

If we intend to have an extra transformation layer before sending to the API then ok.

assets/js/common/SuseManagerSettingsDialog/SuseManagerSettingsModal.stories.jsx
```diff
@@ -0,0 +1,85 @@
...
+    },
+    errors: {
+      description: 'OpenAPI errors coming from backend validation',
+    },
+  },
+};
+
+export const Default = {
+  args: {
+    open: false,
```
nelsonkopliku, line 45: What about a trigger in the stories to open the dialog? Or alternatively `open: true` to all of them, if it does not become too invasive.

nelsonkopliku (changes requested):
Looks good! 
The main change needed imho is adding the ability in stories to actually open the dialog.

The others can be deferred in following PRs, if you wish.

dottorblaster:
@nelsonkopliku Thanks for the review! I was thinking the same about the button, initially I provided `open: true` to the stories but in the documentation page it became really cumbersome with those being modals.

Then I thought we could just use the controls plugin

immagine

What do you think?

nelsonkopliku:
@dottorblaster you're completely right, makes totally sense. It didn't immediately pop to my mind :smile:

dottorblaster:
@nelsonkopliku no problemo :+1: onto the other stuff you spotted :dash:

nelsonkopliku (approved):
Alright, good with storybook controls :rocket: 

For the other things up to you about when to tackle :+1:

jamie-suse (approved):

assets/js/common/SuseManagerSettingsDialog/SuseManagerSettingsModal.jsx
```diff
@@ -0,0 +1,151 @@
...
+import { EOS_LOCK_OUTLINED } from 'eos-icons-react';
+
+import Button from '@common/Button';
+import Modal from '@common/Modal';
+import Input, { Password, Textarea } from '@common/Input';
+import Label from '@common/Label';
+
+import { hasError } from './errors';
+
+const defaultErrors = [];
```
jamie-suse, line 13: Does this have a purpose?

dottorblaster: It does! If you specify empty arrays, empty objects or functions (mainly the first two) as default props without declaring them as `const` you incur into their reference constantly changing when rendering and components just get a new prop ref every time they render, ending up in endless rerendering. Your browser gets easily stuck on those, crashing and looping.

dottorblaster: More info here https://github.com/facebook/react/issues/18123

jamie-suse: ahhh very tricky! Good to know :brain:

PR #2308, opened 2024-02-14
Discover diskless SBD clusters

arbulu89:
# Description

Discover diskless SBD clusters.
A diskless SBD consists of:
- There is not any fencing resource configured
- SBD config file must have `SBD_WATCHDOG_DEV` field
- SBD config file must not have `SBD_DEVICE` field
- SBD must no have any device configured and running

More info: https://documentation.suse.com/sle-ha/15-SP5/html/SLE-HA-all/cha-ha-storage-protect.html#sec-ha-storage-protect-diskless-sbd

## How was this tested?

Tests added

### DIFF
--- a/lib/trento/discovery/policies/cluster_policy.ex
+++ b/lib/trento/discovery/policies/cluster_policy.ex
@@ -22,6 +22,8 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
   @default_hana_scale_out_replication_mode "sync"
   @default_hana_scale_out_operation_mode "logreplay"
   @default_maintenance_mode "false"
+  @unknown_fencing_type "Unknown"
+  @diskless_sbd_fencing_type "Diskless SBD"
 
   def handle(
         %{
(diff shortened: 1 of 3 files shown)


lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -450,16 +452,34 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
...
       %{agent: "stonith:" <> fencing_type} ->
         fencing_type
 
       _ ->
         false
     end)
+    |> do_fencing_type(sbd)
+  end
+
+  defp do_fencing_type(nil, sbd) do
```
CDimonaco, line 467: Can we rename it to `extract_fencing_type`?

lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -450,16 +452,34 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
...
+      @diskless_sbd_fencing_type
+    else
+      @unknown_fencing_type
+    end
   end
 
+  defp do_fencing_type(fencing_type, _), do: fencing_type
+
+  defp diskless_sbd?(%{config: %{"sbd_watchdog_dev" => _} = config, devices: []}) do
+    not Map.has_key?(config, "sbd_device")
```
CDimonaco, line 478: Could we write that condition with pattern matching?

Like config has the key `sbd_device` and the other clause without the key

CDimonaco: We could also has one function that extracts the fencing type and retuurn diskless fencing type or unknown, we can skip the if at line 469

CDimonaco (commented):
Overall LGTM,

I left some comments

lib/trento/discovery/policies/cluster_policy.ex
arbulu89: done. initially i didn't want to put the `unknown` case first, as i wanted to put the positive cases first, but yes, eventually the code is now a bit cleaner

CDimonaco (approved):
Great! LGTM

PR #2310, opened 2024-02-14
Add support to PATCH requests in network client

nelsonkopliku:
# Description

Adding support for PATCH operations in our network client.

## How was this tested?

Automated tests.

### DIFF
--- a/assets/js/lib/network/index.js
+++ b/assets/js/lib/network/index.js
@@ -86,6 +86,15 @@ export const del = function del(url, config = null) {
     });
 };
 
+export const patch = function patch(url, data, config = null) {
+  return networkClient
+    .patch(url, data, config)
+    .then(handleResponseStatus)
+    .catch((error) => {
+      handleError(error);
+    });
+};
+
 export const put = function put(url, data, config = null) {
   return networkClient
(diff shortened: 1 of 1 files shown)


assets/js/lib/network/network.test.js
```diff
@@ -188,4 +188,23 @@ describe('networkClient', () => {
       }
     });
   });
+
+  describe('REST operations', () => {
```
jamie-suse, line 192: Is this test really testing anything?

nelsonkopliku: Nothing SUMA specific. It's pretty naive I know 😂 I can live also without it.

jamie-suse: I guess `axiosMock.onPatch()` is doing the heavy-lifting here and validates that the request is a PATCH request. :muscle:

I'm fine with it either way also, but I guess we should also then test the other verbs in `networkClient`, no?

nelsonkopliku: The test you were looking for is not here :smile:

dottorblaster (approved):
LGTM

PR #2311, opened 2024-02-14
Indexserver & nameserver roles in cluster sites

CDimonaco:
# Description

This PR adds the support to nameserver and indexserver roles in cluster nodes.

image

The roles are extracted from node attributes.
The roles are separated by ":", the role string is splitted on the ":" character.

In scale up

`4:P:master1:master:worker:master`

Nameserver is at index 3, and Indexserver at index 5

In scale out

`master1:master:worker:master`

Nameserver is at index 1 and Indexserver at index 3

## How was this tested?

Automated tests, frontend, backend and e2e

### DIFF
--- a/assets/js/lib/test-utils/factories/clusters.js
+++ b/assets/js/lib/test-utils/factories/clusters.js
@@ -40,6 +40,8 @@ export const hanaClusterDetailsNodesFactory = Factory.define(() => ({
   name: faker.animal.dog(),
   site: faker.location.city(),
   virtual_ip: faker.internet.ip(),
+  indexserver_actual_role: 'master',
+  nameserver_actual_role: 'slave',
   hana_status: hanaStatus(),
   attributes: Array.from({ length: 5 }).reduce(
     (acc, _) => ({
--- a/lib/trento/discovery/policies/cluster_policy.ex
+++ b/lib/trento/discovery/policies/cluster_policy.ex
@@ -233,13 +233,26 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
 
       site = Map.get(attributes, "hana_#{String.downcase(sid)}_site", "")
 
+      hana_roles =
+        attributes
+        |> Map.get("hana_#{String.downcase(sid)}_roles", "")
+        |> String.split(":")
+
+      %{
+        indexserver_actual_role: indexserver_actual_role,
+        nameserver_actual_role: nameserver_actual_role
+      } =
+        parse_nodes_actual_roles(hana_roles, cluster_type)
+
       node = %{
         name: name,
         attributes: attributes,
         resources: node_resources,
         site: site,
         hana_status: "Unknown",
-        virtual_ip: virtual_ip
+        virtual_ip: virtual_ip,
+        nameserver_actual_role: nameserver_actual_role,
+        indexserver_actual_role: indexserver_actual_role
       }
 
       hana_status =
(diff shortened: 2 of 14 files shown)


assets/js/lib/test-utils/factories/clusters.js
```diff
@@ -40,6 +40,8 @@ export const hanaClusterDetailsNodesFactory = Factory.define(() => ({
   name: faker.animal.dog(),
   site: faker.location.city(),
   virtual_ip: faker.internet.ip(),
+  indexserver_actual_role: 'master',
```
arbulu89, line 43: we could maybe have an enum here, but well, it works right now as it is I guess

lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -231,13 +231,26 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
 
       site = Map.get(attributes, "hana_#{String.downcase(sid)}_site", "")
 
+      hana_roles =
+        attributes
```
arbulu89, line 235: you could even pipe this to `parse_nodes_actual_roles`

test/support/factory.ex
```diff
@@ -475,6 +458,27 @@ defmodule Trento.Factory do
     }
   end
 
+  def hana_cluster_node_factory do
+    %HanaClusterNode{
+      attributes: %{"attribute" => Faker.Beer.name()},
+      hana_status: "Secondary",
+      virtual_ip: Faker.Internet.ip_v4_address(),
+      name: Faker.StarWars.character(),
+      indexserver_actual_role: nil,
```
arbulu89, line 467: I would put some value instead of nil, as nil is a corner case. `master` and `master` is the most common scenario

abravosuse (approved):
Code apart, it looks good to me @CDimonaco !. For the sake of authenticity we might want to update the story book for HANA scale-out taking into account:

- in a given site, there can only be one master nameserver, the rest are slaves
- in a given site, there can only be one master indexserver, the rest are slaves or standby
- in the nodes that are not part of a site (other), there is no nameserver nor indexserver.

test/trento_web/controllers/v2/cluster_controller_test.exs
```diff
@@ -22,7 +22,16 @@ defmodule TrentoWeb.V2.ClusterControllerTest do
     end
 
     test "should be compliant with Hana Cluster details schema", %{conn: conn} do
-      insert(:cluster, details: build(:hana_cluster_details))
+      insert(:cluster,
+        details:
+          build(:hana_cluster_details,
+            nodes:
+              build_list(1, :hana_cluster_node, %{
```
arbulu89, line 29: changing the factory would remove the need to apply this change

test/trento_web/views/v1/cluster_view_test.exs
```diff
@@ -15,17 +15,29 @@ defmodule TrentoWeb.V1.ClusterViewTest do
     end
 
     test "should remove HANA cluster V2 fields" do
+      nodes =
```
arbulu89, line 18: Is this to make it explicit?
Because the factory would already create this fields, right?

test/e2e/cypress/fixtures/hana-cluster-details/available_hana_cluster.js
```diff
@@ -168,7 +172,7 @@ export const availableHanaCluster = {
             {
               id: 'rsc_SAPHana_HDP_HDB10',
               type: 'ocf::suse:SAPHana',
-              role: 'Slave',
+              role: 'master',
```
arbulu89, line 175: I think this is a mistaken change. This value doesn't have anything to see with the current implementation.
I don't know if in a real cluster though :see_no_evil:

arbulu89 (approved):
Really nice job @CDimonaco 
I put some comments, but it is great in general. No need for a new review (unless you have some question)

test/trento_web/controllers/v2/cluster_controller_test.exs
CDimonaco: yep done

test/support/factory.ex
CDimonaco: you're right! done

lib/trento/discovery/policies/cluster_policy.ex
CDimonaco: nice catch, done

test/e2e/cypress/fixtures/hana-cluster-details/available_hana_cluster.js
CDimonaco: yeah, too much ctrl+d in vscode, reverted 👀

arbulu89 (approved):
:top:

PR #2317, opened 2024-02-15
Handle concurrent settings saving operations

nelsonkopliku:
# Description

Fix for concurrent calls to save settings operation.

Currently if two, or more, concurrent requests come to save settings, the database says: _ok, I have nothing, you can proceed with insertion_ to both, or all, requests resulting in multiple records being inserted which later breaks logic relying on the fact we do want only one record there.

The new migration adds a uniqueness constraint and sets a default value for the id, so we don't need to really pass it around in the code and takes care of reusing a previously inserted id, if any.

## How was this tested?
Current tests cover desired behavior. Concurrency tested manually.

Reopened as we're trying to figure out what way to go https://github.com/trento-project/web/pull/2327

### DIFF
--- a/lib/trento/software_updates.ex
+++ b/lib/trento/software_updates.ex
@@ -25,12 +25,12 @@ defmodule Trento.SoftwareUpdates do
 
   @spec get_settings :: {:ok, Settings.t()} | {:error, :settings_not_configured}
   def get_settings do
-    case Repo.one(Settings) do
-      nil ->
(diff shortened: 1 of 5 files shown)


dottorblaster:
I don't like this solution since we shouldn't enforce constrants on the dabase layer that really belong to domain. Due to the current implementation this is a compromise that the team is taking in order not to rewrite anything.

LGTM

dottorblaster (approved):

PR #2318, opened 2024-02-15
Initial software updates settings state and saga

nelsonkopliku:
# Description

Adding initial state/selector/saga for software updates settings.
It should be enough to build initial loading of the settings.

```js
  const {
    settings,
    error,
    loading,
  } = useSelector(getSoftwareUpdatesSettings());
```

Note: at the moment no distinction is made on errors.
We might want to iterate over the case we get something different from 404, which has the specific meaning of _no settings are available, yet_

## How was this tested?
Automated tests

### DIFF
--- a/assets/js/state/sagas/index.js
+++ b/assets/js/state/sagas/index.js
@@ -59,7 +59,6 @@ import {
 import { SET_USER_AS_LOGGED } from '@state/user';
 
 import { setEulaVisible, setIsPremium } from '@state/settings';
-import { fetchSoftwareUpdatesSettings } from '@state/softwareUpdatesSettings';
 
 import { watchNotifications } from '@state/sagas/notifications';
 import { watchAcceptEula } from '@state/sagas/eula';
@@ -126,8 +125,6 @@ function* initialDataFetch() {
   const { data: databases } = yield call(get, '/databases');
   yield put(setDatabases(databases));
   yield put(stopDatabasesLoading());
-
-  yield put(fetchSoftwareUpdatesSettings());
 }
 
 function* setupSocketEvents() {
--- a/assets/js/state/sagas/index.js
+++ b/assets/js/state/sagas/index.js
@@ -59,6 +59,7 @@ import {
 import { SET_USER_AS_LOGGED } from '@state/user';
 
 import { setEulaVisible, setIsPremium } from '@state/settings';
+import { fetchSoftwareUpdatesSettings } from '@state/softwareUpdatesSettings';
 
 import { watchNotifications } from '@state/sagas/notifications';
 import { watchAcceptEula } from '@state/sagas/eula';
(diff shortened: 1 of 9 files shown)


assets/js/state/softwareUpdatesSettings.js
```diff
@@ -0,0 +1,50 @@
...
+      state.error = null;
+      state.settings = settings;
+    },
+    setEmptySoftwareUpdatesSettings: (state) => {
+      state.loading = false;
+      state.error = null;
+      state.settings = emptySettings;
+    },
+  },
+});
```
dottorblaster, line 36: I would prefer a more granular structure that allows us to stop loading independently but it's absolutely out of personal preference

assets/js/state/sagas/index.js
```diff
@@ -124,6 +126,8 @@ function* initialDataFetch() {
   const { data: databases } = yield call(get, '/databases');
   yield put(setDatabases(databases));
   yield put(stopDatabasesLoading());
+
+  yield put(fetchSoftwareUpdatesSettings());
```
dottorblaster, line 130: Do we need this when doing initial fetching? Maybe we could just dispatch the refresh when loading the about page, what do you think?

dottorblaster (approved):
LGTM with some minor comments, but we can merge :+1:

assets/js/state/softwareUpdatesSettings.js
nelsonkopliku: Only thing that comes to my mind is that in a saga we might end up in unnecessary repetition imho, for instance we might be doing something like this when handing errors.
```js
try {
  // call the endpoint
  setSoftwareUpdatesSettings(settings);
  stopLoadingSoftwareUpdatesSettings();
} catch (error) {
  setEmptySoftwareUpdatesSettings()
  stopLoadingSoftwareUpdatesSettings();
}
```

Let's keep an eye on that anyway and see if an independent `stopLoadingSoftwareUpdatesSettings` might be helpful.

dottorblaster: No problem about that, absolutely :+1:

assets/js/state/sagas/index.js
nelsonkopliku: I don't actually mind. I just thought that we might not need to always call the endpoint when landing on the settings.

I can remove that and let the settings page dispatch a `fetchSoftwareUpdatesSettings`, ok?

dottorblaster: Yes! :+1:

assets/js/state/softwareUpdatesSettings.js
nelsonkopliku: partially true what I said, we could actually have something like this
```js
try {
  // call the endpoint
  setSoftwareUpdatesSettings(settings);
} catch (error) {
  setEmptySoftwareUpdatesSettings();
}
stopLoadingSoftwareUpdatesSettings();
```
Anyway, keeping the state mutation internally in the slice, feels more robust. If we need to change, we'll change.

PR #2319, opened 2024-02-15
Add tooltip for non-compliant saptune

rtorrero:
# Description

This PR adds a tooltip to add further details when saptune is not compliant

Preview:
Captura desde 2024-02-15 15-37-12

### DIFF
--- a/assets/js/pages/SaptuneDetails/SaptuneTuningState.jsx
+++ b/assets/js/pages/SaptuneDetails/SaptuneTuningState.jsx
@@ -10,7 +10,13 @@ function SaptuneTuningState({ state }) {
     case 'not compliant':
       return (
         <div className="flex">
-          <Tooltip content="Run `saptune note verify` in the host for further details">
+          <Tooltip
+            content={
+              <div>
+                Run `saptune note verify` in <br /> the host for further details
+              </div>
+            }
+          >
             <HealthIcon health="critical" />
           </Tooltip>
           <span className="ml-1">Not compliant</span>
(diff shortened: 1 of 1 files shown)


assets/js/pages/SaptuneDetails/SaptuneTuningState.jsx
```diff
@@ -9,7 +10,12 @@ function SaptuneTuningState({ state }) {
     case 'not compliant':
       return (
         <div className="flex">
-          <HealthIcon health="critical" />
+          <Tooltip
+            content="Run `saptune note verify` in the host for further details"
```
arbulu89, line 14: I don't think markdown works here as it is. You would need to wrap it up, something like:
```
<ReactMarkdown className="markdown" remarkPlugins={[remarkGfm]}>
Run `saptune note verify` in the host for further details
 </ReactMarkdown>
```

arbulu89 (commented):
Just a comment.
If @jagabomb agrees that the tooltip needs to go in the bottom is ok from my side

jagabomb:
> Just a comment. If @jagabomb agrees that the tooltip needs to go in the bottom is ok from my side

Something like this would be good...
Screenshot 2024-02-15 at 17 16 35

arbulu89 (approved):
If @jagabomb is happy, I'm happy.
Jurgen, you could check it by yourself in saptune details view storybook, but you would need to change the tuning value to `not compliant` in the options

rtorrero:
As discussed with @arbulu89 we ended up removing the markdown as it adds very little value and looks quite odd

EMaksy (approved):
LGTM!

CDimonaco (approved):

jagabomb:
@rtorrero is it possible to wrap the text over two lines like screenshot below instead of one long line?
Screenshot 2024-02-15 at 18 14 02

rtorrero:
@jagabomb should be good now: 
Captura desde 2024-02-16 08-54-12

jagabomb (approved):
LGTM!

PR #2320, opened 2024-02-15
Fix HostDetails story

arbulu89:
# Description

HostDetails storybook is broken. 
The fix is basically to pass `chartsEnabled` param from above to make the components pure.

All the stories are with disabled charts by now, as mocking them is not that simple. If needed, we can do it in the future, but I don't find a big value on that.

It supersedes: 
https://github.com/trento-project/web/pull/2197


EMaksy (approved):
Great!

CDimonaco (approved):

PR #2327, opened 2024-02-16
Fix race condition getting software updates settings

dottorblaster:
# Description
Fixes a race condition when getting software updates settings from the database. Putting the read (and the write) inside a transation solves the issue since Postgres is read-committed. By doing this we can avoid enforcing business logic outside of the domain layer.

I also took the chance to replicate the problem using an automatic test.

## How was this tested?
Automatic tests added

### DIFF
--- a/lib/trento/software_updates.ex
+++ b/lib/trento/software_updates.ex
@@ -41,6 +41,8 @@ defmodule Trento.SoftwareUpdates do
   def save_settings(settings_submission, date_service \\ DateService) do
     {:ok, settings} =
       Repo.transaction(fn ->
+        Repo.query!("set transaction isolation level serializable;")
+
         with :ok <- ensure_no_settings_configured() do
           save_new_settings(settings_submission, date_service)
         end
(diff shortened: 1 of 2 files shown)


nelsonkopliku (dismissed):
I guess I like this more than this https://github.com/trento-project/web/pull/2317 :smile: 

Thanks!

nelsonkopliku:
> I guess I like this more than this #2317 😄
> 
> Thanks!

@dottorblaster wait, I am still testing and getting unexpected results.

nelsonkopliku:
As shared privately with this implementation some requests may go 500 and still multiple rows are inserted.

We might need an extra query to set the transaction isolation level `Repo.query!("set transaction isolation level serializable;")` https://www.postgresql.org/docs/current/transaction-iso.html
This should protect from double insertions, but in some cases we still get some 500 from some requests.

I don't know... this PR https://github.com/trento-project/web/pull/2317 does not have such issues

nelsonkopliku:
sad that tests are complaining :cry:

dottorblaster:
This is really going too far, I don't like it either because we'd have to modify how the Ecto sandbox behaves and introducing a new tag for this kind of tests. Really not my cup of tea.

PR #2328, opened 2024-02-16
Cluster details updates

rtorrero:
# Description

This PR adds some of the UI changes proposed by @jagabomb:

 - details button becomes a link
 - details content header changes from "Site details" to "Node details"
 - rounded bottom of tables 
 - top margins in the resources in the node details view
 - stop resources section should go under the site details section
 - stop resources section becomes a fixed section and when there are no stopped resources it shows a text saying "No resources to display"

See attached screenshots for more details:
Captura desde 2024-02-16 13-49-53
Captura desde 2024-02-16 13-49-32
Captura desde 2024-02-16 13-49-46

### DIFF
--- a/assets/js/pages/ClusterDetails/StoppedResources.test.jsx
+++ b/assets/js/pages/ClusterDetails/StoppedResources.test.jsx
@@ -1,0 +1,13 @@
+import React from 'react';
+import { screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import { renderWithRouter } from '@lib/test-utils';
+
+import StoppedResources from './StoppedResources';
+
+it('should display a "No resources to display." message when there are no resources', async () => {
+  renderWithRouter(<StoppedResources resources={[]} />);
+
+  expect(screen.getByText('No resources to display.')).toBeInTheDocument();
+});
--- a/assets/js/pages/ClusterDetails/StoppedResources.test.jsx
+++ b/assets/js/pages/ClusterDetails/StoppedResources.test.jsx
@@ -1,13 +1,11 @@
 import React from 'react';
-import { screen } from '@testing-library/react';
+import { screen, render } from '@testing-library/react';
 import '@testing-library/jest-dom';
-
-import { renderWithRouter } from '@lib/test-utils';
 
 import StoppedResources from './StoppedResources';
 
 it('should display a "No resources to display." message when there are no resources', async () => {
-  renderWithRouter(<StoppedResources resources={[]} />);
+  render(<StoppedResources resources={[]} />);
 
   expect(screen.getByText('No resources to display.')).toBeInTheDocument();
 });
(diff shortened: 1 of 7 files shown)


arbulu89 (commented):
Hey @rtorrero 
Only one thing. With the rounded table thing, the header row shouldn't be rounded when there is an additional header on top of it.
Besides that, all of the table in the app are rounder now, and I don't think this was desired. We should have some rounder option as prop in the Table component I guess to make it variable.
What do you think @jagabomb ?

jagabomb:
> Hey @rtorrero Only one thing. With the rounded table thing, the header row shouldn't be rounded when there is an additional header on top of it. Besides that, all of the table in the app are rounder now, and I don't think this was desired. We should have some rounder option as prop in the Table component I guess to make it variable. What do you think @jagabomb ?

Well spotted @arbulu89, yeah only the top and bottom sections of the **Site Detail** component should be rounded. So in this case the, **Site Name** (NBG/WDF) would be the top part of the component with the top rounded corners. Then the grey **Table Header** (Hostname, Nameserver, etc) without any rounded corners. Followed by the **Table Description** with bottom rounded corners. This styling helps to make the **Sites Detail** nodes feel like single components.

rtorrero:
> > Hey @rtorrero Only one thing. With the rounded table thing, the header row shouldn't be rounded when there is an additional header on top of it. Besides that, all of the table in the app are rounder now, and I don't think this was desired. We should have some rounder option as prop in the Table component I guess to make it variable. What do you think @jagabomb ?
> 
> Well spotted @arbulu89, yeah only the top and bottom sections of the **Site Detail** component should be rounded. So in this case the, **Site Name** (NBG/WDF) would be the top part of the component with the top rounded corners. Then the grey **Table Header** (Hostname, Nameserver, etc) without any rounded corners. Followed by the **Table Description** with bottom rounded corners. This styling helps to make the **Sites Detail** nodes feel like single components.

once https://github.com/trento-project/web/pull/2337 is merged, I'll adjust this and re-request review :+1:

EMaksy (approved):
Nice, you survived the CSS war. 
LGTM!

rtorrero:
@jagabomb @arbulu89 seems to be good now: with header
image

without header:
image

arbulu89 (changes requested):
@rtorrero 
Looks good.
Please add a test to the `StoppedResources` view, for the 0 resources scenario.

assets/js/pages/ClusterDetails/StoppedResources.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import { renderWithRouter } from '@lib/test-utils';
```
arbulu89, line 5: You don't need `renderWithRouter` use `render` coming from `@testing-library/react`
```
import { screen, render } from '@testing-library/react';
```

arbulu89 (approved):
Please change the import.
We are fine to merge afterwards

PR #2336, opened 2024-02-19
Clean tgzs after `service manualrun`

rtorrero:
# Description

This PR fixes the issue discussed in today's review meeting where some of the node_modules tgz's where left in the OBS project. This is due to the fact that we clean them up too early and `osc service manualrun` will recreate the node_modules from the obscpio that where updated since the last build.

By cleaning up after `osc service manualrun` we make sure no tgz are left in the working copy and send to OBS.


arbulu89 (approved):
Should be maybe test by hand first? :see_no_evil:

rtorrero:
> Should be maybe test by hand first? 🙈

Tested in my fork, works fine :+1:

PR #2337, opened 2024-02-19
Add header prop to table

arbulu89:
# Description

Add `header` prop to `Table`. This allows us to include a header to the table using the correct bordering layout.
It will be used in the HanaClusterDetails view, to border properly the site details.

Besides that, the table borders are rounded now on.

## How was this tested?

The styles tested with Storybook

### DIFF
--- a/assets/js/common/Table/Table.jsx
+++ b/assets/js/common/Table/Table.jsx
@@ -62,6 +62,7 @@ function Table({
   setSearchParams,
   emptyStateText = 'No data available',
   withPadding = true,
+  header = null,
   rowKey = defaultRowKey,
 }) {
   const {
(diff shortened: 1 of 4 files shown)


rtorrero (approved):
Thanks @arbulu89 

For some context: this came up during the review for https://github.com/trento-project/web/pull/2328 where we noticed the rounded corners where being applied on the header always, even when the table is contained in another diff, and this looks odd. Check the PR for a screenshot.

EMaksy (approved):
LGTM :+1:

assets/js/common/Table/Table.jsx
```diff
@@ -82,6 +83,8 @@ function Table({
     (c) => c.filter && c.filterFromParams
   );
 
+  const hasFilters = columns.filter(({ filter }) => Boolean(filter)).length > 0;
```
dottorblaster, line 86: Nice oneliner bro

dottorblaster (approved):
LGTM, beware of that onliner! But really up to you if it's something to be refactored :smile:

assets/js/common/Table/Table.jsx
arbulu89: it is the freaking npm formatter, it doesn't like anything that it isn't this single line :see_no_evil:

PR #2342, opened 2024-02-20
Bump httpoison version

nelsonkopliku:
# Description

Bumping the version of the http client we use to the latest available https://github.com/edgurgel/httpoison.

Since version 2 some changes on ssl settings have been made. Since we're going to play with those, better have the up to date lib.

## How was this tested?

No tests.


dottorblaster (approved):
LGTM

PR #2354, opened 2024-02-21
Fix software updates patching idempotency

nelsonkopliku:
# Description

As per the title: fixing idempotency when requesting software updates patching.

Currently if we send the same change requests for suma creds twice sequentially with the same `ca_cert`, the `ca_uploaded_at` gets nullified.

## How was this tested?

Added a test.

### DIFF
--- a/test/trento/software_updates_test.exs
+++ b/test/trento/software_updates_test.exs
@@ -343,6 +343,55 @@ defmodule Trento.SoftwareUpdates.SettingsTest do
                SoftwareUpdates.change_settings(change_submission, Trento.Support.DateService.Mock)
     end
 
+    test "should support idempotent sequential settings updates" do
+      %{
+        url: _initial_url,
+        username: initial_username,
+        password: initial_password,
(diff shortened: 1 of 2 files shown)


test/trento/software_updates_test.exs
```diff
@@ -343,6 +343,50 @@ defmodule Trento.SoftwareUpdates.SettingsTest do
...
+      }
+
+      Enum.each(1..3, fn _ ->
+        assert {:ok,
+                %{
+                  url: ^new_url,
+                  username: ^initial_username,
+                  password: ^initial_password,
+                  ca_cert: ^new_ca_cert,
+                  ca_uploaded_at: ^now
```
jamie-suse, line 381: Wouldn't this check always pass as the `DateService` mock is always going to return `now`, for every run of `change_settings()`?

jamie-suse (approved):
LGTM :rocket:

dottorblaster (approved):
LGTM

PR #2355, opened 2024-02-21
Openapi schemas no additional properties

CDimonaco:
# Description

This PR aims to fix the properties of OpenApi Schemas.
This change involves the schemas, controllers and views.
Now our openapi schema is consistent with web responses.

Controller response views and broadcast views are differentiated, in order to achieve consistency on both the transports.

I introduced a new convention the `broadcast_` prefix in views means that the views are used **ONLY** in the broadcast of messages.

Automated tests

### DIFF
--- a/lib/trento_web/openapi/v1/schema/host.ex
+++ b/lib/trento_web/openapi/v1/schema/host.ex
@@ -33,6 +33,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Host do
       title: "Host",
       description: "A discovered host on the target infrastructure",
       type: :object,
+      additionalProperties: false,
       properties: %{
         id: %Schema{type: :string, description: "Host ID", format: :uuid},
         hostname: %Schema{type: :string, description: "Host name"},
--- a/lib/trento_web/openapi/v1/schema/host.ex
+++ b/lib/trento_web/openapi/v1/schema/host.ex
@@ -6,10 +6,10 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Host do
 
   alias TrentoWeb.OpenApi.V1.Schema.{
     Provider,
+    ResourceHealth,
     SaptuneStatus,
     SlesSubscription,
-    Tags,
-    ResourceHealth
+    Tags
   }
 
   defmodule IPv4 do
--- a/lib/trento_web/openapi/v1/schema/host.ex
+++ b/lib/trento_web/openapi/v1/schema/host.ex
@@ -78,6 +78,12 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Host do
           type: :array,
           items: SlesSubscription
         },
+        selected_checks: %Schema{
+          title: "SelectedChecks",
+          description: "A list of check ids selected for an execution on this host",
+          type: :array,
+          items: %Schema{type: :string}
+        },
         saptune_status: SaptuneStatus,
         deregistered_at: %Schema{
           title: "DeregisteredAt",
--- a/lib/trento_web/openapi/v1/schema/host.ex
+++ b/lib/trento_web/openapi/v1/schema/host.ex
@@ -4,7 +4,13 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Host do
   require OpenApiSpex
   alias OpenApiSpex.Schema
 
-  alias TrentoWeb.OpenApi.V1.Schema.{Provider, SaptuneStatus, SlesSubscription, Tags}
+  alias TrentoWeb.OpenApi.V1.Schema.{
+    Provider,
+    SaptuneStatus,
+    SlesSubscription,
+    Tags,
+    ResourceHealth
+  }
 
   defmodule IPv4 do
     @moduledoc false
@@ -52,6 +58,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Host do
           type: :string,
           description: "Version of the agent installed on the host"
         },
+        health: ResourceHealth,
         cluster_id: %Schema{
           type: :string,
           description: "Identifier of the cluster this host is part of",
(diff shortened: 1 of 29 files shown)


dottorblaster (approved):
I really like the approach of creating dedicated view clauses for broadcasts :+1:

lib/trento_web/openapi/v1/schema/host.ex
```diff
@@ -51,6 +58,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Host do
           type: :string,
           description: "Version of the agent installed on the host"
         },
+        health: ResourceHealth,
```
arbulu89, line 61: Nitpick.
Not about this field.
I have checked a bit the code, and the value we could deprecate is the `deregistered_at`.
It is not used in the frontend anywhere.
So we could remove it in a potential, v2 version.
Anyway, it is not strictly about this PR. Up to you. If you don't do it, it is not a big deal.

lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -17,11 +17,38 @@ defmodule TrentoWeb.V1.SapSystemView do
...
+    |> Map.delete(:deregistered_at)
     |> Map.put(:database_instances, rendered_database_instances)
     |> add_system_replication_status_to_secondary_instance
   end
 
   def render("database_instance.json", %{instance: instance}) do
+    instance
+    |> Map.from_struct()
+    |> Map.delete(:__meta__)
+    |> Map.delete(:absent_at)
```
arbulu89, line 29: We shouldn't remove this field. It is used in the frontend.
The error must be that it is not defined in the schema.

lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -17,11 +17,38 @@ defmodule TrentoWeb.V1.SapSystemView do
...
+    |> Map.delete(:host)
+  end
+
+  def render("broadcast_database.json", %{
+        database:
```
arbulu89, line 34: I think we could render inside this the `database.json` view.
After all, they have the exact same fields, both the database and instance (removing the `deregistered_at` is not a big deal, we don't use it in the frontend).
Anyway, if having it explicit makes sense to you, all good from my side, but they should have the same content

lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -86,6 +113,14 @@ defmodule TrentoWeb.V1.SapSystemView do
   end
 
   def render("application_instance.json", %{instance: instance}) do
+    instance
+    |> Map.from_struct()
+    |> Map.delete(:__meta__)
+    |> Map.delete(:absent_at)
```
arbulu89, line 119: We cannot remove the `absent_at`, it is used in the frontend.
And it would look like exactly as the broadcast

lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -114,6 +149,38 @@ defmodule TrentoWeb.V1.SapSystemView do
...
+      rendered_database_instances
+    )
+    |> Map.put(
+      :application_instances,
+      rendered_application_instances
+    )
+    |> add_system_replication_status_to_secondary_instance
+  end
+
+  def render("broadcast_sap_system.json", %{
```
arbulu89, line 167: You can remove the `deregistered_at` from the broadcast as well

lib/trento_web/views/v2/cluster_view.ex
```diff
@@ -6,11 +6,26 @@ defmodule TrentoWeb.V2.ClusterView do
...
     cluster
     |> Map.from_struct()
+    |> Map.put(:details, cluster_details)
+    |> Map.delete(:deregistered_at)
     |> Map.delete(:__meta__)
   end
 
+  def render("cluster_details.json", %{details: %{"sap_systems" => sap_systems} = details}) do
+    sap_systems =
+      Enum.map(sap_systems, &Map.take(&1, [:distributed, :filesystem_resource_based, :nodes]))
```
arbulu89, line 20: What is this for? to remove the `sid` field?
The sid cannot be removed, it is used in the frontend.
Yet again, most probably the schema is what it is wrong

arbulu89 (changes requested):
Hey @CDimonaco 
Great effort.
Unfortunately, there are many things wrong.
I think you have trusted too much in the current state of the schemas.
Many changes you did in the views are wrong, because you based them in the schema, but the schema itself is what it is wrong.
I think I have caught all  the things.
Let me know if you need more details.

As a nitpick:
The database and sapsystem view changes. The content we send in the controller and in the broadcast should be the same. It you want to keep the duplicated code, to be explicit, so be it, but with the same content

lib/trento_web/views/v1/host_view.ex
```diff
@@ -5,14 +5,39 @@ defmodule TrentoWeb.V1.HostView do
...
+    sles_subscriptions =
+      render_many(sles_subscriptions, __MODULE__, "sles_subscription.json",
+        as: :sles_subscription
+      )
+
+    host
+    |> Map.from_struct()
+    |> Map.put(:sles_subscriptions, sles_subscriptions)
+    |> Map.delete(:fully_qualified_domain_name)
+    |> Map.delete(:selected_checks)
```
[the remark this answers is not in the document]
dottorblaster: You're right, we take this and put it in the hosts slice as it is.

CDimonaco:
I took schema and tests for granted :eyes:

So, back to draft and we need to start back from schemas, good catch, fixing this things will give more confidence, but it's a real mess, we will learn a very valuable lesson in the end, I hope so :eyes:

Anyway some code duplication is completely on purpose, and if we want to avoid we need to rethink all the schemas because rn broadcasted info and controller one are very different

lib/trento_web/views/v1/host_view.ex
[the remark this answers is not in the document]
nelsonkopliku: Correct, both when first loading the `/hosts` api and when some event that gets projected changes the selected checks.

arbulu89 (approved):
Now it looks great @CDimonaco 
+1000

PR #2356, opened 2024-02-21
Connect sagas to the actual SUSE Manager config UI

dottorblaster:
# Description
Connecting the actual settings page to the settings API for SUSE Manager integration.

## How was this tested?
Jest tests updates, added new ones

### DIFF
--- a/assets/js/common/SuseManagerSettingsDialog/SuseManagerSettingsModal.jsx
+++ b/assets/js/common/SuseManagerSettingsDialog/SuseManagerSettingsModal.jsx
@@ -12,6 +12,23 @@ import { hasError } from './errors';
 
 const defaultErrors = [];
 
+const getCertificatePayload = (
+  certUploadDate,
+  editingCertificate,
+  certificate
+) => {
+  if (certUploadDate && editingCertificate && certificate === '') {
+    return { ca_cert: null };
+  } else if (!certUploadDate && editingCertificate && certificate === '') {
+    return {};
+  } else if (certUploadDate && !editingCertificate) {
+    return {};
+  }
+  return {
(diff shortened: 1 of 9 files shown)


nelsonkopliku (changes requested):
Looking great!

Besides the two minor comments the main issues I found are:
- on saving when the certificate is left empty in the ui `ca_cert: ""` is passed to the api which does not pass validation. If the certificate is not wanted then the `ca_cert` property should not be in the api request payload.
- similarly, when patching settings if we want to remove the certificate, we need to provide the api with `ca_cert: null` instead of `ca_cert: ""` otherwise the same validation would not pass.

dottorblaster:
@nelsonkopliku thanks for catching that behavior, I applied the fixes :+1:

nelsonkopliku (changes requested):
Almost there.

Saving is working correctly, ca_cert is not sent if the user did not provide any :+1: 

On update, however, we need an explicit `ca_cert: null` if we want to remove the certificate, currently the field is not sent at all, meaning it remains untouched.

Besides this the Remove action seems to not be retained in the modal when changing creds like cert or password.

dottorblaster:
@nelsonkopliku thanks! It should be okay now.

> Besides this the Remove action seems to not be retained in the modal when changing creds like cert or password.

I didn't quite understand this, but I guess you mean that closing and reopening the modal should display the `Remove` buttons again. I think I did something about that.

nelsonkopliku (approved):
Yes, you got it right! I was referring to that.

Besides the CI, LGTM. Great job!

dottorblaster:
Thank you! :smile:

jamie-suse (approved):

PR #2362, opened 2024-02-23
Add flag for "SUSE Manager config enabled"

jamie-suse:
# Description

This PR adds a flag to enable the SUSE Manager config. When the flag is enabled:
- The API endpoints for SUSE Manager credentials are enabled
- The SUSE Manager config is available in the Settings page

Also added a helper library `getFromConfig()` that retrieves a variable from the global `config` object. This way we don't need to add the line `// eslint-disable-next-line no-undef` everywhere when we need to read values from the global `config`.

## How was this tested?

Add unit tests for the `getFromConfig()` library.

### DIFF
--- a/lib/trento_web/router.ex
+++ b/lib/trento_web/router.ex
@@ -146,9 +146,11 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
-      resources "/settings/suma_credentials", SUMACredentialsController,
-        only: [:show, :create, :update, :delete],
-        singleton: true
+      if Application.compile_env!(:trento, :suse_manager_enabled) do
+        resources "/settings/suma_credentials", SUMACredentialsController,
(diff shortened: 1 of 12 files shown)


lib/trento_web/router.ex
```diff
@@ -146,9 +146,11 @@ defmodule TrentoWeb.Router do
 
       get "/hosts/:id/exporters_status", PrometheusController, :exporters_status
 
-      resources "/settings/suma_credentials", SUMACredentialsController,
-        only: [:show, :create, :update, :delete],
-        singleton: true
+      if Application.compile_env!(:trento, :suse_manager_enabled) do
```
nelsonkopliku, line 149: what about using `Unplug.Predicates.AppConfigEquals`?

nelsonkopliku (approved):
Besides the tiny comment LGTM.

Do you think we can also get rid of fun_with_flags from deps? :eyes:

dottorblaster (approved):
:fire: E DAJE

PR #2363, opened 2024-02-23
Integrate suma get system id

nelsonkopliku:
# Description

This PR adds an integration service for SUMA software updates discovery, as a GenServer.
This service mainly keeps track of settings/credentials and authentication against suma.
Available features/interactions:
- `setup`: sets up the state of the process by loading stored credentials and attempting a login (login is retried 5 times before giving up)
- `get_system_id`: gets the system id for a given fqdn. if the process was never setup and no authentication is available, the setup step is invoked before querying suma for the system id
- process is globally registered as [indicated here](https://hexdocs.pm/elixir/1.15.7/GenServer.html#module-name-registration)
- for demo purposes a mock of the integration is provided as `MockSuma`

Notes: 
- setup comes handy when we change settings/creds for instance and we want to re-set service state with new info
- the name to be passed to the service is mainly for testing purposes, here's [Valim's take on that](https://elixirforum.com/t/exunit-start-supervisor-2-is-persisting-process-state-between-tests/15421/3)

TODO:
- intercept cookie expiration, ie when we get a 401 from an api we try logging in first
- hide sensitive info in process state (password, ca_cert, auth cookie) from inspection. Thanks @dottorblaster for [raising this up](https://gabrielpereira.dev/en/posts/protecting-sensitive-data-in-elixir-gen-servers/)
- handle ca_cert 

## How was this tested?

Automated tests added plus manual testing against a real suma instance.

### DIFF
--- a/lib/trento/infrastructure/software_updates/suma.ex
+++ b/lib/trento/infrastructure/software_updates/suma.ex
@@ -1,0 +1,115 @@
+defmodule Trento.Infrastructure.SoftwareUpdates.Suma do
+  @moduledoc """
+  SUMA Software updates discovery adapter
+  """
+
+  @behaviour Trento.SoftwareUpdates.Discovery.Gen
+
+  use GenServer, restart: :transient
+
+  alias Trento.Infrastructure.SoftwareUpdates.{Suma.State, SumaApi}
+  alias Trento.SoftwareUpdates
+
+  require Logger
+
+  @default_name "suma"
+
+  def start_link([]), do: start_link(@default_name)
+
+  def start_link(server_name),
+    do: GenServer.start_link(__MODULE__, %State{}, name: process_identifier(server_name))
+
+  @impl true
+  def init(%State{} = state), do: {:ok, state}
+
+  def identify(server_name \\ @default_name),
+    do:
+      server_name
+      |> identificaton_tuple
+      |> :global.whereis_name()
+
+  def setup(server_name \\ @default_name),
+    do:
+      server_name
+      |> process_identifier
+      |> GenServer.call(:setup)
+
+  @impl true
+  def get_system_id(fully_qualified_domain_name, server_name \\ @default_name),
+    do:
+      server_name
+      |> process_identifier
+      |> GenServer.call({:get_system_id, fully_qualified_domain_name})
+
+  @impl true
+  def handle_call(:setup, _from, %State{} = state) do
+    case setup_auth(state) do
+      {:ok, new_state} ->
+        {:reply, :ok, new_state}
+
+      {:error, _} = error ->
+        {:reply, error, state}
+    end
+  end
+
+  @impl true
+  def handle_call(request, from, %State{auth: nil} = state),
+    do: {:noreply, state, {:continue, {:setup, from, request}}}
+
+  @impl true
+  def handle_call(
+        {:get_system_id, fully_qualified_domain_name},
+        _from,
+        %{
+          url: url,
+          auth: auth_cookie
+        } = state
+      ) do
+    {:reply, SumaApi.get_system_id(url, auth_cookie, fully_qualified_domain_name), state}
+  end
+
+  @impl true
+  def handle_continue({:setup, from, previous_message}, %State{} = state) do
+    case setup_auth(state) do
+      {:ok, new_state} ->
+        Process.send(self(), {previous_message, reply_to: from}, [:nosuspend, :noconnect])
+        {:noreply, new_state}
+
+      {:error, reason} ->
+        {:stop, reason}
+    end
+  end
+
+  @impl true
+  def handle_info(
+        {{:get_system_id, fully_qualified_domain_name}, reply_to: from},
+        %State{
+          url: url,
+          auth: auth_cookie
+        } = state
+      ) do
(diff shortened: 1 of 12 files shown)


lib/trento/infrastructure/software_updates/suma.ex
```diff
@@ -0,0 +1,188 @@
...
+      {:ok, system_id}
+    else
+      _ = error ->
+        Logger.error(
+          "Failed to get system id for host #{fully_qualified_domain_name}. Error: #{inspect(error)}"
+        )
+
+        {:error, :system_id_not_found}
+    end
+  end
```
dottorblaster, line 171: Could you move this logic inside `suma_http_client.ex`? We could let the pure part of the HTTP client come out with all the data stripped down of the irrelevant details.

lib/trento/infrastructure/software_updates/suma.ex
```diff
@@ -0,0 +1,188 @@
...
+        {{:get_system_id, fully_qualified_domain_name}, reply_to: from},
+        %State{
+          url: url,
+          auth: auth_cookie
+        } = state
+      ) do
+    GenServer.reply(from, get_system_id(url, auth_cookie, fully_qualified_domain_name))
+
+    {:noreply, state}
+  end
```
dottorblaster, line 96: It's me for sure but I'm not fully understanding why do we need this `handle_info` clause if we have the `handle_call` that reacts to the same message

lib/trento/infrastructure/software_updates/suma.ex
```diff
@@ -0,0 +1,188 @@
...
+  defp try_login(url, %{username: username, password: password} = credentials, retry) do
+    case login(url, username, password) do
+      {:ok, _} = successful_login ->
+        successful_login
+
+      {:error, reason} ->
+        Logger.error("Failed to Log into SUSE Manager, retrying...", error: inspect(reason))
+        try_login(url, credentials, retry - 1)
+    end
+  end
```
dottorblaster, line 134: Nice recursion bro

lib/trento/infrastructure/software_updates/suma.ex
```diff
@@ -0,0 +1,188 @@
...
+
+  defp get_system_id(url, auth_cookie, fully_qualified_domain_name) do
+    with {:ok, %HTTPoison.Response{status_code: 200, body: body}} <-
+           client().get_system_id(url, auth_cookie, fully_qualified_domain_name),
+         {:ok, %{"success" => true, "result" => result}} <- Jason.decode(body),
+         {:ok, system_id} <- extract_system_id(result) do
+      {:ok, system_id}
+    else
+      _ = error ->
```
dottorblaster, line 164: You're actually using the error, no need for the black hole

```suggestion
      error ->
```

dottorblaster (changes requested):
Good job! Some change requests but after we address those I'm ok to merge :+1:

lib/trento/infrastructure/software_updates/suma.ex
nelsonkopliku: yo mean `extract_system_id`?

dottorblaster: Both `get_system_id` and `extract_system_id` :+1:

lib/trento/infrastructure/software_updates/suma.ex
nelsonkopliku: because of the login loop. 
If we call `Suma.get_system_id("foobar")` before any setup was done and no auth is in the process state we need to login first.

As a first step we fall [in this case](https://github.com/trento-project/web/pull/2363/files#diff-9e97d75bb08b31078aef739786dbfd385efcf710824c9a21cddb515a58630354R58)

```elixir
def handle_call(request, from, %State{auth: nil} = state),
    do: {:noreply, state, {:continue, {:setup, from, request}}}
```
continuing [here](https://github.com/trento-project/web/pull/2363/files#diff-9e97d75bb08b31078aef739786dbfd385efcf710824c9a21cddb515a58630354R74) and logging in
```elixir
def handle_continue({:setup, from, previous_message}, %State{} = state) do
  case setup_auth(state) do
    {:ok, new_state} ->
      Process.send(self(), {previous_message, reply_to: from}, [:nosuspend, :noconnect])
      {:noreply, new_state}

    {:error, reason} ->
      {:stop, reason}
  end
end
```
and finally by doing `Process.send(self(), {previous_message, reply_to: from}, [:nosuspend, :noconnect])` we make sure the handle info for get_system_id is executed where we actually reply with `GenServer.reply(...)`

I am not a fan of that either and very keen to refactor. 
I believe we will come back to this service again and if you're ok we can let this sink and refactor as we go. Would that be ok?

dottorblaster: Thank you so much for the explanation! So far I have no clue about how we could refactor this, so okay to keep it. I was just curious about the flow :+1:

nelsonkopliku:
@dottorblaster thanks for the feedback exchange, was valuable! I guess I addressed what we agreed :wink:

jamie-suse (approved):
LGTM :rocket:

PR #2366, opened 2024-02-26
Add `managed` field in Cluster resources

rtorrero:
# Description

This PR adds:
 - The managed field for all resources in the clusters API
 - The managed column in the cluster resources (node details) view
 - An 'unmanaged' counter as a tooltip on the site details table
 - Small refactor of existing code to adapt the API for v1 (`adapt_v1` function)

## Frontend Preview: 
Checkout storybook "Unmanaged Node Resources"

### Add count of unmanaged resources when a node is online but has unmanaged resources

image

###  Add Managed column in Node details view
image

## How was this tested?
- Updated existing frontend tests
- Added cypress test to check tooltip message
- Updated view tests for API backwards compatibility
- Updated existing policy tests to include managed field


assets/js/pages/ClusterDetails/AttributesDetails.jsx
```diff
@@ -15,16 +15,45 @@ function AttributesDetails({ attributes, resources, title }) {
...
+              key: resourceKeys[0],
+            },
+            {
+              title: 'ID',
+              key: resourceKeys[1],
+            },
+            {
+              title: 'Managed',
+              key: resourceKeys[2],
+              render: (content) => (content ? 'True' : 'False'),
```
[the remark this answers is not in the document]
EMaksy: content (managed)  is a bool, lodash capitalize requires a string. 
We would need to convert content bool value to a string.

You would prefer something like that instead?: 
`render: (content) => _.capitalize(content.toString()),
`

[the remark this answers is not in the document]
arbulu89: We do this in other parts: https://github.com/trento-project/web/blob/main/assets/js/pages/ClusterDetails/HanaClusterDetails.jsx#L161
It doesn't need to be the same, but even in that context, the value is a bool

assets/js/pages/ClusterDetails/AttributesDetails.jsx
```diff
@@ -19,12 +21,39 @@ function AttributesDetails({ attributes, resources, title }) {
...
+              key: 'role',
+            },
+            {
+              title: 'Status',
+              key: 'status',
+            },
+            {
+              title: 'Managed',
+              key: 'managed',
+
```
arbulu89, line 44: Can we remove this empty line?

assets/js/pages/ClusterDetails/ClusterNodeName.jsx
```diff
@@ -6,31 +6,34 @@ import {
...
+const statusIcons = {
+  Online: <EOS_BOLT_FILLED className="tn-online" />,
+  Offline: <EOS_POWER_OFF_OUTLINED className="tn-offline" />,
+  Maintenance: <EOS_BUILD_OUTLINED className="tn-maintenance" />,
+  Default: <EOS_WARNING_OUTLINED className="tn-unknown" />,
 };
 
-function ClusterNodeName({ status, hostId, children }) {
+const getNodeStatusIconAndMessage = (status, resources) => {
+  const unmanagedResources = filter(resources, { managed: false }).length;
```
arbulu89, line 22: Maybe `countBy(resources, { managed } => managed).true`

assets/js/pages/ClusterDetails/ClusterNodeName.jsx
```diff
@@ -6,31 +6,34 @@ import {
...
+  const message =
+    unmanagedResources > 0
```
arbulu89, line 24: This logic is not correct.
We need to follow the correct precedence to create the message:
```
offline -> warning icon (tooltip offline)
unclean -> warning icon (tooltip unclean)
standby -> warning icon (tooltip standby)
maintenance -> maintenance icon
resource unmanaged -> maintenance icon
```

This means that we need to have this tooltip if the previous states are not met.
That's why most probably the previous `switch/case` was a good idea, as it gives you this ordering.

Besides that, with this new logic the `x resources unmanaged` message wouldn't have the maintenance icon.

arbulu89 (commented):
Hey @EMaksy 
Some comments. We need to rethink some new things you did

assets/js/pages/ClusterDetails/ClusterNodeName.jsx
```diff
@@ -26,11 +28,21 @@ const getNodeStatusIcon = (status) => {
   }
 };
 
-function ClusterNodeName({ status, hostId, children }) {
+const getNodeMessageAndIcon = (status, resources) => {
+  const { unmanaged } = countBy(resources, (resource) =>
```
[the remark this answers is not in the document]
arbulu89: Useful for what?
I would stick to for what we need, that is to count the unmanaged resources

assets/js/pages/ClusterDetails/HanaClusterDetails.test.jsx
```diff
@@ -385,10 +388,20 @@ describe('HanaClusterDetails component', () => {
     expect(screen.getByText('Attributes')).toBeInTheDocument();
     expect(screen.getByText('Resources')).toBeInTheDocument();
 
-    Object.keys(resources[0]).forEach((key) => {
-      expect(screen.getByText(key)).toBeInTheDocument();
-      screen.getAllByText(resources[0][key]).forEach((element) => {
-        expect(element).toBeInTheDocument();
+    tableColumns.forEach(({ key, title }) => {
```
EMaksy, line 391: I had to refactor this test as we use now a custom config, previously we were iterating through the resource key value pairs and used it to render our table.  Now we provide [a config](https://github.com/trento-project/web/pull/2366/files#diff-e440c92bcaacad26fb9ca18f7a4e0186e6af4712f0eaee988ec4c1414dd0b657R9), so it is more  explicit. 

Making this change, effects this part a lot. I tried to refactor it and adjust to the new config. 
not fully happy with it, as it is quite long!

Open for refactoring suggestions.

assets/js/pages/ClusterDetails/AscsErsClusterDetails.test.jsx
```diff
@@ -269,10 +271,20 @@ describe('ClusterDetails AscsErsClusterDetails component', () => {
...
     expect(screen.getByText('Resources')).toBeInTheDocument();
 
-    Object.keys(resources[0]).forEach((key) => {
-      expect(screen.getByText(key)).toBeInTheDocument();
-      screen.getAllByText(resources[0][key]).forEach((element) => {
-        expect(element).toBeInTheDocument();
+    tableColumns.forEach(({ key, title }) => {
+      expect(screen.getByText(title)).toBeInTheDocument();
+      resources.forEach((resource) => {
+        let value;
```
EMaksy, line 277: Maybe there is a better way ? Open for suggestions :+1:

assets/js/pages/ClusterDetails/AttributesDetails.jsx
```diff
@@ -1,9 +1,42 @@
 import React, { useState } from 'react';
 
+import { capitalize } from 'lodash';
+
 import Button from '@common/Button';
 import Modal from '@common/Modal';
 import Table from '@common/Table';
 
+export const resourceTableConfig = {
```
EMaksy, line 9: I'm exporting the config, so it can be reused in tests. So if we add or delete a new key value pair, we don't need to change the tests in AscsErsClusterDetails and HanaClusterDetails. An alternative would be just to hard code the table titles/keys in the test and adjust them every time we change something. 

Any preferences or better options?

test/e2e/cypress/e2e/hana_cluster_details.cy.js
```diff
@@ -193,6 +193,18 @@ context('HANA cluster details', () => {
             .eq(1)
             .should('have.class', host.status);
         });
+
+        it(`${host.hostname} icon should have the expected tooltip message`, () => {
```
EMaksy, line 197: Hey @arbulu89 and @CDimonaco, I added a test, to check the tooltip message. 
We could extend this even and test different scenarios. For this, we would need to enchance the photofinish scenario and add some logic to validate it. 
What do you guys think ?

assets/js/pages/ClusterDetails/HanaClusterDetails.test.jsx
arbulu89: Hardcode the tests in a `it.each` please. For each column. And the same as before. This could go in the `AttributeResources.test.jsx`, and remove this from here

arbulu89 (changes requested):
@rtorrero @EMaksy 
The backend part is ready for me. I did some comments, and I think it could be improved, but well, at least it is correct.
@EMaksy The frontend part could get some additional work

test/trento_web/views/v1/cluster_view_test.exs
```diff
@@ -40,6 +41,14 @@ defmodule TrentoWeb.V1.ClusterViewTest do
       refute Access.get(node, :nameserver_actual_role)
       refute Access.get(node, :indexserver_actual_role)
       refute Access.get(node, :status)
+
+      Enum.each(details.stopped_resources, fn stopped_resource ->
+        refute Map.has_key?(stopped_resource, :managed)
```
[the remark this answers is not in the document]
rtorrero: couldn't `Access.get` return a falsy boolean? (causing the test to pass when it shouldn't)

A `Map.has_key?` seems also semantically more correct to me.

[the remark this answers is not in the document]
arbulu89: It was more about consistency, not having `Access` above and `Map.has_key?` here.
I would be fine having `has_key?` in both.
Anyway, it was a detail

assets/js/pages/ClusterDetails/AttributeDetails.test.jsx
```diff
@@ -100,10 +100,36 @@ describe('AttributesDetails', () => {
...
+      expect(screen.getByText(key)).toBeInTheDocument();
+      expect(screen.getByText(value)).toBeInTheDocument();
+    });
+  };
+
+  const assertResourcesDisplayed = (resources, expectedResource) => {
+    expectedResource.forEach(({ key, title }) => {
+      expect(screen.getByText(title)).toBeInTheDocument();
+      resources.forEach((resource) => {
+        let value;
```
EMaksy, line 114: I put a let instead of const, as otherwise eslint complains about nested ternary operator. Maybe there is a better way ?

assets/js/pages/ClusterDetails/AttributeDetails.test.jsx
```diff
@@ -0,0 +1,167 @@
...
+    {
+      attributes: ascsErsClusterAttributes,
+      resources: ascsErsClusterResources,
+    },
+  ] = ascsErsClusterNodeFactory.buildList(1);
+  const [
+    { attributes: hanaClusterAttributes, resources: hanaClusterResources },
+  ] = hanaClusterDetailsNodesFactory.buildList(1);
+
+  const scenarios = [
```
EMaksy, line 51: Any more ideas for additional scenarios?

arbulu89 (approved):
Thank you @EMaksy @rtorrero 
Ready to merge from my side :rocket:

PR #2368, opened 2024-02-28
Change host's health based on software updates discovery result

nelsonkopliku:
# Description

This PR adds the domain logic to handle software updates discovery.
- a new event `SoftwareUpdatesDiscoveryCompleted` is emitted containing the discovered relevant patches counts. Needed to properly update aggregate state and opens the possibility to project such data if needed and to possibly broadcast via websockets interesting information that might be useful for the UX.
-  based on what has been discovered also a `HostHealthChanged` event might be emitted

I think we might need extra domain interactions related to software updates discovery business process that would allow us to answer the following:
- what happens when a user clears up suma settings? should we invalidate latest discovery data and ignore it in host's aggregated health?
- if we don't do anything when clearing up settings, we might wait for _the watch process_ :tm: next tick, however also in that case we need to decide what to do if a discovery had previously ran, but now we do not have settings anymore.

## How was this tested?

Automated tests added.

### DIFF
--- a/lib/trento/hosts/host.ex
+++ b/lib/trento/hosts/host.ex
@@ -37,6 +37,14 @@ defmodule Trento.Hosts.Host do
   The checks health is obtained from the [Checks Engine executions](https://github.com/trento-project/wanda/).
   Every time a checks execution for a host completes the execution's result is taken into account to determine host's health.
   Checks execution is started either by an explicit user request or periodically as per the scheduler configuration.
+
+  ### Software Updates Discovery health
+
+  The Software Updates Discovery health is obtained from the integration with an external service, SUMA,
+  that provides information about relevant patches and upgradable packages for the host.
+  Presence of relevant patches concurs to the host's aggregated health as follows:
+  - critical if there is at least one security advisory
+  - warning if there are only buxfixes/software enhancements
   """
 
   require Trento.Enums.Provider, as: Provider
@@ -52,12 +60,14 @@ defmodule Trento.Hosts.Host do
     AwsProvider,
     AzureProvider,
     GcpProvider,
+    RelevantPatches,
     SaptuneStatus,
     SlesSubscription
   }
 
   alias Trento.Hosts.Commands.{
     CompleteHostChecksExecution,
+    CompleteSoftwareUpdatesDiscovery,
     DeregisterHost,
     RegisterHost,
     RequestHostDeregistration,
@@ -86,7 +96,8 @@ defmodule Trento.Hosts.Host do
     HostTombstoned,
     ProviderUpdated,
     SaptuneStatusUpdated,
-    SlesSubscriptionsUpdated
+    SlesSubscriptionsUpdated,
+    SoftwareUpdatesDiscoveryCompleted
   }
 
   @required_fields []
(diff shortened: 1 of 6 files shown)


lib/trento/hosts/host.ex
```diff
@@ -738,17 +802,30 @@ defmodule Trento.Hosts.Host do
...
+  defp compute_software_updates_discovery_health(%RelevantPatches{total: 0}),
+    do: Health.passing()
+
+  defp compute_software_updates_discovery_health(%RelevantPatches{
+         security_advisories: security_advisories
+       })
+       when security_advisories > 0,
+       do: Health.critical()
+
+  defp compute_software_updates_discovery_health(_), do: Health.warning()
```
jamie-suse, line 814: I'm just curious as to when `Health.warning()` triggers?

nelsonkopliku: > warning if there are only buxfixes/software enhancements
- no patches -> passing
- some patches and at least one security advisory -> critical
- some patches, no security advisories but only buxfixes/software enhancements -> warning

lib/trento/hosts/host.ex
```diff
@@ -507,6 +524,39 @@ defmodule Trento.Hosts.Host do
...
+          relevant_patches: %RelevantPatches{
+            total: total,
+            security_advisories: security_advisories,
+            bug_fixes: bug_fixes,
+            software_enhancements: software_enhancements
+          }
+        }
+      )
+      when security_advisories + bug_fixes + software_enhancements != total,
+      do: {:error, :inconsistent_software_updates_discovery}
```
dottorblaster, line 540: Is this something that we care about? I would get rid of this clause, I'm pretty sure that the total you're getting is the sum of the lengths of all the lists :smile:

dottorblaster (changes requested):
Just a tiny change request, I think it would simplify the tests as well, let me know what do you think. Great job, keep it up!

lib/trento/hosts/host.ex
```diff
@@ -110,12 +121,18 @@ defmodule Trento.Hosts.Host do
...
+      values: Health.values(),
+      default: Health.unknown()
+
     field :health, Ecto.Enum, values: Health.values(), default: Health.unknown()
     field :rolling_up, :boolean, default: false
     field :selected_checks, {:array, :string}, default: []
     field :deregistered_at, :utc_datetime_usec, default: nil
 
     embeds_one :saptune_status, SaptuneStatus
+    embeds_one :relevant_patches, RelevantPatches
```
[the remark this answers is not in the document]
nelsonkopliku: not having the relevant patches in the aggregate like that would mean mainly moving the logic about determining whether the health outside of the aggregate, however I do believe that it is domain logic fitting very well here.

lib/trento/hosts/host.ex
nelsonkopliku: well, I was doubtful myself, I could just get rid of the total. I was just thinking of making it extremely simple for someone consuming that possibily projected/broadcasted information to not have to do sums, bot i can live wiothout that

lib/trento/hosts/host.ex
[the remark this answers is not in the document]
nelsonkopliku: I see now! Yes I will remove from aggregate's state :+1:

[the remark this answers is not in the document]
nelsonkopliku: here we go https://github.com/trento-project/web/pull/2368/commits/bf6e275c4fc976f5bab4a6c40b30b3d659c60d0e

dottorblaster (approved):
You have my sword

jamie-suse (approved):

jamie-suse:
gimli

PR #2369, opened 2024-02-28
Fix :nxdomain crashing when SMTP server was wrongly set up

dottorblaster:
# Description
Fixing the alerting module crashing when the SMTP server is wrongly configured because String casting is not implemented by default for tuples. Applying a little bit of pattern magic to rescue in that specific case.

## How was this tested?
Unit test added to replicate the behavior and test that the fix stays in place over time


nelsonkopliku (approved):
LGTM! Thanks! :email:

PR #2370, opened 2024-02-28
Add expect enum usage

arbulu89:
# Description

Add `expect_enum` results usage. This means that the host expectations are now of `expect` and `expect_enum`, and the results in the result detail view could be of warning type.

## How was this tested?

Tested with UT


assets/js/lib/model/index.js
```diff
@@ -1,13 +1,21 @@
 export const EXPECT = 'expect';
+export const EXPECT_ENUM = 'expect_enum';
 export const EXPECT_SAME = 'expect_same';
 
+const hostExepectations = [EXPECT, EXPECT_ENUM];
```
EMaksy, line 5: Just a quick question here, we have multiple files in model directory.  
Like cluster.js, sapSystem.js and index.js

What kind of data should we always put in index.js?

For now, it is mostly cluster details view related.

EMaksy (approved):
Nice jooob! 
LGTM

assets/js/lib/model/index.js
arbulu89: well, that's subjective, as long as it doesn't go out of control having one file is convenient

PR #2376, opened 2024-02-29
Add a scheduled software updates discovery job

nelsonkopliku:
# Description

Fixes # (issue)

## How was this tested?


lib/trento/software_updates/discovery.ex
```diff
@@ -13,5 +21,90 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
+
+  defp track_relevant_patches(%{advisory_type: advisory_type}, %{
+         security_advisories: security_advisories,
+         bug_fixes: bug_fixes,
+         software_enhancements: software_enhancements
+       }) do
+    advisory_type = AdvisoryType.from_string(advisory_type)
+
+    %{
+      security_advisories: increment_security_advisories(advisory_type, security_advisories),
```
jamie-suse, line 88: Is there a reason to abstract this into a function? Also why does the function take an `AdvisoryType`?

nelsonkopliku: Addressed differently here https://github.com/trento-project/web/pull/2376/commits/dcf5f14774bd12c7a19e0f445ab2e3432b920e0e

If it helps, I am happy :smile:

jamie-suse (approved):
beautiful :ok_hand:

lib/trento/software_updates/discovery.ex
```diff
@@ -13,5 +21,87 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
+  @spec discover_software_updates :: {:ok, {list(), list()}}
+  def discover_software_updates,
+    do:
+      {:ok,
+       Hosts.get_all_hosts()
+       |> Enum.map(&discover_host_software_updates/1)
+       |> Enum.split_with(fn
+         {:ok, _, :discovered, _, _} -> true
+         _ -> false
+       end)}
```
dottorblaster, line 33: Can't you just return `[]` over an erroring discovery? you can keep the logging and you just have to `Enum.flatten` this list instead of playing with tagged tuples.

nelsonkopliku: When testing I want to rely on the output of the function, rather than logging :sweat_smile: That's the reason behind it.

We could alternatively return just one list with all the discovery results, successful and errored. However we'd need to do filtering of successful vs errored results in the test to make meaningful assertions. We get that distinction for free by splitting here. I find it convenient.

I simplified a bit the tuple returned in success case from `{:ok, host_id, :discovered, system_id, relevant_patches}` to `{:ok, host_id, system_id, relevant_patches}` :+1:

PR #2381, opened 2024-03-04
Redact sensitive data in SUMA GenServer state

jamie-suse:
# Description

This PR redacts sensitive data (`password`, `auth`, `ca_cert`) in the SUMA GenServer state when the GenServer is inspected or logged. More concretely, a custom `Inspect` protocol is implemented for the GenServer's `State` struct, thereby redacting said fields when inspections such as `:sys.get_state/1` or `:sys.get_status/1` are called.

## How was this tested?

Added unit test


nelsonkopliku (approved):
Super :rocket: 

we could also test against `:sys.get_status()`, but don't bother, I tested manually and it works as expected.

dottorblaster (approved):
:fire:

PR #2386, opened 2024-03-06
Add a clear operation to suma service

nelsonkopliku:
# Description

Adding a clear feature to the suma integration service that wipes out its internal state.

We might end up in adding the `clear` function in the discovery behaviour, but I'll add that later if actually needed (I guess so :smile: )

## How was this tested?

Automated tests.


dottorblaster (approved):
LGTM

jamie-suse (approved):
:rocket:

nelsonkopliku:
@dottorblaster @jamie-suse thanks for the review. 
Just for everyone to be aware I am having doubts whether this operation should be a `call` or a `cast`, fire and forget.
As we keep going on it we'll discover and eventually refactor, if needed.
Merging for now :+1:

PR #2389, opened 2024-03-06
Add wrap prop to tooltip

dottorblaster:
# Description
Our tooltip component wraps elements in a span by default. In order not to mess up with CSS styles in some flexbox/grid context this behavior needs to be turned off on demand.

## How was this tested?
Jest tests added


nelsonkopliku (approved):
LGTM

jamie-suse (approved):
:rocket:

PR #2393, opened 2024-03-07
Clear suma discovery domain

nelsonkopliku:
# Description

This PR introduces domain logic to handle clearing up of software updates discovery results.
Basically when a user clears up suma credentials, whatever has been previously discovered, gets ignored.
Specifically the health of the software updates discovery gets ignored from the host's aggregated health.

This is going to happen in a loop for all the hosts meaning that a bunch of `HostHealthChanged` could be emitted upon creds clear up, with related ui notifications.

## How was this tested?

Automated tests


dottorblaster (approved):
LGTM

I was thinking maybe we could leverage already existing events but in the end it doesn't matter and when having events involved shortcuts fall short pretty quickly :+1:

jamie-suse (approved):

PR #2391, opened 2024-03-06
Add ability to provide a CA cert when making requests to SUMA

jamie-suse:
# Description

This change adds ability to provide a CA cert when making requests to SUMA.

Because the [HTTPoison](https://github.com/edgurgel/httpoison) library uses Erlang library [hackney](https://github.com/benoitc/hackney) underneath, whose API demands that we provide a path to a file containing the certificate; we need to save a certificate file on the filesystem somewhere. Note that if we save some new credentials that _don't_ contain a certificate (i.e. the user _clears_ the certificate), then the old certificate file is removed.

## How was this tested?

Added unit tests for this new functionality.

PR #2395, opened 2024-03-07
Don't return error when deleting non-existent CA cert file

jamie-suse:
# Description

Relates to https://github.com/trento-project/web/pull/2391

If the SUMA GenServer attempts to delete a certificate file that doesn't exist, it will return an error status. This should not be considered an error case for our use-case and so we should use `File.rm_rf/1` instead, which specifically ignores this.

This behaviour/symptom also surfaced with some flakiness in the test `"should not save CA certificate file if no cert is provided"`, where _sometimes_ `Suma.setup(@test_integration_name)` returned the aforementioned `{:error, :enoent}`, due to the nondeterministic nature of the order of operations.

## How was this tested?

Flakiness of unit test `"should not save CA certificate file if no cert is provided"` should be fixed.

### DIFF
--- a/lib/trento/infrastructure/software_updates/suma.ex
+++ b/lib/trento/infrastructure/software_updates/suma.ex
@@ -129,7 +129,12 @@ defmodule Trento.Infrastructure.SoftwareUpdates.Suma do
     end
   end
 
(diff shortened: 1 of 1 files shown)


dottorblaster (approved):
Fair enough

nelsonkopliku (approved):

PR #2396, opened 2024-03-07
Use derived state `use_ca_cert` in SUMA GenServer

jamie-suse:
# Description

Arises from [comment](https://github.com/trento-project/web/pull/2391#discussion_r1515712747) in https://github.com/trento-project/web/pull/2391

> If we decide to have `SumaApi` functions accept a `use_ca_cert` as mentioned [here](https://github.com/trento-project/web/pull/2391/files#r1515701175) we could also think of centralizing that `ca_cert =! nil` in the genserver state that we set in this function 👀

This change adds some derived state `use_ca_cert` into the SUMA GenServer, to abstract away the logic of checking `ca_cert =! nil` when the GenServer calls the SUMA API functions.

## How was this tested?

Updated existing unit tests.

### DIFF
--- a/lib/trento/infrastructure/software_updates/state.ex
+++ b/lib/trento/infrastructure/software_updates/state.ex
@@ -9,7 +9,8 @@ defmodule Trento.Infrastructure.SoftwareUpdates.Suma.State do
     :username,
     :password,
     :ca_cert,
-    :auth
+    :auth,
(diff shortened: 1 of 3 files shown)


nelsonkopliku (approved):
On fire :fire:

dottorblaster (approved):

PR #2397, opened 2024-03-07
Remove duplicated redux state `database_instances` field from the SAP system reducer

arbulu89:
# Description

This PRs remove a lot of duplicated code in the redux state usage.
Having the `database_instances` information the sap system reducer is not needed, as we are not using this field information anywhere. 
When we need to merge get a SAP system database instances, we just get them in a selector code, matching the `sap_system_id`.

I have checked all the options, and we should be fine.
In the other hand, I have left that information in the backend. Even though we don't use it in the frontend, I find it useful for any other usecase (to know the database instances belonging to a SAP system).

## How was this tested?

Current tests are enough to see that anything breaks

### DIFF
--- a/assets/js/state/sagas/databases.js
+++ b/assets/js/state/sagas/databases.js
@@ -24,16 +24,6 @@ import {
   unsetDatabaseInstanceDeregistering,
 } from '@state/databases';
 
-import {
-  upsertDatabaseInstancesToSapSystem,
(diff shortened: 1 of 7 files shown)


jamie-suse (approved):
Makes sense :rocket:

CDimonaco (approved):

EMaksy (approved):
Nice job! 
LGTM!

PR #2403, opened 2024-03-11
Api Key settings frontend

CDimonaco:
# Description

This PR adds the Api Key settings frontend.
It includes the generation of new api key with custom expiration

New settings page

image

API Key settings modal

image

## How was this tested?

Automated tests

### DIFF
--- a/assets/css/input-number.css
+++ b/assets/css/input-number.css
@@ -1,0 +1,170 @@
+.rc-input-number {
+  display: inline-block;
+
+  height: 26px;
+
+  margin: 0;
+
+  padding: 0;
+
+  font-size: 12px;
+
+  line-height: 26px;
+
+  vertical-align: middle;
+
+  border: 1px solid #d9d9d9;
+
+  border-radius: 4px;
+
+  transition: all 0.3s;
+}
+
+.rc-input-number-focused {
+  border-color: #1890ff;
+
+  box-shadow: 0 0 0 2px rgba(24, 144, 255, 0.2);
+}
+
+.rc-input-number-out-of-range input {
+  color: red;
+}
+
+.rc-input-number-handler {
+  display: block;
+
+  /* height: 12px; */
+
+  overflow: hidden;
+
+  line-height: 12px;
+
+  text-align: center;
+
+  touch-action: none;
+}
+
+.rc-input-number-handler-active {
+  background: #ddd;
+}
+
+.rc-input-number-handler-up-inner,
+.rc-input-number-handler-down-inner {
+  color: #666666;
+
+  -webkit-user-select: none;
+
+  user-select: none;
+}
+
+.rc-input-number:hover {
+  border-color: #1890ff;
+}
+
+.rc-input-number:hover .rc-input-number-handler-up,
+.rc-input-number:hover .rc-input-number-handler-wrap {
+  border-color: #1890ff;
+}
+
+.rc-input-number-disabled:hover {
+  border-color: #d9d9d9;
+}
+
+.rc-input-number-disabled:hover .rc-input-number-handler-up,
+.rc-input-number-disabled:hover .rc-input-number-handler-wrap {
+  border-color: #d9d9d9;
+}
+
+.rc-input-number-input-wrap {
+  height: 100%;
+
+  overflow: hidden;
+}
+
+.rc-input-number-input {
+  width: 100%;
+
+  height: 100%;
+
+  padding: 0;
+
+  color: #666666;
+
+  line-height: 26px;
+
+  text-align: center;
+
+  border: 0;
+
+  border-radius: 4px;
+
+  outline: 0;
+
+  transition: all 0.3s ease;
+
+  transition: all 0.3s;
+
+  -moz-appearance: textfield;
+}
+
+.rc-input-number-handler-wrap {
+  float: right;
+
+  width: 20px;
+
+  height: 100%;
+
+  border-left: 1px solid #d9d9d9;
+
+  transition: all 0.3s;
+}
+
+.rc-input-number-handler-up {
+  padding-top: 1px;
+
+  border-bottom: 1px solid #d9d9d9;
+
+  transition: all 0.3s;
+}
+
+.rc-input-number-handler-up-inner:after {
+  content: '+';
+}
+
+.rc-input-number-handler-down {
+  transition: all 0.3s;
+}
+
+.rc-input-number-handler-down-inner:after {
+  content: '-';
+}
+
+.rc-input-number-handler-down-disabled,
+.rc-input-number-handler-up-disabled {
+  opacity: 0.3;
+}
+
+.rc-input-number-handler-down-disabled:hover,
+.rc-input-number-handler-up-disabled:hover {
+  color: #999;
+
+  border-color: #d9d9d9;
+}
+
+.rc-input-number-disabled .rc-input-number-input {
+  background-color: #f3f3f3;
+
+  cursor: not-allowed;
+
+  opacity: 0.72;
+}
+
+.rc-input-number-disabled .rc-input-number-handler {
+  opacity: 0.3;
+}
+
+.rc-input-number-disabled .rc-input-number-handler:hover {
+  color: #999;
+
+  border-color: #d9d9d9;
+}
--- a/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
+++ b/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
@@ -105,6 +105,7 @@ function ApiKeySettingsModal({
               className="h-3/4"
               type="number"
               disabled={!timeFormEnabled}
+              min="0"
               error={quantityError}
               onChange={({ target: { value } }) => {
                 setTimeQuantity(parseInt(value, 10));
--- a/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
+++ b/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
@@ -1,14 +1,24 @@
-import { addDays, addMonths, addYears, setHours, setMinutes } from 'date-fns';
+import {
+  addDays,
+  addMonths,
+  addYears,
+  format,
+  parseISO,
+  setHours,
+  setMinutes,
+} from 'date-fns';
 import React, { useState } from 'react';
-import { noop } from 'lodash';
-
+import { noop, truncate } from 'lodash';
+import { EOS_CONTENT_COPY, EOS_INFO_OUTLINED } from 'eos-icons-react';
 import Button from '@common/Button';
 import Modal from '@common/Modal';
-import Input, { Password, Textarea } from '@common/Input';
+import Input from '@common/Input';
 import Label from '@common/Label';
 import Select from '@common/Select';
+import Switch from '@common/Switch';
 
-const normalizeExpiration = (expiration) => setMinutes(setHours(0), 0);
+const normalizeExpiration = (expiration) =>
+  setMinutes(setHours(expiration, 0), 0);
 
 const availableTimeOptions = [
   {
@@ -31,6 +41,7 @@ function ApiKeySettingsModal({
   onSave = noop,
   onClose = noop,
   generatedApiKey,
+  generatedApiKeyExpiration,
 }) {
   const timeOptions = availableTimeOptions.map((o) => o.type);
   const [timeQuantity, setTimeQuantity] = useState(0);
@@ -39,42 +50,123 @@ function ApiKeySettingsModal({
   const [quantityError, setQuantityError] = useState(false);
   const [timeFormEnabled, setTimeFormEnabled] = useState(true);
 
+  const saveFormState = () => {
+    if (!timeFormEnabled) {
+      onSave({ apiKeyExpiration: null });
+      return;
+    }
+    if (timeQuantity === 0 || !timeQuantity) {
+      setQuantityError(true);
+      return;
+    }
+    const timeQuantitySettings = availableTimeOptions.find(
+      (q) => q.type === timeQuantityType
+    );
+    const apiKeyExpiration = timeQuantitySettings.timeGenerator(timeQuantity);
+
+    onSave({ apiKeyExpiration: normalizeExpiration(apiKeyExpiration) });
+  };
+
   return (
     <Modal
       title="API Key Settings"
-      className="!w-2/4"
+      className="!w-3/4 !max-w-3xl"
       open={open}
       onClose={onClose}
     >
-      <div className="flex items-center my-5 space-x-2">
-        <div className="w-1/3">
-          <Label>Key Expiration</Label>
-        </div>
+      <div className="flex flex-col my-5">
+        <span className="my-2 mb-4 text-gray-500">
+          {' '}
+          By generating a new key, you will need to replace the API key on all
+          hosts.{' '}
+        </span>
 
-        <div className="w-2/4">
-          <Input
-            value={timeQuantity}
-            className="h-4/5"
-            type="number"
-            disabled={!timeFormEnabled}
-            error={quantityError}
-            onChange={({ target: { value } }) => {
-              setTimeQuantity(value);
+        <div className="flex space-x-1">
+          <div className="w-1/5">
+            <Label>Never Expires</Label>
+          </div>
+
+          <Switch
+            selected={!timeFormEnabled}
+            onChange={() => {
+              setTimeFormEnabled((enabled) => !enabled);
               setQuantityError(false);
             }}
           />
         </div>
-        <div className="w-2/4 pt-4">
-          <Select
-            optionsName=""
-            options={timeOptions}
-            disabled={!timeFormEnabled}
-            value={timeQuantityType}
-            onChange={(value) => setTimeQuantityType(value)}
-          />
+        <div className="flex items-center my-2 space-x-2">
+          <div className="w-1/3">
+            <Label>Key Expiration</Label>
+          </div>
+
+          <div className="w-2/4">
+            <Input
+              value={timeQuantity}
+              className="h-3/4"
+              type="number"
+              disabled={!timeFormEnabled}
+              error={quantityError}
+              onChange={({ target: { value } }) => {
+                setTimeQuantity(parseInt(value, 10));
+                setQuantityError(false);
+              }}
+            />
+          </div>
+          <div className="w-2/4 pt-4">
+            <Select
+              optionsName=""
+              options={timeOptions}
+              disabled={!timeFormEnabled}
+              value={timeQuantityType}
+              onChange={(value) => setTimeQuantityType(value)}
+            />
+          </div>
+          <div className="w-1/6 h-4/5">
+            <Button
+              onClick={() => saveFormState()}
+              disabled={quantityError || loading}
+            >
+              Generate
+            </Button>
+          </div>
         </div>
+        {quantityError && (
+          <span className="my-1 mb-4 text-red-500">
+            {' '}
+            Key expiration value needs to be greater than 0{' '}
+          </span>
+        )}
+        {generatedApiKey && (
+          <div className="flex flex-col my-1 mb-4">
+            <div className="flex space-x-2">
+              <div className="w-full break-words p-2 pr-2 rounded-lg bg-white border-gray-300 border">
+                <code> {truncate(generatedApiKey, { length: 65 })} </code>
+              </div>
+              <button
+                type="button"
+                aria-label="generate api key"
+                onClick={() => {
+                  window.navigator.clipboard.writeText(generatedApiKey);
+                }}
+              >
+                <EOS_CONTENT_COPY role="button" size="25" />
+              </button>
+            </div>
+            <div className="flex space-x-2">
+              <EOS_INFO_OUTLINED size="20" className="mt-2" />
+
+              <div className="mt-2 text-gray-600 text-sm">
+                {' '}
+                Key Will Expire{' '}
+                {format(parseISO(generatedApiKeyExpiration), 'd LLL yyyy')}
+              </div>
+            </div>
+          </div>
+        )}
         <div className="w-1/6 h-4/5">
-          <Button>Generate</Button>
+          <Button type="primary-white" onClick={onClose}>
+            Close
+          </Button>
         </div>
       </div>
     </Modal>
--- a/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
+++ b/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
@@ -40,7 +40,12 @@ function ApiKeySettingsModal({
   const [timeFormEnabled, setTimeFormEnabled] = useState(true);
 
   return (
-    <Modal title="API Key Settings" open={open} onClose={onClose}>
+    <Modal
+      title="API Key Settings"
+      className="!w-2/4"
+      open={open}
+      onClose={onClose}
+    >
       <div className="flex items-center my-5 space-x-2">
         <div className="w-1/3">
           <Label>Key Expiration</Label>
--- a/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
+++ b/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import {
   addDays,
   addMonths,
@@ -46,6 +46,7 @@ function ApiKeySettingsModal({
   const timeOptions = availableTimeOptions.map((o) => o.type);
   const [timeQuantity, setTimeQuantity] = useState(0);
   const [timeQuantityType, setTimeQuantityType] = useState(timeOptions[0]);
+  const [keyGenerated, setKeyGenerated] = useState(false);
 
   const [quantityError, setQuantityError] = useState(false);
   const [apiKeyNeverExpires, setapiKeyNeverExpires] = useState(false);
@@ -53,6 +54,7 @@ function ApiKeySettingsModal({
   const generateApiKeyExpiration = () => {
     if (apiKeyNeverExpires) {
       onGenerate({ apiKeyExpiration: null });
+      setKeyGenerated(true);
       return;
     }
     if (timeQuantity === 0 || !timeQuantity) {
@@ -65,7 +67,16 @@ function ApiKeySettingsModal({
     const apiKeyExpiration = timeQuantitySettings.timeGenerator(timeQuantity);
 
     onGenerate({ apiKeyExpiration: normalizeExpiration(apiKeyExpiration) });
+    setKeyGenerated(true);
   };
+
+  useEffect(() => {
+    setapiKeyNeverExpires(false);
+    setQuantityError(false);
+    setTimeQuantityType(timeOptions[0]);
+    setTimeQuantity(0);
+    setKeyGenerated(false);
+  }, [open]);
 
   return (
     <Modal
@@ -137,7 +148,7 @@ function ApiKeySettingsModal({
             Key expiration value needs to be greater than 0{' '}
           </span>
         )}
-        {generatedApiKey && (
+        {generatedApiKey && keyGenerated && !loading && (
           <div className="flex flex-col my-1 mb-4">
             <div className="flex space-x-2">
               <div className="w-full break-words p-2 pr-2 rounded-lg bg-white border-gray-300 border">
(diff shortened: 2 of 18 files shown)


assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,176 @@
+import {
+  addDays,
+  addMonths,
+  addYears,
+  format,
+  parseISO,
+  setHours,
+  setMinutes,
+} from 'date-fns';
+import React, { useState } from 'react';
```
arbulu89, line 10: ocd: we usually put the `React` import as first line in the file, so it is obvious that this is a react component

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+  {
+    type: 'years',
+    timeGenerator: (quantity) => addYears(new Date(), quantity),
+  },
+];
+
+function ApiKeySettingsModal({
+  open = false,
+  loading = false,
+  onSave = noop,
```
arbulu89, line 41: What is `onSave`? when you click the `Generate` button?
Should we maybe renamed to `onGenerate` ?

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+  };
+
+  return (
+    <Modal
+      title="API Key Settings"
+      className="!w-3/4 !max-w-3xl"
+      open={open}
+      onClose={onClose}
+    >
+      <div className="flex flex-col my-5">
```
arbulu89, line 77: Checking the modal in `storybook`, I would say that some margins don't follow the mock up.
Can you confirm this with @jagabomb ?
I'm fine, but I want Jurgen's approval there hehe

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+  onClose = noop,
+  generatedApiKey,
+  generatedApiKeyExpiration,
+}) {
+  const timeOptions = availableTimeOptions.map((o) => o.type);
+  const [timeQuantity, setTimeQuantity] = useState(0);
+  const [timeQuantityType, setTimeQuantityType] = useState(timeOptions[0]);
+
+  const [quantityError, setQuantityError] = useState(false);
+  const [timeFormEnabled, setTimeFormEnabled] = useState(true);
```
arbulu89, line 51: Maybe `expirationEnabled` ?
I mean, the variable is used to check if we have an infinite key or not.
Anyway, it is just a nitpick

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -24,28 +25,52 @@ import {
   getSoftwareUpdatesSettingsErrors,
 } from '@state/selectors/softwareUpdatesSettings';
 
+import { dismissNotification } from '@state/notifications';
+import { API_KEY_EXPIRATION_NOTIFICATION_ID } from '@state/sagas/settings';
+
 import SuseManagerConfig from '@common/SuseManagerConfig';
+import { format, parseISO } from 'date-fns';
```
arbulu89, line 32: I would put this `import` in the top block, as it is an external library import

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -24,28 +25,52 @@ import {
...
-      })
+  const fetchApiKeySettings = () =>
+    get('/settings/api_key')
+      .then(
+        ({ data: { generated_api_key: newApiKey, expire_at: expireAt } }) => {
+          setApiKey(newApiKey);
+          setApiKeyExpiration(expireAt);
+          setLoading(false);
+        }
+      )
```
arbulu89, line 52: Wouldn't be better to include a `finally` clause here, after the `catch` where we don `setLoading(false)`?
This one, we just do it once.
The same for the `saveApiKeySettings`, that if the `patch` fails, we don't set the loading value otherwise

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -24,28 +25,52 @@ import {
...
       });
+
+  const saveApiKeySettings = (expiration) => {
+    setLoading(true);
+    patch('/settings/api_key', { expire_at: expiration }).then(
+      ({ data: { generated_api_key: newApiKey, expire_at: expireAt } }) => {
+        setApiKey(newApiKey);
+        setApiKeyExpiration(expireAt);
+        setLoading(false);
+        setApiKeySettingsModalOpen(false);
```
arbulu89, line 65: Is this `setApiKeySettingsModalOpen` here correct?
I mean, when we click `Generate` button, we still need to continue in the modal, and show the new key/expiration date, etc
The modal should be closed only when we click `Close` as far as I know

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -79,18 +104,8 @@ function SettingsPage() {
...
-                  size="big"
-                  disabled={loading}
-                  onClick={() => setShowApiKey(true)}
-                >
-                  Get your key now!
-                </Button>
-              </div>
               <Transition
-                show={showApiKey}
+                show={Boolean(apiKey)}
```
arbulu89, line 108: I think we can entirely remove this `Transition`.
It is not used

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -119,10 +134,42 @@ function SettingsPage() {
...
+                        : 'Key will never expire'}
+                    </div>
+                    <div className="!ml-auto">
+                      <Button
+                        onClick={() => {
+                          window.navigator.clipboard.writeText(apiKey);
+                        }}
+                        type="primary-white"
+                      >
+                        Copy Key
```
arbulu89, line 156: Let's use the same `copy` icon here, as we do in the modal.
The same thing in both at least. What do you think @jagabomb ?

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -119,10 +134,42 @@ function SettingsPage() {
                     )}
                   </code>
                 </div>
+                {apiKey && (
+                  <div className="flex space-x-2 my-4">
+                    <EOS_INFO_OUTLINED size="20" className="mt-2" />
+
+                    <div className="mt-2 text-gray-600 text-sm">
+                      {apiKeyExpiration
+                        ? `Key Will Expire ${format(
```
arbulu89, line 143: `Key will expire ...` in small case

assets/js/pages/SettingsPage/SettingsPage.test.jsx
```diff
@@ -10,10 +10,25 @@ import {
...
 import { softwareUpdatesSettingsFactory } from '@lib/test-utils/factories/softwareUpdatesSettings';
+import { networkClient } from '@lib/network';
+import MockAdapter from 'axios-mock-adapter';
 
 import SettingsPage from './SettingsPage';
 
+const axiosMock = new MockAdapter(networkClient);
+
 describe('Settings Page', () => {
+  afterEach(() => {
```
arbulu89, line 21: It would be nice to have a text to check if the api_key is displayed

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+          hosts.{' '}
+        </span>
+
+        <div className="flex space-x-1">
+          <div className="w-1/5">
+            <Label>Never Expires</Label>
+          </div>
+
+          <Switch
+            selected={!timeFormEnabled}
```
arbulu89, line 90: Why this negation?
I would expect that if the `timeFormEnabled` is true the switch being selected

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+  generatedApiKeyExpiration,
+}) {
+  const timeOptions = availableTimeOptions.map((o) => o.type);
+  const [timeQuantity, setTimeQuantity] = useState(0);
+  const [timeQuantityType, setTimeQuantityType] = useState(timeOptions[0]);
+
+  const [quantityError, setQuantityError] = useState(false);
+  const [timeFormEnabled, setTimeFormEnabled] = useState(true);
+
+  const saveFormState = () => {
```
arbulu89, line 53: I would rename this to `generate\save` or similar, because at the end, it does more than saving the form state.

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,176 @@
...
+          </div>
+        </div>
+        {quantityError && (
+          <span className="my-1 mb-4 text-red-500">
+            {' '}
+            Key expiration value needs to be greater than 0{' '}
+          </span>
+        )}
+        {generatedApiKey && (
+          <div className="flex flex-col my-1 mb-4">
```
arbulu89, line 140: Maybe this `div` could be subcomponent itself, as we use it (or should) the same content here in the modal and in the settings page.
What do you think?

arbulu89 (commented):
Hey @CDimonaco ,
Great job.
I have noted many comments, excluding by now the `Input/Switch` components.
Let me know what do you think

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
arbulu89: Anyway, if Jurgen agrees that they should be different, so be it

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
arbulu89: Okey, forget about this. The logic is a negation in fact

assets/js/pages/SettingsPage/SettingsPage.jsx
CDimonaco: The transition is used to gently display the api key when it's fetched, instead of displaying "directly" in the page, it's a nice visual touch

assets/js/pages/SettingsPage/SettingsPage.jsx
CDimonaco: Nice catch, fixing rn

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: done

assets/js/pages/SettingsPage/SettingsPage.jsx
CDimonaco: done

assets/js/pages/SettingsPage/SettingsPage.jsx
CDimonaco: done

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: done, nice catch, it's more readable

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: done

assets/js/pages/SettingsPage/SettingsPage.jsx
CDimonaco: done

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: done

CDimonaco:
PR ENV
https://2403.prenv.trento.suse.com

abravosuse (commented):
Beautiful, Carmine!

Just two questions/remarks:

1. In the number field in the generate modal, when you write a number you have two sets of icons to increase/decrease the number. Shouldn't there be only one set, the one of the field itself?
image

2. Every time you enter the generate modal, the previous settings are cleared out. This is more a question for @jagabomb . Wouldn't it be better that the last set of settings is displayed instead? For instance, if the user turned on the the non expiration toggle, then, when you open the generate modal, it is still turned on. Or if the user set 1 year as expiration date then, when you open the modal, it is set to 1 year. Just wondering. The way it works now is also fine.

CDimonaco:
> Beautiful, Carmine!
> 
> Just two questions/remarks:
> 
>     1. In the number field in the API Key Settings modal, when you write a number you have two sets of icons to increase/decrease the number. Shouldn't there be only one set, the one of the field itself?
> 
> 
> image
> 
>     2. Every time you enter the API Key Settings modal, the previous settings are cleared out. This is more a question for @jagabomb . Wouldn't it be better that the last set of settings is displayed instead? For instance, if the user turned on the the non expiration toggle, then, when you open the API Key Settings modal, it is still turned on. Or if the user set 1 year as expiration date then, when you open the modal, it is set to 1 year. Just wondering. The way it works now is also fine.

For the first question is a browser compatibility issues, I tested on some different browser and I did not find the issue, the library we are using is one of the best but these glitch could appear

For the second question, is technically feasible to do, but it's not easy because we should have compatibility with the previous key it's almost impossible to come back to the form with the expiration set, for example we know the date of the expiration but we don't know if you did 1 month or 30 days for example, I think it's better to leave the modal as is, resettinng it's state when the user close the modal

This keep consistency and usability imho

abravosuse:
> For the first question is a browser compatibility issues, I tested on some different browser and I did not find the issue, the library we are using is one of the best but these glitch could appear

OK. But, for the record, I'm not using anything out of ordinary (Chrome Version 122.0.6261.112).

> For the second question, is technically feasible to do, but it's not easy because we should have compatibility with the previous key it's almost impossible to come back to the form with the expiration set, for example we know the date of the expiration but we don't know if you did 1 month or 30 days for example, I think it's better to leave the modal as is, resettinng it's state when the user close the modal
> 
> This keep consistency and usability imho

OK. I thought the data in the data entered in the modal was persistent. If it adds complexity to the implementation, then forget about it.

CDimonaco:
> > For the first question is a browser compatibility issues, I tested on some different browser and I did not find the issue, the library we are using is one of the best but these glitch could appear
> 
> OK. But, for the record, I'm not using anything out of ordinary (Chrome Version 122.0.6261.112).
> 
> > For the second question, is technically feasible to do, but it's not easy because we should have compatibility with the previous key it's almost impossible to come back to the form with the expiration set, for example we know the date of the expiration but we don't know if you did 1 month or 30 days for example, I think it's better to leave the modal as is, resettinng it's state when the user close the modal
> > This keep consistency and usability imho
> 
> OK. I thought the data in the data entered in the modal was persistent. If it adds complexity to the implementation, then forget about it.

The modal data is "persistent" across page reloads, but in the end the modal will generate an expiration timestamp.

We could remove the modal "reset" so the user can see the previous settings if they open and close the modal in the same page, but I think it's not really useful and can lead to some confusion about the usage

assets/js/pages/SettingsPage/SettingsPage.test.jsx
CDimonaco: done

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
jagabomb: API Key Default State
Hi @CDimonaco, here is my feedback regarding the spacing, and number input for the default state:

1. Reduce and standardise the spacing between the heading and between line of content (highlighted in red block with pixel height dimensions). Ideally they were 16px in the design.
2. Remove the margin below the close button. There is already a spacing which is provided by the padding of the modal.
3. Resolve this UI addition and subtraction indicator for the numeric field since there is a native component (Chrome) which creates a duplication of the same function.

jagabomb:
> 2\. Every time you enter the API Key Settings modal, the previous settings are cleared out. This is more a question for @jagabomb . Wouldn't it be better that the last set of settings is displayed instead? For instance, if the user turned on the the non expiration toggle, then, when you open the API Key Settings modal, it is still turned on. Or if the user set 1 year as expiration date then, when you open the modal, it is set to 1 year. Just wondering. The way it works now is also fine.

I agree with you here @abravosuse, although I think this Storybook version doesn't really need to showcase the logic like that. I would say that behaviour should be in the AC so that we implement the logic in the product. Unless the team feels there is value to implement logic like that in the Storybook too?

assets/css/input-number.css
```diff
@@ -0,0 +1,170 @@
+.rc-input-number {
```
arbulu89, line 1: I would appreciate if this changes were included in a seperated PR, to find them easily when you look them in PRs.
But well, let's move forward

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,191 @@
...
+  generatedApiKey,
+  generatedApiKeyExpiration,
+}) {
+  const timeOptions = availableTimeOptions.map((o) => o.type);
+  const [timeQuantity, setTimeQuantity] = useState(0);
+  const [timeQuantityType, setTimeQuantityType] = useState(timeOptions[0]);
+  const [keyGenerated, setKeyGenerated] = useState(false);
+
+  const [quantityError, setQuantityError] = useState(false);
+  const [apiKeyNeverExpires, setapiKeyNeverExpires] = useState(false);
```
arbulu89, line 52: `setApiKeyNeverExpires` looks better

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,191 @@
...
+          <span className="my-1 mb-4 text-red-500">
+            {' '}
+            Key expiration value needs to be greater than 0{' '}
+          </span>
+        )}
+        {generatedApiKey && keyGenerated && !loading && (
+          <div className="flex flex-col my-1 mb-4">
+            <div className="flex space-x-2">
+              <div className="w-full break-words p-2 pr-2 rounded-lg bg-white border-gray-300 border">
+                <code> {truncate(generatedApiKey, { length: 65 })} </code>
```
arbulu89, line 155: The space after and before code tags are intentional?

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,191 @@
...
+              >
+                <EOS_CONTENT_COPY role="button" size="25" />
+              </button>
+            </div>
+            <div className="flex space-x-2">
+              <EOS_INFO_OUTLINED size="20" className="mt-2" />
+
+              <div className="mt-2 text-gray-600 text-sm">
+                {generatedApiKeyExpiration
+                  ? `Key Will Expire ${format(
```
arbulu89, line 172: `Key will expire...` all words small case except the first one

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.stories.jsx
```diff
@@ -0,0 +1,74 @@
...
+      description: 'Whether the settings are loading or submitting',
+      control: {
+        type: 'boolean',
+      },
+    },
+    generatedApiKeyExpiration: {
+      description:
+        'The new generated api key expiration expressed in ISO8601 Timestamp',
+    },
+    generatedApiKey: {
```
arbulu89, line 26: you should include the `onGenerated` here with some action. this way, you don't really need to mock it in the args, as storybook will add the action in its record:
```
onGenerate: {
   action: 'Generate key',
   description: 'New key is generated',
}
```

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.stories.jsx
```diff
@@ -0,0 +1,74 @@
...
+      description: 'The new generated api key',
+    },
+  },
+  args: {
+    loading: false,
+    open: false,
+    generatedApiKeyExpiration: new Date().toISOString(),
+    generatedApiKey:
+      'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJ0cmVudG9fYXBpX2tleSIsImV4cCI6MTcxMjE1MzE3MCwiaWF0IjoxNzA5NzMzOTcxLCJpc3MiOiJodHRwczovL2dpdGh1Yi5jb20vdHJlbnRvLXByb2plY3Qvd2ViIiwianRpIjoiYmZmMjA0YjUtMzJmMS00YmVlLThiMGItY2IxZGQwNTlmNGRjIiwibmJmIjoxNzA5NzMzOTcxLCJ0eXAiOiJCZWFyZXIifQ.0Lz0MwZaFpIGbSohnkiJ6AN5FFb5Vg5ZVhqM3fdUf3M',
+    onSave: ({ apiKeyExpiration }) =>
```
arbulu89, line 36: With the previous change, this shouldn't be needed...
PD: but in any case, it should `onGenerated`

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.test.jsx
```diff
@@ -0,0 +1,245 @@
...
+      await user.type(screen.getByRole('spinbutton'), '20');
+
+      await user.click(screen.getByRole('button', { name: 'Generate' }));
+
+      expect(
+        screen.queryByText('Key expiration value needs to be greater than 0')
+      ).not.toBeInTheDocument();
+    });
+
+    it('should return on onGenerate the correct expiration date when months are selected with a valid quantity', async () => {
```
arbulu89, line 86: I think you could wrap the next 3 tests in one with `it.each`, but it is fine they way it is as well

assets/js/pages/SettingsPage/SettingsPage.jsx
arbulu89: apparently not hehe

arbulu89 (approved):
Excellent work @CDimonaco 
Codewise, ready to merge from my side (with some small nitpicks XD)

jagabomb:
API Key Generated State

1. I wonder if we should make this sentence case instead of title case, "**Key will expires 12 Mar 2024**"? What are you thoughts @abravosuse?
2. Can we have a hover state for the "**Copy to Clipboard**" button and after you click it show a tooltip that says "**Copied to clipboard**". I have included button states for reference.

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
jagabomb: If it helps us have consistency in the layout and implementation then I am happy for it to be a subcomponent.

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: For the point 1 and 2, I will standardize the space, for the last point, this is a custom component from the same component library we use for other inputs, if we fallback to the native component, we should design all the style and create a custom component to match all the others input design.

Why this component is bad? We could of course fallback to the native component but we need to create a custom component which I think it's a big effort and out the scope of this pr.

I tested on multiple browser and I cannot reproduce the "dual" indicators on the number component, but again, if we need to write custom logic on top of the custom or native component it's not an easy task and out of the scope of this pr.

cc @dottorblaster maybe my difficulty estimation is wrong, wdyt?

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -0,0 +1,191 @@
...
+  return (
+    <Modal
+      title="API Key Settings"
+      className="!w-3/4 !max-w-3xl"
+      open={open}
+      onClose={onClose}
+    >
+      <div className="flex flex-col my-5">
+        <span className="my-2 mb-4 text-gray-500">
+          {' '}
```
EMaksy, line 90: Just curious, you had to add this empty spaces because otherwise CSS was not formatted properly, right?

CDimonaco: It's a prettier thingy :D

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
jagabomb: @CDimonaco I agree, if the effort is too large for the numeric input to be improved then we can re-look at this in another PR. I just wanted to note the input behaviour since Alberto also raised it too.

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: done

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: done, I was copying right from the mockups xD

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.stories.jsx
CDimonaco: done

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.stories.jsx
CDimonaco: done with the action

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: removed

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
dottorblaster: Point 1 and 2 are perfectly addressable in this PR, when it comes to point 3 I didn't quite get your feedback @jagabomb, my advice is to merge and proceed incrementally to do this kind of work on the design system, which we couldn't foresee.

The alternative is to adopt a text input for the moment, but I would suggest otherwise since with a number input we can ensure nothing strange is submitted.

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: If it's a subcomponent, we will have the same copy button?

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.stories.jsx
EMaksy: Also, onClose should be added to cover all props right?

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
CDimonaco: Space standardized, you can check in storybook but this is the screenshot, there are also no more margin under the button, only the modal margins.

image

assets/js/common/Input/InputNumber.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import RcInputNumber from 'rc-input-number';
+
+import classNames from 'classnames';
+
+function InputNumber({
```
EMaksy, line 6: Hey man, I'm just curious why you went for a component wrapper instead of directly using it?

EMaksy (approved):
Hey man great work, what a big beautiful pr. 
I wrote some comments and questions, but the code looks great.
For me LGTM!

assets/js/common/Input/InputNumber.jsx
CDimonaco: for injecting classes in a more atomic way

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
jagabomb: @dottorblaster It is easier to see the issue with the duplicated numeric addition and subtraction (arrow up and arrow down) buttons below. But as I said before, I am happy to defer this to another PR and revisit it when we have time. 😊 
Screenshot 2024-03-13 at 09 34 54

CDimonaco:
Merging as is

Creating 2 tech debts, one for the copy button, the other one for the numeric input.

PR #2416, opened 2024-03-13
Add Redux Sagas for getting SUSE Manager software updates

jamie-suse:
# Description

Add Redux Sagas to allow retrieval of software updates for a given host.

This PR also includes Selectors to derive the number of relevant patches and upgradable packages from the state.

## How was this tested?

Added unit tests.


assets/js/state/sagas/softwareUpdates.test.js
```diff
@@ -0,0 +1,88 @@
...
+      const axiosMock = new MockAdapter(networkClient);
+      const hostId = faker.string.uuid();
+      const responses = [
+        { status: 404, body: { message: '404 Not found' } },
+        { status: 500, body: { message: 'java.lang.NullPointerException' } },
+      ];
+
+      /* eslint-disable no-await-in-loop */
+      /* eslint-disable no-restricted-syntax */
+      for (const { status, body } of responses) {
```
jamie-suse, line 71: If there's a way to refactor this loop, it would be welcome.
The main problem is looping with an `await`. I tried with `responses.forEach()` and also `Promise.all()` and neither worked. However ESLint does not like the presence of the `await` in the loop...

assets/js/state/softwareUpdates.js
```diff
@@ -0,0 +1,44 @@
+import { createAction, createSlice } from '@reduxjs/toolkit';
+
+const initialState = {
+  loading: false,
+  softwareUpdates: [],
```
jamie-suse, line 5: The structure of `softwareUpdates`is an array of 
```javascript
softwareUpdates: [
    { hostId: 'example_id', softwareUpdates: [ ...sumaPatches ]},
    ...
]
```

What do you think?

assets/js/state/sagas/softwareUpdates.js
```diff
@@ -0,0 +1,28 @@
...
+    const response = yield call(getSoftwareUpdates, hostId);
+    yield put(setSoftwareUpdates({ hostId, softwareUpdates: response.data }));
+  } catch (error) {
+    yield put(setEmptySoftwareUpdates());
+    const errors = get(error, ['response', 'data'], []);
+    yield put(setSoftwareUpdatesErrors(errors));
+  }
+}
+
+export function* watchSoftwareUpdates() {
```
nelsonkopliku, line 26: Not sure whether you intend to to this in a followup, but just for clarity
`watchSoftwareUpdates` needs to be registered in `assets/js/state/sagas/index.js`

assets/js/state/softwareUpdates.js
nelsonkopliku: we could consider an alternative where we keep a dictionary like `hostId -> discoveredSoftwareUpdates`

```
softwareUpdates: {
  "host_id": {relevant_patches: [...], upgradable_packages: [...]}
}
```
it might be helpful to a direct access by host id.

assets/js/state/sagas/softwareUpdates.js
```diff
@@ -0,0 +1,28 @@
...
+} from '@state/softwareUpdates';
+
+export function* fetchSoftwareUpdates({ payload: { hostId } }) {
+  yield put(startLoadingSoftwareUpdates());
+
+  try {
+    const response = yield call(getSoftwareUpdates, hostId);
+    yield put(setSoftwareUpdates({ hostId, softwareUpdates: response.data }));
+  } catch (error) {
+    yield put(setEmptySoftwareUpdates());
```
nelsonkopliku, line 20: I guess we should continue keeping track of the host id also on failure, so empty only for the specific host. Otherwise we end up in emptying the whole state, for all the hosts.

assets/js/state/softwareUpdates.js
```diff
@@ -0,0 +1,44 @@
...
+    startLoadingSoftwareUpdates: (state) => {
+      state.loading = true;
+    },
+    setSoftwareUpdates: (state, { payload: { hostId, softwareUpdates } }) => {
+      state.loading = false;
+      state.softwareUpdates = state.softwareUpdates
+        .filter((su) => su.hostId !== hostId)
+        .concat({ hostId, softwareUpdates });
+    },
+    setEmptySoftwareUpdates: (state) => {
```
nelsonkopliku, line 22: related to the comment in the saga: in this way we are clearing up the whole state, regardless of the host.

We either keep in this state the available software updates for one host at a time or if we intend to keep the available software updates for all the hosts, we might want to have `setEmptySoftwareUpdates` empty only the host relevant data.

assets/js/state/sagas/softwareUpdates.test.js
```diff
@@ -0,0 +1,88 @@
...
+} from '@state/softwareUpdates';
+
+import { fetchSoftwareUpdates } from './softwareUpdates';
+
+describe('Software Updates saga', () => {
+  describe('Fetching Software Updates', () => {
+    it('should successfully fetch software updates', async () => {
+      const axiosMock = new MockAdapter(networkClient);
+      const hostId = faker.string.uuid();
+      const successfulResponse = [
```
nelsonkopliku, line 22: just to keep track: a successful response would have relevant patches and upgradable packages splitted like:
```js
{
  relevant_patches: [],
  upgradable_packages: [],
}
```

assets/js/state/sagas/softwareUpdates.test.js
```diff
@@ -0,0 +1,88 @@
...
+        {
+          date: '2023-05-18',
+          advisory_name: 'SUSE-15-SP4-2023-2245',
+          advisory_type: 'Bug Fix Advisory',
```
nelsonkopliku, line 26: If I recall it correctly, we do a little transformation in the backend for this field only.
Advisory type would be one of:
- `security_advisory`
- `bugfix`
- `enhancement`

but needs to be double checked against the actual api response, when ready.

assets/js/state/softwareUpdates.test.js
```diff
@@ -0,0 +1,174 @@
...
+      loading: false,
+      softwareUpdates: [
+        { hostId: 'host1', softwareUpdates: [] },
+        { hostId: 'host2', softwareUpdates: newSoftwareUpdates },
+      ],
+      errors: [],
+    });
+  });
+
+  it('should empty software updates', () => {
```
nelsonkopliku, line 101: ah cool, in this test it should be more evident the opportunity of having a consistent state/behavior relation. As mentioned either one hosts' available software updates or all hosts' software updates, with related actions.

nelsonkopliku (commented):
Good job, Jamie. 
It's a good starting point, however I think we need to iterate again in order to make things click.
There might be edge cases I've overlooked, but let's make it work first, then we polish what needs fine tuning :wink: 

Hope the comments are somewhat useful. Happy to help.

assets/js/state/sagas/softwareUpdates.test.js
jamie-suse: Nice! Yeah `test.each()` solves the problem quite nicely! Merci!

assets/js/state/sagas/softwareUpdates.test.js
```diff
@@ -0,0 +1,90 @@
...
+            date: '2023-05-18',
+            advisory_name: 'SUSE-15-SP4-2023-2245',
+            advisory_type: 'bugfix',
+            advisory_status: 'stable',
+            id: 2192,
+            advisory_synopsis: 'Recommended update for libzypp, zypper',
+            update_date: '2023-05-18',
+          },
+        ],
+        software_updates: [
```
nelsonkopliku, line 34:
```suggestion
        upgradable_packages: [
```

nelsonkopliku (commented):
Cool, good enhancements!

I guess we just need to change `software_updates` to `upgradable_packages` :stuck_out_tongue_closed_eyes: 

What I am actually missing is the clicking part on error occurrence, cause I am not sure yet what the backed would return.  https://github.com/trento-project/web/pull/2415#discussion_r1523536747

I am afraid we need to keep this a bit on hold.

assets/js/state/softwareUpdates.js
```diff
@@ -0,0 +1,56 @@
...
+        },
+      };
+    },
+    setEmptySoftwareUpdates: (state, { payload: { hostId } }) => {
+      state.loading = false;
+      state.softwareUpdates = Object.entries(state.softwareUpdates).reduce(
+        (newState, [id, updates]) =>
+          id !== hostId ? { ...newState, [id]: updates } : newState,
+        {}
+      );
```
dottorblaster, line 36:
```suggestion
     state.softwareUpdates = {
        ...state.softwareUpdates,
        [hostId]: {},
      };
```

What do you think?

dottorblaster (approved):
No blockers from my side, LGTM

nelsonkopliku (approved):
I guess we can give this a go.
Error handling should catch 404s and 422s.

assets/js/state/softwareUpdates.js
jamie-suse: Sure, I think it's good. It got me thinking though; what do we think about clearing setting `[hostId]: { relevant_patches: [], upgradable_packages: [] }` as opposed to `[hostId]: {}`?

jamie-suse: On second thought, it might be bad as the selector would report `relevant_patches`/`upgradable_packages == 0` instead of `undefined`. But as @nelsonkopliku mentioned, error handling should handle this before it even hits the selector :thinking:

dottorblaster: No strong opinion, let's ship whatever you prefer :+1:

jamie-suse: I like having a semantic difference between `undefined` vs `0`. Sure, the error handling should handle this but it still gives a more semantically correct representation of the state of the system.

nelsonkopliku (approved):
LGTM

PR #2429, opened 2024-03-18
Set docker as install method in `env` CI job

arbulu89:
# Description

Set `install_method: docker` explicitily, as `rpm` is the default option and e2e depends on docker containers.
Related to: https://github.com/trento-project/ansible/pull/30


dottorblaster (approved):
Fair enough

PR #2430, opened 2024-03-18
Display SUMA updates in Host Details page

jamie-suse:
# Description

This change mounts the `AvailableSoftwareUpdates` component in the _Host Details_ page, allowing users to see the available updates and patches for a host, as instructed by SUMA.

## How was this tested?

Added React component tests to validate that the _Available Software Updates_ component is correctly being displayed in the _Host Details_ page.

### DIFF
--- a/assets/js/pages/HostDetailsPage/HostDetails.jsx
+++ b/assets/js/pages/HostDetailsPage/HostDetails.jsx
@@ -66,6 +66,7 @@ function HostDetails({
   relevantPatches,
   upgradablePackages,
   softwareUpdatesLoading,
+  softwareUpdatesTooltip,
   cleanUpHost,
   requestHostChecksExecution,
   navigate,
@@ -223,11 +224,7 @@ function HostDetails({
           className="mx-0 my-4"
           relevantPatches={relevantPatches}
           upgradablePackages={upgradablePackages}
-          tooltip={
-            relevantPatches === undefined && upgradablePackages === undefined
-              ? 'SUSE Manager is not available'
-              : undefined
-          }
+          tooltip={softwareUpdatesTooltip}
           loading={softwareUpdatesLoading}
         />
         <ChartsFeatureWrapper chartsEnabled={chartsEnabled}>
--- a/assets/js/pages/HostDetailsPage/HostDetails.jsx
+++ b/assets/js/pages/HostDetailsPage/HostDetails.jsx
@@ -4,6 +4,7 @@ import classNames from 'classnames';
 import { EOS_CLEAR_ALL, EOS_PLAY_CIRCLE, EOS_SETTINGS } from 'eos-icons-react';
 
 import { agentVersionWarning } from '@lib/agent';
+import { getFromConfig } from '@lib/config';
 
 import BackButton from '@common/BackButton';
 import Button from '@common/Button';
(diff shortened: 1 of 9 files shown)


assets/js/pages/HostDetailsPage/HostDetailsPage.jsx
```diff
@@ -92,6 +103,9 @@ function HostDetailsPage() {
       selectedChecks={hostSelectedChecks}
       slesSubscriptions={host.sles_subscriptions}
       catalog={catalog}
+      relevantPatches={numRelevantPatches}
+      upgradablePackages={numUpgradablePackages}
+      softwareUpdatesLoading={softwareUpdatesLoading}
```
dottorblaster, line 108: Maybe we wanna add a tooltip text as a prop here? I really don't know, and I really don't wanna block this PR just for this nit

jamie-suse: Done, do you think it's worth testing the tooltip behaviour from `HostDetails.test.jsx`, as it's already being tested in `AvailableSoftwareUpdates.test.jsx`... :thinking:

dottorblaster: No, I think we can iterate over that, for example in the future we can have a selector to determine tooltip's text

assets/js/pages/HostDetailsPage/HostDetailsPage.jsx
```diff
@@ -66,6 +78,7 @@ function HostDetailsPage() {
     getExportersStatus();
     refreshCatalog();
     dispatch(updateLastExecution(hostID));
+    dispatch(fetchSoftwareUpdates(hostID));
```
nelsonkopliku, line 81: this needs to be like the following
```suggestion
    dispatch(fetchSoftwareUpdates({hostId: hostID}))
```
unless we change that action

jamie-suse: Hmm just had a look through our codebase and we use a mix of `hostID` and `hostId` :thinking: I guess `hostID` is *more* popular though, so I'll change the action

nelsonkopliku: up to you, but the thing is we need to pass an object, not the plain id (whatever is called).
unless we do transformation in the action

```
export const createAction = createAction(
  FETCH_SOFTWARE_UPDATES,
  (hostId) => ({ payload: { hostId } })
);
```

assets/js/pages/HostDetailsPage/HostDetails.jsx
```diff
@@ -215,6 +220,13 @@ function HostDetails({
             />
           </div>
         </div>
+        <AvailableSoftwareUpdates
```
nelsonkopliku, line 223: let's gate this behind a check against `getFromConfig('suseManagerEnabled')`

assets/js/pages/HostDetailsPage/HostDetailsPage.jsx
jamie-suse: Ah I see what you mean. I thought I was using the action-creator in `@state/softwareUpdates`. Instead, I was calling the Saga directly :flushed:

dottorblaster (approved):

PR #2436, opened 2024-03-19
Clear software updates discovered health on host deregistration

nelsonkopliku:
# Description

This PR makes sure a `SoftwareUpdatesDiscoveryCleared` event is emitted when deregistering a host.

I also considered the alternative to hook in the deregistration process manager by dispatching a `ClearSoftwareUpdatesDiscovery` command when appropriate, but that would have meant possibily emitting also a `HostHealthChanged` event which is irrelevant for the usecase.

Gotta double check if rollup needs also any intervention.

## How was this tested?

Automated tests.

@arbulu89 you might have good points here, that's why I am asking your review as well.

### DIFF
--- a/lib/trento/hosts/host.ex
+++ b/lib/trento/hosts/host.ex
@@ -410,6 +410,9 @@ defmodule Trento.Hosts.Host do
         %DeregisterHost{deregistered_at: deregistered_at}
       ) do
     [
+      %SoftwareUpdatesDiscoveryCleared{
+        host_id: host_id
+      },
       %HostDeregistered{
         host_id: host_id,
         deregistered_at: deregistered_at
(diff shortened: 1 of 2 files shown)


arbulu89 (commented):
I'm out of context.
But, why do you need this?
I mean, if the host is deregistered, do we care if the setting are cleared or not?
I guess that when it is restored back again, events related to sofware events will come again

nelsonkopliku:
> I'm out of context. But, why do you need this? I mean, if the host is deregistered, do we care if the setting are cleared or not? I guess that when it is restored back again, events related to sofware events will come again

the point is that when it comes back, we want to start form a clean sheet, with regards to possibly previously discovered software updates, and let the next discovery tick update with fresh data.

nelsonkopliku:
btw @arbulu89 we can't give for granted that after a host deregistration/restoration SUMA integration will still be active, meaning that fresh discoveries might not come in later, so if we don't clean up it might happen we have stale state that affects the overall host state.

jamie-suse (approved):

PR #2445, opened 2024-03-20
Split sap system database aggregates

arbulu89:
# Description

Split the sap system and database aggregate logic.
Basically it moves the logic related to the database to a new aggregate.
The related adjustments to make this initially work are done as well.

I will move the projections/read models/events, etc in a 2nd PR

## How was this tested?

Tests has been moved. I have adapted the e2e tests that now work. Some of them still fail as we don't have any feature parity yet.


lib/trento/databases/lifespan.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Trento.Databases.Lifespan do
```
nelsonkopliku, line 1: This is currently just using the DefaultLifespan. Is it going to be changed with regards to rollup event for databases?

CDimonaco: Yeah, rollup will come in a separate pr later on

CDimonaco (approved):

EMaksy (approved):
Man splitting the sap system aggregate is quite a lot, thanks for taking care!
I did not find anything wrong in it. LGTM!

PR #2452, opened 2024-03-21
Move database events

arbulu89:
# Description

Move the `database` events to its own folder, and create the `databases` context file.
Simple easy file movements.

I don't take care of `upcasting` the events to change `sap_system_id` to `database_id`. We will take care of this when we handle the old events, as we might even ditch them.
I have included the `legacy` events by now, because we need them. Again, we might remove them when we take care of old events.

If you still see some `sap_system_id` fields in the database code, don't worry. I will remove them when I modify the projections code.

Next to come:
- Projectors
- Controllers/views/frontend

## How was this tested?

Tests passing as before

### DIFF
--- a/lib/trento/databases/events/legacy/database_deregistered.ex
+++ b/lib/trento/databases/events/legacy/database_deregistered.ex
@@ -5,7 +5,7 @@ defmodule Trento.SapSystems.Events.DatabaseDeregistered do
 
   use Trento.Support.Event
 
-  defevent do
+  defevent superseded_by: Trento.Databases.Events.DatabaseDeregistered do
     field :sap_system_id, Ecto.UUID
     field :deregistered_at, :utc_datetime_usec
   end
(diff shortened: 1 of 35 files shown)


lib/trento/databases/events/legacy/database_deregistered.ex
```diff
@@ -5,7 +5,7 @@ defmodule Trento.SapSystems.Events.DatabaseDeregistered do
 
   use Trento.Support.Event
 
-  defevent do
+  defevent superseded_by: Trento.Databases.Events.DatabaseDeregistered do
```
EMaksy, line 8: Why do we need to add superseded_by again?

arbulu89: Because the name of the module is changed from `Trento.SapSystem.Events` to `Trento.Databases.Events`

EMaksy (approved):
Hey man, lgtm!
I could not find any issues.

CDimonaco (approved):

PR #2454, opened 2024-03-21
Move database projections

arbulu89:
# Description

Following the database/sap system split. Split the projections and read models code to the database aggregate.

The unique new piece of code that this adds is the migration to rename `sap_system_id` field to `database_id` in the database instance read model.

You might still see some `sap_system_id` usages in the database projector, but they are part of the views, which I will change in the next PR.

Still pending:
- Controllers/views/frontend

## How was this tested?

Previous tests passing

### DIFF
--- a/lib/trento/databases.ex
+++ b/lib/trento/databases.ex
@@ -5,7 +5,7 @@ defmodule Trento.Databases do
 
   import Ecto.Query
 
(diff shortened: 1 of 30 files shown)


EMaksy (approved):
LGTM!

PR #2457, opened 2024-03-21
Make `Software Updates` service handle no SUMA settings saved

jamie-suse:
# Description

This changes allows the `Software Updates` service to return an error with a response status of `422` and provides the message _"SUSE Manager settings not configured."_.

## How was this tested?

Updated existing unit tests to validate new behaviour.

### DIFF
--- a/lib/trento/software_updates.ex
+++ b/lib/trento/software_updates.ex
@@ -108,9 +108,8 @@ defmodule Trento.SoftwareUpdates do
 
   @spec get_software_updates(Ecto.UUID.t()) ::
           {:ok, map()}
-          | {:error, :unable_to_get_software_updates,
-             :settings_not_configured
-             | :system_id_not_found
+          | {:error, :settings_not_configured,
+             :system_id_not_found
              | :not_found
              | :fqdn_not_found
              | :error_getting_patches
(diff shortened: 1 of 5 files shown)


nelsonkopliku (approved):
LGTM.

dottorblaster (changes requested):
Great job! Just a couple things then we can merge :+1:

nelsonkopliku:
hey @dottorblaster, Jamie and I paired a bit on this, so let me share some rationale:
there was an initial requirement of differentiating in the UI _SUMA not available_ vs _host not found on SUMA_, so we thought of contextualizing the failure as `:unable_to_get_software_updates` in order to support differentiating the _reason why_ the operation of getting software updates for a host failed.

Since it seems (@jamie-suse could you confirm, please?) that this differentiation is not strictly needed now (and it can be done in many ways). I believe we could defer any further fine grained error handling for now.

That said, we surely can just add `:settings_not_configured` as a possible failure, but since `:settings_not_configured` is already handled somewhere else in the fallback controller (operations on the credentials resource), without some extra contextualization we'd get on the _get software updates for a host_ endpoint a `404 The requested resource cannot be found.` both in case settings were not configured and if the requested host id is not in trento. 
In all other cases a 422 with a tiny more specific error message is returned.
And I'd be fine with it.

In any case I guess we have room to improve error handling, and this relates also to the discussion you raised some days ago that ended up in the possibility to have contextual fallback controller. But let's of course defer that :smile:

dottorblaster (approved):
@nelsonkopliku all clear, thank you! Of course the status quo always has room for improvement, let's keep this on the radar, thank you so much!

PR #2469, opened 2024-03-25
Database rollup

arbulu89:
# Description

Implement the database rollup logic. Basically, it copies what we do with the sap system rollup.
Nothing new

## How was this tested?

UT added (i have done some manual testing as well)

### DIFF
--- a/lib/trento/databases/database.ex
+++ b/lib/trento/databases/database.ex
@@ -22,7 +22,8 @@ defmodule Trento.Databases.Database do
   alias Trento.Databases.Commands.{
     DeregisterDatabaseInstance,
     MarkDatabaseInstanceAbsent,
-    RegisterDatabaseInstance
+    RegisterDatabaseInstance,
+    RollUpDatabase
   }
 
   alias Trento.Databases.Events.{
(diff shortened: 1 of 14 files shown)


lib/trento/databases/database.ex
```diff
@@ -602,6 +635,12 @@ defmodule Trento.Databases.Database do
 
   defp maybe_emit_database_deregistered_event(_, _), do: nil
 
+  defp maybe_emit_database_tombstoned_event(%Database{database_id: database_id, instances: []}) do
```
arbulu89, line 638: This is my unique doubt point.
I initially thought to emit this event once the database is deregistered, but it created some "strange" scenarios, when after that, the secondary instances is deregistered, tombstoning yet again, and rolled up stream.
For this reason, I prefer just to tombstone and rollup once all instances are gone (which is a normal thing to happen I think)

CDimonaco: I think this is the correct way to do it, on main we have this scenario, tombstoning only when we don't have instances (0 instances on both applications and database), when splitting aggregates this behavior could just be replicated on both aggregates

CDimonaco (approved):

EMaksy (approved):
Great work!

PR #2472, opened 2024-03-26
Move database controller views

arbulu89:
# Description

Move the database related controllers and views code to its respective files.
In next PR:
- Adapt frontend

## How was this tested?

Previously created UTs moved from file to file

PR #2473, opened 2024-03-26
Adapt frontend to use the new database_id field

arbulu89:
# Description

Adapt the frontend code to use the new `database_id`.
The unique new code is the `getInstanceID` function, used in places where both application and database instances are coming together.

Depends on: https://github.com/trento-project/web/pull/2472

## How was this tested?

Previous tests are passing


EMaksy (approved):
LGTM!

PR #2475, opened 2024-03-26
Dynamic catalog filters

nelsonkopliku:
# Description

This PR allows filters in the catalog page to be populated based on the whole catalog content, while the actual list below is based on a filteredCatalog.

This requires making a differentiation in the intent: loading the whole catalog vs getting a filtered version of it, and I ended up adding this tiny seed `filteredCatalog: true` in the `updateCatalog` action.

However, I am not totally happy with it.
I attempted polishing catalog state and saga even further, but that would require rationalizing everything catalog related and updating many other places where the catalog is queried, like the checks selection, checks result page, etc... then I preferred keeping the change at minimum and defer further refactoring.

## How was this tested?

Automated tests

### DIFF
--- a/assets/js/state/sagas/catalog.test.js
+++ b/assets/js/state/sagas/catalog.test.js
@@ -3,14 +3,14 @@ import { catalogCheckFactory } from '@lib/test-utils/factories';
 import { recordSaga } from '@lib/test-utils';
 
 import { networkClient } from '@lib/network';
-import { updateCatalog } from './catalog';
-
 import {
   setCatalogLoading,
   setCatalogData,
   setCatalogError,
   setFilteredCatalog,
-} from '../catalog';
+} from '@state/catalog';
+
+import { updateCatalog } from './catalog';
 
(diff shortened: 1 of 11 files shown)


assets/js/state/sagas/catalog.test.js
```diff
@@ -5,7 +5,12 @@ import { recordSaga } from '@lib/test-utils';
 import { networkClient } from '@lib/network';
 import { updateCatalog } from './catalog';
 
-import { setCatalogLoading, setCatalogData, setCatalogError } from '../catalog';
+import {
+  setCatalogLoading,
+  setCatalogData,
+  setCatalogError,
+  setFilteredCatalog,
+} from '../catalog';
```
jamie-suse, line 13:
```suggestion
} from '@state/catalog';
```

jamie-suse (approved):
tiny one-line fix for the imports (that you didn't even introduce :stuck_out_tongue:) but otherwise LGTM

PR #2476, opened 2024-03-26
Add CopyButton component

CDimonaco:
# Description

This PR adds the copy to clipboard component.
Recording 2024-03-26 at 15 28 23

## How was this tested?

Manual testing, the component relies on an external library.

### DIFF
--- a/assets/css/tooltip.scss
+++ b/assets/css/tooltip.scss
@@ -53,6 +53,7 @@ as mentioned in issue https://github.com/react-component/tooltip/issues/445
   */
   @apply block absolute visible opacity-100;
 
+  z-index: 99999;
   &.rc-tooltip.rc-tooltip-fade-enter,
   &.rc-tooltip.rc-tooltip-fade-appear {
     animation-duration: 0.2s;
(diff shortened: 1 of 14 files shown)


arbulu89 (commented):
Nice!!
@CDimonaco we still have the other copy button in the settings page, which most probably has the same error this had.
We should take care of that.
Replacing to use this component or with something else.
Or do you plan to work on that in other PR? in that case I would approve this

jagabomb (commented):
@CDimonaco This looks good to me. I do have a small request... Is it possible to keep the tooltip (Copied to clipboard) visible, after the button is selected and when the mouse moves away from the button's hover area?

CDimonaco:
> Nice!! @CDimonaco we still have the other copy button in the settings page, which most probably has the same error this had. We should take care of that. Replacing to use this component or with something else. Or do you plan to work on that in other PR? in that case I would approve this

Copy button in settings is now fixed, the button is still the same because we need to thing a little bit more the change to ui in the settings page

jagabomb (approved):
LGTM!

CDimonaco:
> Code wise everything looks great! Whether we use the same copy icon in all the pages goes a bit out of scope of my review. It is super easy to replace, so we can even do it in a next PR

In the end we changed also in the settings, rn the style for apikey and the copy button is unified
image

PR #2480, opened 2024-03-27
Update frontend to correlate SAP systems and databases properly

EMaksy:
# Description

This pr aims to:
- Update the frontend to use database ID to create the links instead of sap system ID
- Update the health_summary_service.ex and health_overview_view.ex to use the new database_id and database healt
- Updated sapSystems factory to use the new databaseId

## How was this tested: 
Updated existing test in front end and backend

### DIFF
--- a/assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
+++ b/assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
@@ -46,8 +46,8 @@ describe('SapSystemsOverviews component', () => {
         db_host: dbAddress,
         application_instances: applicationInstances,
         database_instances: databaseInstances,
+        database_id: databaseID,
       } = sapSystem;
-
       renderWithRouter(
         <SapSystemsOverview
           sapSystems={[sapSystem]}
@@ -69,7 +69,7 @@ describe('SapSystemsOverviews component', () => {
       );
       expect(mainRow.querySelector('td:nth-child(3) > a')).toHaveAttribute(
         'href',
-        `/databases/${sapSystemID}`
+        `/databases/${databaseID}`
       );
       expect(mainRow.querySelector('td:nth-child(4)')).toHaveTextContent(
         tenant
@@ -107,6 +107,7 @@ describe('SapSystemsOverviews component', () => {
 
       const enrichedDatabaseInstances = databaseInstances.map((instance) => {
         const host = hostFactory.build({ id: instance.host_id });
+
         return {
           ...instance,
           host: {
@@ -118,7 +119,6 @@ describe('SapSystemsOverviews component', () => {
           },
         };
       });
-
       renderWithRouter(
         <SapSystemsOverview
           sapSystems={[sapSystem]}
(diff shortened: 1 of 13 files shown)


test/trento/sap_systems/services/health_summary_service_test.exs
```diff
@@ -34,6 +34,8 @@ defmodule Trento.SapSystems.Services.HealthSummaryServiceTest do
         host_id: a_host_id
       )
 
+      insert(:database, id: database_id)
```
arbulu89, line 37: I would change the order of things.
Put this insert first, after the host read model, and get the database id from it.
Use this add in the `insert(:sap_system)` and the other `database_instance_without_host`.
This is the correct order of things in real life as well

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
```diff
@@ -46,8 +46,8 @@ describe('SapSystemsOverviews component', () => {
         db_host: dbAddress,
         application_instances: applicationInstances,
         database_instances: databaseInstances,
+        database_id: databaseID,
```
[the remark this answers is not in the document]
EMaksy: So to what do we want to stick for the future :)

lib/trento_web/openapi/v1/schema/database.ex
```diff
@@ -55,6 +49,12 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Database do
           description: "System Replication Status"
         },
         health: ResourceHealth,
+        absent_at: %Schema{
```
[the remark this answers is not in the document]
EMaksy: You already added  the [database_id in line 24](https://github.com/trento-project/web/blob/9e116c22e2db1a091c8e5654cf0c3dc4154b8db6/lib/trento_web/openapi/v1/schema/database.ex#L24)

arbulu89 (approved):
Alright! Thanks @EMaksy 
A couple of nitpicks, ready to merge from my side

test/trento_web/controllers/v1/sap_system_controller_test.exs
```diff
@@ -15,7 +15,8 @@ defmodule TrentoWeb.V1.SapSystemControllerTest do
 
   describe "list" do
     test "should list all sap_systems", %{conn: conn} do
-      %{id: sap_system_id, database_id: database_id} = build(:sap_system)
+      %{id: sap_system_id, database_id: database_id} = insert(:sap_system)
```
CDimonaco, line 18: Could we add a real database and provide the real database_id when inserting the sap system?

CDimonaco (approved):
LGTM, just minor comments + what @arbulu89 already pointed out

In general try to use a real database in the tests, and then provide the real database_id to the sap system factory.

We don't have an explicit foreign key on the database, so as long is a uuid when can add whatever we want ad database_id, but try to keep things consistent and "real". we don't have foreign key constraint due to the eventual consistence of our application and this can be an issue, but we should always remember when testing to get the things as close to real scenario

test/trento_web/controllers/v1/sap_system_controller_test.exs
arbulu89: @EMaksy for more info, initially @CDimonaco created a refence between this `database_id` and the `id` on the database read model.
This was to assure that there couldn't be any sap system without an existing database.
Due some reasons, we removed that constraint initially, for example, to make the tests a bit more flexible.
But if we implement them properly, we could at some point get back that reference, and the tests should continue passing

PR #2485, opened 2024-04-02
Handle old database events

arbulu89:
# Description

Handle old SAP system events happening in the new database aggregate.
As the aggregate ID belongs to the database, old events that now are not used must be handled.
In some of them, we just need to upcast the event, and replace `sap_system_id` by `database_id`.
In the rollup case, we need to slightly modify the content and send a new event (`DatabaseRolledUp`), as the data belonged to the database in the first place.

I used a macro to avoid a lot of boilerplate code in some upcast methods.

## How was this tested?

The code changes are tested with UT.
I have tested the general usage with manual testing.
We can maybe try to have some e2e testing for this.


lib/trento/sap_systems/sap_system.ex
```diff
@@ -88,11 +88,18 @@ defmodule Trento.SapSystems.SapSystem do
     field :deregistered_at, :utc_datetime_usec, default: nil
 
     embeds_many :instances, Instance
+
+    # The current steam is composed by legacy events, coming from database events
```
arbulu89, line 92: This is the part that I dislike a bit, but we are forced to have such a thing.
The deregistration process manager doesn't have upcast/upgrade option, so we cannot change its inner state.
This means, that when old events populate the `application_instances` list with IDs that now belong to databases, when the deregistration request arrives, the `DeregisterApplicationInstance` is sent with an ID that now belongs to a database.
Anyway, `commanded` doesn't differenciate the events per type, so it tries to apply the events for that database to the `SapSystem` aggregate, so we need to handle them. Otherwise, we have the classic pattern matching error in the `apply` function.
This `legacy` fields is used to see if the aggregate receives some database related event, and if this is the case, as the aggregate is "invalid" we basically discard all the incoming commands now on.

This only happens when we have old events

arbulu89: At the end I have changed this, so the aggregate stays intact.
The issue is now handled in the enrichment protocol, which takes care of not sending deregistration commands if the ID belongs now to a database.
It is cleaner, and it looks better.
And if in the future, we clean up the event store in some migration, it is as simple as removing the module itself

test/trento/databases/events/database_rolled_up_test.exs
```diff
@@ -0,0 +1,76 @@
+defmodule Trento.Databases.Events.DatabaseRolledUpTest do
+  use Trento.AggregateCase, aggregate: Trento.Databases.Database, async: true
+
+  alias Trento.Databases.Database
+  alias Trento.Databases.Events.DatabaseRolledUp
+
+  describe "DatabaseRolledUp event upcasting, version 2" do
+    test "should upcast the current snapshot format when the snapshot has legacy data" do
+      database_id = Faker.UUID.v4()
```
EMaksy, line 9: What do you think about using the database factory here ?

EMaksy (approved):
Hey man, great job, fascinating how old events can be transformed. 
I just left a minor comment, but besides this lgtm!

test/trento/databases/events/database_rolled_up_test.exs
arbulu89: As far as I know we don't have a database aggregate state factory, right?
Maybe you refer to the database read model, which is a different thing.
I mean, we could have it, but well, as we just use it once, i found it not necessary (or maybe is just lazyness hehe)

lib/trento/sap_systems/events/sap_system_rolled_up.ex
```diff
@@ -4,9 +4,19 @@ defmodule Trento.SapSystems.Events.SapSystemRolledUp do
   It is used to trigger the stream archiving process and it contains the snapshot of the sap system aggregate.
   """
 
+  alias Trento.Databases.Events.DatabaseRolledUp
+
   use Trento.Support.Event
 
-  defevent resource: "sap_system" do
+  # Supersede the event to DatabaseRolleUp if the snapshot has the old format
```
rtorrero, line 11: DatabaseRolledUp

rtorrero (approved):
@arbulu89 thanks for the PR and the online explanation today, lgtm

rtorrero (commented):
I'm revoking my approval, I want to have a chat with you, I've noticed with my regression tests that the dashboard seems to be empty after using this branch and upcasting events from the pgsql dump from before the split. Did you notice that? is that to be expected?

arbulu89:
> I'm revoking my approval, I want to have a chat with you, I've noticed with my regression tests that the dashboard seems to be empty after using this branch and upcasting events from the pgsql dump from before the split. Did you notice that? is that to be expected?

Yes, this is normal. The sap systems and application instances are removed in the migration.
Once you start the agents again (with photofinish for example) you will see them again, with the new IDs

CDimonaco (approved):

PR #2488, opened 2024-04-03
Fix suma url overlap

nelsonkopliku:
# Description

Truncating longer suma urls with an ellipsis.
image

## How was this tested?

Storybook is enough.


jagabomb (approved):
LGTM

PR #2492, opened 2024-04-04
Fix Incorrect RDBMS Value Display on SAP Systems Overview Page

EMaksy:
# Description

This PR addresses an issue on the SAP systems overview page, where the displayed RDBMS and Tenant values are identical. The Tenant value displayed is correct, but the attached Rdbms value  uses the Tenant value.

Now the displayed RDBMS reflect the database sid as in future the tenant and database sid could be different as multi tenancy support is currently work in progress.

## How was this tested?

WIP


lib/trento_web/views/v1/sap_system_view.ex
```diff
@@ -25,7 +25,8 @@ defmodule TrentoWeb.V1.SapSystemView do
         sap_system:
           %{
             application_instances: application_instances,
-            database_instances: database_instances
+            database_instances: database_instances,
+            database: database
```
arbulu89, line 29: Destruct this:
```
database: %{
   sid: database_sid
}
```

test/trento/sap_systems/projections/sap_system_projector_test.exs
```diff
@@ -335,7 +335,7 @@ defmodule Trento.SapSystems.Projections.SapSystemProjectorTest do
       projection =
       SapSystemReadModel
       |> Repo.get(sap_system_id)
-      |> Repo.preload([:tags])
+      |> Repo.preload([:tags, :database])
```
arbulu89, line 338: we should change the assert on this test, right?
the `database_sid` should be there, in the broadcast

test/trento_web/views/v1/sap_system_view_test.exs
```diff
@@ -2,15 +2,18 @@ defmodule TrentoWeb.V1.SapSystemViewTest do
   use TrentoWeb.ConnCase, async: true
 
   import Phoenix.View
-
+  import Trento.Factory
   alias Trento.Databases.Projections.DatabaseInstanceReadModel
   alias Trento.SapSystems.Projections.SapSystemReadModel
 
   test "should add the system replication status to the secondary instance and should remove it from the primary one" do
+    database = insert(:database)
```
arbulu89, line 10: Use `build`. The insert will create a big mess otherwise.
In general, don't use `insert` unless you using the `DataCase` template.
Without this, the database is not sandboxed and you will start seeing a lot of errors

assets/js/lib/test-utils/factories/sapSystems.js
```diff
@@ -36,6 +36,8 @@ export const sapSystemFactory = Factory.define(({ params }) => {
   const id = params.id || faker.string.uuid();
   const sid = params.sid || faker.string.alphanumeric(3, { casing: 'upper' });
   const databaseId = params.database_id || faker.string.uuid();
+  const databaseSid =
```
arbulu89, line 39: maybe it is time at some point to have a sid generator function eheh
nah, forget about it in this PR at least. 
if you want, you can work afterwards on that as small tech-debt.
But it is not a big deal anyway

arbulu89 (changes requested):
Looking great!
My only concern is the `sap_system_registered` event, which most probably is not sending the `database_id` in the view function, as it doesn't use the generic `sap_system` rendering function.
I think that if we fix that, everything else looks great

test/trento_web/views/v1/sap_system_view_test.exs
```diff
@@ -2,41 +2,130 @@ defmodule TrentoWeb.V1.SapSystemViewTest do
...
-               system_replication: "Secondary",
-               system_replication_status: "ACTIVE"
-             } ->
-               true
-
-             _ ->
-               false
-           end)
+
+      assert ^expected_sap_system_json =
```
arbulu89, line 88: Some comments about this. If you do a simple pattern matching `=` test, you don't really need to have all the fields of this `expected_sap_system_json`.
If you want to do a real "everything is equal", you should use the `==` comparison.

Besides that, I encourage you to now have a `expect_**` variable. It hides things.
Put just the expected values there, and you see all of the pinned things.
But with `==` you don't need to pin anything anyway

test/trento_web/views/v1/sap_system_view_test.exs
```diff
@@ -2,41 +2,130 @@ defmodule TrentoWeb.V1.SapSystemViewTest do
...
+        )
+
+      expected_sap_system_json =
+        [
+          %{
+            id: sap_system.id,
+            tags: [],
+            inserted_at: nil,
+            updated_at: nil,
+            health: sap_system.health,
```
arbulu89, line 39: Destruct things.
I know that it might be more code, but it is much more elegant and the recommended way.
I know that we have this in other places, but the more you do the other, the more habit you will have to do that way.

arbulu89 (changes requested):
Some additional comments.
But we are going in the correct direction

test/trento/sap_systems/projections/sap_system_projector_test.exs
```diff
@@ -62,6 +63,7 @@ defmodule Trento.SapSystems.Projections.SapSystemProjectorTest do
     assert event.db_host == projection.db_host
     assert event.health == projection.health
     assert event.ensa_version == projection.ensa_version
+    assert event.database_id == projection.database_id
```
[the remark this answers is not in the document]
EMaksy: Yeah, you are right, added database_sid assertion!

rtorrero (approved):
Looks perfect @EMaksy, thanks!

PR #2494, opened 2024-04-04
Fix checking existence of certain checks in the catalog

nelsonkopliku:
# Description

Fixes determination check existence in the catalog based on the target type or cluster type.

## How was this tested?

Automated tests.

### DIFF
--- a/assets/js/lib/model/checks.js
+++ b/assets/js/lib/model/checks.js
@@ -1,27 +1,23 @@
-import { pipe, some, castArray, filter } from 'lodash/fp';
+import { castArray } from 'lodash';
 import { TARGET_CLUSTER } from '.';
-
-const hasMetadata = ({ metadata }) => !!metadata;
-const hasClusterTypeMetadata = ({ metadata }) =>
-  hasMetadata({ metadata }) && !!metadata.cluster_type;
 
 const isCheckForTarget =
   (targetType) =>
   ({ metadata }) =>
-    !hasMetadata({ metadata }) || metadata.target_type === targetType;
+    !metadata || metadata.target_type === targetType;
+
+const supportsClusterType =
+  (clusterType) =>
+  ({ metadata }) =>
+    !metadata?.cluster_type ||
+    castArray(metadata.cluster_type).some(
+      (metadataClusterType) => metadataClusterType === clusterType
+    );
 
 export const hasChecksForTarget = (catalog, targetType) =>
-  some(isCheckForTarget(targetType))(catalog);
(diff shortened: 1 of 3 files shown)


assets/js/lib/model/checks.js
```diff
@@ -0,0 +1,28 @@
...
+
+export const hasChecksForTarget = (catalog, targetType) =>
+  some(isCheckForTarget(targetType))(catalog);
+
+export const hasChecksForClusterType = (catalog, clusterType) =>
+  pipe(
+    filter(isCheckForTarget(TARGET_CLUSTER)),
+    some(
+      ({ metadata }) =>
+        !hasMetadata({ metadata }) ||
```
jamie-suse, line 21: Isn't this check already done in the `filter()` step, and also later in `hasClusterTypeMetadata()`?

assets/js/lib/model/checks.js
```diff
@@ -0,0 +1,28 @@
...
+        !hasClusterTypeMetadata({ metadata }) ||
+        pipe(
```
jamie-suse, line 23: Isn't this creating an array of size 1?

assets/js/lib/model/checks.js
nelsonkopliku: yes, but it is a different check.
`hasClusterTypeMetadata` checks whether there are metadata AND there is a `cluster_type` in it.

`!hasMetadata({ metadata }) ||` usage hear means that if there are not metadata we consider the check applicable for the given `clusterType`

assets/js/lib/model/checks.js
nelsonkopliku: you mean `castArray`?

If you're referring to it, it creates an array of size 1 if cluster_type is let's say a string like `ascs_ers`, but it covers cases where we have:
```yaml
metadata:
  target_type: cluster
  cluster_type:
    - hana_scale_up
    - hana_scale_out
```
maybe not frequent, but the metadata in the check work like this :smile:

assets/js/lib/model/checks.js
nelsonkopliku: nevermind, that seems to be safely removable 1408a32d91f6ba4b83ad3ccc95cabb38f4f7588f

assets/js/lib/model/checks.js
jamie-suse: Oh okay, so the type of `clusterType` is `String | [String]`?

nelsonkopliku: potentially yes

jamie-suse (approved):

PR #2499, opened 2024-04-08
Fix get systems by cluster selector

arbulu89:
# Description
Fix regression we had during the sap system/database split.

image

The sap system/database URL was not being populated.
The more I check, i don't know how this happened, since the UT looks totally correct.
Maybe the update to redux toolkit 2 has changed something, and the lodash set does something that it doesn't like...

## How was this tested?

E2E test improved to cover this case


EMaksy (approved):
Nice catch! 
Adding e2e is super helpful for the future!

CDimonaco (approved):

rtorrero (approved):

PR #2504, opened 2024-04-09
Add upcasting regression e2e test to the CI

rtorrero:
# Description

This PR uses a snapshot from a trento server instance prior to the split of the DB and SapSystem aggregates to test the upcasting to the new events for regressions. With this we should be confident that the upgrade from an existing system goes smoothly when existing data from the eventstore is migrated to the new events.

This can of course be used in general to test other upcasting of events and can be extended to include more tests (as of now, only the sap systems overview and the database overview is beeing tested)

>NOTE: For the tests to run, the label "regression" needs to be added to the PR

>NOTE: I wanted to have the CI code separate in a different file but if I do so, it forces me to repeat many of the required jobs, ending up in re-running a lot of steps to generate the assets required to run trento.

## How was this tested?

In a separate fork, creating a PR and adding the label.


.github/workflows/ci.yaml
```diff
@@ -15,6 +16,8 @@ env:
   OTP_VERSION: 26
   MIX_ENV: test
   NODE_VERSION: "16"
+  MANTAINERS: '["cdimonaco", "dottorblaster", "fabriziosestito", "rtorrero", "nelsonkopliku", "arbulu89","jagabomb","emaksy","jamie-suse"]'
```
arbulu89, line 19: Could we have all of this logic in an additional file? Like the `pr_env.yaml`
At the end, as it is triggered by a label, so a different flow, to have it separated

.github/workflows/ci.yaml
```diff
@@ -436,6 +439,112 @@ jobs:
...
+            assets/node_modules
+          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-${{ hashFiles('assets/package-lock.json') }}
+      - name: Retrieve E2E NPM Cached Dependencies
+        uses: actions/cache@v4
+        id: npm-e2e-cache
+        with:
+          path: |
+            test/e2e/node_modules
+          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-${{ hashFiles('test/e2e/package-lock.json') }}
+      - name: Check Eslint and JS Code Format
```
arbulu89, line 500: I don't think we need this `eslint` here. It is done in the normal flow

.github/workflows/ci.yaml
```diff
@@ -436,6 +439,112 @@ jobs:
...
+        run: cd test/e2e && npm run lint && npm run format:check
+      - name: "Docker compose dependencies"
+        uses: isbang/compose-action@v1.5.1
+        with:
+          compose-file: "./docker-compose.yaml"
+          down-flags: "--volumes"
+          compose-flags: "-f docker-compose.regression.yaml"
+      - name: Mix setup
+        run: mix setup
```
arbulu89, line 509: `mix setup`, are you sure of this?
In all the tests I did this was not necessary, in fact, it created a lot of issues because it tries to modify the database with entries that are already modified

test/fixtures/regression/sap_system_split/trento_eventstore_dev.dump
arbulu89, line 1: The dump is done with `postgresql16`, right?
We just pinned version 15 in the docker-compose, so we should use that one

arbulu89, line 1: I miss some events in the dump.
I just see the normal 27 node events, and the sap system rollup.
If we run the database overview and sap system overview e2e tests, we should have more.
What I would do:
- Start the app
- Run 27 nodes photofinish job
- Rollup a SAP system. NWD for example (deregister app and db instance)
- Run database over E2E tests
- Run sap system overview E2E tests
Once all of them passed (they should), create the dump

arbulu89 (commented):
Personally, I don't like the "script + docker exec" strategy, with a custom docker compose.
It looks really rigid and requires more maintenance (you need to keep up with the docker compose postgres version).

I think that it would be much more convinient to run the restores from outside, with some github CI plugin (i just found this in 1 sec search: https://hub.docker.com/r/actions/pg_restore) or just running `pg_restore` in a normal run stage.

With this, everything is obvious in the CI itself, and you don't need to look for external files.
Besides that, it will be easy to use new fixture files in the future, as you just need to update a variable name, and nothing else.

I don't know what other thing though about this

.github/workflows/ci.yaml
rtorrero: I wrote an update in the PR description but you probably wrote this earlier: wouldn't this force us to repeat many of the steps that are already performed in the regular run? I supposedly could use the cache between different CI yamls but if I can't add dependencies I have no way of knowing if the assets in the cache are updated to the current commit or not. Maybe I'm missing something?

.github/workflows/ci.yaml
rtorrero: I'm not sure but I got in troubles running the restores without it. I didn't look much into it as I thought it was needed, I'll double check :+1:

test/fixtures/regression/sap_system_split/trento_eventstore_dev.dump
rtorrero: the dump is done with the postgres container that our docker-compose used before the changes, with is postgres:latest (which as you say probably means postgresql16. I'll use the pin too and update it :+1:

rtorrero: I missed the last step, when we spoke I assumed the cypress run you meant was the one we do once the restore is performed. Will add :+1: .

.github/workflows/ci.yaml
rtorrero: The `mix setup` was there as a way of avoiding the creation of the database (mix setup would do this for us), but I've replaced it with a saner approach: when creating the `.sql` files with `pg_dump`, I used `-C` option which results in the `.sql` files also including `CREATE DATABASE` statements. This allows that during restoration, the databases are not required to be there in the first place (and allows to remove this `mix setup`)

test/fixtures/regression/sap_system_split/trento_eventstore_dev.dump
rtorrero: I've updated it to use postgres15

.github/workflows/ci.yaml
```diff
@@ -436,6 +439,118 @@ jobs:
           name: e2e-screenshots
           path: test/e2e/cypress/screenshots/
 
+  check-regression-label:
+    name: Check if the environment creation criteria are met, store in the job output
+    runs-on: ubuntu-latest
```
arbulu89, line 444: Let's avoid `ubuntu-latest`. I know that it hardly affects here, but better pin it

.github/workflows/ci.yaml
```diff
@@ -436,6 +439,118 @@ jobs:
...
+        id: npm-e2e-cache
+        with:
+          path: |
+            test/e2e/node_modules
+          key: ${{ runner.os }}-${{ env.NODE_VERSION }}-${{ hashFiles('test/e2e/package-lock.json') }}
+      - name: "Docker compose dependencies"
+        uses: isbang/compose-action@v1.5.1
+        with:
+          compose-file: "./docker-compose.yaml"
+          down-flags: "--volumes"
```
arbulu89, line 505: Is this `volumes` required?

arbulu89 (approved):
Nice @rtorrero 
Thank you!
I would've prefer the job in other file, but I understand the rationale behind, so it is good.
Just check that we don't have any sensitive data in the database dumps. I don't think so, but anyway

.github/workflows/ci.yaml
EMaksy: I guess we can also add our newcomer: 

janvhs

.github/workflows/ci.yaml
EMaksy: I'm with xabi on this, fixed version is extra save :+1:

EMaksy (approved):
Hey nice, I left just two small comments. 
LGTM!

CDimonaco (approved):
LGTM!

.github/workflows/ci.yaml
```diff
@@ -436,6 +439,117 @@ jobs:
...
+          spec: |
+            cypress/e2e/databases_overview.cy.js
+            cypress/e2e/sap_systems_overview.cy.js
+          wait-on-timeout: 30
+          config: baseUrl=http://localhost:4000
+      - name: Upload cypress test screenshots
+        uses: actions/upload-artifact@v4
+        if: failure()
+        with:
+          name: e2e-screenshots
```
arbulu89, line 549: Hey @rtorrero ,
Please change this `name` and `path`. 
They are a copy paste from e2e I guess. They could conflict with e2e artifcats

.github/workflows/ci.yaml
arbulu89: In fact, you could remove fabrizio as well

PR #2507, opened 2024-04-10
Handle requests for software updates discoveries from Host aggregate

nelsonkopliku:
# Description

> Yet another alternative to https://github.com/trento-project/web/pull/2505 and https://github.com/trento-project/web/pull/2506

This PR moves the logic for software updates discovery inside the host aggregate which is going to emit a `SoftwareUpdatesDiscoveryRequested` in the following scenarios:

- when a host is registered
- when a host is restored
- when the fqdn of a host changes

Additionally when the fqdn of the host transitions from a meaningful value to `null` we take care of dispatching a `ClearSoftwareUpdatesDiscovery` for clean up purposes.

Following this we will have a `SoftwareUpdatesDiscoveryEventHandler` listening to `SoftwareUpdatesDiscoveryRequested` getting information from suma and dispatching back a `CompleteSoftwareUpdatesDiscovery` command

When implementing the new event handler some minor changes to `Trento.SoftwareUpdates.Discovery` might be needed. Here's [some reference](https://github.com/trento-project/web/pull/2506/files#diff-873065796eb0fae979370c63b209d8c71480e0d3e4b14bf74474da07f3638c81)

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/hosts/host.ex
+++ b/lib/trento/hosts/host.ex
@@ -158,7 +158,7 @@ defmodule Trento.Hosts.Host do
 
   # New host registered
   def execute(
-        %Host{host_id: nil} = host,
+        %Host{host_id: nil},
         %RegisterHost{
           host_id: host_id,
           hostname: hostname,
@@ -172,9 +172,7 @@ defmodule Trento.Hosts.Host do
           installation_source: installation_source
         }
       ) do
-    host
-    |> Multi.new()
-    |> Multi.execute(fn _ ->
+    [
       %HostRegistered{
         host_id: host_id,
         hostname: hostname,
@@ -188,8 +186,7 @@ defmodule Trento.Hosts.Host do
         fully_qualified_domain_name: fully_qualified_domain_name,
         heartbeat: :unknown
       }
-    end)
-    |> Multi.execute(&maybe_emit_software_updates_discovery_requested_event/1)
+    ] ++ maybe_emit_software_updates_discovery_events(host_id, nil, fully_qualified_domain_name)
   end
 
   # Reject all the commands, except for the registration ones when the host_id does not exists
@@ -214,7 +211,7 @@ defmodule Trento.Hosts.Host do
           fully_qualified_domain_name: fully_qualified_domain_name,
           installation_source: installation_source,
           deregistered_at: deregistered_at
-        } = host,
+        },
         %RegisterHost{
           hostname: hostname,
           ip_addresses: ip_addresses,
@@ -228,19 +225,17 @@ defmodule Trento.Hosts.Host do
         }
       )
       when not is_nil(deregistered_at) do
-    host
-    |> Multi.new()
-    |> Multi.execute(fn _ ->
+    [
       %HostRestored{host_id: host_id}
-    end)
-    |> Multi.execute(&maybe_emit_software_updates_discovery_requested_event/1)
+    ] ++ maybe_emit_software_updates_discovery_events(host_id, nil, fully_qualified_domain_name)
   end
 
   def execute(
         %Host{
           host_id: host_id,
-          deregistered_at: deregistered_at
-        } = host,
+          deregistered_at: deregistered_at,
+          fully_qualified_domain_name: current_fully_qualified_domain_name
+        },
         %RegisterHost{
           hostname: hostname,
           ip_addresses: ip_addresses,
@@ -249,17 +244,13 @@ defmodule Trento.Hosts.Host do
           total_memory_mb: total_memory_mb,
           socket_count: socket_count,
           os_version: os_version,
-          fully_qualified_domain_name: fully_qualified_domain_name,
+          fully_qualified_domain_name: new_fully_qualified_domain_name,
           installation_source: installation_source
         }
       )
       when not is_nil(deregistered_at) do
-    host
-    |> Multi.new()
-    |> Multi.execute(fn _ ->
-      %HostRestored{host_id: host_id}
-    end)
-    |> Multi.execute(fn _ ->
+    [
+      %HostRestored{host_id: host_id},
       %HostDetailsUpdated{
         host_id: host_id,
         hostname: hostname,
@@ -269,11 +260,15 @@ defmodule Trento.Hosts.Host do
         total_memory_mb: total_memory_mb,
         socket_count: socket_count,
         os_version: os_version,
-        fully_qualified_domain_name: fully_qualified_domain_name,
+        fully_qualified_domain_name: new_fully_qualified_domain_name,
         installation_source: installation_source
       }
-    end)
-    |> Multi.execute(&maybe_emit_software_updates_discovery_requested_event/1)
+    ] ++
+      maybe_emit_software_updates_discovery_events(
+        host_id,
+        current_fully_qualified_domain_name,
+        new_fully_qualified_domain_name
+      )
   end
 
   def execute(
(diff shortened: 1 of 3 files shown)


lib/trento/hosts/host.ex
```diff
@@ -172,19 +172,24 @@ defmodule Trento.Hosts.Host do
...
+        cpu_count: cpu_count,
+        total_memory_mb: total_memory_mb,
+        socket_count: socket_count,
+        os_version: os_version,
+        installation_source: installation_source,
+        fully_qualified_domain_name: fully_qualified_domain_name,
+        heartbeat: :unknown
+      }
+    end)
+    |> Multi.execute(&maybe_emit_software_updates_discovery_requested_event/1)
```
arbulu89, line 192: What if we group all of these maybes at once:

```
# args are, host_id, new_fqdn, old_fqdn
def maybe_emit_software_updates_discovery_events(host_id, nil, nil),do: nil

def maybe_emit_software_updates_discovery_events(host_id, nil, _old_fqdn) do
  %SoftwareUpdatesDiscoveryCleared{
    host_id: host_id
  }
end

def maybe_emit_software_updates_discovery_events(host_id, new_fqdn, nil) do
  %SoftwareUpdatesDiscoveryRequested{
    host_id: host_id,
    fully_qualified_domain_name: new_fqdn
  }
end

def maybe_emit_software_updates_discovery_events(host_id, fqdn, fqdn) do: nil

def maybe_emit_software_updates_discovery_events(host_id, new_fqdn, _old_fqdn) do
  %SoftwareUpdatesDiscoveryRequested{
    host_id: host_id,
    fully_qualified_domain_name: new_fqdn
  }
end
```

No `is_nil` functions, just standard easy pattern matching.

lib/trento/hosts/host.ex
```diff
@@ -262,7 +272,8 @@ defmodule Trento.Hosts.Host do
         fully_qualified_domain_name: fully_qualified_domain_name,
         installation_source: installation_source
       }
-    ]
+    end)
+    |> Multi.execute(&maybe_emit_software_updates_discovery_requested_event/1)
```
arbulu89, line 276: In this scenario, we have the chance of having a `fqdn` before, but the new resister that restores sends a nil `fqdn`, in which case, i guess we should emit `SoftwareUpdatesDiscoveryCleared`.

PD: with the proposal I did, this should be covered, as we only have one function

test/trento/hosts/host_test.exs
```diff
@@ -1899,6 +2014,110 @@ defmodule Trento.Hosts.HostTest do
       )
     end
 
+    test "should restore a deregistered host and trigger software updates discovery" do
```
arbulu89, line 2017: I miss a `should restore a deregistered host and trigger the software updates cleaned event` or similar, if the new coming discovery has fqdn nil, and it was set before

arbulu89 (changes requested):
Hey @nelsonkopliku ,
The approach looks correct.
Just a couple of things:
- we don't need to use the `multi` as we don't need to get any change from previous events in this event
- i think we can simply the new `maybe_` functions, solution proposed

Besides that, all good

lib/trento/hosts/host.ex
nelsonkopliku: during a restoration, the transition of the fqdn to null does not really require a clean up, because things were cleared on deregistration. makes sense?

test/trento/hosts/host_test.exs
nelsonkopliku: same as here https://github.com/trento-project/web/pull/2507/files#r1560611137

lib/trento/hosts/host.ex
arbulu89: yes, you are right, it makes sense.

lib/trento/hosts/host.ex
nelsonkopliku: This proposal always takes into account a transition of the fqdn, but sometimes we just care about the end result in the aggregate state other times we need to check the transition, as we discussed in another comment.

This is why the multi and the functions as they have been implemented now. 

What could we do differently?

arbulu89: As far as I see using the `multi` is not necessary.
If the software update events must be emitted depends simply on the old and new `fqdn` values, and we don't need a state update to get that, they are already there, in the initial state and event.

For example, if you substitute 1:1 the functions i wrote there by the ones you used, all the tests are green. Not a big deal, but by default, it is better to not use `when` guards if you can solve things with simple pattern matching, and using less functions looks ideal (my opinion).

It is true that not using the `multi` is a bit inconvenient on how you need to use this functions. As you cannot send a `nil` event in a list of events, you need to something like this:

```
def execute(
        %Host{host_id: nil},
        %RegisterHost{
          host_id: host_id,
          ...
        }
      ) do
    [
      %HostRegistered{
        host_id: host_id,
         ...
      }
    ] ++ maybe_emit_software_updates_discovery_events(host_id, fully_qualified_domain_name, nil)
  end
```

and later on change slightly the functions I wrote to return lists:
```
  def maybe_emit_software_updates_discovery_events(_host_id, nil, nil), do: []

  def maybe_emit_software_updates_discovery_events(host_id, nil, _old_fqdn) do
    [
      %SoftwareUpdatesDiscoveryCleared{
        host_id: host_id
      }
    ]
  end
```
and so on. For me this looks perfectly fine.
Tests are green in this case as well.

If it was up to me, I would choose this option, as you don't use a multi that you don't really need, and from my perspective, it is cleaner and precise (by the way, using multi is not harmful, simply, we don't need it in my opinion).

Now, this is my opinion, your solution works (as does mine), and as this is your PR, I leave the decision up to you.

PD: i base my opinion seeing the tests coming in this PR are all green

arbulu89 (approved):
Hey @nelsonkopliku ,
As the solution you propose is correct, I'm giving a green light.
I proposed other solution that I think is better (my opinion).
But in any case, it is ok to move forward

nelsonkopliku:
> Hey @nelsonkopliku , As the solution you propose is correct, I'm giving a green light. I proposed other solution that I think is better (my opinion). But in any case, it is ok to move forward

Here we go @arbulu89 https://github.com/trento-project/web/pull/2507/commits/772f280799a9ea27bdca2237207387ec0e814747, thanks for the feedback

lib/trento/hosts/host.ex
```diff
@@ -259,10 +260,15 @@ defmodule Trento.Hosts.Host do
...
         os_version: os_version,
-        fully_qualified_domain_name: fully_qualified_domain_name,
+        fully_qualified_domain_name: new_fully_qualified_domain_name,
         installation_source: installation_source
       }
-    ]
+    ] ++
+      maybe_emit_software_updates_discovery_events(
+        host_id,
+        current_fully_qualified_domain_name,
```
arbulu89, line 269: Shouldn't this 2nd argument be `nil` here?
I mean, I understood that in this circunstance we don't want to emit the `Cleared` event, but if the new fqdn is `nil`, and we had a value before (current), it would be emitted, or not?

arbulu89 (commented):
Thank you @nelsonkopliku ,
Good job

lib/trento/hosts/host.ex
nelsonkopliku: yep, I guess I missed that test.
Here we go https://github.com/trento-project/web/pull/2518
Thanks!

PR #2509, opened 2024-04-11
Revert wrongly changed path

rtorrero:
# Description

Revert the change of the path. This path is a variable used to decide what artifacts to upload and it can't change the path where cypress outputs these files.


arbulu89 (approved):
Thank you!
And if by chance, you remove one of the 2 new lines you have after, my ocd would appreciate id :see_no_evil:

PR #2512, opened 2024-04-11
Removed eula popup notification

chrisb1s:
# Description

Completely removed the Eula popup message from the web app. This includes:

- the elexir files
- the redux javascript part
- the migration of the database

## How was this tested?

Tested with the tests provided by the module/tests.


stefanotorresi:
I think we need to leverage api versioning for the endpoint removal, right @arbulu89 @CDimonaco ?

assets/js/state/sagas/index.js
```diff
@@ -95,14 +92,6 @@ function* initialDataFetch() {
...
 
-  const {
-    data: { eula_accepted, premium_subscription },
-  } = yield call(get, '/settings');
-
-  if (!eula_accepted) {
-    yield put(setEulaVisible());
-  }
-
   if (premium_subscription) {
```
nelsonkopliku, line 95: `premium_subscription` is never available. I guess also this condition can go with also `yield put(setIsPremium());` and the related piece of state/saga/selector if any.

assets/js/state/settings.js
```diff
@@ -1,29 +1,20 @@
...
-    acceptEula: (state) => {
-      state.eulaVisible = false;
-    },
   },
 });
 
-export const ACCEPT_EULA = 'ACCEPT_EULA';
-
-export const { setEulaVisible, setIsPremium, acceptEula } =
+export const { setIsPremium  } =
```
nelsonkopliku, line 17: As per the previous comment this might not be needed anymore.

lib/trento_web/openapi/v1/schema/platform.ex
```diff
@@ -13,10 +13,6 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Platform do
       type: :object,
       additionalProperties: false,
       properties: %{
-        eula_accepted: %Schema{
```
nelsonkopliku, line 16: Removing this field breaks API backward compatibility. We need to add a v2 version of it.

nelsonkopliku (commented):
There is some extra eula related frontend items that might be removed.

Besides this `GET /api/v1/settings` needs to be bumped to v2.
Not sure if the removal of the other two APIs is breaking BC (guess so)

```
==========================================================================
==                            API CHANGE LOG                            ==
==========================================================================
                                  Trento                                  
--------------------------------------------------------------------------
--                            What's Deleted                            --
--------------------------------------------------------------------------
- POST   /api/v1/accept_eula
- POST   /api/v1/settings/accept_eula

--------------------------------------------------------------------------
--                            What's Changed                            --
--------------------------------------------------------------------------
- GET    /api/v1/settings
  Return Type:
    - Changed 200 OK
      Media types:
        - Changed application/json
          Schema: Broken compatibility
          Missing property: eula_accepted (boolean)
--------------------------------------------------------------------------
--                                Result                                --
--------------------------------------------------------------------------
                 API changes broke backward compatibility                 
--------------------------------------------------------------------------
```

arbulu89 (commented):
Hey @chrisb1s ,
The removal looks good so far. Some comments from my side
I'm not that sure if we can just remove a endpoint without bumping the controller or schemas version.
We can take that risk, but users using it might start getting failures.
I guess we need to discuss that.

assets/js/state/settings.js
arbulu89: I think it is. Being premium or not was not only about the eula.
We still have this `Premium: true/false` thing, based on the artifacts creation and if there are sles subscriptions active

nelsonkopliku: you're right :+1:

arbulu89 (approved):
It looks good now!
Thank you @chrisb1s

PR #2515, opened 2024-04-12
Software updates event handler

nelsonkopliku:
# Description

This PR introduces a `SoftwareUpdatesDiscoveryEventHandler` reacting to `SoftwareUpdatesDiscoveryRequested` events, getting information from SUMA and dispatching back in the system a `CompleteSoftwareUpdatesDiscovery` possibly resulting in a host's health change.

What's next:
- make the discovery tick dispatch `DiscoverSoftwareUpdates` command so that `SoftwareUpdatesDiscoveryRequested` + event listener is triggered (currently we'd only have a completion event without an initiation event)
- trigger software updates discovery process for all the registered hosts when saving/changing suma settings
- improve consistency between health in hosts list and what gets returned in GET `/hosts/:id/software_updates` 

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
+++ b/lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
@@ -8,8 +8,6 @@ defmodule Trento.Infrastructure.Commanded.EventHandlers.SoftwareUpdatesDiscovery
     application: Trento.Commanded,
     name: "software_updates_discovery_event_handler"
 
-  require Logger
-
   alias Trento.Hosts.Events.SoftwareUpdatesDiscoveryRequested
 
   alias Trento.SoftwareUpdates.Discovery
--- a/lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
+++ b/lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
@@ -1,0 +1,28 @@
+defmodule Trento.Infrastructure.Commanded.EventHandlers.SoftwareUpdatesDiscoveryEventHandler do
+  @moduledoc """
+  Event handler for software updates discovery related events.
+  Here is where the actual integration with the external system happens and relevant changes in Host are triggered.
+  """
+
+  use Commanded.Event.Handler,
+    application: Trento.Commanded,
+    name: "software_updates_discovery_event_handler"
+
+  require Logger
+
+  alias Trento.Hosts.Events.SoftwareUpdatesDiscoveryRequested
+
+  alias Trento.SoftwareUpdates.Discovery
+
+  def handle(
+        %SoftwareUpdatesDiscoveryRequested{
+          host_id: host_id,
+          fully_qualified_domain_name: fully_qualified_domain_name
+        },
+        _
+      ) do
+    Discovery.discover_host_software_updates(host_id, fully_qualified_domain_name)
+
+    :ok
+  end
+end
--- a/lib/trento/software_updates/discovery.ex
+++ b/lib/trento/software_updates/discovery.ex
@@ -37,24 +37,24 @@ defmodule Trento.SoftwareUpdates.Discovery do
     do: adapter().get_upgradable_packages(system_id)
 
   @spec discover_software_updates :: {:ok, {list(), list()}}
-  def discover_software_updates,
-    do:
-      {:ok,
-       Hosts.get_all_hosts()
-       |> Enum.map(fn
-         %HostReadModel{id: host_id, fully_qualified_domain_name: fully_qualified_domain_name} ->
-           case discover_host_software_updates(host_id, fully_qualified_domain_name) do
-             {:error, error} ->
-               {:error, host_id, error}
+  def discover_software_updates do
+    {:ok,
+     Hosts.get_all_hosts()
+     |> Enum.map(fn
+       %HostReadModel{id: host_id, fully_qualified_domain_name: fully_qualified_domain_name} ->
+         case discover_host_software_updates(host_id, fully_qualified_domain_name) do
+           {:error, error} ->
+             {:error, host_id, error}
 
-             _ = success ->
-               success
-           end
-       end)
-       |> Enum.split_with(fn
-         {:ok, _, _, _} -> true
-         _ -> false
-       end)}
+           {:ok, _, _, _} = success ->
+             success
+         end
+     end)
+     |> Enum.split_with(fn
+       {:ok, _, _, _} -> true
+       _ -> false
+     end)}
+  end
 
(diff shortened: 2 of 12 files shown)


lib/trento/software_updates/discovery.ex
```diff
@@ -66,18 +75,16 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
     Logger.info("Host #{host_id} does not have an fqdn. Skipping software updates discovery")
-    {:error, host_id, :host_without_fqdn}
+    {:error, :host_without_fqdn}
   end
 
-  defp discover_host_software_updates(%HostReadModel{
-         id: host_id,
-         fully_qualified_domain_name: fully_qualified_domain_name
-       }) do
+  @spec discover_host_software_updates(String.t(), String.t()) ::
```
arbulu89, line 85: We don't need to have the signature twice. With the one above should be enough

lib/trento/software_updates/discovery.ex
```diff
@@ -41,7 +41,16 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
       {:ok,
        Hosts.get_all_hosts()
-       |> Enum.map(&discover_host_software_updates/1)
+       |> Enum.map(fn
+         %HostReadModel{id: host_id, fully_qualified_domain_name: fully_qualified_domain_name} ->
+           case discover_host_software_updates(host_id, fully_qualified_domain_name) do
+             {:error, error} ->
+               {:error, host_id, error}
+
+             _ = success ->
```
arbulu89, line 50: I think that having here `:ok` pattern matching is more visual. What do you think?
```
{:ok, host_id, system_id, relevant_patches} ->
   {:ok, host_id, system_id, relevant_patches}

# or
{:ok, _, _, _} = success ->
  success
```

It is more obvious, and if for some reason we add any other `:error` return to the `discovery_host_software_updates`, we would still be safe

lib/trento/software_updates/discovery.ex
```diff
@@ -41,7 +41,16 @@ defmodule Trento.SoftwareUpdates.Discovery do
     do:
       {:ok,
        Hosts.get_all_hosts()
-       |> Enum.map(&discover_host_software_updates/1)
+       |> Enum.map(fn
```
arbulu89, line 44: Nothing for this PR, but I find having such a long functions as one liners a bit ugly hehe
Even though it is a piped one lines, I would definitely add a `do ... end` block.

lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
```diff
@@ -0,0 +1,28 @@
...
+  def handle(
+        %SoftwareUpdatesDiscoveryRequested{
+          host_id: host_id,
+          fully_qualified_domain_name: fully_qualified_domain_name
+        },
+        _
+      ) do
+    Discovery.discover_host_software_updates(host_id, fully_qualified_domain_name)
+
+    :ok
```
arbulu89, line 26: i find this always `:ok` return not really safe or robust.
I would personally check what the discovery returns, and based on that return some error.

We can implement the `error` function later on, even if it does a simple `:skip`.
This would give the look that we are dealing with the error (even though initially is doing nothing).

Maybe in the future, we could have some retry strategy, or broadcast some message to the frontend, or anything like that.

I find it a elegant solution at least.

PD: The [error](https://hexdocs.pm/commanded/Commanded.Event.Handler.html#callbacks) function

lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
```diff
@@ -0,0 +1,28 @@
...
+  @moduledoc """
+  Event handler for software updates discovery related events.
+  Here is where the actual integration with the external system happens and relevant changes in Host are triggered.
+  """
+
+  use Commanded.Event.Handler,
+    application: Trento.Commanded,
+    name: "software_updates_discovery_event_handler"
+
+  require Logger
```
arbulu89, line 11: We don't need this `require Logger`, right?

test/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler_test.exs
```diff
@@ -0,0 +1,72 @@
...
+
+  setup [:set_mox_from_context, :verify_on_exit!]
+
+  test "should discover software updates when a SoftwareUpdatesDiscoveryRequested is emitted" do
+    %SoftwareUpdatesDiscoveryRequested{
+      host_id: host_id,
+      fully_qualified_domain_name: fully_qualified_domain_name
+    } = event = build(:software_updates_discovery_requested_event)
+
+    system_id = 100
```
arbulu89, line 22: Could we maybe use `UUID` here?

lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
arbulu89: Anyway, as we are not going to do anything with the failure by now, I'm okey to leaving as it is.
Maybe just putting a comment like `The error is not being handled, as there is not any strategy for it`

test/trento/software_updates/discovery_test.exs
```diff
@@ -252,6 +256,74 @@ defmodule Trento.SoftwareUpdates.DiscoveryTest do
     end
   end
 
+  defp fail_on_getting_system_id(fully_qualified_domain_name, discovery_error) do
```
arbulu89, line 259: Maybe better put this private helper functions at the end of the file, instead of the middle?

arbulu89 (approved):
Good job!
I dropped some cosmetic related comments mostly.

PD: I commented on the option of handling some errors in the discovery. It is true that it would just be boilerplate code.
I was just wondering how we will notify the user when a discovery fails, because until now, we just print some messages in the logs.

I guess this discussion goes beyond of this PR. Would it make sense to have a discussion about it?

lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
nelsonkopliku: I agree, some more robust error handling might be beneficial.
I just wanted to get the flow and iterate on errors as per the needs

lib/trento/software_updates/discovery.ex
nelsonkopliku: you mean the function in the `map` or the whole `discover_software_updates`?

nelsonkopliku: I guess `discover_software_updates`. Fixed :smile:

arbulu89: Yes! thank you!

PR #2516, opened 2024-04-12
Remove sid duplication from view

EMaksy:
# Description

This Pr will remove duplicated SID when multiple instances of the same SAP system are running on the same host. 
SID shows only once in the SID column of the Host's Overview.

## How was this tested?

Added unit test

### DIFF
--- a/assets/js/pages/HostsList/HostsList.jsx
+++ b/assets/js/pages/HostsList/HostsList.jsx
@@ -117,21 +117,22 @@ function HostsList() {
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
         render: (sids, { sap_systems }) => {
-          const sidsArray = sap_systems.map((instance, index) => {
-            const instanceID = getInstanceID(instance);
-            return [
-              index > 0 && ', ',
-              <SapSystemLink
-                key={`${instanceID}-${instance?.id}`}
-                systemType={instance?.type}
-                sapSystemId={instanceID}
-              >
-                {instance?.sid}
-              </SapSystemLink>,
-            ];
-          });
-          const extractSid = (item) => item[1]?.props?.children || null;
-          return uniqBy(sidsArray, extractSid);
+          const sidsArray = uniqBy(sap_systems, getInstanceID).map(
+            (instance, index) => {
+              const instanceID = getInstanceID(instance);
+              return [
+                index > 0 && ', ',
+                <SapSystemLink
+                  key={`${instanceID}-${instance?.id}`}
+                  systemType={instance?.type}
+                  sapSystemId={instanceID}
+                >
+                  {instance?.sid}
+                </SapSystemLink>,
+              ];
+            }
+          );
+          return sidsArray;
         },
       },
       {
--- a/assets/js/pages/HostsList/HostsList.jsx
+++ b/assets/js/pages/HostsList/HostsList.jsx
@@ -3,6 +3,7 @@ import React, { useState } from 'react';
 import { useSearchParams, useNavigate } from 'react-router-dom';
 import { useSelector, useDispatch } from 'react-redux';
 import { EOS_WARNING_OUTLINED } from 'eos-icons-react';
+import { uniqBy } from 'lodash';
 
 import CleanUpButton from '@common/CleanUpButton';
 import HealthIcon from '@common/HealthIcon';
@@ -15,7 +16,6 @@ import Table from '@common/Table';
 import Tags from '@common/Tags';
 import Tooltip from '@common/Tooltip';
 
-import { uniqBy } from 'lodash';
 import { post, del } from '@lib/network';
 import { agentVersionWarning } from '@lib/agent';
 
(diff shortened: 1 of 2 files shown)


assets/js/pages/HostsList/HostsList.jsx
```diff
@@ -129,8 +130,8 @@ function HostsList() {
               </SapSystemLink>,
             ];
           });
-
-          return sidsArray;
+          const extractSid = (item) => item[1]?.props?.children || null;
```
arbulu89, line 133: Let's use `piping`, right?
in Line 120:  `const sidsArray = uniqBy(sap_systems, getInstanceID).map((instance, index) => {`

assets/js/pages/HostsList/HostsList.test.jsx
```diff
@@ -121,6 +124,30 @@ describe('HostsLists component', () => {
...
+        sid: duplicateSID,
+        host_id: host.id,
+      });
+
+      const state = {
+        ...defaultInitialState,
+        hostsList: {
+          hosts: [host],
+        },
+        sapSystemsList: {
```
arbulu89, line 141: let's add some database instances with the same SID as well, as we need to have the same logic for bot

arbulu89 (changes requested):
Hey @EMaksy ,
A suggestion :wink:

assets/js/pages/HostsList/HostsList.jsx
```diff
@@ -15,6 +15,7 @@ import Table from '@common/Table';
 import Tags from '@common/Tags';
 import Tooltip from '@common/Tooltip';
 
+import { uniqBy } from 'lodash';
```
arbulu89, line 18: Put this import in line 6 or so (after EOS_WARNING_...).
We put external imports first

assets/js/pages/HostsList/HostsList.test.jsx
```diff
@@ -121,6 +125,37 @@ describe('HostsLists component', () => {
...
+        host_id: host.id,
+      });
+      const state = {
+        ...defaultInitialState,
+        hostsList: {
+          hosts: [host],
+        },
+        sapSystemsList: {
+          applicationInstances: [...sapInstances],
+          databasesList: {
```
arbulu89, line 149: This looks wrong. `databasesList` shouldn't be inside `sapSystemsList`, it should be in the same level

assets/js/pages/HostsList/HostsList.test.jsx
```diff
@@ -121,6 +125,37 @@ describe('HostsLists component', () => {
...
+      const host = hostFactory.build();
+      const duplicateSID = faker.string.alpha({ casing: 'upper', count: 3 });
+      const id = faker.string.uuid();
+      const sapInstances = sapSystemApplicationInstanceFactory.buildList(2, {
+        sap_system_id: id,
+        sid: duplicateSID,
+        host_id: host.id,
+      });
+      const databaseInstances = databaseInstanceFactory.buildList(2, {
+        sid: duplicateSID,
```
arbulu89, line 139: This should have the same `id`, right?
As we are doing the filtering by id:
`database_id: id`

arbulu89 (changes requested):
Some little corrections.

assets/js/pages/HostsList/HostsList.test.jsx
EMaksy: Ah yeah, thanks you are right!

assets/js/pages/HostsList/HostsList.test.jsx
```diff
@@ -121,6 +125,38 @@ describe('HostsLists component', () => {
         ).toBeVisible()
       );
     });
+
+    it('should show only unique SIDs', async () => {
+      const host = hostFactory.build();
+      const duplicateSID = faker.string.alpha({ casing: 'upper', count: 3 });
```
arbulu89, line 131: It looks more appropriate `duplicatedSID` with the final `d`

assets/js/pages/HostsList/HostsList.test.jsx
```diff
@@ -121,6 +125,38 @@ describe('HostsLists component', () => {
...
+        host_id: host.id,
+        sid: duplicateSID,
+      });
+      const state = {
+        ...defaultInitialState,
+        hostsList: {
+          hosts: [host],
+        },
+        sapSystemsList: {
+          applicationInstances: [...sapInstances],
```
arbulu89, line 149: you don't need to expand values in a list, it is already a list hehe
`applicationInstances: sapInstances` (i would call it as it is in fact, applicationInstances).
The same with `databaseInstances. 

In fact, most probably, you can just remove them after the `:`, as they have the same name, and they should be just used that way.
Give a try (it might be prettied by the formatter)

```
sapSystemsList: { applicationInstances },
databasesList: { databaseInstances },
```

arbulu89 (approved):
Good!
A couple other cosmetic suggestions

PR #2518, opened 2024-04-15
Fix fqdn nil transition on host restoration and host detail changes

nelsonkopliku:
# Description

Tiny fix taking care of the proper nil transitions of the fqdn on host restoration+host detail changes.

## How was this tested?

Automated tests


arbulu89 (approved):
Nice!!!

PR #2519, opened 2024-04-15
Issue software updates discoveries on save/change settings

nelsonkopliku:
# Description

This PR adds the ability to trigger software updates discovery process for all the registered hosts when saving or changin settings.
It is implemented in a best effort fashion: if dispatching of `DiscoverSoftwareUpdates` for any of the host fails (unlikely) the overall operation does not fail and saving credentials is not affected.

This is needed when:
- there are previously registered hosts, but credentials  haven't been saved yet
- settings (and previous discoveries) has been cleaned and then new credentials are saved 
- currently saved settings are changed

An extra clause in the host aggregate has been added that does not emit any `SoftwareUpdatesDiscoveryRequested` when a `DiscoverSoftwareUpdates` is dispatched for a host without an fqdn (did not see yet a usecase to return an error)

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/hosts/host.ex
+++ b/lib/trento/hosts/host.ex
@@ -549,6 +549,12 @@ defmodule Trento.Hosts.Host do
   # Software Updates Discovery
 
   def execute(
+        %Host{fully_qualified_domain_name: nil},
+        %DiscoverSoftwareUpdates{}
+      ),
+      do: []
+
+  def execute(
         %Host{
           host_id: host_id,
           fully_qualified_domain_name: fully_qualified_domain_name
--- a/lib/trento/software_updates/discovery.ex
+++ b/lib/trento/software_updates/discovery.ex
@@ -7,8 +7,7 @@ defmodule Trento.SoftwareUpdates.Discovery do
 
   alias Trento.Hosts.Commands.{
     ClearSoftwareUpdatesDiscovery,
-    CompleteSoftwareUpdatesDiscovery,
-    DiscoverSoftwareUpdates
+    CompleteSoftwareUpdatesDiscovery
   }
 
   alias Trento.Hosts.Projections.HostReadModel
@@ -37,14 +36,24 @@ defmodule Trento.SoftwareUpdates.Discovery do
   def get_upgradable_packages(system_id),
     do: adapter().get_upgradable_packages(system_id)
 
-  @spec discover_software_updates :: :ok
+  @spec discover_software_updates :: {:ok, {list(), list()}}
   def discover_software_updates do
-    Enum.each(Hosts.get_all_hosts(), fn
-      %HostReadModel{id: host_id} ->
-        %{host_id: host_id}
-        |> DiscoverSoftwareUpdates.new!()
-        |> commanded().dispatch()
-    end)
+    {:ok,
+     Hosts.get_all_hosts()
+     |> Enum.map(fn
+       %HostReadModel{id: host_id, fully_qualified_domain_name: fully_qualified_domain_name} ->
+         case discover_host_software_updates(host_id, fully_qualified_domain_name) do
+           {:error, error} ->
+             {:error, host_id, error}
+
+           {:ok, _, _, _} = success ->
+             success
+         end
+     end)
+     |> Enum.split_with(fn
+       {:ok, _, _, _} -> true
+       _ -> false
+     end)}
   end
 
   @spec clear_software_updates_discoveries :: :ok | {:error, any()}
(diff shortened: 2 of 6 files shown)


lib/trento/software_updates/discovery.ex
```diff
@@ -36,24 +37,14 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
-           {:ok, _, _, _} = success ->
-             success
-         end
-     end)
-     |> Enum.split_with(fn
-       {:ok, _, _, _} -> true
-       _ -> false
-     end)}
+    Enum.each(Hosts.get_all_hosts(), fn
+      %HostReadModel{id: host_id} ->
```
arbulu89, line 43: Do we need to go through a command/aggregate on this?
I mean, it already was doing what we need.
I understand that this change is done to have the `SoftwareUpdatesDiscoveryRequested` event, but honestly, I think that is better not to have these kind of events in user interactions. We discarded this for wanda for example. At the end, the event is not changing anything in the aggregate, and we are adding extra steps to the process for something that already works fine.

arbulu89 (commented):
I think we already talked about this offline.
Anyway, I don't know if changes in the `SoftwareUpdatesDiscoveryCompleted` to `SoftwareUpdatesHealthUpdated` does impact on this PR

lib/trento/software_updates/discovery.ex
nelsonkopliku: To me the process is 
`DiscoverSoftwareUpdates` -> `SoftwareUpdatesDiscoveryRequested` 
-> event handler -> `CompleteSoftwareUpdatesDiscovery` (or `UpdateSoftwareUpdatesDiscoveryHealth`) 
-> `SoftwareUpdatesDiscoveryCompleted` (or `SoftwareUpdatesHealthUpdated`/`Changed`)
regardless of where it is triggered. 
It is just consistent. Also in the history of events.

> the event is not changing anything in the aggregate
>
correct, for now we do not change any state.
However we're about to change `software_updates_discovery_health` to support also `not_set` and we might also have the opportunity to set the state to something like `running` maybe and have that information broadcasted to interested parties (the frontend for example).

Besides this the code is much simplified.

The main rationale is consistency in the process and simplification of that piece of code.
The extra opportunity for the `running` state is just a nice side effect.
What do you think?

nelsonkopliku:
> I think we already talked about this offline. Anyway, I don't know if changes in the `SoftwareUpdatesDiscoveryCompleted` to `SoftwareUpdatesHealthUpdated` does impact on this PR

@arbulu89 Thinking of addressing that in a followup PR. Would it be ok?

lib/trento/software_updates/discovery.ex
arbulu89: For me it just makes thinks more complex. Instead of a direct path (1 function), we use a function, command, event, event listener, function. (from command on, multiplied by the number of hosts).

All of this, just to add a event in the store, which actually I would prefer not to have it.

Imagine that you have 100 hosts, and you start saving SUMA settings from the frontend. Every single change emits 100 events. And if you fail to put the correct data in the form (which we allow), more.
For a thing, that is not a real event (from my point of view), it is just a user intervention.
More over, we had the same discussion for wanda, and decided not to add such events.

We don't necessarily need to go through the same process to reach the same goal, discover udpates.

nelsonkopliku: the problem with many events hosts is that we'd need to do many requests to suma, sequentially.

Yes we could add whatever library do do concurrent things, but we'd get that concurrency for free (or almost free with an extra event in the history) by just leveraging current architecture.

nelsonkopliku: but still, I'll revert the change and see what happens

lib/trento/hosts/host.ex
```diff
@@ -548,6 +548,12 @@ defmodule Trento.Hosts.Host do
 
   # Software Updates Discovery
 
+  def execute(
```
arbulu89, line 551: We don't need this now, right?

arbulu89 (approved):
Hey @nelsonkopliku ,
I wouldn't personally put the discovery in the `save_settings` functions. At the, end `save_settings` should save settings. I would put it in the controller itself (which is the unique point where we do it).

Either way, the end result is the same, so I'm good with the code

lib/trento/hosts/host.ex
nelsonkopliku: well does not harm and keeps away any possibility of weird things happening :smile:

PR #2520, opened 2024-04-15
Fix empty relevant patches visualization

arbulu89:
# Description

Fix empty relevant patches visualization. When relevant patches is coming as an empty list (all green!), a unknown icon was being shown.

image

Some type missmatch between string and int (thank you js).
In fact, storybook and the tests were done in a way that actual failure was masked hehe

Check the PR env to see the `0` value and other value for hosts `vmdrbddev02` (unknown) and `vmhdbdev01` (2 patches).
Damn: the PR env doesn't show what I want, as it doesn't mock suma calls as expected, as it is released as `prod`

## How was this tested?

Tests updated, and manually confirmed


nelsonkopliku (approved):
Thanks!

jamie-suse (approved):
Makes sense, good spot! :eye:

PR #2521, opened 2024-04-15
Add photofinish scenario for multi-tenant case

rtorrero:
# Description

This PR:
 - adds a new photofinish scenario from a system that can take advantage of the new multi-tenant support.
 - makes use of this scenario in the demo deployment (CI) and the PRENV feature


arbulu89 (approved):
Hey @rtorrero ,
All good.
My unique doubt is about the host `agent_version`, this shows as `.gitxx` version.
@abravosuse what do you think? should we maybe better show `2.2.0` or `2.3.0` even?
You can see it working in https://2521.prenv.trento.suse.com

abravosuse:
> Hey @rtorrero , All good. My unique doubt is about the host `agent_version`, this shows as `.gitxx` version. @abravosuse what do you think? should we maybe better show `2.2.0` or `2.3.0` even? You can see it working in https://2521.prenv.trento.suse.com

Since this is a 2.3.0 feature, I think we should go for '2.3.0'.

PR #2524, opened 2024-04-16
Fetch developers from GH API instead of hardcoded list

rtorrero:
# Description

This PR aims to make it easier to maintain the list of maintainers by using the GH API to fetch the developers group from our organization instead of having two separate hardcoded lists in the CI yaml.

As a test, I've added the two relevant labels to see if they would run, and they run correctly.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,6 @@ env:
   OTP_VERSION: 26
   MIX_ENV: test
   NODE_VERSION: "16"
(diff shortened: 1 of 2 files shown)


EMaksy (approved):
Hey man, seems good to me :)

arbulu89 (commented):
Honestly, i don't like it hehe
Do we need to add this to any place where we want to look for maintaners?
There is not this solved be GH already somehow.
I would stick to the previous hardcoded list if it was me hehe

Let's see what others say

rtorrero:
@arbulu89 I tested the GH action you shared on slack and it doesn't seem to work. Returns a 404 internally when using the API apparently.

I'll let this stall here for a bit unless someone else wants to add anything. We can close it down afterwards, no hard feeling :+1:

PR #2525, opened 2024-04-16
Parallelize SUMA queries using ParallelStream

arbulu89:
# Description

Parallelize SUMA queries during the discovery. Using [ParallelStream](https://github.com/beatrichartz/parallel_stream/tree/main) we can easily achieve that.
It uses a typical workers pool behind the scene to balance the concurrency. The defaults values are here: https://github.com/beatrichartz/parallel_stream/blob/main/lib/parallel_stream/defaults.ex

## How was this tested?

Tests update. I have tested a bit manually, adding some sleeps here and there to see that all the tasks are done in parallel, which are


nelsonkopliku (approved):
Thanks @arbulu89 :rocket:

abravosuse (approved):
Both following test pass:

- empty trento + empty settings-> save settings -> new hosts discovered + software updates discovery happening
- populated trento + empty settings -> save settings -> software updates discovery happening

jamie-suse (approved):
nice :rocket:

PR #2526, opened 2024-04-16
Add a generate sid function and fix bug

EMaksy:
# Description

This pr fixes a small tech debt, by adding a helper function for generating SIDs. 
This will improve the code readability for tests and factories, and also we used faker wrongly, instead of generating three upper-cased letters it just generated one.

### DIFF
--- a/assets/js/lib/test-utils/factories/sapSystems.js
+++ b/assets/js/lib/test-utils/factories/sapSystems.js
@@ -2,7 +2,7 @@
 import { faker } from '@faker-js/faker';
 import { Factory } from 'fishery';
 
-import { databaseInstanceFactory } from './databases';
+import { databaseInstanceFactory, generateSid } from '.';
 
 const ensaVersion = () =>
   faker.helpers.arrayElement(['no_ensa', 'ensa1', 'ensa2']);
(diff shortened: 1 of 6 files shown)


assets/js/lib/test-utils/factories/sapSystems.js
```diff
@@ -26,26 +26,25 @@ export const sapSystemApplicationInstanceFactory = Factory.define(() => ({
...
+  const databaseSid = params.database_sid || generateSid();
   return {
     application_instances: sapSystemApplicationInstanceFactory.buildList(2, {
       sap_system_id: id,
       sid,
     }),
     database_instances: databaseInstanceFactory.buildList(2, {
       database_id: databaseId,
-      sid: faker.string.alphanumeric(3, { casing: 'upper' }),
+      sid: generateSid(),
```
arbulu89, line 47: Shouldn't this and the next one be `databaseSid`, generated on top?
I mean, a SAP system all database instances should have the same SID, right?

arbulu89 (approved):
Good!
Just one comment.
Did you check if in any other place we generate the SID with UUID or any other faker function? I guess yes, but it wouldn't harm to recheck hehe

assets/js/lib/test-utils/factories/sapSystems.js
EMaksy: Yeah, you are fully right with this!

EMaksy:
Thanks @arbulu89, I took a look and think I caught them all :+1:

rtorrero (approved):
Nice :-)! looks cleaner indeed :+1:

CDimonaco (approved):

PR #2527, opened 2024-04-16
Add missed permission check

rtorrero:
# Description

This PR adds a missed permission check in the CI for the regression test


arbulu89 (approved):
:rocket:

PR #2528, opened 2024-04-17
Change `software_updates_discovery_health` default value

nelsonkopliku:
# Description

In this PR  
- the default state for the host's `software_updates_discovery_health` is changed from `unknown` to `not_set` (required in followup changes setting `software_updates_discovery_health` and consequently host's health to unknown on discovery errors)
- the calculation of the resulting `software_updates_discovery_health` based on discovered patches is moved from the Host aggregate to the Discovery
- `SoftwareUpdatesDiscoveryCompleted` is changed to `SoftwareUpdatesHealthChanged` carrying only the computed health

## How was this tested?

Automated tests.


lib/trento/hosts/commands/complete_software_updates_discovery.ex
```diff
@@ -3,14 +3,14 @@ defmodule Trento.Hosts.Commands.CompleteSoftwareUpdatesDiscovery do
   Complete the software updates discovery with the detected info
   """
 
-  alias Trento.Hosts.ValueObjects.RelevantPatches
+  require Trento.SoftwareUpdates.Enums.SoftwareUpdatesHealth, as: SoftwareUpdatesHealth
```
arbulu89, line 6: Maybe we can change the docstring on top

lib/trento/software_updates/enums/software_updates_health.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Trento.SoftwareUpdates.Enums.SoftwareUpdatesHealth do
+  @moduledoc """
+  Type that represents the possible health values for the software updates discovery process.
+  """
+
+  use Trento.Support.Enum, values: [:passing, :warning, :critical, :unknown, :not_set]
```
arbulu89, line 6: Maybe `unset` is more appropriate that `not_set`.
What does our anglo native @jamie-suse say hehe ?

lib/trento/hosts/host.ex
```diff
@@ -571,22 +571,25 @@ defmodule Trento.Hosts.Host do
...
           host_id: host_id,
-          relevant_patches: relevant_patches
+          health: health
         }
       ) do
     host
     |> Multi.new()
     |> Multi.execute(fn _ ->
-      %SoftwareUpdatesDiscoveryCompleted{
+      %SoftwareUpdatesHealthChanged{
```
arbulu89, line 580: We miss the `if` here:
```
# current_health coming from the aggregate `software_updates_discovery_health` field
if current_health != health do
  %SofwareUpdatesHealthChanged{...
end
```
Like we do with other healths

arbulu89 (approved):
Looking good!
My unique concern is that we don't check for the current software discovery health when we receive the command, because if the health is the same, we don't need to emit the event.
Besides that, (except `no_set/unset` term, which I hope @jamie-suse can help), everything looks good.

PD: Let's not merge this in a hurry. We need to see the currently changed things in demo (like tags, selected checks, etc), as we will need to recreate the db and event store

PR #2530, opened 2024-04-17
Properly clear suma genserver state when clearing software updates settings

nelsonkopliku:
# Description

This makes sure the state of suma genserver is cleared up always when clearing software updates settings, otherwise might happen that genserver state is kept, causing discoveries to happen when they should not.

## How was this tested?

Automated tests and on a real infrastructure.

### DIFF
--- a/lib/trento/software_updates/discovery.ex
+++ b/lib/trento/software_updates/discovery.ex
@@ -58,9 +58,7 @@ defmodule Trento.SoftwareUpdates.Discovery do
 
   @spec clear_software_updates_discoveries :: :ok | {:error, any()}
   def clear_software_updates_discoveries do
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):
Yes, I think it is better to clean in any scenario.

Anyway, the clear simply cleans the auth cookie, right?
I don't know how that could affect doing not needed discoveries, but well, still think it is a good change. I guess we spare one potentially useless query to suma, that would fail.

nelsonkopliku:
@arbulu89 the problem was that when clearing up suma settings when there were no hosts in the system did clear the entry in the database but not the genserver state :see_no_evil:

nelsonkopliku:
correct, if there was any authentication. In case what was in memory was good, no change needed.

PR #2532, opened 2024-04-18
Handle software updates discovery errors

nelsonkopliku:
# Description

This PR handles software updates discovery errors by dispatching an `unknown` health.

Cleaned a bit up error propagation.

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/software_updates/discovery.ex
+++ b/lib/trento/software_updates/discovery.ex
@@ -80,32 +80,54 @@ defmodule Trento.SoftwareUpdates.Discovery do
   end
 
   def discover_host_software_updates(host_id, fully_qualified_domain_name) do
-    with {:ok, system_id} <- get_system_id(fully_qualified_domain_name),
-         {:ok, relevant_patches} <- get_relevant_patches(system_id),
-         :ok <-
-           host_id
-           |> build_discovery_completion_command(relevant_patches)
-           |> commanded().dispatch() do
-      {:ok, host_id, system_id, relevant_patches}
-    else
-      error ->
+    case discover_relevant_patches(fully_qualified_domain_name) do
+      {:ok, system_id, relevant_patches} ->
+        determine_health_and_dispatch_completion_command(host_id, system_id, relevant_patches)
+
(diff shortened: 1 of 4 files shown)


lib/trento/software_updates/discovery.ex
```diff
@@ -80,32 +80,54 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
+      {:ok, system_id, relevant_patches} ->
+        determine_health_and_dispatch_completion_command(host_id, system_id, relevant_patches)
+
+      {:error, discovery_error} = error ->
         Logger.error(
           "An error occurred during software updates discovery for host #{host_id}:  #{inspect(error)}"
         )
 
-        {:error, error}
+        dispatch_completion_command_on_discovery_error(host_id, discovery_error)
```
arbulu89, line 92: I don't understand a thing :sweat_smile: 
Wasn't as simple as putting the the dispatch with the `unknown` state in the `else` side?

```
commanded().dispatch(
      CompleteSoftwareUpdatesDiscovery.new!(%{
        host_id: host_id,
        health: SoftwareUpdatesHealth.unknown()
    })
)

error
```

arbulu89 (commented):
Hey @nelsonkopliku ,
I find the change overly complex.
Why not simply dispatch the command and returning the error?

In fact, the dispatch error case looks a bit over defensive. Our system won't send this command to any host that is not already registered.

I mean, if in the `unknown` state, if we receive `host_not_registered` (I don't know when), who cares, we still need to return the original error. At the end, we just want to know if there was an error or not. We don't use the specifics anywhere. I mean, what use has this `dispatching error` (that why the way, you don't send with the same format for the correct flow of it fails)?

nelsonkopliku:
@arbulu89 thanks for the review, I applied the simplified version. 
I maybe have some PTSD on over-simplifying things and unwanted side effects :sweat_smile:

arbulu89 (approved):
Thank you!
PD: This will conflict with my other PR:https://github.com/trento-project/web/pull/2531
But we can fix the conflicts afterwards

PR #2535, opened 2024-04-19
Adjust suma timeouts

nelsonkopliku:
# Description

Adjusting timeouts for suma requests.

By default all requests to suma have been set to `[timeout: 500, recv_timeout: 1_500]`

> https://hexdocs.pm/httpoison/HTTPoison.Request.html
> 
> `:timeout` - timeout for establishing a TCP or SSL connection, in milliseconds. Default is 8000
> `:recv_timeout` - timeout for receiving an HTTP response from the socket. Default is 5000

Also the SUMA genserver call timeout has been increased to `15s` which covers the worst case scenarios where:
- all 5 retrials of login fail responding in time ~10s or less
- login succeeds only on the last try plus discovery request: 4 login failures + 1 login success + 1 specific api for a total amount of ~12s or less

This would benefit also applied in this case https://github.com/trento-project/web/pull/2531. 
And besides that, as discussed with @arbulu89, we could also consider breaking the loop of login retrial if we detect a 401 during a login attempt.

## How was this tested?

Manual testing.

### DIFF
--- a/lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
+++ b/lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
@@ -51,7 +51,7 @@ defmodule Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor do
       "#{base_url}/auth/login",
       payload,
       [{"Content-type", "application/json"}],
-      ssl_options(use_ca_cert) ++ [timeout: 1_000]
+      ssl_options(use_ca_cert) ++ timeout_options()
     )
   end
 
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):
I guess that this reduces the authentication failure time.
Anyway, we still have a `raise` error only with this change.
I guess https://github.com/trento-project/web/pull/2531 (or whatever other solution) would fix this

nelsonkopliku:
Not only auth, but every call. And it makes timeouts kinda predictable.
And yes, we still need to adjust where the synchronization point happens.

PR #2540, opened 2024-04-23
Track software update discoveries

nelsonkopliku:
# Description

This PR introduces tracking of software updates discovery results.

- on a successful discovery `relevant_patches` and `upgradable_packages` of a host are tracked
- on a failed discovery the `failure reason` is tracked (issues when authenticating/system id retrieval/relevant patches retrieval/upgradable packages retrieval)
- on clear up, relevant previously tracked discoveries are removed

Note that there can be at most one tracked discovery per host, that gets replaced by newer discoveries.

What's next: the tracked discoveries can be queried in `/hosts/:host_id/software_updates` and possibly later also on overviews.

## How was this tested?

Automated tests and manually.


dottorblaster (approved):
LGTM

The approach is fine, the naming of some functions doesn't sound so good but I think we can revisit this later

PR #2541, opened 2024-04-23
Fix hana scale out no hooks

arbulu89:
# Description

Fix 2 things in HANA scale out clusters discovery:
- Split properly the secondary site name. Using simply `_` caused errors if the site name has this character
- Use a more global attribute to get the secondary sync state

## How was this tested?

New test added and tested with scenario given by Alberto

### DIFF
--- a/lib/trento/discovery/policies/cluster_policy.ex
+++ b/lib/trento/discovery/policies/cluster_policy.ex
@@ -656,7 +656,7 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
       ^secondary_site ->
         parse_crm_cluster_property(
           cluster_properties,
-          "hana_#{String.downcase(sid)}_glob_srHook",
+          "hana_#{String.downcase(sid)}_glob_sync_state",
           "Unknown"
         )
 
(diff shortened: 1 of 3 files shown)


abravosuse (approved):
Tested on HANA scale-out cluster with no hooks. It works.

lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -494,7 +494,7 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
     |> Enum.filter(fn
       %{name: name} -> String.starts_with?(name, "hana_#{String.downcase(sid)}_site_lss_")
     end)
-    |> Enum.map(fn %{name: name} -> name |> String.split("_") |> Enum.at(-1) end)
+    |> Enum.map(fn %{name: name} -> name |> String.split("_site_lss_") |> Enum.at(-1) end)
```
rtorrero, line 497: I guess we are using now instead `_site_lss_` as the string that we use to mark the split, wouldn't it make sense to define this as a constant better?

rtorrero (commented):
1 small comment, when clarified we are good to go :+1:

lib/trento/discovery/policies/cluster_policy.ex
arbulu89: well, we have this string parts spread all around the policy, and they are not used anywhere else.
i find it acceptable

abravosuse:
@arbulu89 confirmed that the discovery of clusters with srHooks enabled still works with these changes.

rtorrero (approved):

PR #2543, opened 2024-04-23
Create Abilities and UserAbilities schema

CDimonaco:
# Description

Create Abilities and UsersAbilities schema and context

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/abilities.ex
+++ b/lib/trento/abilities.ex
@@ -1,0 +1,104 @@
+defmodule Trento.Abilities do
+  @moduledoc """
+  The Abilities context.
+  """
+
+  import Ecto.Query, warn: false
+  alias Trento.Repo
+
+  alias Trento.Abilities.Ability
+
+  @doc """
+  Returns the list of abilities.
+
+  ## Examples
+
+      iex> list_abilities()
+      [%Ability{}, ...]
+
+  """
+  def list_abilities do
+    Repo.all(Ability)
+  end
+
+  @doc """
+  Gets a single ability.
+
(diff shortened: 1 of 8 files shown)


lib/trento/abilities/users_abilities.ex
```diff
@@ -0,0 +1,24 @@
+defmodule Trento.Abilities.UsersAbilities do
+  @moduledoc """
+  Many to main table schema used to connect users and abilities.
```
arbulu89, line 3: `Many to main` -> `Many to many`.

And the next maybe better:
`We have a dedicated schema to implement the association between users and abilities, to just enable read only operation. Using the default ecto schema, declaring a user with a not existing ability would trigger the creation of this second, and we don't want to allow that. Abilities are just read only.

lib/trento/abilities/ability.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Trento.Abilities.Ability do
+  @moduledoc """
```
arbulu89, line 2: Let's add a docstring or put `@moduledoc false`.
I would maybe document that the `label` is just an informative field to describe what the ability does

lib/trento/abilities.ex
```diff
@@ -0,0 +1,104 @@
+defmodule Trento.Abilities do
```
arbulu89, line 1: Thank you phoenix :rocket:

arbulu89 (approved):
Perfect! So cool

lib/trento/abilities/users_abilities.ex
CDimonaco: done

PR #2544, opened 2024-04-23
Add New User Management View

EMaksy:
# Description

This pr adds a user management view. (Uses only crud) 
## Preview: 
User View

Currently, the view fetches all existing users from the backend and renders them using our Table component. 
An admin user is created when installing Trento and must not be deleted by the user, this means: 

- Delete button is disabled for the admin user. (a tooltip on hover will inform the user)
- Delete button is enabled for all other users.

By clicking on the Delete Button, a modal is opened to ask the user if he is sure that he wants to delete the displayed user. 
Modal open

Pressing the cancel button closes the modal and the user is not deleted.
By pressing the deleted button a second time, the user is **soft deleted** in the backend and not shown anymore in the user's view. 
Pressing on the username will redirect to the user edit page with the `users/:id/edit `route
Pressing on Create User will redirect to the user creation page with `/users/new `route

## What was added?
- adds new user's management view
- new css button classes to meet design requirements
- added api calls for listing all and deleting users
- added user Factory to create users and admin user data
- adds an icon on the left side nav bar that navigates to the users view
- adds new routes to trento index page
- adds  storybook story

## How was this tested?

Added test for Users view: 
- Testing empty table
- Testing table with admin user
- Testing table with multiple users

## What is missing? 
- E2E test will come in a separate pr.


assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
...
+const USER_CREATE_ROUTE = '/users/new';
+
+function Users({
+  handleDeleteUser = () => {},
+  navigate = () => {},
+  setModalOpen = () => {},
+  setDeleteUserId = () => {},
+  deleteUserId = 0,
+  modalOpen = false,
+  users = [],
```
arbulu89, line 20: As far as I remember, we don't usually put an empty list as is as default.
There was some error with it.
We just define a constant on top (`defaultUsers`) and assign it.
I don't remember why, but we do it everywhere

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
...
+import PageHeader from '@common/PageHeader';
+import Modal from '@common/Modal';
+import Tooltip from '@common/Tooltip';
+import Banner from '@common/Banners/Banner';
+import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
+
+const USER_CREATE_ROUTE = '/users/new';
+
+function Users({
+  handleDeleteUser = () => {},
```
arbulu89, line 14: Let's use `lodash` `noop` for declaring defaults

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import Button from '@common/Button';
+import Table from '@common/Table';
+import PageHeader from '@common/PageHeader';
+import Modal from '@common/Modal';
+import Tooltip from '@common/Tooltip';
+import Banner from '@common/Banners/Banner';
+import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
```
arbulu89, line 9: External imports usually go before the internal ones (`common`)

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
...
+              >
+                Delete
+              </Button>
+            ) : (
+              <Tooltip content="Admin user can not be deleted">
+                <Button type="danger-op-50-no-border">Delete</Button>
+              </Tooltip>
+            )}
+
+            {modalOpen && deleteUserId === item.id && (
```
arbulu89, line 77: Let's not put the modal in the table configuration. We should put it in the main `return`
(or even other file if you prefer, but i find alright to have it in this)

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
...
+        title: 'Created',
+        key: 'created',
+      },
+
+      {
+        title: 'Actions',
+        key: 'actions',
+        render: (content, item) => (
+          <>
+            {item.id !== 1 ? (
```
arbulu89, line 61: I don't think we need this `if/else` case.
We should use the `disabled` prop for the button, and simply choose the styles based on the user id

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
...
+import { Link } from 'react-router-dom';
+import Button from '@common/Button';
+import Table from '@common/Table';
+import PageHeader from '@common/PageHeader';
+import Modal from '@common/Modal';
+import Tooltip from '@common/Tooltip';
+import Banner from '@common/Banners/Banner';
+import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
+
+const USER_CREATE_ROUTE = '/users/new';
```
arbulu89, line 11: I would put this simply in the `navigate`, we use it once, and it is even easier to see what happens.
What do you think?

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
...
+            )}
+          </>
+        ),
+      },
+    ],
+  };
+
+  return (
+    <div className="flex flex-wrap">
+      <div className="flex w-1/2 h-auto overflow-hidden overflow-ellipsis break-words">
```
arbulu89, line 125: Do we need this div, with the ellipsis, etc?
I mean, this is a fixed header, just `PageHeader` would make it, or not?
Or maybe we need to for alignment?

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
...
+          <Button
+            className="inline-block mx-1 border-green-500 border"
+            size="small"
+            onClick={() => navigate(USER_CREATE_ROUTE)}
+          >
+            Create User
+          </Button>
+        </div>
+      </div>
+      {loading ? (
```
arbulu89, line 139: Do we want this `loading` stage?
I mean, we don't use it in any other table.
We would have the table `empty` for a sec

assets/js/pages/Users/Users.jsx
arbulu89, line 14: Better `onDeleteUser`

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,154 @@
...
+import Tooltip from '@common/Tooltip';
+import Banner from '@common/Banners/Banner';
+import { EOS_LOADING_ANIMATED } from 'eos-icons-react';
+
+const USER_CREATE_ROUTE = '/users/new';
+
+function Users({
+  handleDeleteUser = () => {},
+  navigate = () => {},
+  setModalOpen = () => {},
```
arbulu89, line 16: Why do we need to pass this `setModalOpen`, `setDeleteUserId`, `deleteUserId`, and `modalOpen` from outside. They look inherent and pure for this page.

assets/js/pages/Users/Users.stories.jsx
```diff
@@ -0,0 +1,111 @@
...
+};
+export const EmptyUsersTable = {
+  args: {
+    users: [],
+  },
+};
+
+export const UsersOverview = {
+  args: {
+    users: [
```
arbulu89, line 92: we have factories for this hehe

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,82 @@
...
+  }, [userUpdateTrigger]);
+
+  useEffect(() => {
+    if (error) {
+      toast.error(error);
+      setError(null);
+    }
+  }, [error]);
+
+  const usersTableData = userData.map(
```
arbulu89, line 56: I would say this mapping belongs to the table code itself.
It shouldn't exist, to the trasformations in the table itself

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,82 @@
...
+  const [error, setError] = useState(null);
+
+  const navigate = useNavigate();
+
+  const fetchUsers = () => {
+    setLoading(true);
+    listUsers()
+      .then((response) => {
+        setUserData(response.data);
+        setLoading(false);
```
arbulu89, line 25: If we still use the loading, you don't need to do it twice. 
You should have a `finally` entry to do it at the end, for both positive and error cases

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,82 @@
...
+      })
+      .catch((_error) => {
+        setError('An error occurred during loading users');
+        setUserData([]);
+        setLoading(false);
+      });
+  };
+
+  const handleDeleteUser = (userId) => {
```
arbulu89, line 34: `onDeleteUser`

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,82 @@
...
+    deleteUser(userId)
+      .then(() => {
+        setUserUpdateTrigger(true);
```
arbulu89, line 37: Do we want to fetch users again here?
It is not simpler just removing the entry from the list.
Less operations...
And we know that it succedded.

arbulu89 (changes requested):
A lot of comments :sweat_smile: 

Other generic comment.
I think we should have UT for the table itself, which are super easy to do, just passing users as props, and other simpler one for the page, where you mock fetch and delete, so you only have 2 tests here

jagabomb:
Thanks for this PR @EMaksy. This looks good to me so far. I have two minor pieces of feedback for now:

1. For the **Delete User** pop-up/modal, can we change the warning banner word "_action_" to lowercase so that it reads as follows, "_This action cannot be undone._"
2. For the loading state of the **Users** component, should the **Create** button also not be hidden (@arbulu89 what do you think, screenshot is supplied below)?
Screenshot 2024-04-25 at 09 31 46

arbulu89:
@jagabomb yes, i would hide it, or at least disable it.
Anyway, I would be still fine seeing the transition from `Empty data` in the table to a populated data..

assets/js/lib/api/users.js
```diff
@@ -0,0 +1,5 @@
+import { networkClient } from '@lib/network';
+
+export const listUsers = () => networkClient.get('/users');
+
+export const deleteUser = (userId) => networkClient.delete(`/users/${userId}`);
```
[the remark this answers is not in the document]
EMaksy: We can have a tech debt tracked and finally stick with one haha, I hope we don't need an ADR for that xD

I stick with `userID `:+1:

assets/js/pages/Users/Users.jsx
CDimonaco: +1 The visual impact of the component, without any external side effect should be contained in the component, with the proper state, the external actions like perform the api call for delete should be injected.

assets/js/pages/Users/Users.jsx
CDimonaco: +1, also the modal should be in the dom, it's the open prop that triggers the opening, don't wrap the modal into another clause, it has it's own clause for opening/closing

assets/js/pages/Users/Users.jsx
CDimonaco: I proposed the loading tbh, I think it's better to have even if it's little time, maybe we could add a more smooth transition with a fixed delay

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,82 @@
...
+import { listUsers, deleteUser } from '@lib/api/users';
+import { format, parseISO } from 'date-fns';
+
+import { toast } from 'react-hot-toast';
+
+import Users from './Users';
+
+function UsersPage() {
+  const [modalOpen, setModalOpen] = useState(false);
+  const [deleteUserId, setDeleteUserId] = useState(null);
```
CDimonaco, line 12: Like said before, the visual thing of the inner component should be in the inner component.

We should follow a dummy/container components approach


This UsersPage is the container, that contains logic, external dependencies, api call, the inner component is dummy, but controlled in it's visual things and use the prop passed from outside just for side effects, like api calls

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+  const fetchUsers = () => {
+    setLoading(true);
+    listUsers()
+      .then((response) => {
+        setUserData(response.data);
+        setLoading(false);
+      })
+      .catch((_error) => {
+        setError('An error occurred during loading users');
```
CDimonaco, line 28: Please the text as const

assets/js/pages/Users/UsersPage.jsx
CDimonaco: +1

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,82 @@
...
+      })
+      .catch((_error) => {
+        setError('An error occurred during deleting user');
+      });
+  };
+
+  useEffect(() => {
+    fetchUsers();
+    setUserUpdateTrigger(false);
+  }, [userUpdateTrigger]);
```
CDimonaco, line 47: What's this user update trigger?

What we want to trigger? Simply fetch the users initiall on mount, and then , call explicitely the fetch users function to refetch.

Please don't use use effects like that. We don't have any trigger, effects react when things changes and they are meaningful for the domain.

assets/js/pages/Users/UsersPage.jsx
CDimonaco: +1

CDimonaco (dismissed):
Everything @arbulu89 said, plus some other comments,

Overall good job but please take attention to some details.

The updateTrigger in the use effect is the most important to tackle

assets/js/pages/Users/UsersPage.jsx
CDimonaco: I think refetching the users after a delete it's a better choice, we don't have inconsistencies and the call is cheap, manipulating. It's better to have the data api immutable and display it

Just don't use the trigger and call the function directly

assets/js/pages/Users/Users.jsx
EMaksy: It is required for alignment, I would stick with it, as we reuse it also in other parts of the app

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,165 @@
...
+        key: 'enabled',
+        render: (content, item) => (
+          <span>{item.enabled ? 'Enabled' : 'Disabled'}</span>
+        ),
+      },
+      {
+        title: 'Created',
+        key: 'created_at',
+        render: (content, item) => (
+          <span> {format(parseISO(item.created_at), 'MMMM dd, yyyy')} </span>
```
arbulu89, line 63: We could put `'PPpp'` as format to show the hour and minute.
Anyway, what Jurgen says :stuck_out_tongue_closed_eyes:

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,165 @@
...
+        title: 'Actions',
+        key: 'actions',
+        render: (content, item) => (
+          <div>
+            <Tooltip
+              content="Admin user can not be deleted"
+              isEnabled={item.id === 1}
+            >
+              <Button
+                className="text-red-500"
```
arbulu89, line 76: Do we use the same `className` for admin as well?
I thought we were using a more ofuscated version

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,165 @@
...
+        title="Delete User"
+      >
+        <div className="flex flex-col my-2">
+          <Banner type="warning">
+            <span className="text-sm">This action cannot be undone.</span>
+          </Banner>
+          <span className="my-1  text-gray-500">
+            Are you sure you want to delete the following user account?
+          </span>
+          {user ? (
```
arbulu89, line 127: Why this if condition?
Is there any circunstance where the user doesn't exist?
I mean, how are we trying to delete in that case?
What you could do is instead of storing just the `deleteUserID`, you could store the whole user in that `setState`, this way, the username must always exist, so you would remove this `if` and the `getUserByID` function and its usage

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,67 @@
+import React, { useState, useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { listUsers, deleteUser } from '@lib/api/users';
+
+import { toast } from 'react-hot-toast';
+
+import Users from './Users';
+
+const successDeleteMessage = 'User deleted successfully';
```
arbulu89, line 9: I would put this constants as upper case `SUCCESS_DELETE_MESSAGE` and so on.
At the end they are real consts

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,67 @@
...
+const errorLoadingMessage = 'An error occurred during loading users';
+const errorDeletingMessage = 'An error occurred during deleting user';
+
+function UsersPage() {
+  const [loading, setLoading] = useState(false);
+  const [userData, setUserData] = useState([]);
```
arbulu89, line 15: Better maybe `users`  and `setUsers` ?

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,67 @@
...
+import { toast } from 'react-hot-toast';
+
+import Users from './Users';
+
+const successDeleteMessage = 'User deleted successfully';
+const errorLoadingMessage = 'An error occurred during loading users';
+const errorDeletingMessage = 'An error occurred during deleting user';
+
+function UsersPage() {
+  const [loading, setLoading] = useState(false);
```
arbulu89, line 14: Not a big deal, but maybe we should put loading to `true` at the beginning, so we don't have any possibility that we pass through any strange state

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,67 @@
...
+      .catch((_error) => {
+        setError(errorLoadingMessage);
+        setUserData([]);
+      })
+      .finally(() => {
+        setLoading(false);
+      });
+  };
+
+  const onDeleteUser = (userId) => {
```
arbulu89, line 35: `userID` :stuck_out_tongue_closed_eyes:

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,67 @@
...
+      .catch((_error) => {
+        setError(errorDeletingMessage);
+      });
+  };
+
+  useEffect(() => {
+    fetchUsers();
+  }, []);
+
+  useEffect(() => {
```
arbulu89, line 50: Do we need this? 
Not better to just `toast.error` in the 2 places we have it.
You would remove the `error` state

arbulu89 (commented):
Much better @EMaksy 
The prod code is looking good

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,165 @@
...
+                Delete
+              </Button>
+            </Tooltip>
+          </div>
+        ),
+      },
+    ],
+  };
+
+  const user = getUserByID(users, deleteUserID);
```
CDimonaco, line 94: Please use a proper state, this is not a good thing for the component lifecycle

assets/js/pages/Users/Users.jsx
CDimonaco: Yeah, store whatever in the state, don't use the const, and we don't need the if, in any case the user should no exist, better to crash

assets/js/pages/Users/UsersPage.jsx
CDimonaco: +1

assets/js/pages/Users/UsersPage.jsx
CDimonaco: +1

assets/js/pages/Users/UsersPage.jsx
CDimonaco: I suggested to have the error state for consistency in all the fetching sequences we do and to have it, i think it's better for consistency, having an error state could give us more flexibility even in the future

assets/js/pages/Users/UsersPage.jsx
EMaksy: Actually, it is a very good idea…
The API call is fast, loading state does not look that good but if loading is per default true, it makes the transition better for giving it a little bit more time !

assets/js/pages/Users/Users.jsx
EMaksy: Removing this helper, added selectedUser state, with this the whole component can work with user data  and this becomes obsolete :)

jagabomb (approved):
LGTM!

assets/js/lib/test-utils/factories/users.js
```diff
@@ -0,0 +1,21 @@
...
+  actions: 'Delete',
+  enabled: faker.datatype.boolean(),
+  fullname: faker.internet.displayName(),
+  email: faker.internet.email(),
+}));
+
+export const adminUser = userFactory.params({
+  id: 1,
+  username: 'admin',
+  created_at: faker.date.past().toISOString(),
```
rtorrero, line 17: Do you need to specify this again? it also does `faker.date.past().toISOString()` like the default `userFactory`, right?

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,159 @@
...
+      </div>
+      <Modal
+        open={modalOpen}
+        className="!w-3/4 !max-w-3xl"
+        onClose={() => setModalOpen(false)}
+        title="Delete User"
+      >
+        <div className="flex flex-col my-2">
+          <Banner type="warning">
+            <span className="text-sm">This action cannot be undone.</span>
```
rtorrero, line 113: I'd use the more formal `can't`

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,67 @@
...
+import { useNavigate } from 'react-router-dom';
+import { listUsers, deleteUser } from '@lib/api/users';
+
+import { toast } from 'react-hot-toast';
+
+import Users from './Users';
+
+const SUCCESS_DELETE_MESSAGE = 'User deleted successfully';
+const ERROR_LOADING_MESSAGE = 'An error occurred during loading users';
+const ERROR_DELETING_MESSAGE = 'An error occurred during deleting user';
```
rtorrero, line 11: "while loading users" or "during the loading of users"
"while deleting user" or "during the deletion of user"

assets/js/pages/Users/UsersPage.test.jsx
```diff
@@ -0,0 +1,92 @@
...
+    await userEvent.click(deleteButtons[1]);
+    expect(await screen.getByText(modalWarningText)).toBeVisible();
+    expect(await screen.getByText(bannerText)).toBeVisible();
+    expect(screen.getAllByText(user.username)[1]).toBeVisible();
+    deleteButtons = screen.getAllByText('Delete');
+    await userEvent.click(deleteButtons[2]);
+    axiosMock.onDelete(`/api/v1/users/${user.id}`).reply(200);
+    await waitFor(() => {
+      expect(axiosMock.history.delete.length).toBe(1);
+      expect(axiosMock.history.delete[0].url).toBe(`/users/${user.id}`);
```
rtorrero, line 89: my linter complains here that `screen.getAllByText()` doesn't return a promise which would make the `await` unneeded

rtorrero (commented):
Good stuff! few comments

assets/js/lib/test-utils/factories/users.js
EMaksy: I will go with the formatISO thingy we use at other parts of the app. 
```import { formatISO } from 'date-fns';
.....
  created_at: formatISO(faker.date.past()),
.....
```

assets/js/pages/Users/Users.jsx
EMaksy: @jagabomb what do you think?

assets/js/pages/Users/UsersPage.test.jsx
```diff
@@ -0,0 +1,92 @@
...
+const axiosMock = new MockAdapter(networkClient);
+
+describe('UsersPage', () => {
+  afterEach(() => {
+    axiosMock.reset();
+  });
+
+  it('should render users page with a table without data.', async () => {
+    axiosMock.onGet('/api/v1/users').reply(200, []);
+    await act(async () => {
```
arbulu89, line 23: We do you need this `act`?
This test doesn't really look an `async` test, bu well, maybe I'm wrong :sweat_smile:

assets/js/pages/Users/UsersPage.test.jsx
```diff
@@ -0,0 +1,92 @@
...
+    expect(await screen.findByText(toolTipText)).toBeVisible();
+
+    await userEvent.click(deleteButtons[1]);
+    expect(await screen.getByText(modalWarningText)).toBeVisible();
+    expect(await screen.getByText(bannerText)).toBeVisible();
+    expect(screen.getAllByText(user.username)[1]).toBeVisible();
+    deleteButtons = screen.getAllByText('Delete');
+    await userEvent.click(deleteButtons[2]);
+    axiosMock.onDelete(`/api/v1/users/${user.id}`).reply(200);
+    await waitFor(() => {
```
arbulu89, line 87: I don't think we need this `waitFor`, the `axiosMock.onDelete` takes care of asserting this thing

arbulu89 (changes requested):
@EMaksy 
It looks good, but I don't really like the tests :sweat_smile: 
You did an effort to have a totally pure component, to not use it afterwards.
I think we need 2 test files, one to test the `Users.jsx` and other to test the `UsersPage.jsx`.
Testing the 1st one is super simple, because you don't need need to mock any axios call.

About the 2nd, you just need to tests.
- Table is populated properly
- Used deletion works (or doesn't work and you raise a toast).
In fact, you don't test things that you could in the `Page` element, like what happens in error cases etc.

Having pure components is not for storybook only, it is to have easy testing

assets/js/pages/Users/Users.jsx
EMaksy: Yeah by disabling the button we achieve the same visual effect as in the mockup :+1:

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,159 @@
...
+            className="text-red-500 text-left w-auto"
+            size="small"
+            type="transparent"
+            disabled={item.id === 1}
+            onClick={() => {
+              setModalOpen(true);
+              setSelectedUser(item);
+            }}
+          >
+            <Tooltip
```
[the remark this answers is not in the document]
arbulu89: Could this be maybe solved using `size = 'fit'` ?
With this, you would removing that padding, and after all you don't need it

assets/js/pages/Users/Users.stories.jsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+import { action } from '@storybook/addon-actions';
+
+import { adminUser, userFactory } from '@lib/test-utils/factories/users';
+
+import Users from './Users';
+
+const mockedNavigate = action('navigate');
```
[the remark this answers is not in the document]
EMaksy: Yeah, I will remove it!

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+  return (
+    <div className="flex flex-wrap">
+      <div className="flex w-1/2 h-auto overflow-hidden overflow-ellipsis break-words">
+        <PageHeader className="font-bold">Users</PageHeader>
+      </div>
+      <div className="flex w-1/2 justify-end">
+        <div className="flex w-fit whitespace-nowrap">
+          <Button
+            className="inline-block mx-1 border-green-500 border"
```
jagabomb, line 96: @EMaksy Following up on our conversation on Friday, I realised that the `border-green-500` and `border` classes can be removed here since they clearly visible when the button is disabled in the Loading state.

assets/js/pages/Users/UsersPage.test.jsx
EMaksy: I thought we should use act when the state changes, played around it fails without xD

assets/js/pages/Users/Users.stories.jsx
```diff
@@ -0,0 +1,86 @@
...
+      control: { type: 'object' },
+    },
+    loading: {
+      description: 'Display loading state of the component',
+      control: { type: 'boolean' },
+    },
+  },
+};
+
+export const Default = {
```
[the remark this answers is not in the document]
EMaksy: This is odd, it seems happening only in storybook :/

assets/js/pages/Users/Users.jsx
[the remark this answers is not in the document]
EMaksy: Sadly no, the Table component assigns the width automatically to the table items, the tooltip wraps around the whole button persists popping up on an empty place. 
Open for suggestions  :+1:

assets/js/pages/Users/Users.jsx
```diff
@@ -0,0 +1,157 @@
...
+import Table from '@common/Table';
+import Tooltip from '@common/Tooltip';
+
+const defaultUsers = [];
+
+function Users({
+  onDeleteUser = noop,
+  navigate = noop,
+  users = defaultUsers,
+  loading = true,
```
arbulu89, line 19: I would put the default value for loading to `false`. 
You just manage the real value in the `UsersPage`.
Otherwise you are forced to put `loading: false` in all storybook entries

assets/js/pages/Users/Users.jsx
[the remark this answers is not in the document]
arbulu89: Ok, I think there is some way to have it "right", but well, the current solution works as expected, so it is good

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,69 @@
+import React, { useState, useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { listUsers, deleteUser } from '@lib/api/users';
+
+import { toast } from 'react-hot-toast';
+
+import Users from './Users';
+
+export const SUCCESS_DELETE_MESSAGE = 'User deleted successfully';
+export const ERROR_DELETE_MESSAGE = 'An error occurred: User not found';
```
arbulu89, line 10: Better `USER_NOT_FOUND_ERROR` maybe
By the way, I wouldn't export them, just for test convenience, you can just write the same text in the tests

assets/js/pages/Users/UsersPage.jsx
```diff
@@ -0,0 +1,69 @@
...
+      });
+  };
+
+  const onDeleteUser = (userID) => {
+    deleteUser(userID)
+      .then(() => {
+        toast.success(SUCCESS_DELETE_MESSAGE);
+      })
+      .catch((_error) => {
+        setError(ERROR_DELETE_MESSAGE);
```
arbulu89, line 41: By now, this could be alright, but I would be careful of catching all the errors as they were `user not found`.
You could have the forbidden error as well.
Any case, i don't think it can happen in reallity

assets/js/pages/Users/UsersPage.test.jsx
arbulu89: I still think we don't need it, and there are ways to do it without (like useing `findByText` which has a default timeout), but well, it is alright

assets/js/pages/Users/UsersPage.test.jsx
```diff
@@ -0,0 +1,106 @@
...
+      renderWithRouter(<UsersPage />);
+    });
+    expect(toast.error).toHaveBeenCalledWith(ERROR_FETCH_MESSAGE);
+  });
+
+  it('should render toast with success message when deleting was successfully', async () => {
+    const admin = adminUser.build();
+    const user = userFactory.build();
+
+    axiosMock.onGet('/api/v1/users').reply(200, [admin, user]);
```
arbulu89, line 68: Pipe my friend
```
axiosMock
.onGet('/api/v1/users')
.reply(200, [admin, user])
.onDelete(`/api/v1/users/${user.id}`)
.reply(204)
```

assets/js/pages/Users/UsersPage.test.jsx
```diff
@@ -0,0 +1,106 @@
...
+    axiosMock.onDelete(`/api/v1/users/${user.id}`).reply(204);
+
+    await act(async () => {
+      renderWithRouter(<UsersPage />);
+    });
+
+    let deleteButtons = screen.getAllByText('Delete');
```
arbulu89, line 75: Why this `let` and a next reassignment?
It doesn't look like needed.
If you need to get the `Delete` button from the modal, use other `const` as they are different things

assets/js/pages/Users/Users.test.jsx
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
+import '@testing-library/jest-dom';
+import { screen, render } from '@testing-library/react';
+import { adminUser, userFactory } from '@lib/test-utils/factories/users';
+import { renderWithRouter } from '@lib/test-utils';
+import { userEvent } from '@testing-library/user-event';
+import Users from './Users';
+
+describe('Users', () => {
+  it('should render a loading table with a disabled create user button', async () => {
```
arbulu89, line 10: why `async`? (in most of the tests in fact)

assets/js/pages/Users/Users.test.jsx
```diff
@@ -0,0 +1,98 @@
...
+    expect(button).not.toBeDisabled();
+    expect(screen.getByText('No data available')).toBeVisible();
+  });
+
+  it('should render a table with users', async () => {
+    const creationTime = [
+      '2024-03-22T16:20:57.801758Z',
+      '2024-04-22T16:20:57.801758Z',
+    ];
+    const exptedCreationTime = ['March 22, 2024', 'April 22, 2024'];
```
arbulu89, line 42: typo `expted` to `expected`

arbulu89 (approved):
Thank you @EMaksy 
More comments, but ready to merge so far

assets/js/pages/Users/UsersPage.jsx
EMaksy: I get your point, we could validate the error message and display different error messages, but I also think that currently only the not found error is possible. 
Maybe we can come back to it later when we add user permissions.
What do you think ?

PR #2547, opened 2024-04-24
Avoid SUMA health complete on settings not configured error

arbulu89:
# Description

Avoid sending `CompleteSoftwareUpdatesDiscovery` during the discovery if the SUMA settings are not configured.
This is needed in scenarios like when the `SoftwareUpdatesDiscoveryEventHandler` handler runs the discovery.
Without this change, the health of the software updates is set to `unknown`, even though we shouldn't be changing it, as SUMA settings are not configured.

So the unique difference is that the `settings_not_configured` error doesn't dispatch the command, nothing else.

## How was this tested?

UT added and tested manually

### DIFF
--- a/lib/trento/software_updates/discovery.ex
+++ b/lib/trento/software_updates/discovery.ex
@@ -90,6 +90,9 @@ defmodule Trento.SoftwareUpdates.Discovery do
            |> commanded().dispatch() do
       {:ok, host_id, system_id, relevant_patches}
     else
+      {:error, :settings_not_configured} ->
+        {:error, :settings_not_configured}
+
       {:error, discovery_error} = error ->
         Logger.error(
           "An error occurred during software updates discovery for host #{host_id}:  #{inspect(error)}"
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):
Thanks @arbulu89. 
Have you considered implementing a command middleware doing this check every time a `CompleteSoftwareUpdatesDiscovery` is dispatched?

I would have implemented it that way, however I can live with this as well.

arbulu89:
@nelsonkopliku No, I didn't consider that.
How would you do that? Querying the database every time we have this command?
It looks like less efficient. As we already have the value, why not just simply use it?

nelsonkopliku:
Yes, maybe some extra tiny query when that command would be dispatched :see_no_evil: 

This is just fine, at the end when that command gets dispatched a discovery (read attempt to call suma) have already happened, so we wouldn't be blocking anything interesting in such a middleware.

dottorblaster (approved):
LGTM :+1:

PR #2552, opened 2024-04-25
Policy enforcing on users

CDimonaco:
# Description

This pr creates the bodyguard policy for user management endpoints.
We create default abilities and attach them to default admin in the release task.
The /api/v1/users endpoints are protected. Only accessible to `all:all` or `users:all`

## How was this tested?

Automated tests


lib/trento/users/policy.ex
```diff
@@ -0,0 +1,27 @@
...
+  def authorize(action, %User{} = user, User) when action in [:index, :show],
+    do: has_read_ability?(user)
+
+  def authorize(action, %User{} = user, User) when action in [:update, :delete, :create],
+    do: has_write_ability?(user)
+
+  def authorize(_, _, _), do: false
+
+  defp has_write_ability?(user), do: has_global_ability?(user) or has_users_all_ability?(user)
+  defp has_read_ability?(user), do: has_global_ability?(user) or has_users_all_ability?(user)
```
arbulu89, line 20: `has_read_ability?` should be on top of `has_write_ability?`

lib/trento/users/policy.ex
```diff
@@ -0,0 +1,27 @@
...
+
+  defp has_global_ability?(%User{abilities: abilities}),
```
arbulu89, line 22: I was having a look to this and I thought that it would be cool to have something like this, that is more declarative and you can extend in the future easier with other abilities:
```
defp has_read_ability?(user), do: user_has_any_abilities?(user, [%Ability{name: "all", resource: "all"}, %Ability{name: "users", resource: "all"}])

def user_has_any_abilitities(%User{abilities: abilities}, expected_abilities), 
    do: Enum.any?(expected_abilities, fn expected_ability -> expected_ability in abilities end)
```

Just my head thinking, feel free to discard hehe

lib/trento_web/controllers/fallback_controller.ex
```diff
@@ -138,6 +138,13 @@ defmodule TrentoWeb.FallbackController do
     |> render(:"403")
   end
 
+  def call(conn, {:error, :forbidden}) do
```
arbulu89, line 141: I would personally remove the `operation_not_permitted` from above, and only have one.

priv/repo/migrations/20240425131500_add_basic_permissions.exs
```diff
@@ -0,0 +1,14 @@
...
+  use Ecto.Migration
+
+  def up do
+    execute "INSERT INTO abilities(id, name, resource, label, inserted_at, updated_at) VALUES (1, 'all', 'all', 'Permits all operations on all the resources', NOW(), NOW())"
+
+    execute "INSERT INTO abilities(id, name, resource, label, inserted_at, updated_at) VALUES (2, 'all', 'users', 'Permits all operations on user resource', NOW(), NOW())"
+  end
+
+  def down do
+    execute "DELETE FROM abilities where id = 1"
```
arbulu89, line 11: I would be careful with `id = 1` kind of comparisons.
I have seen ecto skipping some ids for unknown reasons.
Having something like `DELETE FROM abilities where name = 'all' and resrouce = 'all';` looks more stable, and easier to copy paste for the future. Because otherwise you need to keep of the count of entries, etc

test/trento_web/controllers/v1/users_controller_test.exs
```diff
@@ -1,4 +1,5 @@
-defmodule TrentoWeb.V1.UserControllerTest do
+defmodule TrentoWeb.V1.UsersControllerTest do
+  alias Trento.Abilities
```
arbulu89, line 2: Better the `alias` after the `import`

test/trento_web/controllers/v1/users_controller_test.exs
```diff
@@ -11,11 +12,54 @@ defmodule TrentoWeb.V1.UserControllerTest do
...
+    %{id: ability_id} = insert(:ability, name: "all", resource: "all")
+    insert(:users_abilities, user_id: user_id, ability_id: ability_id)
+
+    conn =
+      Pow.Plug.assign_current_user(conn, %{"user_id" => user_id}, Pow.Plug.fetch_config(conn))
+
     {:ok, conn: put_req_header(conn, "accept", "application/json"), api_spec: api_spec}
   end
 
+  describe "forbidden response" do
```
arbulu89, line 36: I have my shared feelings on this.
I think that the policy should be tested in isolation and pure in `test/policy_test` kind of file. Where you can test any possible combination.
But i understand that you want to test that the policy is applied properly.
I see that for example we don't test that the `users:all` doesn't allow the call, which for me is a bad thing.
Maybe we should test both?

arbulu89 (commented):
Hey @CDimonaco 
The code looks great in general.
My unique real concern is that we don't test the policy completely.
For example, the `users:all` ability is not tested.
I think we should have pure tests for the policy itself (together with the controller if you want).
They are so easy to test that it doesn't look a big deal.

test/trento_web/controllers/v1/users_controller_test.exs
CDimonaco: Yes, you're right, I will test both, full policy with unit tests, and policy controller integration with controller tests

priv/repo/migrations/20240425131500_add_basic_permissions.exs
CDimonaco: It's true, but in this migration we are forcing the sequence so we put id 1 and 2 in the up insert and we are sure of these ids, we can have them hardcoded

arbulu89: True :+1:

test/trento_web/controllers/v1/users_controller_test.exs
CDimonaco: done

test/trento_web/controllers/v1/users_controller_test.exs
CDimonaco: done

lib/trento_web/controllers/fallback_controller.ex
CDimonaco: yes, done and refactored

lib/trento/users/policy.ex
CDimonaco: I abstracted the ability checking, thx!

lib/trento/users/policy.ex
CDimonaco: done

test/trento/users/policy_test.exs
```diff
@@ -0,0 +1,47 @@
...
+
+  test "should allow write operations if the user has all:users ability" do
+    user = %User{abilities: [%Ability{name: "all", resource: "all"}]}
+
+    Enum.each([:update, :create, :delete], fn action ->
+      assert true == Policy.authorize(action, user, User)
+    end)
+  end
+
+  test "should disallow other abilities" do
```
arbulu89, line 40: What is this test for? I would expect some `refute` as the test talks about `disallow`, but checking it, i think we can just simply remove it

arbulu89 (approved):
Nice @CDimonaco 
Thanks

test/trento/users/policy_test.exs
CDimonaco: I refactored the test, for some reason I had the working copy not committed lol

priv/repo/migrations/20240425131500_add_basic_permissions.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Trento.Repo.Migrations.AddBasicPermissions do
+  use Ecto.Migration
+
+  def up do
+    execute "INSERT INTO abilities(id, name, resource, label, inserted_at, updated_at) VALUES (1, 'all', 'all', 'Permits all operations on all the resources', NOW(), NOW())"
+
+    execute "INSERT INTO abilities(id, name, resource, label, inserted_at, updated_at) VALUES (2, 'all', 'users', 'Permits all operations on user resource', NOW(), NOW())"
+  end
```
gagandeepb, line 8: I wonder what were the considerations to make it necessary to load abilities from the db? For instance, I'm curious why can't there be static abilities, i.e. at the code level? (Maybe I'm thinking too much along a role-based approach, and there is a different/more granular approach that has been agreed here for this.)

arbulu89: Because using migrations is really convenient to execute/rollback the actions.
Otherwise, we would need to add seeds/release actions to add them. And the same would happen if you want to remove. You would need to explicitily remove them in the code, which would become error prone

gagandeepb: I see, thanks for the explanation. My comment was a bit broader, though, on the overall approach of loading/storing abilities in the db in the first place, but I guess maybe that is necessitated due to the access control meta-approach considered (RBAC vs granular ability based, etc.). 
On a somewhat side note: I have been reading a bit into [Cedar](https://github.com/cedar-policy/cedar) recently, and they have some interesting ideas about checking/validating static policies that are decoupled from the application code completely, using some formal methods/tools that give better assurances. It might be interesting to evaluate too, at some point.

EMaksy (approved):

PR #2560, opened 2024-04-29
Remove `DiscoverSoftwareUpdates` command

nelsonkopliku:
# Description

As above: removing `DiscoverSoftwareUpdates` command as it is never actually dispatched. 

It was initially introduced in this PR https://github.com/trento-project/web/pull/2519.


dottorblaster (approved):
I have to admit I was wondering wether we needed this :sweat_smile: 

Thanks!

jamie-suse (approved):

PR #2561, opened 2024-04-29
Read tracked software updates discoveries

nelsonkopliku:
# Description

This PR changes the data source of the `GET /hosts/:host_id/software_updates` endpoint.
It now reads from previously tracked discovery result.

- if a discovery result for the given host id is not found the endpoint ends up in a 404 (either the host id is invalid or a discovery for the host has not been tracked, very likely because of a missing fqdn*)
- if there is a failure reason in the discovery result, it gets returned as part of an `{:error, reason}` tuple. The endpoint in this case is going to return what defined in the fallback controller (very likely a 422 on `system_id_not_found`, `error_getting_patches`, `error_getting_packages`)

Note* the `fqdn_not_found` error becomes a tricky one. 
It would be great to track it as a possible failure reason for a discovery result, however since the discovery process is not always triggered the same way, it might happen that when detected null in the domain, a discovery for the specific host is not attempted at all, making it impossible to track where we track results currently. 
We could instead track it when discovery is initiated via the ticker or when saving settings. However I preferred keeping at least some consistency by _not tracking_ it at all.

If we want to keep `fqdn_not_found` explicit we could keep the `get_host_fqdn` function, but we'd be querying the hosts unnecessarily, imho.
So since we are currently only really interested in `:settings_not_configured` error (which is still properly handled) and all other error fall into a generic bucket, I believe we can just defer further changes to when we refine error handling.

## How was this tested?

Automated + manual.

### DIFF
--- a/lib/trento/software_updates/discovery.ex
+++ b/lib/trento/software_updates/discovery.ex
@@ -273,8 +273,8 @@ defmodule Trento.SoftwareUpdates.Discovery do
        }) do
     {
       :ok,
-      normalize_discovered_result_list(relevant_patches),
-      normalize_discovered_result_list(upgradable_packages)
+      keys_to_atoms(relevant_patches),
+      keys_to_atoms(upgradable_packages)
     }
   end
 
@@ -284,7 +284,7 @@ defmodule Trento.SoftwareUpdates.Discovery do
   defp failure_reason_to_atom("max_login_retries_reached"), do: :max_login_retries_reached
   defp failure_reason_to_atom(_), do: :unknown_discovery_error
 
-  defp normalize_discovered_result_list(discovered_result_list),
+  defp keys_to_atoms(discovered_result_list),
     do:
       discovered_result_list
       |> Jason.encode!()
--- a/lib/trento/software_updates/discovery.ex
+++ b/lib/trento/software_updates/discovery.ex
@@ -81,6 +81,13 @@ defmodule Trento.SoftwareUpdates.Discovery do
     :ok
   end
 
+  @spec get_discovery_result(String.t()) :: {:ok, list(), list()} | {:error, any()}
+  def get_discovery_result(host_id) do
+    DiscoveryResult
+    |> Repo.get(host_id)
+    |> handle_discovery_result()
+  end
+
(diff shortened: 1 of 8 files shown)


lib/trento/software_updates/discovery.ex
```diff
@@ -254,6 +261,35 @@ defmodule Trento.SoftwareUpdates.Discovery do
 
   defp compute_software_updates_discovery_health(_), do: SoftwareUpdatesHealth.warning()
 
+  defp handle_discovery_result(nil), do: {:error, :not_found}
+
+  defp handle_discovery_result(%DiscoveryResult{failure_reason: failure_reason})
+       when not is_nil(failure_reason),
+       do: {:error, failure_reason_to_atom(failure_reason)}
```
dottorblaster, line 268: I'm not clear why we're squashing errors this way. At this point we could just return whatever result we have and just have the frontend recognizing if `failure_reason` is not nil.

I suppose this was done because otherwise we'd have to rework the frontend? If this was the purpose, I'm fine with it. Maybe we can revisit it in a second time.

lib/trento/software_updates/discovery.ex
```diff
@@ -254,6 +261,29 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
+         upgradable_packages: upgradable_packages
+       }) do
+    {
+      :ok,
+      normalize_discovered_result_list(relevant_patches),
+      normalize_discovered_result_list(upgradable_packages)
+    }
+  end
+
+  defp normalize_discovered_result_list(discovered_result_list),
```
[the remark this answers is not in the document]
dottorblaster: I'm like meh myself.

Could you at least rename this just to `keys_to_atoms`?

dottorblaster (approved):
A couple minor nitpicks, but overall LGTM

lib/trento/software_updates/discovery.ex
nelsonkopliku: I wanted to keep the endpoint unchanged in its input/output. Not changing the frontend is indeed a consequence.

PR #2566, opened 2024-04-30
Edit user form

arbulu89:
# Description

Add user edition form. It is pretty straightforward.
One thing that is missing is the `headers` usage in the edit user `patch` query, as we need to send some value for the optimistic locking.

## How was this tested?

Tested with UT and manually

### DIFF
--- a/assets/js/lib/test-utils/factories/users.js
+++ b/assets/js/lib/test-utils/factories/users.js
@@ -10,7 +10,8 @@ export const userFactory = Factory.define(() => ({
   enabled: faker.datatype.boolean(),
   fullname: faker.internet.displayName(),
   email: faker.internet.email(),
-  password: faker.internet.password(),
+  created_at: formatISO(faker.date.past()),
+  updated_at: formatISO(faker.date.past()),
 }));
 
 export const adminUser = userFactory.params({
--- a/assets/js/lib/test-utils/factories/users.js
+++ b/assets/js/lib/test-utils/factories/users.js
@@ -5,7 +5,6 @@ import { formatISO } from 'date-fns';
 export const userFactory = Factory.define(() => ({
   id: faker.number.int(),
   username: faker.internet.userName(),
-  created_at: formatISO(faker.date.past()),
   actions: 'Delete',
   enabled: faker.datatype.boolean(),
   fullname: faker.internet.displayName(),
--- a/assets/js/pages/Users/UserForm.jsx
+++ b/assets/js/pages/Users/UserForm.jsx
@@ -1,5 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import { capitalize, noop } from 'lodash';
+import { format, parseISO } from 'date-fns';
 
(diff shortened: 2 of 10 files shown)


assets/js/pages/Users/UserForm.jsx
```diff
@@ -100,11 +103,14 @@ function UserForm({
     const user = {
       fullname: fullNameState,
       email: emailAddressState,
-      username: usernameState,
-      password: passwordState,
-      password_confirmation: confirmPasswordState,
       enabled: statusState === USER_ENABLED,
+      ...(!editing && { username: usernameState }),
```
EMaksy, line 107: This is a cool usage of the spread operator xD

assets/js/pages/Users/UserForm.stories.jsx
```diff
@@ -35,18 +41,24 @@ export default {
...
+    createdAt: {
+      description: 'User creation time',
       control: {
         type: 'text',
       },
     },
-    confirmPassword: {
-      description: 'Password confirmation',
+    udpatedAt: {
+      description: 'User last edition time',
```
EMaksy, line 51: Just an idea: 
Timestamp of last user data update

EMaksy (approved):
Great job, just some small comments :+1:

assets/js/lib/test-utils/factories/users.js
```diff
@@ -5,12 +5,12 @@ import { formatISO } from 'date-fns';
 export const userFactory = Factory.define(() => ({
   id: faker.number.int(),
   username: faker.internet.userName(),
-  created_at: formatISO(faker.date.past()),
   actions: 'Delete',
   enabled: faker.datatype.boolean(),
   fullname: faker.internet.displayName(),
   email: faker.internet.email(),
-  password: faker.internet.password(),
```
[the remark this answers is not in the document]
arbulu89: At the end, we don't need it.
The password is never returned by our user api's, because it is a sensitive information.
for that, we don't really need in the factory

assets/js/pages/Users/UserForm.jsx
arbulu89: basically, in patches, we don't want to send all the fields, so this is a nice way of conditionally adding them

assets/js/pages/Users/UserForm.stories.jsx
arbulu89: I will just leave with `User last update timestamp` to be more aligned with the `createdAt: 'User creation timestamp'`

assets/js/lib/test-utils/factories/users.js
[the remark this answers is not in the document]
EMaksy: Yeah, makes sense, it was just handy for tests, but you are fully right, we should stick with the backend truth :)

jagabomb (approved):
LGTM!

CDimonaco (approved):
:chef_kiss: LGTM!

PR #2567, opened 2024-04-30
Remove "SUSE Manager connection failed" error message

jamie-suse:
# Description

Changes to the error handling in the `AvailableSoftwareUpdates` component, when a connection error happens while requesting software updates:

- Replaces message "_SUSE Manager connection failed_" with "_Unknown_"
- Replaces tooltip message "_**SUSE Manager** was not able to retrieve the requested data_" with "_**Trento** was not able to retrieve the requested data_"

These new requirements obviate the need for `connectionError` in the Redux state, originally added in https://github.com/trento-project/web/pull/2538

## How was this tested?

Updated unit tests

### DIFF
--- a/assets/js/common/AvailableSoftwareUpdates/AvailableSoftwareUpdates.jsx
+++ b/assets/js/common/AvailableSoftwareUpdates/AvailableSoftwareUpdates.jsx
@@ -21,13 +21,12 @@ const containerClassNames = classNames(
 function AvailableSoftwareUpdates({
   className,
   settingsConfigured = false,
+  softwareUpdatesSettingsLoading,
+  onBackToSettings = noop,
+  softwareUpdatesLoading,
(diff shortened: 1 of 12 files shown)


nelsonkopliku (approved):
LGTM!

So the thing now is that whatever error the available software updates endpoint might return, we'll get `Unknown` in the indicators.

As long as @abravosuse is fine with it, I am fine as well.

abravosuse:
> LGTM!
> 
> So the thing now is that whatever error the available software updates endpoint might return, we'll get `Unknown` in the indicators.
> 
> As long as @abravosuse is fine with it, I am fine as well.

For the time being, I'm good with this. Once we differentiate other scenarios (host not found / connection error) we can further tune the messages.

dottorblaster (approved):
LGTM

PR #2576, opened 2024-05-02
Add a generate password button for the UserForm

EMaksy:
# Description

This pr will add a generate password button to the User creation/edit form. 
As we can't use faker to produce some random password in production, I added a password generation helper function. 

After some research, using [crypto.getRandomValues](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) seems like a valid solution to get cryptographically strong random values. 

Per default, we generate a 16 char long password which includes digits, uppercase/lowercase letters and special characters

### Preview

[Screencast from 2024-05-03 12-29-28.webm](https://github.com/trento-project/web/assets/54111255/b046f9b0-bd3c-475f-b10c-d85611b95acf)

## How was this tested?

User Create test updated and added new test
Storybook

## What is missing ? 

Add test for user edit after merge

### DIFF
--- a/assets/js/common/Input/Password.jsx
+++ b/assets/js/common/Input/Password.jsx
@@ -12,6 +12,7 @@ function Password({
   id,
   name,
   value,
+  initialValue,
   placeholder = 'Password',
   error = false,
   disabled = false,
@@ -24,7 +25,8 @@ function Password({
       className={classNames(className)}
       id={id}
       name={name}
-      initialValue={value}
+      value={value}
+      initialValue={initialValue}
       placeholder={placeholder}
       type={inputType}
       error={error}
--- a/assets/js/pages/Users/generatePassword.js
+++ b/assets/js/pages/Users/generatePassword.js
@@ -1,0 +1,15 @@
+const DIGITS = '0123456789';
+const UPPERCASE_LETTERS = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
+const LOWERCASE_LETTERS = 'abcdefghijklmnopqrstuvwxyz';
+const SPECIAL_CHARACTERS = '~!@#$%^&*()_+={}[]|:;<>,.?/-\'"`';
+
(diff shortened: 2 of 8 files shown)


CDimonaco (approved):

arbulu89 (approved):
Hey @EMaksy ,
The change looks good.

Please make sure the `Password` component change doesn't break anything.

assets/js/pages/Users/generatePassword.js
```diff
@@ -0,0 +1,15 @@
...
+const SPECIAL_CHARACTERS = '~!@-#$%^&*()_+={}[]|:;<>,.?/';
+
+const PASSWORD_CHARACTERS =
+  DIGITS + UPPERCASE_LETTERS + LOWERCASE_LETTERS + SPECIAL_CHARACTERS;
+
+export const generatePassword = (
+  length = 16,
+  characters = PASSWORD_CHARACTERS
+) =>
+  Array.from(crypto.getRandomValues(new Uint32Array(length)))
```
rtorrero, line 13: Does this guarantee that we will always get a password that  meets our requirements?

rtorrero (approved):
LGTM, just left you a question :+1:

assets/js/pages/Users/generatePassword.js
EMaksy: No good point @rtorrero, there is a minimal chance that we have char repetition. 
I would tackle this in a follow-up pr

assets/js/common/Input/Password.jsx
```diff
@@ -12,6 +12,7 @@ function Password({
   id,
   name,
   value,
+  initialValue,
```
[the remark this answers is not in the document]
EMaksy: We had a quick review with @nelsonkopliku on this.
Checked and tested it out on main. Everything seems to be fine besides a broken Password sub story.

I will update the story and with this we should be good to go.

PR #2581, opened 2024-05-03
Validation of CA certificate inside SUSE Manager settings

dottorblaster:
# Description
Introducing validation for the CA certificate uploaded in SUSE Manager settings.

Took advantage of this to also migrate the settings schema to the Single Table Inheritance pattern implemented across the rest of our settings.

## How was this tested?

Existing unit tests

### DIFF
--- a/lib/trento/software_updates/settings.ex
+++ b/lib/trento/software_updates/settings.ex
@@ -4,6 +4,7 @@ defmodule Trento.SoftwareUpdates.Settings do
   """
 
   use Ecto.Schema
+  use Trento.Support.Ecto.STI, sti_identifier: :suse_manager_settings
 
   import Ecto.Changeset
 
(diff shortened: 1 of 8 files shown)


priv/repo/migrations/20240502105156_migrate_suma_settings_sti.exs
```diff
@@ -0,0 +1,15 @@
...
+  def change do
+    alter table(:settings) do
+      add :suse_manager_settings_url, :string
+      add :suse_manager_settings_username, :string
+      add :suse_manager_settings_password, :binary
+      add :suse_manager_settings_ca_cert, :binary
+      add :suse_manager_settings_ca_uploaded_at, :utc_datetime_usec
+    end
+
+    drop table(:software_update_settings)
```
nelsonkopliku, line 13: I was about to say that we maybe need to migrate the content from `software_update_settings` to `settings`, but since we never released this yet, I guess this should be just fine.

CDimonaco (approved):

lib/trento/software_updates/settings.ex
```diff
@@ -43,14 +47,47 @@ defmodule Trento.SoftwareUpdates.Settings do
...
+
+    if never_before |> X509.DateTime.to_datetime() |> DateTime.before?(now) and
+         never_after |> X509.DateTime.to_datetime() |> DateTime.after?(now) do
+      :ok
+    else
+      {:error, [ca_cert: {"the X.509 certificate is not valid", validation: :ca_cert_validity}]}
     end
   end
 
+  defp maybe_validate_ca_cert(changeset, %{ca_cert: nil}), do: changeset
```
nelsonkopliku, line 81: Nothing major. Would you mind moving these `maybe_validate_ca_cert` before `validate_ca_cert`, please?
Reading would just flow smoother :smile:

nelsonkopliku (changes requested):
Cool! Thanks for addressing this.

Besides just a minor nitpick in functions appearance order, I am getting a 500 when the following ca cert is provided

```
-----BEGIN CERTIFICATE-----
foobar
-----END CERTIFICATE-----
```
I guess some extra check is needed :eyes:

dottorblaster:
@nelsonkopliku apparently we were getting a raise in that case. I added a test just for that, thanks.

nelsonkopliku (approved):

jamie-suse (approved):
:rocket:

PR #2582, opened 2024-05-03
Manage npm dependencies

janvhs:
# Description

These patches removes `canvas` as direct dependency, as it is only used indirectly by `jsdom`. Furthermore, it moves `postcss-import` and `@storybook/client-api` to the development dependency section, as they are used in testing and compilation.

Eventually, as part of a npm audit warning, it increments the patch version of `eslint-config-airbnb-base` from 6.3.0 to 6.3.1.

More information are available in the commit history.

## How was this tested?

To verify that nothing broke, I ran the jest test suite and ensured that storyblock is still working as expected.


dottorblaster (approved):
I don't remember actually why `postcss-import` ended up in the dependencies array instead of just devdeps.

Thank you!

PR #2583, opened 2024-05-03
User profile routes

CDimonaco:
# Description

Routes for updating/showing a logged user profile.
Like the normal user updates, the default admin could not update profile.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/users/user.ex
+++ b/lib/trento/users/user.ex
@@ -70,7 +70,7 @@ defmodule Trento.Users.User do
   end
 
   defp validate_current_password(changeset, %{password: _password} = attrs),
-    do: changeset |> pow_current_password_changeset(attrs)
+    do: pow_current_password_changeset(changeset, attrs)
 
   defp validate_current_password(changeset, _), do: changeset
 
--- a/lib/trento/users/user.ex
+++ b/lib/trento/users/user.ex
@@ -55,6 +55,7 @@ defmodule Trento.Users.User do
 
   def profile_update_changeset(user, attrs) do
     user
+    |> validate_current_password(attrs)
     |> pow_password_changeset(attrs)
     |> pow_extension_changeset(attrs)
     |> validate_password()
@@ -67,6 +68,11 @@ defmodule Trento.Users.User do
     |> validate_required(:deleted_at)
     |> put_change(:username, "#{username}__#{deleted_at}")
   end
+
+  defp validate_current_password(changeset, %{password: _password} = attrs),
+    do: changeset |> pow_current_password_changeset(attrs)
+
+  defp validate_current_password(changeset, _), do: changeset
 
   defp lock_changeset(user, attrs) do
     cast(user, attrs, [:locked_at])
--- a/lib/trento/users/user.ex
+++ b/lib/trento/users/user.ex
@@ -53,6 +53,14 @@ defmodule Trento.Users.User do
     |> lock_changeset(attrs)
   end
 
+  def profile_update_changeset(user, attrs) do
+    user
+    |> pow_password_changeset(attrs)
+    |> pow_extension_changeset(attrs)
+    |> validate_password()
+    |> custom_fields_changeset(attrs)
+  end
+
   def delete_changeset(%__MODULE__{username: username} = user, %{deleted_at: deleted_at} = attrs) do
     user
     |> cast(attrs, [:deleted_at])
--- a/lib/trento_web/controllers/v1/profile_controller.ex
+++ b/lib/trento_web/controllers/v1/profile_controller.ex
@@ -1,0 +1,39 @@
+defmodule TrentoWeb.V1.ProfileController do
+  use TrentoWeb, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  alias TrentoWeb.OpenApi.V1.Schema
+  alias Trento.Users
+
+  plug OpenApiSpex.Plug.CastAndValidate, json_render_error_v2: true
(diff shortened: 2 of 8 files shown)


lib/trento/users/user.ex
```diff
@@ -53,6 +53,14 @@ defmodule Trento.Users.User do
     |> lock_changeset(attrs)
   end
 
+  def profile_update_changeset(user, attrs) do
+    user
```
arbulu89, line 57: Shouldn't we have `pow_current_password_changeset` here?
How do we enforce otherwise that the `current_password` is coming?
Or `pow_password_changeset` does this?  But in that the case, it would happen in the current update, which doesn't

lib/trento_web/controllers/v1/profile_controller.ex
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  operation :update,
+    summary: "Update the current user profile",
+    tags: ["Platform"],
+    request_body:
+      {"UserProfileUpdateRequest", "application/json", Schema.User.UserProfileUpdateRequest},
+    responses: [
+      ok: {"Profile updated successfully", "application/json", Schema.User.UserProfile},
+      unprocessable_entity: Schema.UnprocessableEntity.response()
```
arbulu89, line 33: We should have `forbidden` for the admin user here I guess

lib/trento_web/openapi/v1/schema/user.ex
```diff
@@ -4,6 +4,61 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
...
 
+  defmodule UserProfile do
+    @moduledoc false
+
+    @schema %Schema{
+      title: "UserProfile",
+      description: "Trento User profile",
+      type: :object,
+      additionalProperties: false,
+      properties: %{
```
arbulu89, line 15: Shouldn't we have the `abilities` here?
Or do we include them in a next PR?
Yes, we can do later on, once the `abilities` schemas and code are ready. Related to: https://github.com/trento-project/web/pull/2568

lib/trento_web/openapi/v1/schema/user.ex
```diff
@@ -4,6 +4,61 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
...
+    @moduledoc false
+
+    @schema %Schema{
+      title: "UserProfileUpdateRequest",
+      description: "Request body to update a user profile",
+      additionalProperties: false,
+      type: :object,
+      properties: %{
+        fullname: %Schema{type: :string, description: "User full name", nullable: false},
+        email: %Schema{type: :string, description: "User email", nullable: false, format: :email},
```
arbulu89, line 49: I'm missing `current_password` here

lib/trento_web/router.ex
```diff
@@ -86,6 +86,9 @@ defmodule TrentoWeb.Router do
     scope "/v1", TrentoWeb.V1 do
       pipe_through [:api_v1]
 
+      get "/profile", ProfileController, :show
```
arbulu89, line 89: I would personally put these entries close to `/users`.
Just to have some logical order

arbulu89 (commented):
Hey @CDimonaco 
It looks good, but I'm missing the `current_password` usage.
The user is asked to give the current password to change it.
I think we can use that simply with `pow_current_password_changeset` in the user changeset as it requires to send `current_password`.
What do you think?

EMaksy (commented):
Hey man, good work!
+1 to @arbulu89 questions/comments and also some questions from my side :)

test/trento/users_test.exs
```diff
@@ -5,6 +5,49 @@ defmodule Trento.UsersTest do
   alias Trento.Users.User
   import Trento.Factory
 
+  describe "user profile" do
+    test "update_user_profile does not update the profile when fields are not present" do
```
rtorrero, line 9: I'm not sure that the description of the test matches what the test does, am I missing something?

lib/trento/users/user.ex
CDimonaco: you're right, done!

lib/trento_web/openapi/v1/schema/user.ex
CDimonaco: Let's include when the abilities pr code is ready

lib/trento_web/openapi/v1/schema/user.ex
CDimonaco: done

lib/trento_web/router.ex
CDimonaco: done

lib/trento_web/views/v1/profile_view.ex
```diff
@@ -0,0 +1,23 @@
+defmodule TrentoWeb.V1.ProfileView do
+  use TrentoWeb, :view
+
+  def render("profile.json", %{
```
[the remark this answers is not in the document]
CDimonaco: Yes, we will have different data, so we can differentiate between data sent/returned

test/trento/users_test.exs
CDimonaco: It's for testing that the update is idempontent, no field passed, no field updated, this should match with the patch verb we want to implement

lib/trento_web/controllers/v1/profile_controller.ex
CDimonaco: done

arbulu89 (approved):
Really nice @CDimonaco !!
:heart:

PR #2584, opened 2024-05-03
Add soft delete for email and update test

EMaksy:
# Description
The backend user management supports soft deletion.  
When a user is deleted, we apply soft deletion on the username.
With this we unlock the username which must be unique.

This pr will also soft delete the mail of a user, not only username.

## How was this tested?
Manual testing + existing test was updated to check for mail


CDimonaco (approved):

arbulu89 (approved):
Great!!

janvhs (approved):
Looks great!

PR #2591, opened 2024-05-06
Enable user status update

arbulu89:
# Description

Enable user status selection in frontend, it was disabled by now.
It requires a small modification in the backend to set the `locked_at` value during creation.

## How was this tested?

Tests added


CDimonaco (approved):
LGTM

PR #2593, opened 2024-05-06
Make SUSE Manager section in Host Details page reactive

jamie-suse:
# Description

This change adds the ability for Trento to send a WebSocket message to indicate that the software update discovery for a given host has been completed (`HOST_SOFTWARE_UPDATES_DISCOVERY_COMPLETED`).

When a client receives the `HOST_SOFTWARE_UPDATES_DISCOVERY_COMPLETED` message, the client then requests the _software updates settings_ information, and then also requests the _software updates_ of a given host.

## How was this tested?

- Added checks to ensure that the new WebSocket message is being sent.
- Added tests for the new Redux Saga to ensure the correct behaviour.


lib/trento/software_updates/discovery.ex
```diff
@@ -177,6 +177,16 @@ defmodule Trento.SoftwareUpdates.Discovery do
       |> Multi.run(:command_dispatching, fn _, _ ->
         dispatch_completion_command(host_id, discovered_health)
       end)
+      |> Multi.run(:notify_client, fn _, _ ->
+        with :ok <-
+               TrentoWeb.Endpoint.broadcast(
```
nelsonkopliku, line 182: The appropriate place to do this would be when a `SoftwareUpdatesHealthChanged` (possibly happening when a software updates discovery compeltes) event is emitted from the aggregate (I am hating myself for not having kept `SoftwareUpdatesDiscoveryCompleted` as a name... anyway) so I would move this broadcasting in `SoftwareUpdatesDiscoveryEventHandler`.
Similarly to what we do when broadcasting in projectors.

nelsonkopliku (changes requested):
Good job!

We just need to leverage `SoftwareUpdatesDiscoveryEventHandler` event handler and do relevant things when specific other relevant things happen :smile: 

Besides this, what happens when a host gets deregistered or settings are cleared?
We might want that reactiveness also in these cases.

lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
```diff
@@ -27,11 +28,21 @@ defmodule Trento.Infrastructure.Commanded.EventHandlers.SoftwareUpdatesDiscovery
...
 
-  def handle(
-        %SoftwareUpdatesDiscoveryCleared{
-          host_id: host_id
-        },
-        _
-      ),
-      do: Discovery.clear_tracked_discovery_result(host_id)
+  def handle(%SoftwareUpdatesHealthChanged{host_id: host_id}, _) do
+    TrentoWeb.Endpoint.broadcast(
```
nelsonkopliku, line 32: What about wrapping this broadcast in a function? And use that in both `handle`. Yes, just a nit :smile:

dottorblaster (approved):
LGTM, great work!

lib/trento/infrastructure/commanded/event_handlers/software_updates_discovery_event_handler.ex
jamie-suse: I disagree in the case of this abstraction. In my view `Endpoint.broadcast/3` is already a nice abstraction provided to us to use - `broadcast(topic, event, payload)`. On top of this, we still need to provide `host_id` to our wrapper function anyway, therefore I'm not sure what work it's really helping us avoid.

nelsonkopliku: Up to you, I am fine eitherway :wink:

nelsonkopliku (approved):
Alright, great improvements :rocket:

It would be great avoiding unnecessary requests to the APIs, but I guess we can just go with this now. 
We at least have a saner behavior, now.

PR #2594, opened 2024-05-06
Update installation instructions

janvhs:
# Description

This patchset documents the preferred version of the required dependencies
in the documentation, documents additional dependencies, which might be
required for some users and synchronises the compiler version in the `mix.exs` file
with the version managed by `asdf`.

These are mainly inconveniences I encountered during the first setup on a MacBook.

An explanation and reasoning for these changes is available as part of the individual
commit descriptions. Those are accessible by pressing on the `...` next to the commit
headline below.

### DIFF
--- a/guides/development/hack_on_the_trento.md
+++ b/guides/development/hack_on_the_trento.md
@@ -13,13 +13,13 @@ In order to run the Trento Web application, the following software must be insta
 ### Additional requirements
 
 Some platforms might not be able to use pre-built versions of some dependencies.
-Therefore, some additional dependencies might be required. This will not effect
-most users and should be referred to when installation issues come up.
-For these dependencies, the distro packaged version will usually be sufficient.
+Therefore, some additional dependencies might be required. This does not effect
+most users and can be referred to, when installation issues come up.
+For these dependencies, the distro packaged version is usually sufficient.
 
 1. [Python3](https://www.python.org/)
 2. [setuptools](https://setuptools.pypa.io/en/latest/index.html)
-3. [gcc](https://gcc.gnu.org/) or [clang](https://clang.llvm.org/)
+3. [gcc](https://gcc.gnu.org/)
 4. [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/)
 
 ### Ensure Compatibility with asdf
--- a/guides/development/hack_on_the_trento.md
+++ b/guides/development/hack_on_the_trento.md
@@ -4,9 +4,9 @@
 
 In order to run the Trento Web application, the following software must be installed:
 
-1. [Elixir](https://elixir-lang.org/)
-2. [Erlang OTP](https://www.erlang.org/)
-3. [Node.js](https://nodejs.org/en/)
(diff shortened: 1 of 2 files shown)


guides/development/hack_on_the_trento.md
```diff
@@ -4,12 +4,24 @@
...
+### Additional requirements
+
+Some platforms might not be able to use pre-built versions of some dependencies.
+Therefore, some additional dependencies might be required. This will not effect
+most users and should be referred to when installation issues come up.
+For these dependencies, the distro packaged version will usually be sufficient.
+
+1. [Python3](https://www.python.org/)
+2. [setuptools](https://setuptools.pypa.io/en/latest/index.html)
+3. [gcc](https://gcc.gnu.org/) or [clang](https://clang.llvm.org/)
```
jamie-suse, line 22: We don't really need to add Clang here, do we? We just use GCC according to the Ansible Playbook

guides/development/hack_on_the_trento.md
```diff
@@ -4,12 +4,24 @@
...
+1. [Elixir](https://elixir-lang.org/) - 1.15.7 preferred
+2. [Erlang OTP](https://www.erlang.org/) - 26.1.2 preferred
+3. [Node.js](https://nodejs.org/en/) - 16.16.0 preferred
 4. [Docker](https://docs.docker.com/get-docker/)
 5. [Docker Compose](https://docs.docker.com/compose/install/)
 
+### Additional requirements
+
+Some platforms might not be able to use pre-built versions of some dependencies.
+Therefore, some additional dependencies might be required. This will not effect
```
arbulu89, line 16: Try to use present term in the text: `This does not affect most users...`.
And so on

guides/development/hack_on_the_trento.md
arbulu89: True

arbulu89 (approved):
Looks good!
Only try to use the present term in the writing instead of future

guides/development/hack_on_the_trento.md
```diff
@@ -4,12 +4,24 @@
...
-3. [Node.js](https://nodejs.org/en/)
+1. [Elixir](https://elixir-lang.org/) - 1.15.7 preferred
+2. [Erlang OTP](https://www.erlang.org/) - 26.1.2 preferred
+3. [Node.js](https://nodejs.org/en/) - 16.16.0 preferred
 4. [Docker](https://docs.docker.com/get-docker/)
 5. [Docker Compose](https://docs.docker.com/compose/install/)
 
+### Additional requirements
+
+Some platforms might not be able to use pre-built versions of some dependencies.
```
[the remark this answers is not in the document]
janvhs: I was using macOS, these dependencies can be required for the npm package _canvas_. These are only required, if the platform can not download pre-built binaries from the GitHub release. This problem can occur, when the platform is too esoteric, this is the case with aarch64-apple-darwin, or GitHub can not be reached.

guides/development/hack_on_the_trento.md
janvhs: Thanks for pointing that out :D

PR #2596, opened 2024-05-06
Use plain string certificates

nelsonkopliku:
# Description

This PR enables using SUMA ca certificate as a plain string instead of a reference to a file in the filesystem.

Took inspiration from this post https://elixirforum.com/t/using-client-certificates-from-a-string-with-httposion/8631/7

## How was this tested?

Current tests updated + manual tests.

### DIFF
--- a/lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
+++ b/lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
@@ -38,7 +38,7 @@ defmodule Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor do
   @behaviour Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor
 
   @impl true
-  def login(base_url, username, password, ca_cert \\ nil) do
+  def login(base_url, username, password, ca_cert) do
     payload =
       Jason.encode!(%{
         "login" => username,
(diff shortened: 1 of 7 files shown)


jamie-suse (approved):
wow, so much nicer :rocket:

lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
```diff
@@ -3,44 +3,42 @@ defmodule Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor do
...
-              use_ca_cert :: boolean()
+              ca_cert :: String.t() | nil
             ) ::
               {:ok, HTTPoison.Response.t()} | {:error, HTTPoison.Error.t()}
 
   @behaviour Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor
 
   @impl true
-  def login(base_url, username, password, use_ca_cert \\ false) do
+  def login(base_url, username, password, ca_cert \\ nil) do
```
dottorblaster, line 41: Do we need the `\\ nil`? I mean, we can just have `ca_cert`, it doesn't really convert to some default if it's `nil` :+1:

dottorblaster (changes requested):
Just a comment about default arguments but I think this is ready to ship

dottorblaster (approved):
LGTM

PR #2600, opened 2024-05-07
Multi select component

arbulu89:
# Description

Add a new `MultiSelect` common component. It is based in [react-select](https://react-select.com/home) package.
Reusing the currently existing `Select` didn't look simple enough, as it does a lot of assumptions about single selection.
I have included a tooltip option, in both the menu option in the selected value.

The code is pretty simple and it follows the same styling system that we use for `Select`.
In fact, the most annoying thing is the styling. We have to put the `unstyled` prop together with custom styles to override the default classes, as per default they are applied after the initial baseline.
(if somebody wants to undestand more: https://github.com/JedWatson/react-select/blob/master/storybook/stories/ClassNamesWithTailwind.stories.tsx

multi-select

## How was this tested?

Tested and storybook added

### DIFF
--- a/assets/js/common/MultiSelect/MultiSelect.jsx
+++ b/assets/js/common/MultiSelect/MultiSelect.jsx
@@ -1,0 +1,92 @@
+import React from 'react';
+
+import Select, { components as selectComponents } from 'react-select';
+import { ChevronUpDownIcon, XMarkIcon } from '@heroicons/react/20/solid';
+
+import { noop } from 'lodash';
+import classNames from 'classnames';
+
+import Tooltip from '@common/Tooltip';
+
+const option = (props) => (
+  <Tooltip content={props.data.tooltip} isEnabled={!!props.data.tooltip}>
+    <selectComponents.Option {...props} />
+  </Tooltip>
+);
+
+const multiValueContainer = (props) => (
+  <Tooltip content={props.data.tooltip} isEnabled={!!props.data.tooltip}>
+    <selectComponents.MultiValueContainer {...props} />
+  </Tooltip>
+);
+
+const dropdownIndicator = (props) => (
+  <selectComponents.DropdownIndicator {...props}>
+    <ChevronUpDownIcon
+      className="w-5 h-5 text-gray-400 cursor-pointer"
+      aria-hidden="true"
+    />
+  </selectComponents.DropdownIndicator>
+);
+
+const clearIndicator = (props) => (
+  <selectComponents.ClearIndicator {...props}>
+    <XMarkIcon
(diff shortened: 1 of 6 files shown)


rtorrero (approved):
@arbulu89 looks good! my only remark is that when interacting with the component in storybook, there doesn't always seem to be enough space:
Captura desde 2024-05-07 14-42-23

Could it be fixed using a ContainerWrapper where enough space is asigned beforehand?

arbulu89:
@rtorrero well, don't use the `docs` for such a thing hehe
Just go to the appropriate views, the `docs` has some effect in the layout.
Anyway, I'm going to adjust a bit the story for the width (which won't change what you see in any case)

assets/js/common/MultiSelect/MultiSelect.jsx
```diff
@@ -0,0 +1,94 @@
...
+  multiValueLabel: () => 'ml-1',
+  control: () =>
+    'relative w-full py-2 px-3 text-left bg-white rounded-lg cursor-default border border-gray-300 sm:text-sm',
+  menu: () =>
+    'absolute w-full py-1 mt-1 overflow-auto text-base bg-white rounded-md shadow-lg max-h-60 ring-1 ring-black ring-opacity-5 focus:outline-none sm:text-sm z-[1]',
+  option: ({ isFocused }) =>
+    classNames(
+      isFocused && 'text-green-900 bg-green-100',
+      'cursor-default select-none relative py-2 px-3'
+    ),
```
dottorblaster, line 66: I don't know if you went with this on purpose but another way you could declare it is:

```suggestion
    classNames(
      {
        'text-green-900': isFocused,
        'bg-green-100': isFocused,
      },
      'cursor-default select-none relative py-2 px-3'
    ),
```

dottorblaster (changes requested):
Just one tiny nitpick but really great job! :heart:

assets/js/common/MultiSelect/MultiSelect.jsx
arbulu89: just followed `react-select` examples hehe
I will change anyway, it is the way we use in other places indeed

arbulu89: done!
thank you mate

CDimonaco (approved):
LGTM

:chefkiss:

EMaksy (approved):
LGTM!

jagabomb (commented):
Hey @arbulu89, this looks great so far. Can we make the selections the same green as the tag on the Host Detail screen?
Screenshot 2024-05-08 at 11 06 48

PR #2590, opened 2024-05-06
Add top-right corner profile button & menu

rtorrero:
# Description
This PR adds the profile icon menu in the top right corner.

Captura desde 2024-05-06 07-25-41
Captura desde 2024-05-06 07-25-33

NOTE: The state changes are just to prove that, the usage of the /me endpoint will be gone before the PR is merged in favour of a newer implementation that @CDimonaco is working on that include these fields plus many more
NOTE: Still adding storybook changes

## How was this tested?

Tests added

PR #2601, opened 2024-05-07
Use profile endpoint

rtorrero:
# This is in draft pending PR#2590

### DIFF
--- a/assets/js/pages/Layout/Layout.jsx
+++ b/assets/js/pages/Layout/Layout.jsx
@@ -1,6 +1,10 @@
 /* eslint-disable jsx-a11y/anchor-is-valid */
 import React, { useState, useCallback } from 'react';
+import { useSelector } from 'react-redux';
 import { NavLink, Outlet } from 'react-router-dom';
+
+import { clearCredentialsFromStore } from '@lib/auth';
+import { selectUserEmail, selectUserName } from '@state/selectors/user';
 
 import {
   EOS_HOME_OUTLINED,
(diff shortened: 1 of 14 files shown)


assets/js/pages/Layout/Layout.jsx
```diff
@@ -84,9 +84,8 @@ function Layout() {
       : localStorage.setItem('sidebar-collapsed', true);
   }, [isCollapsed]);
 
-  const email = useSelector(selectUserEmail);
-  const username = useSelector(selectUserName);
-
+  const userProfile = useSelector(getUserProfile);
```
EMaksy, line 87: We could destruct user Profile for username and email and pass it as props for profile menu
``` const { username, email } = useSelector(getUserProfile);```

File: assets/js/pages/Layout/Layout.jsx [hunk 2]
```diff
@@ -175,8 +174,8 @@ function Layout() {
             <div className="relative flex flex-col justify-end h-full px-8 md:w-full">
               <div className="relative p-5 flex items-center w-full space-x-8 justify-end mr-20">
                 <ProfileMenu
-                  username={username}
-                  email={email}
+                  username={userProfile.username}
```
EMaksy, line 177: If we destruct userProfile we `could` do something like this
```
      <ProfileMenu
        username={username}
        email={email}
        logout={logout}
      />
    </div>
```

what do you think ?

rtorrero:
Closing, changes got added to pr #2590

PR #2602, opened 2024-05-07
Ensure SUMA related controllers return 403 when credentials were not set

nelsonkopliku:
# Description

Following up an offline discussion about avoiding returning 401 for non trento related auth.

## How was this tested?

Tests updated.


jamie-suse (approved):

dottorblaster (approved):
LGTM

PR #2607, opened 2024-05-08
Fix users/profile controller tests

arbulu89:
# Description

Fix a test that now fails due most probably bad conflict resolution.


EMaksy (commented):
Oh, so this was breaking yesterday. 
Thanks for taking care!

PR #2608, opened 2024-05-08
Switch to use link in ProfileMenu

rtorrero:
# Description

The profile menu was wrongly using an <a> element for linking. This should fix it. Also:
  - test have been adjusted
  - storybook now includes a way to prevent navigation when clicking on the profile link

### DIFF
--- a/assets/js/common/ProfileMenu/ProfileMenu.jsx
+++ b/assets/js/common/ProfileMenu/ProfileMenu.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Menu } from '@headlessui/react';
+import { Link } from 'react-router-dom';
 
 import { EOS_ACCOUNT_CIRCLE_OUTLINED } from 'eos-icons-react';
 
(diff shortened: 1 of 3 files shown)


assets/js/common/ProfileMenu/ProfileMenu.jsx
```diff
@@ -43,12 +44,12 @@ function ProfileMenu({ username, email, logout }) {
           </span>
         </Menu.Item>
         <Menu.Item>
-          <a
+          <Link
```
arbulu89, line 47: Doesn't `Link` use `to` instead of `href`?

rtorrero: yes, I went a bit too fast :sweat_smile:

arbulu89 (approved):
Thank you!
Just test at least manually that everything works hehe

CDimonaco (approved):

EMaksy (approved):

PR #2609, opened 2024-05-08
Add optimistic lock profile

arbulu89:
# Description

Add optimistic locking to the profile patch operation, for the same reason we do for the users patch.

PD: It duplicates some functions that we use in the `users_controller`. I think by now it is OK, but at some point, if we continue having this feature expanded, maybe we want to have something more generic

PD2: This PR doesn't change the frontend. I don't know how we will implement the `etag` usage, if we will query `/api/v1/profile` everytime we go to the page, or we just get it from redux. I would vote for the first, but now i'm wondering why did we store values like `createdAt` and `updatedAt` in the store

## How was this tested?

Tested with UT

### DIFF
--- a/lib/trento/users.ex
+++ b/lib/trento/users.ex
@@ -72,6 +72,8 @@ defmodule Trento.Users do
     user
     |> User.profile_update_changeset(attrs)
     |> Repo.update()
+  rescue
+    Ecto.StaleEntryError -> {:error, :stale_entry}
   end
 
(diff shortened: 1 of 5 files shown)


arbulu89:
Closing the PR, as in fact, we don't need this feature.
If a user is in the profile page, and some admin changes its values, this will trigger an automatic page refresh for that user, so we wouldn't be in any inconsistent state.

PR #2611, opened 2024-05-08
Users abilities frontend

arbulu89:
# Description

Load abilities and user specific abilities in the users creation and edit views.
Now, we can update them.

It uses the new [MultiSelect](https://github.com/trento-project/web/pull/2600) component
multi-select

## How was this tested?

UT added together with some manual testing

### DIFF
--- a/assets/js/pages/Users/UserForm.jsx
+++ b/assets/js/pages/Users/UserForm.jsx
@@ -5,6 +5,7 @@ import { format, parseISO } from 'date-fns';
 import Button from '@common/Button';
 import Input, { Password } from '@common/Input';
 import Label from '@common/Label';
+import MultiSelect from '@common/MultiSelect';
 import Select from '@common/Select';
 import Tooltip from '@common/Tooltip';
 import { getError } from '@lib/api/validationErrors';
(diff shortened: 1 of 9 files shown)


CDimonaco (approved):
LGTM

assets/js/pages/Users/UserForm.jsx
```diff
@@ -58,6 +69,9 @@ function UserForm({
   const [confirmPasswordState, setConfirmPassword] = useState('');
   const [confirmPasswordErrorState, setConfirmPasswordError] = useState(null);
   const [statusState, setStatus] = useState(status);
+  const [selectedAbilities, setAbilities] = useState(
+    userAbilities.map(({ id }) => id)
```
rtorrero, line 73: Wouldn't be something like "effectiveAbilities" more appropriate? I'm asking because this seems to reflect the abilities that are effective at a certain moment and not the ones that have been selected in terms of UI, but not yet submitted. (Hope i'm not misunderstanding)

rtorrero (approved):
LGTM

assets/js/pages/Users/UserForm.jsx
arbulu89: `selectedAbilities` indeed reflects the abilities that are selected for that user in that moment in the UI.
Initially, the abilities that are currently stored in the DB of course.
When you change them using the UI, the values are changed.
If they are submitted or not is irrelevant for the state.

arbulu89: And if you mean the `userAbilities`, this are the initial user abilities.
It could be named with many other ways yes, like `currentAbilities`, `userCurrentAbilities`, etc
I think it is quite explicit what it means

PR #2612, opened 2024-05-08
Forbidden guard for user entries

arbulu89:
# Description
Add the new `ForbiddenGuard` component that takes care of permitting or forbidding certain user related views and entries, in this case the `Users` entry in the side bar and any user related view.

The `ForbiddenGuard` has 2 modes:
- "Normal", which just removes the guarded children if they are not permitted
- Outlet, where it forbids the complete outlet entry showing a forbidden text

forbidden_guard

## How was this tested?

Tests added, and some manual testing

### DIFF
--- a/assets/js/pages/ForbiddenGuard/ForbiddeGuard.test.jsx
+++ b/assets/js/pages/ForbiddenGuard/ForbiddeGuard.test.jsx
@@ -1,0 +1,126 @@
+import React from 'react';
+
+import { render, screen } from '@testing-library/react';
+import 'intersection-observer';
+import '@testing-library/jest-dom';
+import { Route, Routes } from 'react-router-dom';
+import { renderWithRouter, withState } from '@lib/test-utils';
+import ForbiddenGuard from './ForbiddenGuard';
+
+describe('ForbiddenGuard component', () => {
+  describe('Normal guard', () => {
+    it('should permit children if the user is authorized', () => {
+      const [StatefulGuard] = withState(
+        <ForbiddenGuard permitted={['show:test']}>
+          <div>Permitted</div>
+        </ForbiddenGuard>,
+        {
+          user: {
+            abilities: [{ name: 'show', resource: 'test' }],
+          },
+        }
+      );
+
+      render(StatefulGuard);
+
+      expect(screen.getByText('Permitted')).toBeVisible();
+    });
+
+    it('should permit for user with all abilities', () => {
+      const [StatefulGuard] = withState(
+        <ForbiddenGuard permitted={[]}>
+          <div>Permitted</div>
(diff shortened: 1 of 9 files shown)


jagabomb (approved):
From a UX point of view, I like how this works. From a UI point of view, I wonder if we should make the **Access Denied** view more bolder... this is just a thought for a future PR/change.

arbulu89:
> From a UX point of view, I like how this works. From a UI point of view, I wonder if we should make the **Access Denied** view more bolder... this is just a thought for a future PR/change.

@jagabomb @abravosuse Feel free to suggest any content there, for this PR (if it is simple) or for a future one if it is more complex

CDimonaco (approved):

assets/js/pages/ForbiddenGuard/ForbiddeGuard.test.jsx
```diff
@@ -0,0 +1,126 @@
...
+        {
+          user: {
+            abilities: [],
+          },
+        }
+      );
+
+      renderWithRouter(StatefulGuard);
+
+      expect(
```
rtorrero, line 121: Maybe it could be added a expectation that the forbidden content is not rendered?

assets/js/trento.jsx
```diff
@@ -116,7 +114,18 @@ function App() {
                     path="hosts/:targetID/executions/last/:checkID/:resultTargetType/:resultTargetName"
                     element={<CheckResultDetailPage targetType={TARGET_HOST} />}
                   />
-                  <Route path="users/new" element={<CreateUserPage />} />
+                  <Route
```
rtorrero, line 117: I'm not fully understanding what this does, I guess it hides all other routes behind the ForbiddenGuard but why does it need to be a Route element without a path? (sorry for the potentially dumb question :sweat_smile: )

rtorrero (approved):
LGTM

assets/js/trento.jsx
arbulu89: Without using the `path`, we implement a layout kind of route:
https://reactrouter.com/en/main/route/route#layout-routes

It applies the guard to the routes that are within this element, and returns the outlet if needed.
It is the same way we protect all the routes with the login, so the not logged users are redirected to the guarded view.

assets/js/pages/ForbiddenGuard/ForbiddeGuard.test.jsx
arbulu89: I don't get it...
In this case the forbidden content is always rendered.
If you mean that we should have other test case for that, i think the test above does that.
Both cannot coexist, it is not compatible with how the outlet works

PR #2613, opened 2024-05-08
User edit create toast

EMaksy:
# Description

This pr will enrich the CreateUser and EditUser pages.
When a user was created or edited, a success toast will pop up to inform the user.
Similar like in the Users Page. 

Also added a toast to inform the user when something went wrong.
The biggest case would be if a network error happens.

### Create user:
[create_toast.webm](https://github.com/trento-project/web/assets/54111255/50c05c2f-4d57-4482-8c1b-0b8b5ab7fa6a)

### Edit user
[edit_toast.webm](https://github.com/trento-project/web/assets/54111255/e2e3ad6c-8956-433d-8694-bdbaf9d5b4ab)

### Network Error on EditPage when web server crashed
[network_error_edit.webm](https://github.com/trento-project/web/assets/54111255/4941bcb2-ecee-4626-875a-5f4913236017)

## How was this tested?
Enriched existing test
Added network test

### DIFF
--- a/assets/js/pages/Users/CreateUserPage.jsx
+++ b/assets/js/pages/Users/CreateUserPage.jsx
@@ -11,6 +11,8 @@ import { createUser } from '@lib/api/users';
 import UserForm from './UserForm';
 
 const ERROR_GETTING_ABILITIES = 'Error getting user abilities';
+const SUCCESS_USER_CREATION = 'User created succesfuly';
+const UNEXPECTED_ERROR_MESSAGE = 'Unexpected error occurred, refresh the page';
 
 export const fetchAbilities = (setAbilities) => {
   listAbilities()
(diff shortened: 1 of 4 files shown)


arbulu89 (approved):
Looks good @EMaksy !
Don't you want to put a toast message if the `getUser` fails in the edit view?

Otherwise, we would get a simple `Not found` view, but without the same warning that an unexpected error happened.

Just think about it hehe

I guess it would be as simple as in the get pipeline:

```
.catch((error) => {
  if (!error.response) {
    toast.error(ERROR)
  }
})
```

CDimonaco (approved):

rtorrero (approved):
No comments from my side, great work :-)

PR #2620, opened 2024-05-13
Add disabled prop to multi-select

arbulu89:
# Description

Add the `disabled` prop to `MultiSelect`. It is true that we could've used the `isDisabled` passed as `prop`, but as we have `disabled` all over the code, it looks more appropriate.

Now, the background color is changed to `gray-50` when disabled.

To use it together with an empty placeholder, just do:
```
<MultiSelect
  ...
  disabled
  placeholder=""
/>
```

image

## How was this tested?

Tested


CDimonaco (approved):
LGTM

PR #2622, opened 2024-05-13
Rescue StaleError in user update function when abilities are present

CDimonaco:
# Description

Fix optimistic locking handling in the users context when we update a user with abilities.


arbulu89 (approved):
Thank you!
Thank you git for you conflicts :trollface:

dottorblaster (approved):
OUCH

PR #2623, opened 2024-05-13
Add PatchList component

janvhs:
# Description

Add a new `PatchList` common component. It is based on the already implemented `Table` and enhances it.
The icons have the same colour, already present in the standalone icon components.

## How was this tested?

Storybook

### DIFF
--- a/assets/js/common/PatchList/PatchList.jsx
+++ b/assets/js/common/PatchList/PatchList.jsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { Link } from 'react-router-dom';
 import {
   EOS_SHIELD_OUTLINED,
   EOS_CRITICAL_BUG_OUTLINED,
@@ -71,9 +70,13 @@ const patchListConfig = {
       title: 'Advisory',
       key: 'advisory_name',
       render: (content, item) => (
-        <Link className="text-jungle-green-500 hover:opacity-75" to={item.url}>
+        <button
+          type="button"
+          className="text-jungle-green-500 hover:opacity-75"
+          onClick={() => item?.onNavigate?.()}
+        >
           {content}
-        </Link>
+        </button>
       ),
     },
     {
--- a/assets/js/common/PatchList/PatchList.jsx
+++ b/assets/js/common/PatchList/PatchList.jsx
@@ -1,0 +1,96 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import {
+  EOS_SHIELD_OUTLINED,
+  EOS_CRITICAL_BUG_OUTLINED,
+  EOS_ADD_BOX_OUTLINED,
+} from 'eos-icons-react';
+import { format as formatDate } from 'date-fns';
+import classNames from 'classnames';
+
+import Table from '@common/Table';
+import { computedIconCssClass } from '@lib/icon';
+
+function iconFromAdvisoryType(
+  advisoryType,
+  centered = false,
+  hoverOpacity = true,
+  size = 'l'
+) {
+  const hoverOpacityClass = {
+    'hover:opacity-75': hoverOpacity,
+    'hover:opacity-100': !hoverOpacity,
+  };
+
+  switch (advisoryType) {
+    case 'security_advisory':
+      return (
+        <EOS_SHIELD_OUTLINED
+          className={classNames(
+            hoverOpacityClass,
+            computedIconCssClass('fill-red-500', centered)
+          )}
+          size={size}
+        />
+      );
+    case 'bugfix':
+      return (
+        <EOS_CRITICAL_BUG_OUTLINED
+          className={classNames(
+            hoverOpacityClass,
+            computedIconCssClass('fill-yellow-500', centered)
+          )}
+          size={size}
+        />
+      );
+    case 'enhancement':
+      return (
+        <EOS_ADD_BOX_OUTLINED
+          className={classNames(
+            hoverOpacityClass,
+            computedIconCssClass('fill-yellow-500', centered)
+          )}
+          size={size}
+        />
+      );
+    default:
+      return null;
+  }
+}
+
+const patchListConfig = {
+  usePadding: false,
+  pagination: true,
+  columns: [
+    {
+      title: 'Type',
+      key: 'advisory_type',
+      render: (content, _) => {
+        const PrimaryIcon = iconFromAdvisoryType(content);
+        return PrimaryIcon;
+      },
+    },
+    {
+      title: 'Advisory',
+      key: 'advisory_name',
+      render: (content, item) => (
+        <Link className="text-jungle-green-500 hover:opacity-75" to={item.url}>
+          {content}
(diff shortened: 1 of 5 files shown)


assets/js/common/PatchList/PatchList.jsx
```diff
@@ -0,0 +1,96 @@
...
+import Table from '@common/Table';
+import { computedIconCssClass } from '@lib/icon';
+
+function iconFromAdvisoryType(
+  advisoryType,
+  centered = false,
+  hoverOpacity = true,
+  size = 'l'
+) {
+  const hoverOpacityClass = {
```
dottorblaster, line 20:
```suggestion
  const hoverOpacityClassName = {
```

assets/js/common/PatchList/PatchList.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+import Table from '@common/Table';
+import { computedIconCssClass } from '@lib/icon';
+
+function iconFromAdvisoryType(
+  advisoryType,
+  centered = false,
+  hoverOpacity = true,
+  size = 'l'
+) {
```
dottorblaster, line 19: Try to use arrow functions!

```suggestion
const iconFromAdvisoryType = (
  advisoryType,
  centered = false,
  hoverOpacity = true,
  size = 'l'
) => {
```

assets/js/lib/test-utils/factories/relevantPatches.js
```diff
@@ -0,0 +1,16 @@
+import { faker } from '@faker-js/faker';
+import { Factory } from 'fishery';
+
+// Ported values from ${PROJECT_ROOT}/lib/trento/software_updates/enums/advisory_type.ex
```
dottorblaster, line 4: I think you can strip this comment :+1:

assets/js/lib/test-utils/factories/relevantPatches.js
```diff
@@ -0,0 +1,16 @@
...
+const advisoryType = ['security_advisory', 'bugfix', 'enhancement'];
+
+// Ported version of relevant_patch_factory in ${PROJECT_ROOT}/test/support/factory.ex
```
dottorblaster, line 7: I think you can strip this comment :+1:

dottorblaster (changes requested):
Some tiny changes to do, but definitely good job :+1: 

Can you write a Jest test for this?

assets/js/common/PatchList/PatchList.jsx
```diff
@@ -0,0 +1,96 @@
...
+
+const patchListConfig = {
+  usePadding: false,
+  pagination: true,
+  columns: [
+    {
+      title: 'Type',
+      key: 'advisory_type',
+      render: (content, _) => {
+        const PrimaryIcon = iconFromAdvisoryType(content);
```
CDimonaco, line 69: Please return directly the function content without intermediate variables

assets/js/common/PatchList/PatchList.stories.jsx
```diff
@@ -0,0 +1,50 @@
...
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
+    ),
+  ],
+  render: (args) => <PatchList {...args} />,
+};
+
+export const Default = {
+  args: fakePatchArgs,
```
CDimonaco, line 33: Could we avoid the intermediate variable?

It's better to have the structure right here to see, like { patches: yucallthefactory}

assets/js/common/PatchList/PatchList.stories.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const NoPatches = {
+  args: {
+    patches: undefined,
+  },
+};
+
+export const AllStates = {
+  args: {
+    patches: [
```
CDimonaco, line 44: That's good

CDimonaco (changes requested):
Everything @dottorblaster mentioned plus some other comments

But a good job, just some nitpicks 🚀

assets/js/common/PatchList/PatchList.jsx
janvhs: Just wondering, what's the reason for this?

janvhs: Should the function be uppercased to, since it returns a compoenent instance?

assets/js/common/PatchList/PatchList.stories.jsx
janvhs: yes, was an orphan from eralier code :D

assets/js/common/PatchList/PatchList.jsx
dottorblaster: We usually prefer arrow functions since they have lower memory footprint, even tho they get transpiled :smile: No need for an uppercase!

assets/js/common/PatchList/PatchList.jsx
```diff
@@ -0,0 +1,93 @@
...
+      render: (content, _) => iconFromAdvisoryType(content),
+    },
+    {
+      title: 'Advisory',
+      key: 'advisory_name',
+      render: (content, item) => (
+        <Link className="text-jungle-green-500 hover:opacity-75" to={item.url}>
+          {content}
+        </Link>
+      ),
```
dottorblaster, line 77: Move this to `onNavigate` like other components do. This way we don't depend directly from `react-router-dom` and we can test the component without wrapping it in a `MemoryRouter` :+1:

dottorblaster (changes requested):
Last thing is having an `onNavigate` prop instead of a plain `react-router-dom`'s `<Link />`.

Sorry, I didn't spot this before! :heart: Usually we provide `navigate` from above using `useNavigate()` in the `Page` component :+1:

assets/js/common/PatchList/PatchList.jsx
janvhs: @dottorblaster I changed the patches data, so it can hold an onNavigate function.

```jsx
patches = [
  {
    ....
    onNavigate = () => navigate(`/patch/${id}`)
  }
]

<PatchList patches={patches} />
```

The other alternative would have been to do something like this:

```jsx
<PatchList patches={[{...}]} onNavigage = {(patch) => navigate(`/patch/${patch.id}`)} />
```

I think I prefer the first solution, because it leaves more flexibility, but the latter API is cleaner. What do you think?

dottorblaster: @janvhs usually we pass an onNavigate function to the whole component and then encapsulate the logic inside, since it's really something tied to the rendering layer. This way we can compose the data we have from e.g. the Redux store or whatever external source with the actual implementation detail of the rest of the frontend.

The on-click navigation is really something that only matters for the rendering.

dottorblaster (commented):
LGTM, when the merge freeze is over I'll put the green here :+1: :fire:

dottorblaster (approved):

PR #2625, opened 2024-05-14
Password change request frontend

arbulu89:
# Description

Frontend part for the password change request feature.
The most relevant thing is the new `customNotify` and the `ToastNotificactions` pseudo react component.
We need to have it apart, as redux doesn't recommend to put non serializable data into actions (like react data in this case): https://redux.js.org/style-guide/#do-not-put-non-serializable-values-in-state-or-actions

In this case, using them is not that horrible, as they don't end up in any reducer state (which is something we shouldn't never do).

We could "ignore" the raised error (https://redux-toolkit.js.org/usage/usage-guide#working-with-non-serializable-data), and it would perfectly work, and the code would be way simpler, as we could simply put the component in the saga code itself.

With this `customNotify` we get the chance to simply send the id, and get the component from a predefined map.
It is a bit more complex, but we follow the redux recommendation by the book.

We can go with the "ignore" option as well, I don't see it that bad.

PD: More I see, more I prefer the "ignore" warning option. Ignoring means just putting the `NOTIFICATION` and `DISMISSABLE_NOTIFICATION` in a ignored list for the middleware. In fact, right now, the API key notifications are not having this warning. What do you think @CDimonaco @dottorblaster ?

Some other changes:
- Change the profile menu index so it is always in top of the toasts
- Move the toaster inside the router. Otherwise we cannot use `Link` component, for example.
- Fix `dismissionNotification` action. It was not working well. The `id` was not used properly, and the `toast.dismiss` was getting `undefined` as id, which made to dismiss all the existing toasts

Fixes # (issue)

## How was this tested?

### DIFF
--- a/assets/js/common/ProfileMenu/ProfileMenu.jsx
+++ b/assets/js/common/ProfileMenu/ProfileMenu.jsx
@@ -6,7 +6,7 @@ import { EOS_ACCOUNT_CIRCLE_OUTLINED } from 'eos-icons-react';
 
 function ProfileMenu({ username, email, logout }) {
   return (
-    <Menu as="div" className="relative inline-block text-left z-10">
+    <Menu as="div" className="relative inline-block text-left z-[100]">
       <Menu.Button as="button" className="group">
         <span className="flex items-center">
           <EOS_ACCOUNT_CIRCLE_OUTLINED
--- a/assets/js/state/notifications.js
+++ b/assets/js/state/notifications.js
@@ -19,12 +19,9 @@ export const dismissableNotify = createAction(
   })
 );
 
-export const dismissNotification = createAction(
-  DISMISS_NOTIFICATION,
-  ({ notificationID }) => ({
-    payload: { notificationID },
-  })
-);
+export const dismissNotification = createAction(DISMISS_NOTIFICATION, (id) => ({
+  payload: { id },
+}));
 
 export const customNotify = createAction(
   CUSTOM_NOTIFICATION,
--- a/assets/js/state/notifications.js
+++ b/assets/js/state/notifications.js
@@ -3,6 +3,7 @@ import { createAction } from '@reduxjs/toolkit';
 export const NOTIFICATION = 'NOTIFICATION';
 export const DISMISSABLE_NOTIFICATION = 'DISMISSABLE_NOTIFICATION';
 export const DISMISS_NOTIFICATION = 'DISMISS_NOTIFICATION';
+export const CUSTOM_NOTIFICATION = 'CUSTOM_NOTIFICATION';
 
 export const notify = createAction(
   NOTIFICATION,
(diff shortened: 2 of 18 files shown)


assets/js/state/notifications.js
```diff
@@ -20,7 +21,14 @@ export const dismissableNotify = createAction(
 
 export const dismissNotification = createAction(
   DISMISS_NOTIFICATION,
-  ({ notificationID }) => ({
-    payload: { notificationID },
+  (id) => ({
```
arbulu89, line 24: This change is the mentioned error to fix the bug of dismissing all toasts

assets/js/trento.jsx
```diff
@@ -45,8 +45,11 @@ function App() {
 
   return (
     <Provider store={store}>
-      <Toaster position="top-right" />
       <BrowserRouter>
+        <Toaster
```
arbulu89, line 49: As commented, we need to have this to be able to use `Link` in the toasts (everything router related)

CDimonaco (approved):
Good job!

I think the right choice is to not put any serializable data, it's true that we can have exception, and I worked in the past with these exceptions, it was a no go in the long run.

With this approach we can mantain the redux state in good shape and also achieve the result in a pretty predictable and friend way

LGTM!

arbulu89:
> I think the right choice is to not put any serializable data, it's true that we can have exception, and I worked in the past with these exceptions, it was a no go in the long run.

In this case, we need to fix the notification usages in the `settings.jsx` saga as they send the warning icon, which is not serializable

CDimonaco:
> > I think the right choice is to not put any serializable data, it's true that we can have exception, and I worked in the past with these exceptions, it was a no go in the long run.
> 
> In this case, we need to fix the notification usages in the `settings.jsx` saga as they send the warning icon, which is not serializable
> 
> 

We can send the icon name, the plain eos one and we can grab the real icon from the default export of the eos library, like the storybook notification playground

assets/js/common/ProfileMenu/ProfileMenu.jsx
```diff
@@ -6,7 +6,7 @@ import { EOS_ACCOUNT_CIRCLE_OUTLINED } from 'eos-icons-react';
 
 function ProfileMenu({ username, email, logout }) {
   return (
-    <Menu as="div" className="relative inline-block text-left z-10">
+    <Menu as="div" className="relative inline-block text-left z-[100]">
```
EMaksy, line 9: Nice catch !

EMaksy (approved):
Very interesting, lgtm!

arbulu89:
@CDimonaco alright!
I have added the `isHealthIcon` prop to the notifications. This way, we can just put the string to render the icon.
I chose this option before the EOS strings, because the EOS icons come without any style/size/etc, and to have currently existing health icons would require to pass still much more props for the icons, which looks an overkill.
In general, I would recommend to use "plain" icons instead of EOS ones if possible

CDimonaco:
> @CDimonaco alright! I have added the `isHealthIcon` prop to the notifications. This way, we can just put the string to render the icon. I chose this option before the EOS strings, because the EOS icons come without any style/size/etc, and to have currently existing health icons would require to pass still much more props for the icons, which looks an overkill. In general, I would recommend to use "plain" icons instead of EOS ones if possible

Perfect, green light from me

PR #2629, opened 2024-05-16
Update alerting.md

stefanotorresi:
# Description

Some colleagues from Premium Support wanted a more explicit clarification that SMTP authentication is a strict requirement.

yes

## How was this tested?

This is documentation only, no tests are needed.


nelsonkopliku (approved):
LGTM

PR #2630, opened 2024-05-16
Clean and improve e2e performance

arbulu89:
# Description
Two improvements in the e2e tests.

- Load initial photofinish scenario only if needed. With the generic `before`, we are running it even if we don't need, and for example, the `settings` and `catalog` don't need this, so it reduces some time the generic time of execution
- Clean some unused code

This PR could've gone to `main`, but as we are in a merge freeze...

### DIFF
--- a/test/e2e/cypress/e2e/clusters_overview.cy.js
+++ b/test/e2e/cypress/e2e/clusters_overview.cy.js
@@ -81,7 +81,7 @@ context('Clusters Overview', () => {
           clusterIdByName(healthyClusterScenario.clusterName),
           healthyClusterScenario.checks
         );
-        cy.setMockRunnerExpectedResult(healthyClusterScenario.result);
+        // wip: set expected results
         cy.requestChecksExecution(
           clusterIdByName(healthyClusterScenario.clusterName)
         );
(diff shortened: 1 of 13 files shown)


test/e2e/cypress/e2e/clusters_overview.cy.js
```diff
@@ -81,7 +82,7 @@ context('Clusters Overview', () => {
           clusterIdByName(healthyClusterScenario.clusterName),
           healthyClusterScenario.checks
         );
-        cy.setMockRunnerExpectedResult(healthyClusterScenario.result);
+        // wip: set expected results
```
EMaksy, line 85: I guess you plan here additional rework ?

EMaksy (approved):
Great improvement!

rtorrero (approved):
Looks good, I had the same question that @EMaksy already made :+1:

CDimonaco (approved):

test/e2e/cypress/e2e/clusters_overview.cy.js
arbulu89: I don't know...
This were some test that were `skipped` since long time ago.
I just wanted to remove the dead code.
If somebody wants to work on them, feel free hehe

PR #2631, opened 2024-05-16
e2e user management

arbulu89:
# Description

User management epic e2e tests.
I have updated some commands to enable login for multiple users in a more convenient way.

Depends on: https://github.com/trento-project/web/pull/2630

### DIFF
--- a/assets/js/lib/test-utils/factories/users.js
+++ b/assets/js/lib/test-utils/factories/users.js
@@ -12,7 +12,6 @@ export const abilityFactory = Factory.define(() => ({
 export const userFactory = Factory.define(() => ({
   id: faker.number.int(),
   username: faker.internet.userName(),
-  actions: 'Delete',
   enabled: faker.datatype.boolean(),
   fullname: faker.internet.displayName(),
   email: faker.internet.email(),
--- a/test/e2e/cypress/e2e/users.cy.js
+++ b/test/e2e/cypress/e2e/users.cy.js
@@ -90,8 +90,6 @@ describe('Users', () => {
     cy.url().should('include', '/users');
   });
 
-  describe('List users', () => {});
-
   describe('Create user', () => {
     it('should redirect to user creation form', () => {
       cy.contains('button', 'Create User').click();
--- a/test/e2e/cypress/e2e/users.cy.js
+++ b/test/e2e/cypress/e2e/users.cy.js
@@ -1,0 +1,326 @@
+import { userFactory } from '@lib/test-utils/factories/users';
+
+const PASSWORD = 'password';
+const USER = userFactory.build({ username: 'e2etest' });
+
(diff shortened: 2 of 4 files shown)


assets/js/lib/test-utils/factories/users.js
```diff
@@ -12,7 +12,6 @@ export const abilityFactory = Factory.define(() => ({
 export const userFactory = Factory.define(() => ({
   id: faker.number.int(),
   username: faker.internet.userName(),
-  actions: 'Delete',
```
EMaksy, line 15: Just curious why :) ?

EMaksy (approved):
Hey man, looks great!
Just left one question and one suggestion, but besides this everything top!

rtorrero (approved):
Good stuff @arbulu89, thanks!

assets/js/lib/test-utils/factories/users.js
arbulu89: Because this field doesn't even exist.
I guess you added it at the beginning before knowing how it would've been.
It is `abilities` and it comes below. We should've removed this before I guess

test/e2e/cypress/e2e/users.cy.js
```diff
@@ -0,0 +1,326 @@
...
+    it('should fail editing user password if current password is wrong', () => {
+      cy.contains('button', 'Change Password').click();
+      cy.get('input').eq(4).type('wrong');
+      cy.get('input').eq(5).type(PASSWORD);
+      cy.get('input').eq(6).type(PASSWORD);
+      cy.get('button:contains("Save")').eq(1).click();
+      cy.get('p').contains('Is invalid');
+    });
+
+    it('should edit the user password', () => {
```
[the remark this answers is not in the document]
arbulu89: We didn't include this scenario in our "test list". 
I guess there are many many more scenarios that we don't test.
Anyway, since you want this one specifically, I will add it hehe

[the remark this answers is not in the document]
arbulu89: @EMaksy I couple of new tests added.
Admin user cannot be edited in the users view and admin cannot be edited in the profile view.

test/e2e/cypress/e2e/users.cy.js
```diff
@@ -0,0 +1,341 @@
...
+        method: 'GET',
+        auth: { bearer: accessToken },
+        body: {},
+      })
+      .then(({ body: users }) => {
+        users.forEach(({ id }) => {
+          if (id === 1) {
+            return;
+          }
+          cy.request({
```
CDimonaco, line 25: If we want to wait until all the users are really deleted, we could map on users and accumulate cy.request promises and do then or await on Promise.all

CDimonaco (approved):
LGTM! Good job!

PR #2632, opened 2024-05-17
Add HTTPS using cert-manager for demo env

rtorrero:
# Description

This PR enables the usage of cert-manager during the deployment of our demo environment. This will result in https://demo.trento-project.io having a validly signed certificate after deployment.

## How was this tested?
This was tested manually on the demo env by deploying the helm chart using the same commands that are added here in the CI.

NOTE: Setting this as draft until we merge https://github.com/trento-project/helm-charts/pull/95. That PR is needed before this can work.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -752,6 +752,10 @@ jobs:
         )
     env:
       IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}
+      TRENTO_DOMAIN: ${{ secrets.TRENTO_DEMO_IP }}
+      TRENTO_NAMESPACE: ${{ secrets.TRENTO_NAMESPACE }}
+      TRENTO_ADMIN_EMAIL: ${{ secrets.TRENTO_ADMIN_EMAIL }}
+      TRENTO_INGRESS_CLASS: ${{ secrets.TRENTO_INGRESS_CLASS }}
     needs: [build-demo-img, test-e2e]
     steps:
       - name: Start a local k8s cluster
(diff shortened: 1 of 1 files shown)


rtorrero:
As an alternative to this (or even in addition to this) I've created another helm chart that instead of using the manual cert-manager installation approach, uses the cert-manager helm chart as a subchart, alongside with trento-server. If that is a preferred solution, I could push it under e.g. `trento-project/trento-server-tls`.

arbulu89 (approved):
I like the simple approach.
My unique comment is that we wouldn't really need to install cert-manager all the times.
Doing it once in the machine would be enough. Anyway, if it doesn't take long, not a big deal, and in fact, we remove that need if we change machine.

Besides that, one thing we could do is just to have this as an optional thing, with some other env variable (like `DEPLOY_DEMO`). But this is just a potential improvement.

If others are happy, I'm as well

CDimonaco (approved):
Great job, lgtm

.github/workflows/ci.yaml
```diff
@@ -752,22 +752,40 @@ jobs:
         )
     env:
       IMAGE_REPOSITORY: ghcr.io/${{ github.repository_owner }}
+      TRENTO_DOMAIN: ${{ secrets.TRENTO_DEMO_IP }}
+      TRENTO_NAMESPACE: ${{ secrets.TRENTO_NAMESPACE }}
```
nelsonkopliku, line 756: Hey @rtorrero I don't see a `TRENTO_NAMESPACE` secret here https://github.com/trento-project/web/settings/secrets/actions

Is it intentional?

rtorrero: added!

nelsonkopliku (approved):

PR #2636, opened 2024-05-20
Add upgradable packages list component

jamie-suse:
# Description

This PR add an upgradable packages list component.

Example output:

image

## How was this tested?

Added unit test and Storybook demo.


assets/js/common/UpgradablePackagesList/UpgradablePackagesList.jsx
```diff
@@ -0,0 +1,71 @@
...
+      <h1 className="text-3xl w-4/5 p-4">
+        <span className="font-medium">Upgradable Packages: </span>
+        <span
+          className={classNames(
+            'font-bold truncate w-60 inline-block align-top'
+          )}
+        >
+          {hostname}
+        </span>
+      </h1>
```
dottorblaster, line 63: I think we can move all of this into the page component :+1:

assets/js/common/UpgradablePackagesList/UpgradablePackagesList.jsx
```diff
@@ -0,0 +1,71 @@
...
+                {advisory_name}
+              </div>
+            ))}
+          </div>
+        ),
+      },
+    ],
+  };
+
+  const data = upgradablePackages.map((up) => {
```
dottorblaster, line 41:
```suggestion
  const data = upgradablePackages.map((package) => {
```

dottorblaster (changes requested):
Just a couple things then we can merge, nice job! :muscle:

assets/js/common/UpgradablePackagesList/UpgradablePackagesList.jsx
jamie-suse:
```
Parsing error: The keyword 'package' is reserved

Identifier expected. 'package' is a reserved word in strict mode. Modules are automatically in strict mode.
```

dottorblaster: Frown

dottorblaster (approved):
LGTM

PR #2640, opened 2024-05-21
Add validation when a password is generated

EMaksy:
# Description

This pr aims to improve the password generation in order to support our password rules, as the current implementation has a very unlikely case not to meet the requirements.

 

Password Rules:
- at least have 8 characters --> added validation to check for minimum length
- does not have 3 consecutive repeated numbers or letters (example: 111 or aaa) --> added validation for repetitive chars by using a regex rule, which checks for any character that appears three or more times consecutively in a string.
- does not have 3 consecutive sequential numbers or letters (example: 123 or abc) --> added [ASCII  code validation](https://www.commfront.com/pages/ascii-chart) for sequence of chars ascending and descending, exluding special chars to allow combinations like 89: or yz{

Also, this pr will ensure the unlikely case that e2e test fail if the generated password does not meet the requirements
It is a follow-up to @rtorrero  comment: https://github.com/trento-project/web/pull/2576#discussion_r1590935557
## How was this tested?

Added test for validation functions

### DIFF
--- a/assets/js/pages/Users/generatePassword.js
+++ b/assets/js/pages/Users/generatePassword.js
@@ -6,28 +6,40 @@ const SPECIAL_CHARACTERS = '~!@#$%^&*()_+={}[]|:;<>,.?/-\'"`';
 const PASSWORD_CHARACTERS =
   DIGITS + UPPERCASE_LETTERS + LOWERCASE_LETTERS + SPECIAL_CHARACTERS;
 
-export const hasValidLength = (password, pwdLength = 8) =>
+export const hasValidPwdLength = (password, pwdLength = 8) =>
   password.length >= pwdLength;
 
 export const validateNoRepetitiveCharacters = (password) =>
   !/(.)\1{2,}/.test(password);
 
+const isAlphanumeric = (char) => /[A-Za-z0-9]/.test(char);
+
 export const validateNoSequentialCharacters = (password) => {
   if (password.length < 3) {
     return true;
   }
-  for (let i = 2; i < password.length; i += 1) {
-    const firstCharCode = password.charCodeAt(i - 2);
-    const secondCharCode = password.charCodeAt(i - 1);
-    const thirdCharCode = password.charCodeAt(i);
 
+  for (let i = 2; i < password.length; i += 1) {
+    const firstChar = password[i - 2];
+    const secondChar = password[i - 1];
+    const thirdChar = password[i];
     if (
-      (secondCharCode === firstCharCode + 1 &&
-        thirdCharCode === secondCharCode + 1) ||
-      (secondCharCode === firstCharCode - 1 &&
-        thirdCharCode === secondCharCode - 1)
+      isAlphanumeric(firstChar) &&
+      isAlphanumeric(secondChar) &&
+      isAlphanumeric(thirdChar)
     ) {
-      return false;
+      const firstCharCode = firstChar.charCodeAt(0);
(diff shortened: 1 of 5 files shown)


assets/js/pages/Users/generatePassword.js
```diff
@@ -6,10 +6,58 @@ const SPECIAL_CHARACTERS = '~!@#$%^&*()_+={}[]|:;<>,.?/-\'"`';
...
+export const validateNoSequentialCharacters = (password) => {
+  if (password.length < 3) {
+    return true;
+  }
+
+  for (let i = 2; i < password.length; i += 1) {
+    const firstChar = password[i - 2];
+    const secondChar = password[i - 1];
+    const thirdChar = password[i];
+    if (
```
EMaksy, line 26: Here we exclude the special chars to ensure combinations like 89: or yz{ are still possible

assets/js/pages/Users/generatePassword.js
```diff
@@ -6,10 +6,58 @@ const SPECIAL_CHARACTERS = '~!@#$%^&*()_+={}[]|:;<>,.?/-\'"`';
...
+      isAlphanumeric(firstChar) &&
+      isAlphanumeric(secondChar) &&
+      isAlphanumeric(thirdChar)
+    ) {
+      const firstCharCode = firstChar.charCodeAt(0);
+      const secondCharCode = secondChar.charCodeAt(0);
+      const thirdCharCode = thirdChar.charCodeAt(0);
+
+      if (
+        (secondCharCode === firstCharCode + 1 &&
```
EMaksy, line 36: Checking ASCII  code for sequence ascending then descending

arbulu89 (commented):
Hey @EMaksy 
Some comments.

- I wouldn't export all the auxiliary functions. Just create a `isValidPassword` or similar, which includes all the other sub functions. And you can test this with all the cases.

- The sequential test looks complex. I would try to get inspiration in the code we have in the backend. It looks simpler: https://github.com/trento-project/web/blob/user_management/lib/trento/users/user.ex#L120

assets/js/pages/Users/generatePassword.test.js
```diff
@@ -1,20 +1,50 @@
...
-  test('generates passwords that vary', () => {
-    const firstPassword = generatePassword();
-    const secondPassword = generatePassword();
+  test('should generate passwords that vary', () => {
+    const firstPassword = generateValidPassword();
+    const secondPassword = generateValidPassword();
     expect(firstPassword).not.toBe(secondPassword);
   });
 });
+describe('isValidPassword', () => {
```
arbulu89, line 21: I would put some new lines between `describe` and `test` entries

assets/js/pages/Users/generatePassword.test.js
```diff
@@ -1,20 +1,50 @@
...
+  const passwordList = [
+    {
+      expectedResult: false,
+      password: '1234567',
```
arbulu89, line 25: I would use other examples, because many of them fail in multiple scenarios.
Like: 
- `short`
- `passwordaaa`
- `passwordBBB`
- `password555`
- `password1234`
- `passwordefgh`
- `passwordKLMNO`

Otherwise, many parts of the code are not tested. Remember that `some` returns in the first match, and we want to loop through all the sequences I guess

arbulu89 (approved):
Really good!
Some comments in the tests, but good to go

CDimonaco (approved):

PR #2646, opened 2024-05-23
User totp fields

arbulu89:
# Description

Include `TOTP` multi factor authentication required fields to the user shema.
I'm excluding any change in the users context, as even though I have ideas of how it should be, I think it will be better to work on them when we actually start working on the features, as I might forget something otherwise.

I just added some small changes in the changeset functions. But well, we might change them afterwards as well.

## How was this tested?

No need for tests for this changes

### DIFF
--- a/lib/trento/users/user.ex
+++ b/lib/trento/users/user.ex
@@ -18,6 +18,8 @@ defmodule Trento.Users.User do
     UsersAbilities
   }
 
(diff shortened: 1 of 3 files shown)


EMaksy (approved):
Code looks good!

CDimonaco (approved):

PR #2647, opened 2024-05-23
Login with totp

arbulu89:
# Description

Add TOTP validation for users that have this feature enabled.

otp_login

## How was this tested?

Tests added and tested manually. We will add some e2e once the whole feature is finished.

### DIFF
--- a/assets/js/pages/Login/Login.jsx
+++ b/assets/js/pages/Login/Login.jsx
@@ -7,6 +7,8 @@ import { useNavigate, useSearchParams } from 'react-router-dom';
 import { initiateLogin } from '@state/user';
 import classNames from 'classnames';
 
+import Input from '@common/Input';
+
 export default function Login() {
   const [username, setUsername] = useState('');
   const [password, setPassword] = useState('');
@@ -78,7 +80,7 @@ export default function Login() {
                     Username
                   </label>
                   <div className="mt-1">
-                    <input
+                    <Input
                       id="username"
                       type="text"
                       data-testid="login-username"
@@ -88,9 +90,10 @@ export default function Login() {
                       name="username"
                       autoComplete="username"
                       required
+                      error={isUnauthorized}
                       className={classNames(
-                        'appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-jungle-green-500 focus:border-jungle-green-500 sm:text-sm',
-                        { 'border-red-300': isUnauthorized }
+                        'appearance-none px-3 py-2 text-inherit border-gray-300 shadow-sm focus:ring-jungle-green-500 focus:border-jungle-green-500 sm:text-sm',
+                        { 'disabled:opacity-50': authInProgress }
                       )}
                     />
                   </div>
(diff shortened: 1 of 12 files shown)


assets/js/pages/Login/Login.jsx
```diff
@@ -55,59 +68,91 @@ export default function Login() {
...
-                />
+                <div>
+                  <label
+                    htmlFor="password"
+                    className="block text-sm font-medium text-gray-700"
+                  >
+                    Password
+                  </label>
+                  <div className="mt-1">
+                    <input
```
CDimonaco, line 107: Could we use the design system Input component, instead of the html one?

If this is a tech debt from the past was because originally the login page was not made with react, I think could be a good opportunity to change, or at least add the otp one as react component and then have a dedicated tech debt for the others

assets/js/pages/Login/Login.jsx
```diff
@@ -55,59 +68,91 @@ export default function Login() {
...
+            ) : (
+              <div>
+                <label
+                  htmlFor="otp-code"
+                  className="block text-sm font-medium text-gray-700"
+                >
+                  TOTP code
+                </label>
+                <div className="mt-1">
+                  <input
```
CDimonaco, line 136: Same comment as before

config/config.exs
```diff
@@ -53,6 +53,9 @@ config :logger, :console,
 # Use Jason for JSON parsing in Phoenix
 config :phoenix, :json_library, Jason
 
+# Redact sensitive parameters in phoenix logs
+config :phoenix, :filter_parameters, ["password", "totp_code"]
```
CDimonaco, line 57: <3

lib/trento_web/controllers/session_controller.ex
```diff
@@ -21,26 +21,30 @@ defmodule TrentoWeb.SessionController do
...
          properties: %{
-           username: %Schema{
+           username: %OpenApiSpex.Schema{
+             type: :string
+           },
+           password: %OpenApiSpex.Schema{
              type: :string
            },
-           password: %Schema{
+           totp_code: %OpenApiSpex.Schema{
```
CDimonaco, line 39: Could we add nullable:false here? If we do I think you should also omit the field in request instead of passing a possible null

CDimonaco (commented):
Clean and easy, very good job!

I left some comments, wdyt?

assets/js/pages/Login/Login.jsx
arbulu89: Alright!
Yes, it was a tech debt, and I didn't want to change in this PR, but the change is minimal, so I will finally do it.

lib/trento_web/controllers/session_controller.ex
arbulu89: mmm, are you sure? This value shouldn't be `null` in any case.
But we don't pass `null`, simply we don't send the field. The payload field has a conditional addition.
Here the "wrong" thing is that `username` and `password` should be required fields, and the other not, but if I add them as required, the non backward compatible CI change starts complaining...
And as we are in a feature branch, it would complain in all the PRs.
Better to do it once the `user_management` branch is merged to `main`, once for all

arbulu89: PD: The default value is `false` indeed: https://spec.openapis.org/oas/v3.0.3#schemaNullable

EMaksy (approved):
Hey, code wise it looks good!
I guess just the alignment with naming to @CDimonaco pr is left https://github.com/trento-project/web/pull/2651

Well done!

assets/js/pages/Login/Login.jsx
arbulu89: @CDimonaco Done!
If you notice that I use `Input` component for the password instead of `Password`, this is because if I use this 2nd, the classes are not inherited properly...
This is because the `Password` uses the `suffix` prop to add the "show password icon".
When you do this, the classes are not passed to the correct component, as rc-input creates an intermediate `span` element that gets the classes. A mess...
It is a known limitation.
As far as I researched, there might be some workarounds, but I don't want to start looking at them now.
As we don't have the icon now, not adding it doesn't look like an urgent thing

lib/trento_web/controllers/session_controller.ex
CDimonaco: you're right! It's totallly fine!

assets/js/pages/Login/Login.jsx
CDimonaco: Let's do that in another pr, it's fine, we just need to remember this debt

CDimonaco (approved):
As always good job!

LGTM

PR #2648, opened 2024-05-23
Fix suma http executor empty body

dottorblaster:
# Description

When testing against a real world scenario I realized I was passing the wrong arguments to `HTTPoison.request`.

## How was this tested?

Unit tests already in place


dottorblaster:
(Also, turns out `Integer.parse` doesn't do what I thought it would do :joy: )

jamie-suse (approved):

PR #2649, opened 2024-05-24
Add host patches page

janvhs:
# Description

This patch set adds and mounts the relevant patches page.
I still need to fix up the styling, so it would be great to get some help with that.

image

## How was this tested?

Jira tests and Storybook


dottorblaster (changes requested):
Just one thing, you might want to rename `HostRelevantPatches` to `HostRelevantPatchesPage` since we have this non-written convention, we just went with it :smile:

PR #2654, opened 2024-05-27
Patches for packages controller

dottorblaster:
# Description
Adding an endpoint to retrieve patches that cover a list of upgradable packages.

## How was this tested?
Unit tests


test/trento_web/controllers/v1/suse_manager_controller_test.exs
```diff
@@ -92,4 +95,48 @@ defmodule TrentoWeb.V1.SUSEManagerControllerTest do
...
+
+      %PatchesForPackagesResponse{
+        patches: [
+          %PatchesForPackage{package_id: _, patches: _},
+          %PatchesForPackage{package_id: _, patches: _}
+        ]
+      } =
+        conn
+        |> get(
+          "/api/v1/software_updates/packages?package_ids[]=#{Faker.UUID.v4()}&package_ids[]=#{Faker.UUID.v4()}"
```
jamie-suse, line 115: Just curious, is it Phoenix that makes the URL parameters for a list specified like this?

As opposed to something like `"/api/v1/software_updates/packages?package_ids=#{package_1},#{package_2}"`

jamie-suse (approved):

test/trento_web/controllers/v1/suse_manager_controller_test.exs
dottorblaster: I tried something like that, but it didn't work and Phoenix was seeing all the payload as a string; I tried to use `qs` in our JavaScript setup as well and a serialized array was resulting in a string of this form (with square brackets).

Having a look at the specifications it's pretty much framework specific, but on the other hand this form looks like the one that established as a "soft" standard :+1:

PR #2658, opened 2024-05-28
Allow admin disable totp

rtorrero:
# Description

This PR adds the ability to forcefully disable the TOTP feature for a certain user from the user management view. It should *not* allow to enable it as this requires the TOTP enrollment process.

Screenshots:
Captura desde 2024-05-28 17-03-04
Captura desde 2024-05-28 17-03-10
Captura desde 2024-05-28 17-03-13

## How was this tested?
Tests updated

### DIFF
--- a/assets/js/pages/Users/UserForm.jsx
+++ b/assets/js/pages/Users/UserForm.jsx
@@ -245,7 +245,13 @@ function UserForm({
               }}
             />
           </div>
-          <Label className="col-start-1 col-span-1" htmlFor="totp" aria-label="totp">TOTP</Label>
+          <Label
+            className="col-start-1 col-span-1"
+            htmlFor="totp"
+            aria-label="totp"
+          >
+            TOTP
+          </Label>
           <div className="col-start-2 col-span-3">
             <Select
               optionsName="totp"
--- a/assets/js/pages/Users/UserForm.jsx
+++ b/assets/js/pages/Users/UserForm.jsx
@@ -261,7 +261,7 @@ function UserForm({
               ]}
               value={totpState ? 'Enabled' : 'Disabled'}
               onChange={(value) => {
-                setTotpState(value == 'Enabled');
+                setTotpState(value === 'Enabled');
               }}
             />
           </div>
--- a/assets/js/pages/Users/UserForm.jsx
+++ b/assets/js/pages/Users/UserForm.jsx
@@ -33,7 +33,7 @@ function UserForm({
   userAbilities = defaultAbilities,
   createdAt = '',
   updatedAt = '',
-  totpEnabledAt = null,
+  totpEnabledAt = '',
   errors = defaultErrors,
   saving = false,
   saveEnabled = true,
@@ -53,7 +53,7 @@ function UserForm({
   const [confirmPasswordState, setConfirmPassword] = useState('');
   const [confirmPasswordErrorState, setConfirmPasswordError] = useState(null);
   const [statusState, setStatus] = useState(status);
-  const [totpState, setTotpState] = useState(totpEnabledAt !== null);
+  const [totpState, setTotpState] = useState(Boolean(totpEnabledAt));
   const [selectedAbilities, setAbilities] = useState(
     userAbilities.map(({ id }) => id)
   );
@@ -111,7 +111,7 @@ function UserForm({
         password_confirmation: confirmPasswordState,
       }),
       abilities: abilities.filter(({ id }) => selectedAbilities.includes(id)),
-      ...(totpEnabledAt && !totpState && { totp_enabled: false }),
+      ...(totpEnabledAt && !totpState && { totp_disabled: true }),
     };
 
     onSave(user);
@@ -256,7 +256,7 @@ function UserForm({
             <Select
               optionsName="totp"
               options={[
-                { value: 'Enabled', disabled: totpEnabledAt === null },
+                { value: 'Enabled', disabled: !totpEnabledAt },
                 'Disabled',
               ]}
               value={totpState ? 'Enabled' : 'Disabled'}
--- a/assets/js/pages/Users/UserForm.jsx
+++ b/assets/js/pages/Users/UserForm.jsx
@@ -33,6 +33,7 @@ function UserForm({
   userAbilities = defaultAbilities,
   createdAt = '',
   updatedAt = '',
+  totpEnabledAt = null,
   errors = defaultErrors,
   saving = false,
   saveEnabled = true,
@@ -52,6 +53,7 @@ function UserForm({
   const [confirmPasswordState, setConfirmPassword] = useState('');
   const [confirmPasswordErrorState, setConfirmPasswordError] = useState(null);
   const [statusState, setStatus] = useState(status);
+  const [totpEnabled, setTotpEnabled] = useState(totpEnabledAt !== null);
   const [selectedAbilities, setAbilities] = useState(
     userAbilities.map(({ id }) => id)
   );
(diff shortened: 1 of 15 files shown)


assets/js/pages/Users/UserForm.jsx
```diff
@@ -242,6 +245,26 @@ function UserForm({
...
+            className="col-start-1 col-span-1"
+            htmlFor="totp"
+            aria-label="totp"
+          >
+            TOTP
+          </Label>
+          <div className="col-start-2 col-span-3">
+            <Select
+              optionsName="totp"
+              options={['Enabled', 'Disabled']}
```
arbulu89, line 258: Let's put the `Enabled` option disabled.
Like: `options={[{ value: 'Enabled", disabled: true }, 'Disabled']}`.
This way, the user cannot literally choose the `Enabled` button, and it is only selected if it comes from the backend

arbulu89 (changes requested):
Hey @rtorrero 
Leaving aside the backend comment I did, where I would like to discuss what to do...

I think we have some issues in the frontend that we need to fix.

assets/js/pages/Users/UserForm.jsx
```diff
@@ -52,6 +53,7 @@ function UserForm({
   const [confirmPasswordState, setConfirmPassword] = useState('');
   const [confirmPasswordErrorState, setConfirmPasswordError] = useState(null);
   const [statusState, setStatus] = useState(status);
+  const [totpEnabled, setTotpEnabled] = useState(totpEnabledAt !== null);
```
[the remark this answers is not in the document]
rtorrero: Agreed, it's closer to what we want to represent

lib/trento/users.ex
```diff
@@ -93,24 +93,38 @@ defmodule Trento.Users do
...
       |> maybe_set_password_change_requested_at(false)
+      |> handle_totp_disabled()
 
     do_update(user, updated_attrs)
   end
 
   def update_user(%User{locked_at: locked_at} = user, %{enabled: false} = attrs)
       when not is_nil(locked_at) do
-    do_update(user, maybe_set_password_change_requested_at(attrs, false))
+    updated_attrs =
```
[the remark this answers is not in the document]
rtorrero: After giving it a bit thought, I totally agree, it's a transformation that makes much more sense in the controller beacuse it doesn't reflect the schema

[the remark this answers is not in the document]
arbulu89: @rtorrero I had a offline chat with @CDimonaco , and we agreed that by now using the current approach is the best option.
Even though it is not totally ideal, we postpone the change. 
I have created a tech-debt so we can revisit this in the future with better options.

arbulu89 (commented):
@rtorrero 
Looking better.
I'm afraid that I had some second thoughts with the frontend, and after a chat with @CDimonaco we agreed that having the logic in the context by now is more appropriate than having it in the controller, as we already do this

assets/js/pages/Users/UserForm.test.jsx
```diff
@@ -225,6 +230,82 @@ describe('UserForm', () => {
...
+    expect(mockOnSave).toHaveBeenNthCalledWith(1, {
+      fullname,
+      email,
+      enabled: true,
+      abilities: [],
+      totp_enabled: false,
+    });
+  });
+
+  it('should not allow to re-enable TOTP', async () => {
```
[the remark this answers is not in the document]
rtorrero: It was simply poorly explained in the title, updating the description should do it :+1:

lib/trento/users.ex
```diff
@@ -93,24 +93,36 @@ defmodule Trento.Users do
...
       |> maybe_set_password_change_requested_at(false)
+      |> maybe_disable_totp()
 
     do_update(user, updated_attrs)
   end
 
   def update_user(%User{locked_at: locked_at} = user, %{enabled: false} = attrs)
       when not is_nil(locked_at) do
-    do_update(user, maybe_set_password_change_requested_at(attrs, false))
+    updated_attrs =
```
arbulu89, line 103: I would add some test for this.
One basic test at least

assets/js/pages/Users/UserForm.test.jsx
```diff
@@ -225,6 +230,82 @@ describe('UserForm', () => {
...
+          totpEnabledAt={totpEnabledAt}
+          editing
+        />
+      );
+    });
+
+    await user.click(screen.getByText('Disabled'));
+
+    // This should not change the selected option as it should be disabled
+    await user.click(screen.getAllByText('Enabled')[1]);
```
arbulu89, line 305: I think `expect(screen.getAllByText('Enabled')[1]).toBeDisabled()` is more appropriate

arbulu89 (approved):
@rtorrero 
Looks good to me.
A comment about a FE test and I would add some test to the users.ex.
Green light besides that

lib/trento_web/openapi/v1/schema/user.ex
```diff
@@ -190,7 +190,13 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
           description: "User new password, should be the same as password field",
           nullable: false
         },
-        abilities: AbilityCollection
+        abilities: AbilityCollection,
+        totp_enabled: %Schema{
```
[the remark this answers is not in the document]
CDimonaco: Totally agree with this, it's better to change

assets/js/pages/Users/UserForm.jsx
```diff
@@ -33,6 +33,7 @@ function UserForm({
   userAbilities = defaultAbilities,
   createdAt = '',
   updatedAt = '',
+  totpEnabledAt = null,
```
CDimonaco, line 36: Could we have this as an empty string like the other ones?

assets/js/pages/Users/UserForm.jsx
```diff
@@ -52,6 +53,7 @@ function UserForm({
   const [confirmPasswordState, setConfirmPassword] = useState('');
   const [confirmPasswordErrorState, setConfirmPasswordError] = useState(null);
   const [statusState, setStatus] = useState(status);
+  const [totpState, setTotpState] = useState(totpEnabledAt !== null);
```
CDimonaco, line 56: Having explicitely match between a null could not be a good idea, if you put the totpEnabledAt in the props at `''` empty string by default, you could just use the boolean constructor here in useState default value `Boolean(totpEnabledAt)` this will return true if the prop is not empty

assets/js/pages/Users/UserForm.jsx
```diff
@@ -242,6 +245,26 @@ function UserForm({
...
+            htmlFor="totp"
+            aria-label="totp"
+          >
+            TOTP
+          </Label>
+          <div className="col-start-2 col-span-3">
+            <Select
+              optionsName="totp"
+              options={[
+                { value: 'Enabled', disabled: totpEnabledAt === null },
```
CDimonaco, line 259: Also here the boolean constructor

CDimonaco (commented):
Good job!

I left some cosmetic comments

lib/trento_web/openapi/v1/schema/user.ex
```diff
@@ -190,7 +190,13 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
...
+          type: :boolean,
+          description:
+            "TOTP feature enabled for the user. The only accepted value here is 'false'",
```
EMaksy, line 197: In case you make the suggested changes, don't forget about the description :)

CDimonaco (approved):

PR #2660, opened 2024-05-29
Upgrade Node.js to version 20.14.0 LTS

janvhs:
This commit upgrade the Node.js version to 20.14.0.

I am running this for several weeks on my MacBook Air M2 and it works without any problems. This will allow us to upgrade Storybook and ESLint, once all our ESLint plugins are compatible with ESLint 9.

## How was this tested?

Building booth the Dockerfile in the root, and in packaging with Podman and running frontend tests. 
I did not try asdf, since I don't have that installed. However, I am sure it will accept the version.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ on:
 
 env:
   MIX_ENV: test
-  NODE_VERSION: "16"
+  NODE_VERSION: "20"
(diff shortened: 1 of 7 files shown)


dottorblaster (commented):
When we lift the merge freeze just rebase this on main and we can merge :+1:

dottorblaster (approved):

PR #2663, opened 2024-06-03
Add a docs section to pr template

EMaksy:
# Description

This pr updates the pr template, by adding a docs section.
This change should remind every contributor about updating the documentation. Some parts of the documentation could be used in Trento's official documentation, so it is important that especially the [Manual installation guide](https://github.com/trento-project/docs/blob/main/guides/manual-installation.md) and [Ansible guide](https://github.com/trento-project/ansible/blob/main/README.md) is up-to-date.

Sometimes it would make sense to open a separate pr for a docs update, in this case it would not harm to add a reference in the pr description. 
If docs do not require an update, a simple **no docs update required** would be more than enough, similar like what we write in the "How was this tested" section

## How was this tested?
No tests needed

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,26 +3,24 @@
 Please include a summary of the changes and the related issue, if it exists. 
 Also, include relevant motivation and context for this PR. List any dependencies that are required for this change.
 
-- Fixes # (issue)
+Fixes # (issue)
 
 ## Did you add the right label?
 
 Remember to add the right labels to this PR.
+- [ ] **DONE**
 
 ## How was this tested?
 
-Describe the types of tests that have been added for this functionality
-
+Describe the tests that have been added/changed for this new behavior.
+- [ ] **DONE**
 ## Did you update the documentation?
 
-Remember to ask yourself if your PR introduced changes to the following documentation:
+Remember to ask yourself if your PR requires changes to the following documentation:
 
-- [Manual installation guide](https://github.com/trento-project/docs/blob/main/guides/manual-installation.md)?
-- [Trento Ansible guide](https://github.com/trento-project/ansible/blob/main/README.md)?
-- [Alerting](https://github.com/trento-project/web/blob/main/guides/alerting/alerting.md)?
-- [Authentication](https://github.com/trento-project/web/blob/main/guides/authentication/jwt_specification.md)?
-- [Environment variables](https://github.com/trento-project/web/blob/main/guides/development/environment_variables.md)?
-- [Hack on Trento](https://github.com/trento-project/web/blob/main/guides/development/hack_on_the_trento.md)?
-- [Monitoring](https://github.com/trento-project/web/blob/main/guides/monitoring/monitoring.md)?
+- [Manual installation guide](https://github.com/trento-project/docs/blob/main/guides/manual-installation.md)
+- [Trento Ansible guide](https://github.com/trento-project/ansible/blob/main/README.md)
+- [Trento Web guides](https://github.com/trento-project/web/tree/main/guides)
 
 Add a documentation PR or write that no changes are required for the documentation.
+- [ ] **DONE**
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,15 +1,28 @@
...
 ## How was this tested?
 
-Describe what kind of tests for this functionality have been added.
+Describe the types of tests that have been added for this functionality
+
+## Did you update the documentation?
+
+Remember to ask yourself if your PR introduced changes to the following documentation:
+
+- [Manual installation guide](https://github.com/trento-project/docs/blob/main/guides/manual-installation.md)?
```
EMaksy, line 20: Having this list of docs is a dual edged sword, it is nice to have a list where on every pr we can shortly check out if some changes are required, but on the other hand this list needs to be maintained from time to time.

Open for suggestions if we should keep this list or only the text:

nelsonkopliku (approved):
Hey @EMaksy thanks for this!

I am fine eitherway with the list of links ore just the reminder.
If you wanna keep the list of links maybe we could compact web related links to a generic https://github.com/trento-project/web/blob/main/guides/

.github/pull_request_template.md
rtorrero: I think having the links is fine, in the end, as we will use it frequently it should be obvious fast enough when it gets outdated.

rtorrero (approved):
Thanks @EMaksy! LGTM

stefanotorresi (commented):
The problem with adding links is that our documentation is a bit scattered at the moment, so there is no single place one should look into. Certainly it's not just the installation docs, so if you decide to keep it you need to make it more generic.

.github/pull_request_template.md
```diff
@@ -1,15 +1,28 @@
...
+- Fixes # (issue)
 
 ## Did you add the right label?
 
 Remember to add the right labels to this PR.
 
 ## How was this tested?
 
-Describe what kind of tests for this functionality have been added.
+Describe the types of tests that have been added for this functionality
```
janvhs, line 14:
```suggestion
Describe the types of tests that have been added for this functionality.
```
Just keeping things consistent here

.github/pull_request_template.md
```diff
@@ -1,15 +1,28 @@
...
+
+## Did you update the documentation?
+
+Remember to ask yourself if your PR introduced changes to the following documentation:
```
janvhs, line 18:
```suggestion
Remember to ask yourself if your PR effects or requires changes to the following documentation:
```

Because you're going to _introduce changes_ to the documents :D

.github/pull_request_template.md
janvhs: I agree with @rtorrero. I am likely to forget checking some documents, if I do not have a list available

janvhs (approved):
Just have some minor changes in writing, but that's 100% optional and it LGTM :D

.github/pull_request_template.md
jamie-suse, line 14: to me, this is more succinct and clearer:

```suggestion
Describe the tests that have been added/changed for this new behavior.
```

What do you think?

EMaksy:
Thanks for the feedback!

I refactored the pr a bit and added a `Done` section, as an inspiration from the [uyuni project](https://github.com/uyuni-project/uyuni/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

What I liked about it, it adds a clickable task list with progress see  [Github: About task lists](https://docs.github.com/en/issues/managing-your-tasks-with-tasklists/creating-a-tasklist#creating-tasklists-with-markdown).
It is fully optional, and everybody can delete the template, but it provides a nice visual touch.

gagandeepb (approved):
Maybe a consideration for later, but would (optionally, perhaps) putting/linking ticket numbers in PRs be interesting?

PR #2666, opened 2024-06-04
Audit logging via OTEL

gagandeepb:
# Description

Audit logging PoC WIP


lib/trento/router.ex
```diff
@@ -106,4 +106,6 @@ defmodule Trento.Router do
            ],
            to: Databases.Database,
            lifespan: Databases.Lifespan
+
+  middleware OpentelemetryCommanded.Middleware
```
gagandeepb, line 110: Example of what can be seen in the traces:
image

lib/trento_web/plugs/otel.ex
```diff
@@ -0,0 +1,16 @@
...
+  require OpenTelemetry.Tracer, as: Tracer
+
+  def init(default) do
+    default
+  end
+
+  def call(conn, _default) do
+    current_user = conn.assigns.current_user
+
+    Tracer.add_event("Otel", [{"user", current_user}])
```
gagandeepb, line 12: How this would appear in the traces (more customization/different manners of propagation are also additionally possible):

image

gagandeepb:
There is a search API that can help customize the UI/presentation layer if necessary, among other things.
image

PR #2668, opened 2024-06-04
Upgradable packages view

dottorblaster:
# Description
Added page for the upgradable packages of a host and mounted it. This PR also adds all the sagas to fetch related patches that are covered by a package upgrade.

## How was this tested?
Unit tests added

### DIFF
--- a/assets/js/pages/UpgradablePackagesPage/UpgradablePackagesPage.test.jsx
+++ b/assets/js/pages/UpgradablePackagesPage/UpgradablePackagesPage.test.jsx
@@ -1,0 +1,45 @@
+import React from 'react';
+import { screen } from '@testing-library/react';
+
+import { faker } from '@faker-js/faker';
+
+import {
+  renderWithRouterMatch,
+  withState,
+  defaultInitialState,
+} from '@lib/test-utils';
+import { upgradablePackageFactory } from '@lib/test-utils/factories/upgradablePackage';
+import { patchForPackageFactory } from '@lib/test-utils/factories/relevantPatches';
+
+import UpgradablePackagesPage from './UpgradablePackagesPage';
+
+describe('UpgradablePackagesPage', () => {
+  it('should render correctly', () => {
+    const hostID = faker.string.uuid();
+    const patch = patchForPackageFactory.build();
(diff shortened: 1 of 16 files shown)


assets/js/pages/UpgradablePackagesPage/UpgradablePackagesPage.test.jsx
```diff
@@ -0,0 +1,44 @@
...
+        },
+      },
+    });
+
+    renderWithRouterMatch(StatefulPage, {
+      path: 'hosts/:hostID/packages',
+      route: `/hosts/${hostID}/packages`,
+    });
+
+    expect(screen.getAllByText(name, { exact: false })).toBeTruthy();
```
jamie-suse, line 42: Is it worth validating that the upgradable packages are displayed?

assets/js/state/softwareUpdates.js
```diff
@@ -44,18 +45,51 @@ export const softwareUpdatesSlice = createSlice({
...
+          const packagePatches = get(packageInfo, 'patches', []);
+
+          return {
+            ...currentPackage,
+            patches: packagePatches,
+          };
+        }
+      );
+
+      return state;
```
jamie-suse, line 73: Do we actually need to `return state` in this function?

assets/js/pages/UpgradablePackagesPage/UpgradablePackagesPage.test.jsx
dottorblaster: It is to me, because we are making sure that the wiring between the state and the page is done correctly. Maybe we could do some assertions for the patches as well.

assets/js/state/softwareUpdates.js
dottorblaster: ESLint yelled at me if I didn't :smile: Maybe I could architect this piece of code better, I'm really open to suggestions

jamie-suse: Oh hmm that's weird as pretty much all the other reducer functions in our codebase don't return anything :thinking: is that a newly added rule or something?

jamie-suse (approved):

PR #2669, opened 2024-06-04
Add sorting to `Table` component

jamie-suse:
# Description

This PR adds the ability to sort the rows in a `Table` component by clicking on the header of a table column.

Changes to the `Table` component:

- The `sortBy` prop is the predicate function used to sort `data`.

- Columns in `config.columns` can be marked with `sortable: true`.

- Additionally, a `handleClick()` function needs to be provided in `config.columns` for handling when the column header is clicked.

- Finally, in `config.columns`, `sortDirection` needs to be provided to indicate to the user which column is being sorted, and in which direction. It can be `'asc' | 'desc' | undefined/null`. The `undefined`/`null` case is for the columns not currently being sorted.

Note that all these new config options have default falsy values so that when no sorting is required, none of these values need to be provided.

Some default functions to produce sorting predicates have also been provided in `assets/js/common/Table/sorting.js`

## How was this tested?

Added unit test to verify that `Table` sorts data. Add Storybook _story_ to demonstrate feature.

### DIFF
--- a/assets/js/common/Table/SortingIcon.jsx
+++ b/assets/js/common/Table/SortingIcon.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { EOS_ARROW_UPWARD, EOS_ARROW_DOWNWARD } from 'eos-icons-react';
 
+const styles = 'inline-table relative top-1';
+
 export default function SortingIcon({
   sortable = false,
   sortDirection = undefined,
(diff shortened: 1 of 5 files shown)


dottorblaster (changes requested):
This is great :fire: 

I just have one more request but we can include that in a subsequent PR: could we make adopting of sorting easier by creating a function that just gives you a sorting function out of the box? Something like the one you made for the story, but actually available to developers to import.

dottorblaster (approved):
LGTM :fire:

PR #2670, opened 2024-06-05
Policy support functions

CDimonaco:
# Description

Policy support functions to build bodyguard policies

## How was this tested?

Automated tests already present

### DIFF
--- a/lib/trento/support/policy_helper.ex
+++ b/lib/trento/support/policy_helper.ex
@@ -1,0 +1,12 @@
+defmodule Trento.Support.PolicyHelper do
+  @moduledoc """
+  Helper functions for bodyguard policies
+  """
+  alias Trento.Users.User
+
+  def user_has_ability?(%User{abilities: abilities}, %{name: name, resource: resource}),
+    do: Enum.any?(abilities, &(&1.name == name and &1.resource == resource))
+
+  def has_global_ability?(%User{} = user),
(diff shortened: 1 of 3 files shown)


lib/trento/support/policy_helper.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Trento.Support.PolicyHelper do
+  @moduledoc """
+
```
arbulu89, line 3: I would remove this empty new line

lib/trento/support/policy_helper.ex
```diff
@@ -0,0 +1,13 @@
...
+  Helper functions for bodyguard policies
+  """
+  alias Trento.Users.User
+
+  def user_has_ability?(%User{abilities: abilities}, %{name: name, resource: resource}),
```
arbulu89, line 8: Could we have some basic UT specifically for this?

arbulu89 (approved):
Looks good!
Could we have some test? :stuck_out_tongue:

lib/trento/support/policy_helper.ex
CDimonaco: done!

lib/trento/support/policy_helper.ex
CDimonaco: done :P it was already tested but it's good to have dedicated tests

arbulu89: thank you!

PR #2671, opened 2024-06-05
Force ssl in production

CDimonaco:
# Description

Force ssl in production


nelsonkopliku (approved):
cool, so no force_ssl anymore :eyes:

rtorrero (approved):
LGTM, tested manually using the helm chart both for http and https and seems to work correctly

PR #2672, opened 2024-06-05
Disabled guard

arbulu89:
# Description

Add the `DisabledGuard` component. As the docstring of the file says, it simply adds the `disabled` prop to the guarded element if access for the current user is not authorized.
It includes a message with a tooltip if the user is not authorized (the usage is optional), and it works for elements wrapped by Tooltips as well.

I implemented it as a pure element, so we need to pass the current user abilities as prop.
This will create the need to get them in may places, but it will make the code more explicit, and easier to test all the test cases without using the redux store.

Find some examples in storybook.

PD: I have moved the `ForbiddenGuard` to `common` folder as well.


assets/js/common/DisabledGuard/DisabledGuard.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+  const disabledElement = React.cloneElement(guardedElement, {
+    disabled: true,
+  });
+
+  return (
+    <Tooltip
+      isEnabled={withTooltip}
+      content={tooltipMessage}
+      place="bottom"
```
EMaksy, line 42: Can we land in a situation where we want to change the position of the Tooltip ?
Maybe add prop to `DisabledGuard` ? 
On the other hand it is an easy adjustment if we need this in the future, so no hard opinion on that

EMaksy (approved):
Nice!
Just left a question, besides this LGTM!

CDimonaco (approved):

assets/js/common/DisabledGuard/DisabledGuard.jsx
arbulu89: I will just put it as prop and put `bottom` as default

PR #2673, opened 2024-06-05
Some small improvements in the user forms

arbulu89:
# Description

Basically 2 improvements:
- Remove the TOTP configuration select in the user creation form. It cannot be changed, so better remove it
- Disable the TOTP enrollment option in the user profile view for the admin user. Otherwise, the clicking just fails with a forbidden request error.

This 2nd has forced me to add the `disabled` state in the switch. @jagabomb you can change the style change in storybook.

### DIFF
--- a/assets/js/common/Switch/Switch.jsx
+++ b/assets/js/common/Switch/Switch.jsx
@@ -4,15 +4,21 @@ import classNames from 'classnames';
 
 import { Switch as HeadlessSwitch } from '@headlessui/react';
 
-function Switch({ onChange, selected }) {
+function Switch({ onChange, selected, disabled }) {
   return (
     <HeadlessSwitch.Group as="div" className="flex items-center">
       <HeadlessSwitch
         checked={selected}
         className={classNames(
-          'tn-check-switch relative inline-flex flex-shrink-0 h-6 w-11 border-2 border-transparent rounded-full cursor-pointer focus:outline-none transition-colors ease-in-out duration-200',
-          { 'bg-jungle-green-500': selected, 'bg-gray-200': !selected }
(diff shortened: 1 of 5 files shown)


EMaksy (approved):
LGTM

jagabomb:
@arbulu89 So far so good. I am wondering if we should change the component for the Authenticator/TOTP section in the Storybook (for the Admin, With Abilities, etc) to also use the Switch component. Just so that there is some consistency between the Trento Web and the Storybook. What do you think?

arbulu89:
> @arbulu89 So far so good. I am wondering if we should change the component for the Authenticator/TOTP section in the Storybook (for the Admin, With Abilities, etc) to also use the Switch component. Just so that there is some consistency between the Trento Web and the Storybook. What do you think?

I'm not really following you. Storybook shows the actual code we have, there is not distinction.
If what you say is that the Users view form, and the Profile view form (different views) should use the same element, well, I don't really see it neccesary, both are different views, used by different users.
The `Select` usage looks could fine in my opinion

jagabomb:
> I'm not really following you. Storybook shows the actual code we have, there is not distinction. If what you say is that the Users view form, and the Profile view form (different views) should use the same element, well, I don't really see it neccesary, both are different views, used by different users. The `Select` usage looks could fine in my opinion

@arbulu89 Ah, please ignore my comment. 😅  I realize that I getting confused between the **UserForm** and **ProfileForm** in Storybook. This ProfileForm component is correct since the **Select** component is used to enable the TOTP. Then I see the **Toggle** component is on the UserForm view.

jagabomb (approved):
LGTM

PR #2675, opened 2024-06-06
Fix SMTP usage in runtime configuration

CDimonaco:
# Description

Fix SMTP usage in runtime

## How was this tested?

Manual testing

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -128,7 +128,7 @@ if config_env() in [:prod, :demo] do
     username: System.get_env("SMTP_USER") || "",
     password: System.get_env("SMTP_PASSWORD") || "",
     auth: :always,
-    ssl: :if_available,
+    ssl: false,
     tls: :if_available,
     tls_options: [
       versions: [:"tlsv1.2", :"tlsv1.3"],
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -118,6 +118,9 @@ if config_env() in [:prod, :demo] do
     sender: System.get_env("ALERT_SENDER", "alerts@trento-project.io"),
     recipient: System.get_env("ALERT_RECIPIENT", "admin@trento-project.io")
 
+  :ok = :public_key.cacerts_load()
+  [_ | _] = cacerts = :public_key.cacerts_get()
+
   config :trento, Trento.Mailer,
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):
Great @CDimonaco 

As a matter of fact, we tried to use the [tls_certificate_check](https://hexdocs.pm/tls_certificate_check/readme.html) but you cannot use dependency function in the config files

nelsonkopliku (approved):
Good catch!

I was also debugging myself around that and noticed that what we currently provide as `ssl: :if_available` should actually be a boolean as mentioned here https://github.com/swoosh/swoosh/blob/v1.14.3/lib/swoosh/adapters/smtp.ex#L71

not sure whether we want to do something in that regard.

dottorblaster (approved):

EMaksy (approved):

CDimonaco:
SSL Disabled in favour of TLS encryption. Thanks @nelsonkopliku  for spotting it

gagandeepb:
One question to be addressed/explored, possibly at some later point, is that the Swoosh config provided is a deployment specific concern: IIUC there isn't a one single config that fits all cases (i.e. it depends on the config/settings of an external service, i.e. the SMTP server). Some servers may support STARTTLS and some may support TLS/SSL. I wonder if the observed bug, and consequently the fix here are too dependent on the SMTP server config? Note that Mailtrap, using which the bug was initially reported uses STARTTLS by default ([which requires setting ssl to false or omitting altogether, among other things](https://hexdocs.pm/swoosh/Swoosh.Adapters.SMTP.html#module-note)). I don't know what the server, against which this fix was validated, uses. I would be curious to hear your findings regarding this and/or pair on this (i.e. testing against SMTP server configs/ different OTP versions, etc.). Let me know if there is something that I have misunderstood about this.

arbulu89:
@gagandeepb That's a fair concern.
We tested with `Mailtrap` and google SMTP server (`smtp.gmail.com`, port 587), I guess both `tls`, and that's why putting `ssl: false` helps in both cases.
Either way, we have a couple of options to improve this:
- Require usage of TLS based SMTP server to the user as SSL is deprecated
- Make the configuration available through env variables

PR #2677, opened 2024-06-07
Fix SAP systems policy receiving an empty list of instances

dottorblaster:
# Description
Fix SAP systems policy receiving an empty list of instances and marking that as not valid, thus making impossible in some cases to deregister an existing system.

## How was this tested?
Existing unit tests passing, added a unit test to make sure this doesn't incur again

### DIFF
--- a/lib/trento/discovery/payloads/sap_system_discovery_payload.ex
+++ b/lib/trento/discovery/payloads/sap_system_discovery_payload.ex
@@ -40,7 +40,7 @@ defmodule Trento.Discovery.Payloads.SapSystemDiscoveryPayload do
       required: true
     )
     |> cast_embed(:Databases)
-    |> cast_embed(:Instances, required: true)
+    |> cast_embed(:Instances)
     |> validate_required_fields(@required_fields)
     |> validate_inclusion(:Type, @system_types)
   end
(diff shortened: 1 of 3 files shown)


nelsonkopliku (commented):
Thanks for this @dottorblaster.

The change was introduced here https://github.com/trento-project/web/pull/2214/files, along with other `required: true`.

I am not sure if any other of those would cause issues.

nelsonkopliku:
@dottorblaster, if you don't mind, may I just ask to fix this typo making the CI red, please? https://github.com/trento-project/web/actions/runs/9413405002/job/25930084979?pr=2677

dottorblaster:
@nelsonkopliku YES

nelsonkopliku (approved):
thanks @dottorblaster, LGTM.

As long as @abravosuse's test is successful and @CDimonaco is fine with this, let's ship it.

arbulu89:
@dottorblaster Thank you.
This fix means that our e2e tests are "half" complete, because we were testing this feature, but in a different case.
@abravosuse had this issue because when he "deregisters" a SAP instance, he doesn't really clean the env, he just stops the instance, so the `profile` file is still there.

Our e2e tests check that the cleanup button is there when there is a complete removal of the SAP instance in the agent host, which causes a different payload.

I don't know if we should update the e2e tests, as you added this UT test here, but we have this gap there...

PR #2678, opened 2024-06-07
Fix misspelling of resource in a test

dottorblaster:
I just want the CI to stop complaining about this


gagandeepb (approved):
Thanks for this fix!

PR #2679, opened 2024-06-07
Add sorting in `Upgradable Packages` & `Available Patches` lists

jamie-suse:
# Description

Adds sorting to `UpgradablePackagesList` and `PatchList` tables.

`UpgradablePackagesList` can be sorted by `Latest Package`.

`PatchList` can be sorted by `Advisory`, `Synopsis` or `Updated`.

## How was this tested?

Existing `Table`, `UpgradablePackagesList` and `PatchList` unit tests.

### DIFF
--- a/assets/js/common/PatchList/PatchList.jsx
+++ b/assets/js/common/PatchList/PatchList.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import {
   EOS_SHIELD_OUTLINED,
   EOS_CRITICAL_BUG_OUTLINED,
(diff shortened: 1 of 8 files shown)


assets/js/common/PatchList/PatchList.jsx
```diff
@@ -8,6 +8,10 @@ import { format as formatDate } from 'date-fns';
 import classNames from 'classnames';
 import { noop } from 'lodash';
 
+import {
+  createStringSortingPredicate,
+  createDateSortingPredicate,
+} from '@common/Table/sorting';
```
dottorblaster, line 14: Could you re-export these from `index.js` in `@common/Table` so we can do `import Table, {create...} from '@common/Table'`?

dottorblaster (changes requested):
Just one comment then we can merge :+1:

dottorblaster (approved):
LGTM

PR #2681, opened 2024-06-07
Default empty string for smtp server

CDimonaco:
# Description

Fix smtp server value, default to empty string


EMaksy (approved):
Thanks!

nelsonkopliku (approved):

PR #2684, opened 2024-06-10
Add search to table component

janvhs:
## Description

This commit adds a search capability to the table component. The search is case-insensitive and respects normalises Unicode character differences. However, it is not fuzzy or can accommodate spelling mistakes.

## Tests

Furthermore, the change is supported via a story and several test cases, that describe the behaviour.


jamie-suse:
I'd add a bit more semantic meaning to the variable names in `search.js`, will add some suggestions.

assets/js/common/Table/search.js
```diff
@@ -0,0 +1,23 @@
+const canonicalise = (val) => val.normalize().trim().toLowerCase();
```
jamie-suse, line 1: It's a bit difficult as `normalize()` is already taken :stuck_out_tongue: but maybe something like:

```suggestion
const regularize = (str) => str.normalize().trim().toLowerCase();
```

assets/js/common/Table/search.js
```diff
@@ -0,0 +1,23 @@
...
+
+const stringContains = (subject, predicate) =>
```
jamie-suse, line 3: This is not really a predicate

```suggestion
const stringContains = (string, substring) =>
  regularize(string).includes(regularize(substring));
```

assets/js/common/Table/search.js
```diff
@@ -0,0 +1,23 @@
...
+  canonicalise(subject).includes(canonicalise(predicate));
+
+const arrayContains = (subject, predicate) =>
```
jamie-suse, line 6: Again, this is not a predicate. Also you don't need the reduce here, I think this is a bit more succinct, what do you think?

```suggestion
const arrayContains = (strings, searchString) =>
  strings
    .map((s) => stringContains(s, searchString))
    .includes(true);
```

assets/js/common/Table/search.js
janvhs: I agree. I like the word canonical, but don't mind changing it, since it's essentially the same thing :duck:

assets/js/common/Table/search.js
```diff
@@ -0,0 +1,23 @@
...
+    : stringContains(subject, predicate);
+
+export const searchByKey = (subject, predicate, ...key) =>
+  key
+    .map((it) =>
+      Object.hasOwn(subject, it)
+        ? contains(subject[it] ?? '', predicate)
+        : false
+    )
+    .reduce((accumulator, currentValue) => accumulator || currentValue, false);
```
jamie-suse, line 23:
```suggestion
    .includes(true);
```

assets/js/common/Table/Table.jsx
```diff
@@ -134,7 +136,15 @@ function Table({
     })
     .reduce((d, filterFunction) => d.filter(filterFunction), data);
 
-  const sortedData = sortBy ? [...filteredData].sort(sortBy) : filteredData;
+  const searchKeys = columns
+    .filter(({ searchable = false }) => searchable === true)
```
jamie-suse, line 140:
```suggestion
    .filter(({ searchable = false }) => searchable)
```

assets/js/common/Table/Table.jsx
```diff
@@ -134,7 +136,15 @@ function Table({
...
+    .map(({ key }) => key);
+
+  const searchedData = searchBy
+    ? [...filteredData].filter((it) => searchByKey(it, searchBy, ...searchKeys))
```
jamie-suse, line 144: Curious why we spread on `searchKeys` here?

assets/js/common/Table/search.js
```diff
@@ -0,0 +1,23 @@
...
+  subject
+    .map((it) => stringContains(it, predicate))
+    .reduce((accumulator, currentValue) => accumulator || currentValue, false);
+
+const contains = (subject, predicate) =>
+  Array.isArray(subject)
+    ? arrayContains(subject, predicate)
+    : stringContains(subject, predicate);
+
+export const searchByKey = (subject, predicate, ...key) =>
```
jamie-suse, line 16: What about this? (I changed the names to make them easier to understand, but not super set on them)

Note: Also removed the use of the spread operator.

```suggestion
export const search = (row, searchTerm, columnKeys) =>
  columnKeys
    .map((k) => row[k] ? contains(row[k], searchTerm) : false)
    .includes(true);
```

assets/js/common/Table/Table.jsx
janvhs: Just because I started from the case of only having a single key

assets/js/common/Table/search.js
```diff
@@ -0,0 +1,23 @@
...
+
+const stringContains = (subject, predicate) =>
+  canonicalise(subject).includes(canonicalise(predicate));
+
+const arrayContains = (subject, predicate) =>
+  subject
+    .map((it) => stringContains(it, predicate))
+    .reduce((accumulator, currentValue) => accumulator || currentValue, false);
+
+const contains = (subject, predicate) =>
```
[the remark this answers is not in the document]
janvhs: I think you could cast everything to a string array and search over that. But as long as the API does not change, it can always be adjusted, when needed

assets/js/common/Table/search.js
janvhs: Wow, that's actually such a performance optimisation :D

dottorblaster (changes requested):
Good job! Besides @jamie-suse's comments I'd like if possible to see if we could have something using the filtering on the raw data instead of embedding this into the table component. Could you do that?

assets/js/common/Table/search.js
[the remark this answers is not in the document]
jamie-suse: If we move this search feature outside of `Table`, then I think this problem mostly goes away. Because you will know the shape of the specific data you are dealing with, you no longer have to account for every type

janvhs:
@dottorblaster Yeah, I agree with handling the filtering outside the table component, since the Table feels really really bloated anyway

janvhs:
@dottorblaster @jamie-suse This is basically done. I just need to port the search tests from the table to the individual pages, but if you want to look at this, you could :D

dottorblaster (approved):
Very good job, I think you also have to get rid of `search.js` inside `Table`. Keep going!

assets/js/lib/test-utils/index.jsx
```diff
@@ -86,3 +86,8 @@ export async function recordSaga(saga, initialAction, state = {}) {
 
   return dispatched;
 }
+
+export const inspect = (val) => {
```
jamie-suse, line 90: is this used anywhere?

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.test.jsx
```diff
@@ -104,14 +104,33 @@ describe('HostRelevantPatchesPage', () => {
 
       render(<HostRelevantPatchesPage patches={patches} />);
 
-      const advisorySelect = screen.getByRole('button', { name: 'all' });
```
jamie-suse, line 107: Why would you rename these?

janvhs: By mistake

assets/js/lib/filter/index.js
```diff
@@ -0,0 +1,4 @@
+const regularizeString = (str) => str.normalize().trim().toLowerCase();
+
+export const foundStringNaive = (str = '', substring = '') =>
```
jamie-suse, line 3: I don't find `foundStringNaive()` a very indicative name, maybe something like `containsSubstring()`?

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.jsx
```diff
@@ -32,12 +26,15 @@ function HostRelevantPatches({ hostName, onNavigate, patches }) {
   const [displayedPatches, setDisplayedPatches] = useState(patches);
 
   useEffect(() => {
-    setDisplayedPatches(
-      filterPatchesBySynopsis(
-        filterPatchesByAdvisoryType(patches, displayedAdvisories),
-        search
-      )
+    const filteredByAdvisoryKind = filterPatchesByAdvisoryType(
```
jamie-suse, line 29:
```suggestion
    const filteredByAdvisoryType = filterPatchesByAdvisoryType(
```
?

assets/js/pages/UpgradablePackagesPage/UpgradablePackagesPage.jsx
```diff
@@ -11,9 +11,8 @@ import {
   fetchUpgradablePackagesPatches,
 } from '@state/softwareUpdates';
 
-import PageHeader from '@common/PageHeader';
 import BackButton from '@common/BackButton';
-import UpgradablePackagesList from '@common/UpgradablePackagesList';
+import UpgradablePackages from './UpgradablePackages';
```
jamie-suse, line 15: It seems like this was a missed rename?

assets/js/pages/UpgradablePackagesPage/index.js
```diff
@@ -1,3 +1,3 @@
 import UpgradablePackagesPage from './UpgradablePackagesPage';
 
-export default UpgradablePackagesPage;
+export default UpgradablePackagesPage
```
jamie-suse, line 3: I think this should fail the `lint` step?

assets/js/lib/filter/index.js
janvhs: You're right. That name sounds so much better!

assets/js/pages/UpgradablePackagesPage/index.js
dottorblaster: grrr reactions for out prettier check, why did it green light this? :0

janvhs: Hmm eslint doesn't show this as error. Maybe, because it's formatting?

assets/js/pages/UpgradablePackagesPage/UpgradablePackagesPage.jsx
janvhs: No, it is a separate component. I split the navigation and redux into a seperate thing

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.jsx
janvhs: Just me thinking "wait this isn't a string or number... this is a kind". Changed it back... thanks for catching it :smile:

janvhs:
Thank you for the thorough and really, really helpful review @jamie-suse . I really appreciate it! :D

assets/js/pages/UpgradablePackagesPage/UpgradablePackages.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+import UpgradablePackagesList from '@common/UpgradablePackagesList';
+import PageHeader from '@common/PageHeader';
+import Input from '@common/Input';
+import { containsSubstring } from '@lib/filter';
+
+export default function UpgradablePackages({ hostName, upgradablePackages }) {
+  const [search, setSearch] = useState('');
+
+  const displayedPackages = upgradablePackages.filter(
```
jamie-suse, line 12: nice :sunglasses:

PR #2689, opened 2024-06-10
SUSE Manager views navigation

dottorblaster:
# Description
Enabling navigation from the available software updates widget to detail pages for upgradable packages and relevant patches

## How was this tested?
Jest :+1:

### DIFF
--- a/assets/js/common/AvailableSoftwareUpdates/Indicator.jsx
+++ b/assets/js/common/AvailableSoftwareUpdates/Indicator.jsx
@@ -13,8 +13,8 @@ function Indicator({
   tooltip,
   icon,
   loading,
-  navigate,
   children,
+  onNavigate,
 }) {
   const unknown = children === undefined;
 
@@ -39,10 +39,10 @@ function Indicator({
           'flex flex-row items-center border border-gray-200 p-2 rounded-md grow',
           { 'cursor-pointer': !unknown }
         )}
-        onClick={navigate}
+        onClick={onNavigate}
         onKeyDown={({ code }) => {
           if (code === 'Enter') {
-            navigate();
+            onNavigate();
           }
         }}
       >
--- a/assets/js/common/AvailableSoftwareUpdates/Indicator.jsx
+++ b/assets/js/common/AvailableSoftwareUpdates/Indicator.jsx
@@ -1,13 +1,21 @@
 import React from 'react';
 import classNames from 'classnames';
 import {
-  // EOS_KEYBOARD_ARROW_RIGHT_FILLED,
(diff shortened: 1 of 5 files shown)


assets/js/common/AvailableSoftwareUpdates/Indicator.jsx
```diff
@@ -1,13 +1,21 @@
...
 import Tooltip from '@common/Tooltip';
 
-function Indicator({ title, critical, tooltip, icon, loading, children }) {
+function Indicator({
+  title,
+  critical,
+  tooltip,
+  icon,
+  loading,
+  navigate,
```
janvhs, line 16: Don't have strong opinions about this, but how about using `onClick` instead of `navigate`?

assets/js/common/AvailableSoftwareUpdates/Indicator.jsx
```diff
@@ -24,7 +32,20 @@ function Indicator({ title, critical, tooltip, icon, loading, children }) {
...
+          'flex flex-row items-center border border-gray-200 p-2 rounded-md grow',
+          { 'cursor-pointer': !unknown }
+        )}
+        onClick={navigate}
+        onKeyDown={({ code }) => {
+          if (code === 'Enter') {
+            navigate();
+          }
+        }}
+      >
```
janvhs, line 48: Great that you considered accessibility! You can't use a `<button />` due to styling limitations, right?

janvhs (approved):
LGTM shout out for considering accessibility! :D

jamie-suse (approved):
Majestic :rocket:

assets/js/common/AvailableSoftwareUpdates/Indicator.jsx
dottorblaster: You're right, I'll change it

jamie-suse: Isn't this specifically `navigate` though?

As later you would call it like:

```jsx
      <div
        role="button"
        tabIndex={0}
        className={classNames(
          'flex flex-row items-center border border-gray-200 p-2 rounded-md grow',
          { 'cursor-pointer': !unknown }
        )}
        onClick={onClick}
        onKeyDown={({ code }) => {
          if (code === 'Enter') {
            onClick();
          }
        }}
      >
```

???

dottorblaster: @jamie-suse would you be fine with onNavigate? I'd be fine with onClick but maybe onNavigate is more "neutral" 😄

PR #2690, opened 2024-06-12
Activity log retention time

nelsonkopliku:
# Description

Adds the foundation for the activity log settings, specifically retention time.

- migration to add the required fields into the settings STI
- schema and constraints
- initial default retention time is provided
- context exposing basic read/update operations

The context functions can be then used in a controller to expose desired operations.

### DIFF
--- a/priv/repo/migrations/20240612081138_add_activity_log_settings.exs
+++ b/priv/repo/migrations/20240612081138_add_activity_log_settings.exs
@@ -1,0 +1,9 @@
+defmodule Trento.Repo.Migrations.AddActivityLogSettings do
+  use Ecto.Migration
+
+  def change do
+    alter table(:settings) do
+      add :activity_log_retention_time, :map
+    end
+  end
+end
--- a/priv/repo/migrations/20240612081138_add_activity_log_settings.exs
+++ b/priv/repo/migrations/20240612081138_add_activity_log_settings.exs
@@ -6,4 +6,12 @@ defmodule Trento.Repo.Migrations.AddActivityLogSettings do
       add :activity_log_retention_time, :map
     end
   end
+
+  def down do
+    execute "DELETE from settings WHERE type = 'activity_log_settings'"
+
+    alter table(:settings) do
+      remove :activity_log_retention_time
+    end
+  end
 end
--- a/test/trento/activity_log_test.exs
+++ b/test/trento/activity_log_test.exs
@@ -23,16 +23,16 @@ defmodule Trento.ActivityLogTest do
     test "should return settings" do
       %{
         retention_time: %{
-          retention_period: retention_period,
-          retention_period_unit: retention_period_unit
+          value: value,
+          unit: unit
         }
       } = insert(:activity_log_settings)
 
       assert {:ok,
               %Settings{
                 retention_time: %RetentionTime{
-                  retention_period: ^retention_period,
-                  retention_period_unit: ^retention_period_unit
+                  value: ^value,
+                  unit: ^unit
                 }
               }} = ActivityLog.get_settings()
     end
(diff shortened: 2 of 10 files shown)


priv/repo/migrations/20240612081138_add_auditing_settings.exs
```diff
@@ -0,0 +1,10 @@
+defmodule Trento.Repo.Migrations.AddAuditingSettings do
```
gagandeepb, line 1: Optional: maybe a [constraint](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#constraint/3 ) on making retention period > 0:

priv/repo/migrations/20240612081138_add_auditing_settings.exs
```diff
@@ -0,0 +1,10 @@
...
+  use Ecto.Migration
+
+  def change do
+    alter table(:settings) do
+      add :auditing_retention_period, :integer
+      add :auditing_retention_period_unit, :string
```
gagandeepb, line 7: Something to consider: [default values](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#add/3-options) at the migration level.

gagandeepb (commented):
(Just adding some comments for later consideration.)

nelsonkopliku:
hey @gagandeepb, @balanza, I think I have addressed what we discussed:
- removed the `save retention time` operation as we do not really have to cover that use case
- init default retention time to 1 month at application startup, and in seeds for dev environment
- now the settings table has only one field `retention_time`, being an embedded value object stored as json **\***
- renamed the whole thing to `Activity Log`, as per naming discussions :see_no_evil: 

**\*** Note that this moves possible validation errors one level down in `Trento.ActivityLog.Settings` (we get changeset errors on the embedded `RetentionTime`), and I am not sure our error handling at controller level is ready for this.
It might end up in not being an issue, but we'll discover more when actually implementing the controller.
If anything needs to be changed, we'll do so.

Thanks for the feedbacks!

priv/repo/migrations/20240612081138_add_auditing_settings.exs
nelsonkopliku: let's take this into account if we move back from a one-json-field to separate ones

priv/repo/migrations/20240612081138_add_activity_log_settings.exs
```diff
@@ -0,0 +1,17 @@
+defmodule Trento.Repo.Migrations.AddActivityLogSettings do
+  use Ecto.Migration
+
+  def change do
+    alter table(:settings) do
+      add :activity_log_retention_time, :map
+    end
+  end
+
+  def down do
```
nelsonkopliku, line 10: I realized that while rolling back, only the `activity_log_retention_time` field was removed, but the record typed `activity_log_settings` remained, meaning that migrating forward again would result in having the activity log settings record with a `null` retention time.

In order to avoid unwanted side effects I thought it is reasonable just deleting the whole record.

test/trento/activity_log_test.exs
```diff
@@ -0,0 +1,180 @@
...
+          retention_period_unit: retention_period_unit
+        }
+      } = insert(:activity_log_settings)
+
+      assert {:ok,
+              %Settings{
+                retention_time: %RetentionTime{
+                  retention_period: ^retention_period,
+                  retention_period_unit: ^retention_period_unit
+                }
```
balanza, line 36: 1. **thought:** ‏I just realize that this result may be a little verbose. Should we prefer a more lean structure such as
```
{
  retention_time: {
    value: 10
    unit: "month"
  }
}
```

I don't have a strong opinion nor I am considering the needed renaming in the model, I'm just thinking on that

balanza (approved):
LGTM, I just have some thoughts on naming but they're non blocking

test/trento/activity_log_test.exs
nelsonkopliku: Sure 👍 gonna change shortly. It's actually nicer, thanks!

nelsonkopliku: here we go 9432302d0ff1036ccb247a81b19be8beae95fe1f

PR #2692, opened 2024-06-13
Watch channel events

arbulu89:
# Description

Refactor how we watch websocket events using a [eventChannel](https://redux-saga.js.org/docs/advanced/Channels/#using-the-eventchannel-factory-to-connect-to-external-events).
This is the correct way to listen external events.

Big benefits:
- Remove the direct redux store dependency
- Refactor the usage of events, using correct actions (that are created using the javascript internal event names). So we don't have direct attachment to websocket event names and redux event names

## How was this tested?

Test added. I have created a mocked version of the phoenix sockets. It looks minimalist enough


dottorblaster (approved):
LGTM, this is great!

For the future we might consider splitting `channels.js` into multiple files but on the other hand I just hope we don't have many more events to manage :stuck_out_tongue:

EMaksy (commented):
Nice!

PR #2693, opened 2024-06-14
Support multi line text on tooltip

balanza:
# Description

Allow the `<Tooltip />` component to display multi line text. Extra spaces are collapsed as `white-space: pre-line` is used.

Example:
image


balanza:
@jagabomb the need raised when adding the tooltip on tag validation. Let's check details such as text align.

balanza:
I close the PR as the same result is more easy achievable by using jsx node instead of plain text:

```jsx
content = (
      <>
        the head line
        <br />
        the middle line
        <br />
        the bottom line
      </>
    )
```

PR #2696, opened 2024-06-17
Add validation tooltip on tag creation

balanza:
# Description

Add a validation tooltip on tag creation. When the user type an unsupported character, the tooltip appears to communicate the failure; the tooltip disappears when adding a valid character.

ezgif-3-d916a02f2f

The feature is added to all editable tags in the application.

### DIFF
--- a/assets/js/common/Tags/Tags.stories.jsx
+++ b/assets/js/common/Tags/Tags.stories.jsx
@@ -15,7 +15,3 @@ export function Populated(args) {
 export function Empty(args) {
   return <Tags tags={[]} {...args} />;
 }
-
-export function WithValidation(args) {
-  return <Tags tags={[]} {...args} />;
-}
--- a/assets/js/common/Tags/Tags.stories.jsx
+++ b/assets/js/common/Tags/Tags.stories.jsx
@@ -15,3 +15,7 @@ export function Populated(args) {
 export function Empty(args) {
   return <Tags tags={[]} {...args} />;
 }
+
+export function WithValidation(args) {
+  return <Tags tags={[]} {...args} />;
+}
--- a/assets/js/common/Tooltip/Tooltip.jsx
+++ b/assets/js/common/Tooltip/Tooltip.jsx
@@ -26,6 +26,11 @@ function Tooltip({
   place = 'top',
   isEnabled = true,
   wrap = true,
+  // The visible ternary flag forces the tooltip to show/hide
+  // regardless of the user interaction.
+  //  true -> tooltip is always visible
+  //  false -> tooltip is never visible
+  //  undefined -> tooltip is visible when the user triggers the action (e.g. hover)
   visible,
   ...rest
 }) {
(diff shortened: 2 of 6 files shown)


assets/js/common/Tags/Tags.stories.jsx
```diff
@@ -15,3 +15,7 @@ export function Populated(args) {
 export function Empty(args) {
   return <Tags tags={[]} {...args} />;
 }
+
+export function WithValidation(args) {
```
nelsonkopliku, line 19: As validation is embedded in the tag it would be possible to test it in any other story, so I guess we might not need this.

Besides this this story is the same as the previous `Empty` one :smile:

assets/js/common/Tooltip/Tooltip.jsx
```diff
@@ -26,6 +26,7 @@ function Tooltip({
   place = 'top',
   isEnabled = true,
   wrap = true,
+  visible,
```
nelsonkopliku, line 29: I am wondering whether this prop is somehow doing something similar to what `isEnabled` does.

Maybe we can consolidate in only one of the two. What do you think?

nelsonkopliku (commented):
@balanza thanks for this, it looks good. I noted a couple of things.

Besides that, I also noticed that the validation regex is not protecting against the following `@_-=+,,...:::`.
I know, you did not touch the regex. Up to you whether you want to make it stricter.

assets/js/common/Tooltip/Tooltip.jsx
balanza: **Short answer:** 

`visible` can force the element to show or not while `isEnabled` controls whether is in the virtual dom or not.

**Long answer:**

There is some hidden knowledge here I had to find out. The `isEnabled` flag just don't render the tooltip into the virtual dom. The `visible` flag renders the components in the virtual dom, but hides it with css.

The reason we need both is that using `isEnable` would re-render children and make the input field to lose focus.
Furthermore, `visible` is a flag that forces visibility regardless of the user action (that is, the popup is shown even when the user doesn't hover the control).

The `visible` flag is a prop from our dependency `rc-trigger` ([link](https://github.com/fis-components/rc-trigger/blob/7b659fd7f76ad88129ca09e56522b3419fcf66ea/lib/Popup.js#L78-L79)).

balanza: Btw you made me think it might be worth a comment in https://github.com/trento-project/web/pull/2696/commits/a72d5844bdf57261dbc2af988a3fe115122a9321.

Thanks

assets/js/common/Tags/Tags.stories.jsx
balanza: good catch, I removed the story in https://github.com/trento-project/web/pull/2696/commits/3fb66a886c417616600bcac4ffd98f939b2e0bde

assets/js/common/Tooltip/Tooltip.jsx
nelsonkopliku: Ok, if we need both, so be it. Thanks.

nelsonkopliku (approved):
Great @balanza! I am fine with a possible follow up on the regex, if needed.

PR #2701, opened 2024-06-18
Use css style on component testing

balanza:
# Description

Some component tests rely on css class in order to determine the component properties. For example in https://github.com/trento-project/web/commit/56a978b615e4cdd191f5ddd37500d57ff262d293 the `toBeVisible()` matcher   look for style `display: none` to be applied to the element or one of its parents.

This PR proposes the `<WithStyle>` test utility. `<WithStyle>` is a higher-order component that prepends the application css to the component to be rendered. It's meant to be used only on tests that require such ability (for keeping the test footprint small).

In order to work the application css must be built already, thus a `pretest` script is added to package.json. 
~Furthermore, the package.json has been refactored a bit to make use of npm variables (not required, it can be reverted).~

### DIFF
--- a/assets/js/common/Tooltip/Tooltip.test.jsx
+++ b/assets/js/common/Tooltip/Tooltip.test.jsx
@@ -84,4 +84,25 @@ describe('Tooltip', () => {
       expect(screen.queryByText('This is my tooltip text')).toBeNull()
     );
   });
+
+  it('should hide the text when the mouse go away', async () => {
+    const user = userEvent.setup();
+
+    render(
+      <div>
+        <Tooltip content="This is my tooltip text">This is my anchor</Tooltip>
+      </div>
+    );
+
+    expect(screen.queryByText('This is my tooltip text')).toBeNull();
+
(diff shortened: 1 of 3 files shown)


balanza:
A quick thought on performance based on my laptop

```sh
# running command
time npm run test js/common/Tooltip/Tooltip.test.jsx

# without style
npm run test js/common/Tooltip/Tooltip.test.jsx  3.43s user 0.76s system 180% cpu 2.325 total

# with style, generated just in time (first run)
npm run test js/common/Tooltip/Tooltip.test.jsx  5.33s user 0.96s system 168% cpu 3.727 total

# with style, already generated
npm run test js/common/Tooltip/Tooltip.test.jsx  3.48s user 0.73s system 173% cpu 2.419 total
```

balanza:
Addendum: with @jamie-suse we investigated a bit around the possibility to load CSS once for all JSDOM instances. It seems there is no this possibility: furthermore, it seems this feature is considered out-of-scope by JSDOM itself (see https://github.com/testing-library/jest-dom/issues/70).

nelsonkopliku (approved):
LGTM

PR #2702, opened 2024-06-18
Activity log retention time endpoint

gagandeepb:
# Description

Implements an endpoint for getting and updating activity log retention time settings.

Yes.

## How was this tested?

Unit tests.

### DIFF
--- a/lib/trento_web/controllers/v1/settings_controller.ex
+++ b/lib/trento_web/controllers/v1/settings_controller.ex
@@ -108,7 +108,7 @@ defmodule TrentoWeb.V1.SettingsController do
     ]
 
   def update_activity_log_settings(%{body_params: body_params} = conn, _) do
-    %{retention_period: retention_period, retention_period_unit: retention_period_unit} =
+    %{retention_time: %{value: retention_period, unit: retention_period_unit}} =
       body_params
 
     with {:ok, updated_settings} <-
--- a/lib/trento_web/controllers/v1/settings_controller.ex
+++ b/lib/trento_web/controllers/v1/settings_controller.ex
@@ -137,6 +137,10 @@ defmodule TrentoWeb.V1.SettingsController do
         })
 
       _ ->
+        # Here we rebind the error returned by the ActivityLog.get_settings/0 function
+        # to a "not_found" in order to disambiguate from a similar error being returned by the
+        # ActivityLog.change_retention_period/2 function. This error gets picked up by the FallbackController
+        # and leads to dispatching to the appropriate error response.
         {:error, :not_found}
     end
   end
--- a/lib/trento_web/controllers/v1/settings_controller.ex
+++ b/lib/trento_web/controllers/v1/settings_controller.ex
@@ -118,8 +118,8 @@ defmodule TrentoWeb.V1.SettingsController do
           activity_log_settings: updated_settings
         })
 
-      _ ->
-        {:error, :unprocessable_entity}
+      error ->
+        error
     end
   end
 
@@ -134,8 +134,6 @@ defmodule TrentoWeb.V1.SettingsController do
          Schema.Platform.ActivityLogSettings},
       not_found: Schema.NotFound.response()
     ]
-
-  require Logger
 
   def get_activity_log_settings(conn, _) do
     case ActivityLog.get_settings() do
--- a/lib/trento_web/controllers/v1/settings_controller.ex
+++ b/lib/trento_web/controllers/v1/settings_controller.ex
@@ -111,14 +111,11 @@ defmodule TrentoWeb.V1.SettingsController do
     %{retention_period: retention_period, retention_period_unit: retention_period_unit} =
       body_params
 
-    case ActivityLog.change_retention_period(retention_period, retention_period_unit) do
-      {:ok, updated_settings} ->
-        render(conn, "activity_log_settings.json", %{
-          activity_log_settings: updated_settings
-        })
-
-      error ->
-        error
+    with {:ok, updated_settings} <-
+           ActivityLog.change_retention_period(retention_period, retention_period_unit) do
+      render(conn, "activity_log_settings.json", %{
+        activity_log_settings: updated_settings
+      })
     end
   end
 
--- a/lib/trento_web/controllers/v1/settings_controller.ex
+++ b/lib/trento_web/controllers/v1/settings_controller.ex
@@ -2,6 +2,7 @@ defmodule TrentoWeb.V1.SettingsController do
   use TrentoWeb, :controller
   use OpenApiSpex.ControllerSpecs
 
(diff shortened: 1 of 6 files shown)


lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -93,4 +94,56 @@ defmodule TrentoWeb.V1.SettingsController do
...
+
+      error ->
+        error
+    end
+  end
+
+  operation :get_activity_log_settings,
+    summary: "Fetches the Activity Log settings",
+    tags: ["Platform"],
+    request_body:
```
nelsonkopliku, line 129: I don't think we need a `request_body` on a get operation

test/trento_web/controllers/v1/settings_controller_test.exs
```diff
@@ -110,4 +110,58 @@ defmodule TrentoWeb.V1.SettingsControllerTest do
...
+    } do
+      insert(:activity_log_settings)
+
+      conn
+      |> get("/api/v1/settings/activity_log")
+      |> json_response(200)
+      |> assert_schema("ActivityLogSettings", api_spec)
+    end
+
+    test "should not return activity retention settings without setting up", %{
```
nelsonkopliku, line 131: This should never happen because we make sure to initialize a default retention time, but I guess we can keep this.

test/trento_web/controllers/v1/settings_controller_test.exs
```diff
@@ -110,4 +110,58 @@ defmodule TrentoWeb.V1.SettingsControllerTest do
...
+      |> put_req_header("content-type", "application/json")
+      |> put("/api/v1/settings/activity_log", %{
+        retention_period: 42,
+        retention_period_unit: :years
+      })
+      |> json_response(200)
+      |> assert_schema("ActivityLogSettings", api_spec)
+    end
+
+    test "should not update the activity log settings if it is not already configured",
```
nelsonkopliku, line 155: This should never happen because we make sure to initialize a default retention time, but I guess we can keep this.

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -93,4 +94,56 @@ defmodule TrentoWeb.V1.SettingsController do
...
+         Schema.Platform.ActivityLogSettings},
+      unprocessable_entity: Schema.UnprocessableEntity.response(),
+      not_found: Schema.NotFound.response()
+    ]
+
+  def update_activity_log_settings(%{body_params: body_params} = conn, _) do
+    %{retention_period: retention_period, retention_period_unit: retention_period_unit} =
+      body_params
+
+    case ActivityLog.change_retention_period(retention_period, retention_period_unit) do
```
nelsonkopliku, line 115: We could rely on a `with` and let the error bubble up
```elixir
with {:ok, updated_settings} <- ActivityLog.change_retention_period(retention_period, retention_period_unit) do
  render(conn, "activity_log_settings.json", %{
    activity_log_settings: updated_settings
  })
end
```

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -93,4 +94,56 @@ defmodule TrentoWeb.V1.SettingsController do
...
+    summary: "Updates the Activity Log settings",
+    tags: ["Platform"],
+    request_body:
+      {"ActivityLogSettings", "application/json", Schema.Platform.ActivityLogSettings},
+    responses: [
+      ok:
+        {"Activity Log settings saved successfully", "application/json",
+         Schema.Platform.ActivityLogSettings},
+      unprocessable_entity: Schema.UnprocessableEntity.response(),
+      not_found: Schema.NotFound.response()
```
nelsonkopliku, line 108: As of current implementation this endpoint never returns 404.

nelsonkopliku (commented):
I see this is still a draft, hope you don't mind a few comments anyway :smile: 

Besides what commented I would have probably uses dedicated controller, view and openapi-schema files/modules, for consistency with what we did for suma related settings for instance. 
However in this case there is less operations involved, so I'd be fine with this for the time being.

Additionally, have you considered embedding the retention time in its own property in the API shape?

Something like
```json
{
    "retention_time": {
        "value": 1,
        "unit": "days"
    }
}
```
One advantage I might think about it is that it might be more future-proof in case we'll need extra settings related to activity log.

lib/trento_web/controllers/v1/settings_controller.ex
gagandeepb: Screenshot from 2024-06-18 16-32-04
Single clause `with` leads to a credo failure due to [this community readability default rule](https://hexdocs.pm/credo/Credo.Check.Readability.WithSingleClause.html) that is active.

nelsonkopliku: I guess you have a `with` + `else`, I was suggesting only a `with`.
We are anyway doing so elsewhere and credo is not complying :eyes:

gagandeepb: I see, good catch. Fixed this usage.

nelsonkopliku (approved):
Great @gagandeepb! Thanks for this.

I already shared my thoughts about the alternative of embedding the retention time like this
```json
{
    "retention_time": {
        "value": 1,
        "unit": "days"
    }
}
```
instead of the flat list of props
```json
{
    "retention_period": 1,
    "retention_period_unit": "days"
}
```
I can live with both of them, and if needed we can revisit.
Let's hear @balanza feedback and if he has preferences.

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -93,4 +94,50 @@ defmodule TrentoWeb.V1.SettingsController do
...
+
+  def get_activity_log_settings(conn, _) do
+    case ActivityLog.get_settings() do
+      {:ok, settings} ->
+        render(conn, "activity_log_settings.json", %{
+          activity_log_settings: settings
+        })
+
+      _ ->
+        {:error, :not_found}
```
balanza, line 140: **suggestion:** ‏As far as I understand, here we are mapping the `:activity_log_settings_not_configured` error to `:not_found`. The reason we don't do it in the fallback controller is because we want differentiate the case of this error from receiving the very same in the `update_activity_log_settings` function, right?

Maybe a comment would be worth to explain this to readers?

**nitpick:** Maybe the context is a better place to do such mapping, but it' just about moving code.‏

gagandeepb: Sure, let me add a comment. :+1:

nelsonkopliku (approved):
Great @gagandeepb!

I guess we only have a conflict to fix in `lib/trento_web/controllers/fallback_controller.ex`, then we can merge.

PR #2705, opened 2024-06-20
Include netmasks in CIDR notation where previously only IP was displayed

rtorrero:
# Description
This PR adds a netmask to the places where the hosts IP is displayed. This will allow the users to get a better understand of the network layout that the hosts are running on. 

See the attached screenshot for an example of how this looks:
Captura desde 2024-06-21 10-04-02
Captura desde 2024-06-21 10-03-36

This uses the `ip_addresses_netmasks` added in: https://github.com/trento-project/agent/pull/336

## How was this tested?

Existing tests have been adjusted.


test/trento/hosts/events/host_registered_test.exs
```diff
@@ -15,11 +15,12 @@ defmodule Trento.Hosts.Events.HostRegisteredTest do
       os_version = Faker.App.version()
 
       assert %HostRegistered{
-               version: 4,
+               version: 5,
```
EMaksy, line 18: why again did we need all this version updates again ?

EMaksy (commented):
Hey man overall looks good, just let some small questions :+1:

assets/js/pages/HostsList/HostsList.test.jsx
```diff
@@ -27,7 +27,7 @@ describe('HostsLists component', () => {
     [
       {
         host: 'vmhdbdev01',
-        ip: '10.100.1.1110.100.1.13',
+        ip: '10.100.1.11/2410.100.1.13/24',
```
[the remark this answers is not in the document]
arbulu89: This is how it should be. The test is checking that we have all ip addresses.
Here, visually, it looks strange (and maybe the test could be improved), but it is checking that all IP addresses are shown in sequence

lib/trento/discovery/policies/host_policy.ex
```diff
@@ -231,6 +238,9 @@ defmodule Trento.Discovery.Policies.HostPolicy do
           end)
       })
 
+  @spec remove_netmask(String.t()) :: String.t()
+  defp remove_netmask(address), do: hd(String.split(address, "/"))
```
arbulu89, line 242: maybe better:
```
address
|> String.split("/")
|> hd()
```

lib/trento/discovery/payloads/host_discovery_payload.ex
```diff
@@ -18,6 +18,7 @@ defmodule Trento.Discovery.Payloads.HostDiscoveryPayload do
   deftype do
     field :hostname, :string
     field :ip_addresses, {:array, :string}
+    field :ip_addresses_netmasks, {:array, :string}
```
arbulu89, line 21: I'm still not entirely convinced of having a 2nd field that sends "duplicated" data.
What if have this logic in the policy.
You just check if the coming address has the mask to remove it for the `ip_addresses` field, and leave the `ip_addresses_netmask` as is in the read model.
PD: I'm not saying to not have a new entry in the read model. It is only for the payload itself

lib/trento/discovery/policies/host_policy.ex
```diff
@@ -144,6 +145,12 @@ defmodule Trento.Discovery.Policies.HostPolicy do
            host_id: agent_id,
            hostname: hostname,
            ip_addresses: Enum.filter(ip_addresses, &non_loopback_ipv4?/1),
+           ip_addresses_netmasks:
```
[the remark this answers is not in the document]
arbulu89: I hope it is tested as it is hehe
I find the code pretty simple in fact.
I see some duplication that most probably can be removed though

test/trento/hosts/events/host_registered_test.exs
arbulu89: Because the event is upcasted

arbulu89 (commented):
So far so good.
The unique discussion I would like to have is if we really need to send a 2nd field in the payload "duplicating" the ip addresses.
This simply looks "the easy solution". And I don't even know if it would be backward compatible with agents that don't send this field.
If we add some minor complexity, i think we can handle everything just changing the current `ip_addresses`.
Let's chat about it

PR #2708, opened 2024-06-24
Improve tag validation msg show up

nelsonkopliku:
# Description

Currently the validation message of a tag is kept between interactions with the same tag, which might be misleading.

https://github.com/trento-project/web/assets/8167114/6dbc6cc6-2f13-42cf-89da-78f06c4fa390

This PR makes sure that the state determining whether a validation message needs to be show is reset between tag interactions.

Just OCD :see_no_evil: 

## How was this tested?

Jest


balanza (approved):
Great catch! Thanks

PR #2716, opened 2024-06-26
Navigate on sap system deregistration

arbulu89:
# Description

Navigate to `/sap_systems` from the detail view if the sap system is completely deregistered.
To achieve this, the `router` is added in the sagas context, so we can use the navigation functions there, without passing them to the store (which usually is not a good practice, as functions are not serializable).
Besides that, the redux store is not stored as "global" variable anymore, which was as well not a good practice.

Once this is merged, I will create other PR removing the navigate passing usage in remaining code places.

## How was this tested?

UT added

### DIFF
--- a/assets/js/trento.jsx
+++ b/assets/js/trento.jsx
@@ -1,7 +1,12 @@
 import React from 'react';
 import { createRoot } from 'react-dom/client';
 
-import { Routes, Route, BrowserRouter } from 'react-router-dom';
+import {
+  createRoutesFromElements,
+  createBrowserRouter,
+  Route,
+  RouterProvider,
+} from 'react-router-dom';
 
 import { Toaster } from 'react-hot-toast';
 import { Provider } from 'react-redux';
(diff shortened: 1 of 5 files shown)


assets/js/trento.jsx
```diff
@@ -1,13 +1,17 @@
...
+  createBrowserRouter,
+  Route,
+  RouterProvider,
+  Outlet,
+} from 'react-router-dom';
 
 import { Toaster } from 'react-hot-toast';
 import { Provider } from 'react-redux';
 
-import { ErrorBoundary } from 'react-error-boundary';
```
arbulu89, line 9: I'm just removing the `react-error-boundary` usage, as it can go in the router code itself.
Besides that, we were not using the `onReset` callback, as the `resetErrorBoundary` is not called anywhere.

rtorrero (approved):
Nice @arbulu89 ! I manually tested it and it works nicely. I'd also add it to other sagas, this could allow for example that if a cluster gets deregistered while you are on the cluster details view, it doesn't leave a "Loading..." or a "Not found" text in the view.

CDimonaco (approved):
PERFECT!

:chef_kiss:

LGTM

EMaksy (approved):

PR #2717, opened 2024-06-26
Add migration and schema for ActivityLog

nelsonkopliku:
# Description

Migration and schema for `ActivityLog`.
The idea is that:
- `type` represents the occurence type (ie `resource_tagging`, `resource_untagging`, `api_key_generation` etc...)
- `actor` is the username of the user that initiated the action, or fallback to `system` for automatic operations
- `metadata` a bunch of context for the occurence (ie, the resource that has been tagged/untagged, the tag involved...) 

Indexes have been added for `type`, `actor` and `inserted_at` as they would very likely be used in queries.

Correlation ids and causation ids are intentionally left out from this iteration, and would be added at a later stage.  
## How was this tested?

running the migration

### DIFF
--- a/priv/repo/migrations/20240626103903_create_activity_logs.exs
+++ b/priv/repo/migrations/20240626103903_create_activity_logs.exs
@@ -10,9 +10,5 @@ defmodule Trento.Repo.Migrations.CreateActivityLogs do
 
       timestamps(type: :utc_datetime_usec)
     end
-
-    create index(:activity_logs, [:type])
-    create index(:activity_logs, [:actor])
-    create index(:activity_logs, [:inserted_at])
   end
 end
--- a/priv/repo/migrations/20240626103903_create_activity_logs.exs
+++ b/priv/repo/migrations/20240626103903_create_activity_logs.exs
@@ -1,0 +1,18 @@
+defmodule Trento.Repo.Migrations.CreateActivityLogs do
+  use Ecto.Migration
+
+  def change do
+    create table(:activity_logs, primary_key: false) do
+      add :id, :uuid, primary_key: true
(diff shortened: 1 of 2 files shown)


priv/repo/migrations/20240626103903_create_activity_logs.exs
```diff
@@ -0,0 +1,18 @@
...
+      add :type, :string
+      add :actor, :string
+      add :metadata, :map
+
+      timestamps(type: :utc_datetime_usec)
+    end
+
+    create index(:activity_logs, [:type])
+    create index(:activity_logs, [:actor])
+    create index(:activity_logs, [:inserted_at])
```
balanza, line 16: **thought:** ‏What if we create indexes later? I'm a bit concerned about the improvements vs the added drawbacks. Should we do some performance profiles before tuning the db with indexes?

Maybe it's just me overthinking, though.

_Rationale_
Being logging a _write-intensive_ operation, I'm afraid of the negative impact on performance to the traced activities (that is: the faster the write, the better). Rule of thumb, I think we might afford more a _very slow_ read once in a while than a _slightly slow_ write many times.

balanza (approved):

priv/repo/migrations/20240626103903_create_activity_logs.exs
nelsonkopliku: sure, it makes sense. Let's remove indexes for now

nelsonkopliku: ae9032c86efa85b6b110a9bf1a6266f876c9cfd8

gagandeepb (approved):

PR #2719, opened 2024-06-26
Add migration to restart the abilities sequence from 3

CDimonaco:
# Description

This PR, manually restarts the abilities table primary key sequence from 3.

We forced the ids of the global abilities, with id 1 and id 2.
We need to manually increment the abilities sequence to interact with the records using the postgres sequence mechanism.

Without this change the sequence does not know about record 1 and 2 created, it starts from one, so all the insert will fail.

### DIFF
--- a/priv/repo/migrations/20240626134815_restart_abilities_sequence.exs
+++ b/priv/repo/migrations/20240626134815_restart_abilities_sequence.exs
@@ -1,0 +1,11 @@
+defmodule Trento.Repo.Migrations.RestartAbilitiesSequence do
+  use Ecto.Migration
+
+  def up do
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):
Interesting. Would this fix flakiness in tests as well? like [here](https://github.com/trento-project/web/actions/runs/9678955033/job/26705328941#step:8:326) for instance

EDIT: nevermind, the CI says no.

arbulu89 (approved):
Good!
Bloody postresql hehe

CDimonaco:
FORCE MERGE,

ALL TESTS PASSED, COVERRALS IS DOWN

PR #2720, opened 2024-06-26
Remove navigate usage in redux action dispatching

arbulu89:
# Description
Remove `navigate` function usage in redux action dispatching. The navigate is coming now from the redux saga context

Related to: https://github.com/trento-project/web/pull/2716

PD: And a fix for the toaster location that i have messed during previous PR conflicts resolution...


CDimonaco (approved):
Perfect!

Just a nipick, if we want we can assign the context entry name `router` to a const when we inject first time and import it where we need, but it's really a nitpick

PR #2722, opened 2024-06-27
Switch to 404 for SUSE Manager settings

dottorblaster:
Just switching to 404 since now 403 has a differentic semantic meaning for our frontend :+1:


CDimonaco (approved):
LGTM
 🇮🇹

PR #2726, opened 2024-06-28
Activity logger plug

nelsonkopliku:
# Description

This PR adds a Plug that loads the user for the current request and forwards the connection to an ActivityLogger that might or might not be interested in tracking current activity.

Actual implementation of the activity logger to followup.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento_web/plugs/activity_logging_plug.ex
+++ b/lib/trento_web/plugs/activity_logging_plug.ex
@@ -1,0 +1,33 @@
+defmodule TrentoWeb.Plugs.ActivityLoggingPlug do
+  @moduledoc """
+  This plug is responsible for auditing the requests made to the API.
+  """
+
+  @behaviour Plug
+
+  import Plug.Conn
+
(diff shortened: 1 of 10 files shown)


lib/trento_web/plugs/activity_logging_plug.ex
```diff
@@ -0,0 +1,33 @@
...
+  def init(default), do: default
+
+  def call(%Plug.Conn{} = conn, _default \\ nil), do: register_before_send(conn, &log_activity/1)
+
+  defp log_activity(conn) do
+    Task.Supervisor.start_child(Trento.TasksSupervisor, fn ->
+      conn
+      |> load_user()
+      |> ActivityLogger.log_activity()
+    end)
```
gagandeepb, line 22: This might need to reconsidered at a later point when there is a db based ActivityLogger implementation. (Some sort of bounding of concurrency/back pressure might be necessary.)

gagandeepb (approved):

PR #2732, opened 2024-07-02
Fix non existent path

nelsonkopliku:
# Description

Fixing host checks execution request path that is meant to return 403, but goes 404.

Fixes failing CI https://github.com/trento-project/web/actions/runs/9758779367/job/26934132210#step:8:95


arbulu89 (approved):
Yep! The fix was on the way in other PR.
Thank you

gagandeepb (approved):
Thanks for this!

PR #2733, opened 2024-07-02
Return CVEs from `Errata Details` endpoint

jamie-suse:
# Description

This PR adds the ability to return the list of CVEs for a given advisory queried from the `v1/software_updates/errata_details/:advisory_name` endpoint.

## How was this tested?

Updated related unit tests.

No changes are required for the documentation.

### DIFF
--- a/test/trento_web/controllers/v1/suse_manager_controller_test.exs
+++ b/test/trento_web/controllers/v1/suse_manager_controller_test.exs
@@ -172,7 +172,7 @@ defmodule TrentoWeb.V1.SUSEManagerControllerTest do
         {:ok, errata_details}
       end)
 
-      cves = generate_cves(10)
+      cves = build_list(10, :cve)
 
       expect(Trento.SoftwareUpdates.Discovery.Mock, :get_cves, 1, fn _ ->
         {:ok, cves}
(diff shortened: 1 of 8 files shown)


test/trento_web/controllers/v1/suse_manager_controller_test.exs
```diff
@@ -257,4 +300,15 @@ defmodule TrentoWeb.V1.SUSEManagerControllerTest do
...
+  defp generate_cves(n) do
+    Enum.map(
+      1..n,
+      fn _ ->
+        year = Enum.random(1_991..2_024)
+        id = Enum.random(0..9_999)
+        "CVE-#{year}-#{id}"
+      end
+    )
+  end
```
dottorblaster, line 313: Your call if this is fine like this or we want a factory for this data :+1:

dottorblaster (approved):
One minor nitpick but LGTM

janvhs (approved):
Looks great!

test/trento_web/controllers/v1/suse_manager_controller_test.exs
janvhs: Honestly, I'd just keep it here until we need it somewhere else ^^

dottorblaster: No strong feeling :+1:

jamie-suse: Ah I wanted to and tried to do this but I couldn't get the factory to work because the `build` macro expects a `Map` to be returned, and this is just a `List` of `String`s (`[String.t()]`). Is there a way to do it?

jamie-suse: For future reference if anyone sees this - thanks for the link @dottorblaster!
https://github.com/beam-community/ex_machina?tab=readme-ov-file#non-map-factories

dottorblaster (approved):
LGTM

PR #2727, opened 2024-06-28
Activity log settings UI state

balanza:
# Description

Mount the _Audit Logs_ section in the Settings Page.
The section at startup fetches the settings data from the backend and render them in the UI. To edit, a modal is open with editable form fields.
The database is prefilled with default settings, so there is no empty data scenario.

## Changes 

* rename field `retentionTime` to `retention_time` from #2706 to adhere to API naming
* define reducer, selectors and sagas for handling the fetch and the update of the audit logs settings
* map the api operations
* mount the redux state
* mount the UI components in the `SettingsPage` page.

PR #2743, opened 2024-07-03
Validation global error

balanza:
# Description

Add `getGlobalError` utility to map server validation error that not relate to a field. The use case is to render failures coming from the server when fetching/saving data.

This PR also applies the error handling as an improvement of https://github.com/trento-project/web/pull/2727.

See unit tests for expected usage.

### DIFF
--- a/assets/js/common/ActivityLogsSettingsModal/ActivityLogsSettingsModal.test.jsx
+++ b/assets/js/common/ActivityLogsSettingsModal/ActivityLogsSettingsModal.test.jsx
@@ -4,8 +4,8 @@ import { render, screen, act } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import '@testing-library/jest-dom';
 
-import ActivityLogsSettingsModal from '.';
 import { defaultGlobalError } from '@lib/api/validationErrors';
+import ActivityLogsSettingsModal from '.';
 
 const positiveInt = () => faker.number.int({ min: 1 });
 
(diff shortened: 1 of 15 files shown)


assets/js/common/ActivityLogsSettingsModal/ActivityLogsSettingsModal.test.jsx
```diff
@@ -5,6 +5,7 @@ import userEvent from '@testing-library/user-event';
 import '@testing-library/jest-dom';
 
 import ActivityLogsSettingsModal from '.';
+import { defaultGlobalError } from '../../lib/api/validationErrors';
```
nelsonkopliku, line 8:
```suggestion
import { defaultGlobalError } from '@lib/api/validationErrors';
```

assets/js/state/sagas/activityLogsSettings.js
```diff
@@ -11,6 +11,7 @@ import {
   setEditingActivityLogsSettings,
   setNetworkError,
 } from '@state/activityLogsSettings';
+import { defaultGlobalError } from '../../lib/api/validationErrors';
```
nelsonkopliku, line 14:
```suggestion
import { defaultGlobalError } from '@lib/api/validationErrors';
```

assets/js/state/sagas/activityLogsSettings.test.js
```diff
@@ -17,6 +17,7 @@ import {
   fetchActivityLogsSettings,
   updateActivityLogsSettings,
 } from './activityLogsSettings';
+import { defaultGlobalError } from '../../lib/api/validationErrors';
```
nelsonkopliku, line 20:
```suggestion
import { defaultGlobalError } from '@lib/api/validationErrors';
```

nelsonkopliku (approved):
Super, thanks!

Just a few paths to adjust, if you want, for the rest  LGTM.

I am now realizing that in SUMA credentials modal there is no generic failure handling afaik, so this change could be beneficial there as well.
(I mean, you don't have to do that. Just saying that we now have the opportunity to improve other parts of the app :wink: )

balanza:
@nelsonkopliku super thanks for the suggestion, I actually used the automatic import but it's not that savvy... I'll find out if there are any eslint rules we can enforce, though

PR #2745, opened 2024-07-04
Group abilities frontend

arbulu89:
# Description

Group `abilities` in the frontend using a new `AbilitiesMultiSelect` component.
The component groups some abilities based on hardcoded data for the frontend, but it still saves the individual abilities ids, so they can be updated properly in the backend.

PD: pending to add the settings abilities. We can add them once we have them implemented


CDimonaco (approved):

EMaksy (approved):
Very cool thing, I'm soon finishing up the settings ability and can add it later on.

PR #2747, opened 2024-07-04
Tags abilities e2e tests

CDimonaco:
# Description

Tags abilities e2e testing

## How was this tested?

Automated tests

### DIFF
--- a/test/e2e/cypress/e2e/clusters_overview.cy.js
+++ b/test/e2e/cypress/e2e/clusters_overview.cy.js
@@ -4,6 +4,8 @@ import {
   unhealthyClusterScenario,
 } from '../fixtures/clusters-overview/available_clusters';
 
+import { createUserRequestFactory } from '@lib/test-utils/factories';
+
 const clusterIdByName = (clusterName) =>
   availableClusters.find(({ name }) => name === clusterName).id;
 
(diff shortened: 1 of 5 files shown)


test/e2e/cypress/e2e/clusters_overview.cy.js
```diff
@@ -181,3 +183,42 @@ context('Clusters Overview', () => {
...
+    cy.logout();
+    const user = createUserRequestFactory.build({
+      password,
+      password_confirmation: password,
+    });
+    cy.wrap(user).as('user');
+  });
+
+  const password = 'password';
+  it('it should prevent a tag update when the user abilities are not compliant', () => {
```
arbulu89, line 199: Remove `it` from the text. We never put.
For all the test cases please

arbulu89, line 199: Let's put these tests under the `Tag creation` describe as in the other files

arbulu89 (approved):
It looks good.
Take care of removing the `it`s from the text please.

PD: I think we need to revisit a bit the code on disabled tags with @jagabomb 
I saw now that we have only the cross of the existing tags with less opacity, which looks strange.

And besides that, now I check some code, the `ExistingTag` component name (which I proposed...) doesn't look really appropriate, as the component is the cross of the existing component, not the component itself. Maybe we should rename back to `DeleteTagButton` as you had at the beginning. I think I missunderstood what the code was doing, and thought that this was the whole tag itself, but is just the cross.

test/e2e/cypress/e2e/clusters_overview.cy.js
CDimonaco: done

CDimonaco: done

test/e2e/cypress/e2e/clusters_overview.cy.js
```diff
@@ -4,6 +4,8 @@ import {
   unhealthyClusterScenario,
 } from '../fixtures/clusters-overview/available_clusters';
 
+import { createUserRequestFactory } from '@lib/test-utils/factories';
```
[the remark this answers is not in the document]
CDimonaco: done

PR #2753, opened 2024-07-08
Activity Log settings e2e

balanza:
# Description

Provide e2e tests for the Activity log section on the settings page. 

This PR includes a refactor to the production code (https://github.com/trento-project/web/commit/eace365f6f91965ca4780a4066678fa8d1487d9e) that simplifies focusing on elements by segmenting the HTML with `<section>` (no UI change is involved).

### DIFF
--- a/test/e2e/cypress/e2e/settings.cy.js
+++ b/test/e2e/cypress/e2e/settings.cy.js
@@ -701,7 +701,7 @@ context('Settings page', () => {
         .as('initialRetentionTime');
     });
 
-    describe.only('Changing Settings', () => {
+    describe('Changing Settings', () => {
       it('should change retention time', () => {
         // the initial retention time is read before the settings are changed
         cy.get('@initialRetentionTime').then((text) =>
@@ -759,8 +759,7 @@ context('Settings page', () => {
         modal().get('button').contains('Cancel').click();
 
         // Check the modal is closed
-        cy.get('h3').should('not.have.text', 'Enter Activity Logs Settings');
-
+        modal().then(($modal) => $modal.not.is(':visible'));
         // Check the data is not changed
         cy.get('@initialRetentionTime').then((text) =>
           section()
@@ -793,6 +792,7 @@ context('Settings page', () => {
         modal().get('button').contains('Save Settings').click();
 
         // Check we still show the modal
+        modal().then(($modal) => $modal.is(':visible'));
 
         // Close the modal
         modal().get('button').contains('Cancel').click();
--- a/test/e2e/cypress/e2e/settings.cy.js
+++ b/test/e2e/cypress/e2e/settings.cy.js
@@ -677,4 +677,133 @@ context('Settings page', () => {
       });
     });
   });
+
+  describe('Activity log', () => {
+    // Elements must be filtered to match only the ones relate to the Activity Logs.
+    // Readable texts is contained by the section,
+    // while inputs are contained by the modal.
+    // The use of helpers instead of Cypress alias avoid problems with the React rendering,
+    // as the alias is not updated when the component is re-rendered.
+    const section = () => cy.contains('Activity Logs').parents('section');
+    const modal = () =>
+      cy.contains('Enter Activity Logs Settings').parents('div').first();
+
+    before(() => {
+      cy.reload();
+    });
+
+    beforeEach(() => {
+      // Read the initial retention time to compare with the new value
+      // Also, it checks the retention time is displayed correctly
+      section()
+        .get('[aria-label="retention-time"]')
+        .invoke('text')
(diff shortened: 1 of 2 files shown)


test/e2e/cypress/e2e/settings.cy.js
```diff
@@ -677,4 +677,133 @@ context('Settings page', () => {
...
+        .as('initialRetentionTime');
+    });
+
+    describe.only('Changing Settings', () => {
+      it('should change retention time', () => {
+        // the initial retention time is read before the settings are changed
+        cy.get('@initialRetentionTime').then((text) =>
+          section()
+            .get('[aria-label="retention-time"]')
+            .should('have.text', text)
```
nelsonkopliku, line 710: how does this change from what we do in `beforeEach`? Are they both needed?

test/e2e/cypress/e2e/settings.cy.js
```diff
@@ -677,4 +677,133 @@ context('Settings page', () => {
...
+          .get('[aria-label="retention-time"]')
+          .should('have.text', '5 months');
+      });
+
+      it('should not change settings when cancelling operation', () => {
+        // the initial retention time is read before the settings are changed
+        cy.get('@initialRetentionTime').then((text) =>
+          section()
+            .get('[aria-label="retention-time"]')
+            .should('have.text', text)
```
nelsonkopliku, line 739: as before: is this necessary or is what we do in beforeEach enough?

test/e2e/cypress/e2e/settings.cy.js
```diff
@@ -677,4 +677,133 @@ context('Settings page', () => {
       });
     });
   });
+
+  describe('Activity log', () => {
+    // Elements must be filtered to match only the ones relate to the Activity Logs.
```
nelsonkopliku, line 682: interesting approach.
Another way that I find helpful in order to get specific elements and avoid weird css selectors is using custom classes/labels/names (take a look at `SUMA Credentials Management` section)
No need to change anything, just wanting to share.

test/e2e/cypress/e2e/settings.cy.js
```diff
@@ -677,4 +677,133 @@ context('Settings page', () => {
...
+        );
+        cy.get('@changeSettings').should('not.have.been.called');
+      });
+
+      it('should return validation error on invalid input', () => {
+        // the initial retention time is read before the settings are changed
+        cy.get('@initialRetentionTime').then((text) =>
+          section()
+            .get('[aria-label="retention-time"]')
+            .should('have.text', text)
```
nelsonkopliku, line 778: Same as previous comments.

However I see the need for what we do later in line [801](https://github.com/trento-project/web/pull/2753/files#diff-2daf4a927dc90ea32c59092e95cbd3f017189a193fe4fca0fb2f6595dcf0ebb2R801)

nelsonkopliku (approved):
Looks good!

Just a doubt about whether we are doing the same thing in `beforeEach` and in each test.

test/e2e/cypress/e2e/settings.cy.js
balanza: First and foremost, I was experimenting thus I'm not sure _my_ approach is good.

We have to face 2 orders of problems here:
1. to select the element in the right section
2. to select an element that might have been re-rendered.

Regarding 1, maybe selecting even cypress docs says that selecting by text is not a best practice (https://docs.cypress.io/guides/references/best-practices#Text-Content). Perhaps we can assign a specific class/ID to every section? Quirk: models don't fall under any section :upside_down_face: 
I'm considering a refactor, though.

About 2, I found no other option than to use helper functions. I wonder if it's a huge burden for performance.

test/e2e/cypress/e2e/settings.cy.js
balanza: In the `beforeEach` clause we read the initial text, here we ensure the test starts from there.

It's indeed a step that never fails, I mostly added it for completing the test case (start from a value x, get another value y). Should we remove it? I see no harm, I wonder if it can work as a "safety net" for regressions.

test/e2e/cypress/e2e/settings.cy.js
balanza: Yes exactly. I think is more about having a complete story in the test case. Pragmatically, it could be not very helpful.

test/e2e/cypress/e2e/settings.cy.js
nelsonkopliku: Yes, I like that piece of document. I also find _using highly brittle selectors that are subject to change_ (which is not the case here) a pain :see_no_evil: 
All good from my side

nelsonkopliku (approved):

PR #2757, opened 2024-07-08
Provide explicit ordering for CSV file columns

jamie-suse:
# Description

As per feedback, when a user clicks on the _Download CSV_ button, the generated CSV file will now explicitly provide **only** the columns `id`, `advisory_type`, `advisory_name`, `advisory_synopsis` & `update_date`, in that order.

## How was this tested?

This was manually verified, as we can't assert on the _contents_ of an Object URL. However I did verify and update the size of the new file in the existing unit test.


janvhs (approved):
looks good :D

dottorblaster (approved):
LGTM

PR #2758, opened 2024-07-08
Add activity log default ordering by occurrence date

nelsonkopliku:
# Description

Adds a default ordering of the activity log by occurrence date, descending: newer occurrences first.


gagandeepb (commented):
This implementation would get replaced by the features Flop offers around ordering. But up to you if you need a temporary solution.

nelsonkopliku:
> This implementation would get replaced by the features Flop offers around ordering. But up to you if you need a temporary solution.
>
Thanks @gagandeepb, you're right. I don't mind if it gets replaced when the time comes :smile:

gagandeepb (approved):

lib/trento/activity_log.ex
```diff
@@ -40,7 +42,11 @@ defmodule Trento.ActivityLog do
   @spec list_activity_log() :: list(ActivityLog.t())
   def list_activity_log do
     # This will be made filterable/paginatable in a later PR
-    Repo.all(ActivityLog)
+    query =
+      from activity in ActivityLog,
+        order_by: [desc: activity.inserted_at]
+
+    Repo.all(query)
```
balanza, line 49: **thought:** ‏What if we order by `id`? `id`s are autogenerated UUID, hence monotonic; a decrescent ordering by `id` should be enough for getting the latest entries first.

Being a primary key, the `id` field is indexed, while `inserted_at` is not. Rule of thumb, it is more perfomant.

nelsonkopliku: So, I tried `order_by: [desc: activity.id]`, however I am not getting the order I want :confused: it looks like it just ordering alphabetically the uuids. Not sure whether I am doing something wrong.

Thanks for reminding about the index on `inserted_at`, I was under the assumption we had added it with the migration, but now I remember we deferred that.
It might be worth adding it at this point. wdyt?

You made me realize that we might also want to expose the activity log id as part of the api response.

balanza: Ok, so it was a bad assumption from me. Thanks for trying it out.

PR #2760, opened 2024-07-09
Add permissions to the Settings View

EMaksy:
# Description
This pr adds abilities/permissions for the settings view. 
Main features: 
- Added ` ['all:api_key_settings']; [all:suma_settings];  ['all:activity_logs_settings'];` permissions which can be assigned to a user.
- Disabled the Generate Key button for API creation ` [all:api_key_settings];`
- Disabled Edit Settings and Clear Settings Button if the user does not have `[all: all]` or `[all:suma_settings];`
- Disabled Edit Settings button for activity logs if the user does not have `[all: all]` or  `['all:activity_logs_settings'];`
- Protecting backend API calls with policies
- Added and updated stories for storybook
- Added migration for new abilities

## Demo
User creation with new permissions

Default user with disabled buttons

User with right permissions

- [x] **DONE**

## How was this tested?
Added frontend tests to check if buttons are disabled.
Added policy and controller test in backend.

E2E test will follow in a follow-up pr.

- [x] **DONE**

Added moduledoc in new policies for explanation
- [x] **DONE**

### DIFF
--- a/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.jsx
+++ b/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.jsx
@@ -4,6 +4,7 @@ import { noop } from 'lodash';
 import DisabledGuard from '@common/DisabledGuard';
 import Button from '@common/Button';
 
+const activityLogsSettingsPermittedFor = ['all:activity_logs_settings'];
 /**
  * Simplistic utility to transform a time interval in a readable string
  */
@@ -34,7 +35,6 @@ function ActivityLogsConfig({
   retentionTime,
   onEditClick = noop,
   userAbilities,
-  settingsEditPermittedFor,
 }) {
   return (
     <div className="container max-w-7xl mx-auto p-4 sm:p-6 lg:p-8 bg-white dark:bg-gray-800 rounded-lg">
@@ -43,7 +43,7 @@ function ActivityLogsConfig({
         <span className="float-right">
           <DisabledGuard
             userAbilities={userAbilities}
-            permitted={settingsEditPermittedFor}
+            permitted={activityLogsSettingsPermittedFor}
           >
             <Button type="primary-white-fit" size="small" onClick={onEditClick}>
               Edit Settings
--- a/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.jsx
+++ b/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.jsx
@@ -34,14 +34,17 @@ function ActivityLogsConfig({
   retentionTime,
   onEditClick = noop,
   userAbilities,
-  permitted,
+  settingsEditPermittedFor,
 }) {
   return (
     <div className="container max-w-7xl mx-auto p-4 sm:p-6 lg:p-8 bg-white dark:bg-gray-800 rounded-lg">
       <div>
         <h2 className="text-2xl font-bold inline-block">Activity Logs</h2>
         <span className="float-right">
-          <DisabledGuard userAbilities={userAbilities} permitted={permitted}>
+          <DisabledGuard
+            userAbilities={userAbilities}
+            permitted={settingsEditPermittedFor}
+          >
             <Button type="primary-white-fit" size="small" onClick={onEditClick}>
               Edit Settings
             </Button>
--- a/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.jsx
+++ b/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { noop } from 'lodash';
 
+import DisabledGuard from '@common/DisabledGuard';
 import Button from '@common/Button';
 
 /**
@@ -29,15 +30,22 @@ function TimeLabel({ time }) {
  * object that defines a retention time interval. `unit` is one of `day`, `week`, `month`, `year`.'
  * @param {function} props.onEditClick Handles the edit button click
  */
-function ActivityLogsConfig({ retentionTime, onEditClick = noop }) {
+function ActivityLogsConfig({
+  retentionTime,
+  onEditClick = noop,
+  userAbilities,
+  permitted,
+}) {
   return (
     <div className="container max-w-7xl mx-auto p-4 sm:p-6 lg:p-8 bg-white dark:bg-gray-800 rounded-lg">
       <div>
         <h2 className="text-2xl font-bold inline-block">Activity Logs</h2>
         <span className="float-right">
-          <Button type="primary-white-fit" size="small" onClick={onEditClick}>
-            Edit Settings
-          </Button>
+          <DisabledGuard userAbilities={userAbilities} permitted={permitted}>
+            <Button type="primary-white-fit" size="small" onClick={onEditClick}>
+              Edit Settings
+            </Button>
+          </DisabledGuard>
         </span>
       </div>
       <p className="mt-3 mb-3 text-gray-500">
--- a/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.stories.jsx
+++ b/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.stories.jsx
@@ -16,6 +16,14 @@ export default {
       description: 'Callback when the edit button is clicked',
       control: { type: 'function' },
     },
+    userAbilities: {
+      description: 'Users abilities that allow editing activity logs settings',
+      control: 'array',
+    },
+    permitted: {
+      description: 'Abilities that allow editing activity logs settings',
+      control: 'array',
+    },
   },
 };
 
@@ -23,6 +31,8 @@ export const Default = {
   args: {
     retentionTime: { value: 1, unit: 'month' },
     onEditClick: action('edit clicked'),
+    userAbilities: [{ name: 'all', resource: 'all' }],
+    permitted: ['all:activity_logs_settings'],
   },
 };
 
@@ -30,6 +40,8 @@ export const WithPlurals = {
   args: {
     retentionTime: { value: 2, unit: 'month' },
     onEditClick: action('edit clicked'),
+    userAbilities: [{ name: 'all', resource: 'all' }],
+    permitted: ['all:activity_logs_settings'],
   },
 };
 
@@ -37,5 +49,15 @@ export const WithInvalidValue = {
   args: {
     retentionTime: { is: 'invalid' },
     onEditClick: action('edit clicked'),
+    userAbilities: [{ name: 'all', resource: 'all' }],
+    permitted: ['all:activity_logs_settings'],
+  },
+};
+export const WithoutPermission = {
+  args: {
+    retentionTime: { value: 1, unit: 'month' },
+    onEditClick: action('edit clicked'),
+    userAbilities: [{ name: '', resource: '' }],
+    permitted: ['all:activity_logs_settings'],
   },
 };
--- a/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.stories.jsx
+++ b/assets/js/common/ActivityLogsConfig/ActivityLogsConfig.stories.jsx
@@ -20,7 +20,7 @@ export default {
       description: 'Users abilities that allow editing activity logs settings',
       control: 'array',
     },
-    permitted: {
+    settingsEditPermittedFor: {
       description: 'Abilities that allow editing activity logs settings',
       control: 'array',
     },
@@ -32,7 +32,7 @@ export const Default = {
     retentionTime: { value: 1, unit: 'month' },
     onEditClick: action('edit clicked'),
     userAbilities: [{ name: 'all', resource: 'all' }],
-    permitted: ['all:activity_logs_settings'],
+    settingsEditPermittedFor: ['all:activity_logs_settings'],
   },
 };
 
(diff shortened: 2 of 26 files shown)


assets/js/common/ActivityLogsConfig/ActivityLogsConfig.jsx
```diff
@@ -29,15 +30,22 @@ function TimeLabel({ time }) {
  * object that defines a retention time interval. `unit` is one of `day`, `week`, `month`, `year`.'
  * @param {function} props.onEditClick Handles the edit button click
  */
-function ActivityLogsConfig({ retentionTime, onEditClick = noop }) {
+function ActivityLogsConfig({
+  retentionTime,
+  onEditClick = noop,
+  userAbilities,
+  permitted,
```
CDimonaco, line 37: `settingsEditPermittedFor` as prop name

assets/js/common/SuseManagerConfig/SuseManagerConfig.jsx
```diff
@@ -17,6 +18,8 @@ function SuseManagerConfig({
   onClearSettings = noop,
   onTestConnection = noop,
   onCancel = noop,
+  userAbilities,
+  permitted,
```
CDimonaco, line 22: `configEditPermittedFor` as a prop

lib/trento/settings/policy.ex
```diff
@@ -0,0 +1,36 @@
...
+  alias Trento.Users.User
+
+  def authorize(:update_api_key_settings, %User{} = user, ApiKeySettings),
+    do: has_global_ability?(user) or has_api_key_settings_change_ability?(user)
+
+  def authorize(action, %User{}, ApiKeySettings)
+      when action in [:settings, :get_api_key_settings],
+      do: true
+
+  def authorize(:update_activity_log_settings, %User{} = user, ApiKeySettings),
```
CDimonaco, line 24: ActivityLog settings has a dedicated readmodel, you should pass this, if you want to distinguish how to pass different parameters to the policy based on the action, take a look at the tag controller

lib/trento/settings/policy.ex
```diff
@@ -0,0 +1,36 @@
...
+    do: has_global_ability?(user) or has_activity_logs_settings_change_ability?(user)
+
+  def authorize(:get_activity_log_settings, %User{}, ApiKeySettings), do: true
```
CDimonaco, line 27: Same as before

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -144,4 +153,6 @@ defmodule TrentoWeb.V1.SettingsController do
         {:error, :not_found}
     end
   end
+
+  def get_policy_resource(_), do: Trento.Settings.ApiKeySettings
```
CDimonaco, line 157: Settings are based on different readmodels this should change based on the action, you have conn as first parameter, you can pattern match and provide the right readmodel to the policy based on the action

assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
```diff
@@ -39,6 +40,8 @@ const availableTimeOptions = [
 ];
 
 function ApiKeySettingsModal({
+  userAbilities,
+  permitted,
```
CDimonaco, line 44: You don't need to protect the modal, you protected to button to open the modal and backend is already protected, no guards needed here

lib/trento/software_updates/policy.ex
```diff
@@ -0,0 +1,28 @@
...
+  import Trento.Support.PolicyHelper
+  alias Trento.SoftwareUpdates
+  alias Trento.Users.User
+
+  def authorize(action, %User{} = user, SoftwareUpdates)
+      when action in [:create, :update, :delete] do
+    has_global_ability?(user) or has_suma_settings_change_ability?(user)
+  end
+
+  def authorize(action, %User{}, SoftwareUpdates)
```
CDimonaco, line 20: this should be the last catch all pattern match for all the other actions

lib/trento/software_updates/policy.ex
```diff
@@ -0,0 +1,28 @@
...
+      when action in [:show, :test],
+      do: true
+
+  def authorize(_, %User{} = user, _), do: has_global_ability?(user)
```
CDimonaco, line 24: Please restructure this, read the comment before

lib/trento/settings/policy.ex
CDimonaco, line 24: Please also refactor the matches as the other comments in the software update policy

CDimonaco (changes requested):
Overall a good job,

Please refactor the policies, and in particular the settings controller, the third param of the authorize function, should contain the "context" on which the authorize function is acting, we can't have ApiKeySettings for something not related to the apikeysettings, in general in the controller you can match on actions and on whatever conn related and pass the params accordingly.

lib/trento_web/controllers/v1/settings_controller.ex
arbulu89: Most probably you should pattern match with `body_pararms`.
Something like:
def get_policy_resource(%{body_params: %ApiKeySettingsUpdateRequest{}}), do: Trento.Settings.ApiKeySettings

def get_policy_resource(%{body_params: %ActivityLogSettings{}}), do: Trento.ActivityLog.Settings

lib/trento/settings/api_key_settings.ex
```diff
@@ -8,6 +8,8 @@ defmodule Trento.Settings.ApiKeySettings do
 
   import Ecto.Changeset
 
+  defdelegate authorize(action, user, params), to: Trento.Settings.Policy
```
arbulu89, line 11: Remember to add this `defdelegate` to the `trento/activity_logging/settings.ex` file as well

lib/trento_web/controllers/v1/suma_credentials_controller.ex
```diff
@@ -106,6 +115,8 @@ defmodule TrentoWeb.V1.SUMACredentialsController do
     end
   end
 
+  def get_policy_resource(_), do: Trento.SoftwareUpdates
```
arbulu89, line 118: This should return `Trento.SoftwareUpdates.Settings`

lib/trento/software_updates/policy.ex
```diff
@@ -0,0 +1,28 @@
...
+  User with the ability all:suma_settings can edit suma settings and clear suma settings.
+  """
+
+  @behaviour Bodyguard.Policy
+
+  import Trento.Support.PolicyHelper
+  alias Trento.SoftwareUpdates
+  alias Trento.Users.User
+
+  def authorize(action, %User{} = user, SoftwareUpdates)
```
arbulu89, line 15: The 3rd argument should be `Trento.SoftwareUpdates.Settings`.
Remember that we are protecting the settings, not the updates entries

priv/repo/migrations/20240702125459_add_settings_permissions.exs
```diff
@@ -0,0 +1,37 @@
+defmodule Trento.Repo.Migrations.AddSettingsPermissions do
+  use Ecto.Migration
+
+  def up do
+    execute """
+    INSERT INTO abilities (id, name, resource, label, inserted_at, updated_at)
+    VALUES (DEFAULT, 'all', 'api_key_settings', 'Permits generation of API keys', NOW(), NOW())
```
arbulu89, line 7: I would rewrite the label to `Permits all operations in API keys settings`

test/trento/settings/policy_test.exs
```diff
@@ -0,0 +1,53 @@
...
+    user = %User{abilities: [%Ability{name: "other", resource: "other"}]}
+    refute Policy.authorize(:update_api_key_settings, user, ApiKeySettings)
+  end
+
+  test "should not allow new api key generation when user has no abilities" do
+    user = %User{abilities: []}
+    refute Policy.authorize(:update_api_key_settings, user, ApiKeySettings)
+  end
+
+  test "should allow updating activity logs settings if the user has all:all ability" do
```
arbulu89, line 39: We are missing the most important test, that the update of activity logs settings are not allowed if you don't have the required abilities

test/trento_web/controllers/v1/suma_credentials_controller_test.exs
```diff
@@ -482,4 +482,67 @@ defmodule TrentoWeb.V1.SUMACredentialsControllerTest do
       assert "" == resp
     end
   end
+
+  describe "forbidden response" do
+    test "should return 403 when user tries to create settings without right abilities", %{
```
arbulu89, line 487: `should return forbidded` better, as other tests

assets/js/common/ActivityLogsConfig/ActivityLogsConfig.stories.jsx
```diff
@@ -16,26 +16,48 @@ export default {
...
 
 export const WithInvalidValue = {
   args: {
     retentionTime: { is: 'invalid' },
     onEditClick: action('edit clicked'),
+    userAbilities: [{ name: 'all', resource: 'all' }],
+    permitted: ['all:activity_logs_settings'],
+  },
+};
+export const WithoutPermission = {
```
arbulu89, line 56: Usually we are naming those as `EditUnauthorized` kind of way

assets/js/common/ActivityLogsConfig/ActivityLogsConfig.stories.jsx
```diff
@@ -16,26 +16,48 @@ export default {
...
+  args: {
+    retentionTime: { value: 1, unit: 'month' },
```
arbulu89, line 58: Use better:
```
args: {
  ...Default.args,
   userAbilities: [{ name: '', resource: '' }],
   permitted: ['all:activity_logs_settings'],
}
```

assets/js/common/ActivityLogsConfig/ActivityLogsConfig.test.jsx
```diff
@@ -28,10 +37,38 @@ describe('ActivityLogsConfig', () => {
...
 
     await user.click(screen.getByRole('button', { name: 'Edit Settings' }));
 
     expect(spy).toHaveBeenCalledTimes(1);
   });
+
+  it('should render with a disabled Edit Settings button as the user has not the right permissions', async () => {
+    const user = userEvent.setup();
+    const spy = jest.fn();
+    const userWithoutPermission = [{ name: '', resource: '' }];
```
arbulu89, line 56: Provide as empty list better. Empty name and resource look like antipattern

assets/js/common/SuseManagerConfig/SuseManagerConfig.stories.jsx
```diff
@@ -77,4 +88,16 @@ export const WithVeryLongSUMAUrl = {
...
 
-export const Empty = { args: {} };
+export const Empty = {
+  args: {
+    userAbilities: [{ name: 'all', resource: 'all' }],
+    permitted: ['all:all'],
+  },
+};
+
+export const WithoutPermission = {
```
arbulu89, line 98: Same naming as before. And better use an empty list, empty strings looks strange

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -263,6 +277,8 @@ function SettingsPage() {
         </div>
       </div>
       <ApiKeySettingsModal
+        userAbilities={abilities}
```
arbulu89, line 280: As carmine said we can remove these. We don't need to protect anything within the modal

arbulu89 (changes requested):
Some few comments together with what @CDimonaco said

lib/trento_web/controllers/v1/settings_controller.ex
arbulu89: You cannot, it is used by the policy at the end, which is in other module

assets/js/common/ActivityLogsConfig/ActivityLogsConfig.stories.jsx
```diff
@@ -16,26 +16,47 @@ export default {
...
+    userAbilities: [{ name: 'all', resource: 'all' }],
+    settingsEditPermittedFor: ['all:activity_logs_settings'],
   },
 };
 
 export const WithPlurals = {
   args: {
     retentionTime: { value: 2, unit: 'month' },
     onEditClick: action('edit clicked'),
+    userAbilities: [{ name: 'all', resource: 'all' }],
```
arbulu89, line 43: We can use `...Default.args` here and in the next as well

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -287,6 +301,8 @@ function SettingsPage() {
               onRetry={() => dispatch(fetchSoftwareUpdatesSettings())}
             >
               <SuseManagerConfig
+                userAbilities={abilities}
+                configEditPermittedFor={sumaSettingsPermittedFor}
```
arbulu89, line 305: Do you need to pass the `configEditPermittedFor` to SUMA (the same for activility log`)?
It looks something internal for that component.
You could create the constant there, and avoid this extra parameter, which in reality doesn't add any value

arbulu89 (approved):
Really good improvement @EMaksy 
Nice!
Some few nitpick comments.
The unique one that I would really like you having a look is the passing of `configEditPermittedFor` to SUMA configuration. It looks not neccesary, you can create the constant there and avoid one prop that we don't need. The same for activity log.
Once you change this, ready to merge!

assets/js/pages/SettingsPage/SettingsPage.jsx
EMaksy: No you are fully right, it is much cleaner to have a const in the own component

CDimonaco (approved):
Good job!

LGTM 🚀 

@arbulu89 I proposed Eugen to have also the get action in the get_policy_resource in the settings controller, to have more consistency, but it's true we don't use them, really up to you guys, no strong opinions :D

PR #2762, opened 2024-07-09
Use uuids for usernames in user factory

dottorblaster:
Since we were getting occasional violations of primary keys for usernames in tests, I'm just switching to uuids for those in the dedicated factory.


CDimonaco (approved):
LGTM

Thanks for addressing this issue!

PR #2763, opened 2024-07-09
Affected packages

janvhs:
# Description

Display the affected packages and hosts (WIP) in the advisory details page.

## How was this tested?

Elixir unit tests, Jest tests, Storybook

### DIFF
--- a/assets/js/lib/test-utils/factories/advisoryErrata.js
+++ b/assets/js/lib/test-utils/factories/advisoryErrata.js
@@ -4,7 +4,7 @@ import { advisoryType } from './relevantPatches';
 
 const affectedPackageFactory = Factory.define(() => ({
   name: faker.animal.cat().toLowerCase(),
-  arch_label: faker.helpers.arrayElement(["x86_64", "i586", "aarch64"]),
+  arch_label: faker.helpers.arrayElement(['x86_64', 'i586', 'aarch64']),
   version: faker.system.semver(),
   release: `${faker.number.int({ min: 0, max: 100 })}`,
   epoch: `${faker.number.int({ min: 0, max: 50 })}`,
--- a/assets/js/lib/test-utils/factories/advisoryErrata.js
+++ b/assets/js/lib/test-utils/factories/advisoryErrata.js
@@ -2,6 +2,14 @@ import { faker } from '@faker-js/faker';
 import { Factory } from 'fishery';
 import { advisoryType } from './relevantPatches';
 
+const affectedPackageFactory = Factory.define(() => ({
+  name: faker.animal.cat().toLowerCase(),
+  arch_label: faker.helpers.arrayElement(["x86_64", "i586", "aarch64"]),
+  version: faker.system.semver(),
+  release: `${faker.number.int({ min: 0, max: 100 })}`,
+  epoch: `${faker.number.int({ min: 0, max: 50 })}`,
+}));
+
 const fixMapFactory = Factory.define(({ transientParams }) => {
   const { length = 1 } = transientParams;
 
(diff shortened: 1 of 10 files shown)


janvhs:
Would love to get your feedback, so I can copy paste it for the affected hosts section

assets/js/pages/AdvisoryDetails/AdvisoryDetails.test.jsx
```diff
@@ -7,11 +7,15 @@ import { faker } from '@faker-js/faker';
...
-import AdvisoryDetails from './AdvisoryDetails';
+import AdvisoryDetails, { formatPackage } from './AdvisoryDetails';
 
 describe('AdvisoryDetails', () => {
   it('displays a message, when the CVE, packages or fixes section is empty', () => {
-    const errata = advisoryErrataFactory.build({ cves: [], fixes: {} });
+    const errata = advisoryErrataFactory.build({
+      cves: [],
+      fixes: {},
+      affected_packages: {},
```
dottorblaster, line 17:
```suggestion
      affected_packages: [],
```
If you `.buildList()` the factory in tests and `.map()` this in the implementation you're expeting an array, hence better have an empty array here

assets/js/lib/test-utils/factories/advisoryErrata.js
```diff
@@ -31,6 +39,8 @@ export const advisoryErrataFactory = Factory.define(({ params }) => ({
       { transient: { length: faker.number.int({ min: 1, max: 10 }) } }
     ),
   cves: cveFactory.buildList(10),
+  affected_packages:
+    params.affected_packages || affectedPackageFactory.buildList(10),
```
dottorblaster, line 43:
```suggestion
    affectedPackageFactory.buildList(10),
```

If you use an empty array in tests this will be replaced automatically without the need of this `or` expression :+1:

assets/js/pages/AdvisoryDetails/AdvisoryDetails.jsx
```diff
@@ -5,14 +5,16 @@ import PageHeader from '@common/PageHeader';
 import ListView from '@common/ListView';
 import AdvisoryIcon from '@common/AdvisoryIcon';
 
+export const formatPackage = ({ name, version, epoch, release, arch_label }) =>
+  `${name}-${version}-${epoch}.${release}-${arch_label}`;
```
dottorblaster, line 9: I think you can inline this and avoid testing the implementation detail using an `{ exact: false }` in expectations. Otherwise having a more fine grained testing ends up with you doing `formatPackage(pkg) == formatPackage(pkg)`, which will always be true.

assets/js/lib/test-utils/factories/advisoryErrata.js
```diff
@@ -2,6 +2,14 @@ import { faker } from '@faker-js/faker';
 import { Factory } from 'fishery';
 import { advisoryType } from './relevantPatches';
 
+const affectedPackageFactory = Factory.define(() => ({
+  name: faker.animal.cat().toLowerCase(),
```
dottorblaster, line 6:
```suggestion
const affectedPackageFactory = Factory.define(({ sequence }) => ({
  name: `${faker.animal.cat().toLowerCase()}${sequence}`,
```

Using sequences is sort of mandatory if you build lists out of them and want to use `screen.getByText`, even with `{ exact: false }`, otherwise we risk duplicated generated texts and flaky tests :+1:

dottorblaster (changes requested):
Some preliminary changes but in general good job, thank you for requesting feedback earlier!

assets/js/pages/AdvisoryDetails/AdvisoryDetails.jsx
janvhs: Yeah I was thinking the same thing. I'll do that :D

assets/js/lib/test-utils/factories/advisoryErrata.js
janvhs: Omg yeah it's supposed to be an array anyway... I think the time got to me. Thanks for pointing this out ^^

dottorblaster (approved):
LGTM

PR #2766, opened 2024-07-10
Angi architecture frontend

arbulu89:
# Description

Add architecture type visualization in the frontend, between `classic` and `angi`. Only applicable for HANA clusters.
Basically, a simple icon.
- info icon if classic
- green star icon if angi

The icon is displayed in cluster details view, checks selection view and checks execution results view.
Besides that, the cluster checks selection filters checks with the current architecture (if the cluster is angi, only bring checks with this arch type, filtering out classic specific checks). Even though we don't have such a distinction by now...

angi-icon

## How was this tested?
UT and storybook changes added

### DIFF
--- a/assets/js/pages/ClusterDetails/HanaClusterDetails.test.jsx
+++ b/assets/js/pages/ClusterDetails/HanaClusterDetails.test.jsx
@@ -466,6 +466,53 @@ describe('HanaClusterDetails component', () => {
     }
   );
 
+  it.each([
+    { arch: 'angi', tooltip: 'Angi architecture' },
+    { arch: 'classic', tooltip: 'Classic architecture' },
+  ])(
+    'should show cluster type with $arch architecture',
+    async ({ arch, tooltip }) => {
+      const user = userEvent.setup();
+
+      const {
+        clusterID,
+        clusterName,
+        cib_last_written: cibLastWritten,
+        type: clusterType,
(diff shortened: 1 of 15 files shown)


angelabriel:
I like the implementation. Green Star for Angi :smiley:

assets/js/pages/ClusterDetails/HanaClusterDetails.test.jsx
```diff
@@ -466,6 +466,53 @@ describe('HanaClusterDetails component', () => {
     }
   );
 
+  it.each([
+    { arch: 'angi', tooltip: 'Angi architecture' },
+    { arch: 'classic', tooltip: 'Classic architecture' },
+  ])(
+    'should show cluster type with $arch architecture',
```
EMaksy, line 473: why $ in arch ?  :)

EMaksy (approved):
Well done man, i guess there is only one small typo, but besides this lgtm!

assets/js/pages/ClusterDetails/HanaClusterDetails.test.jsx
arbulu89: To interpolate the value of `arch` value coming in the scenarios

CDimonaco (approved):
lgtm

:chef_kiss:

PR #2772, opened 2024-07-12
Add e2e test for settings view to test abilities/permissions

EMaksy:
# Description

This pr i a follow up pr to https://github.com/trento-project/web/pull/2760. 
Adding cypress e2e test to do what:
- Test if a user has no abilities/permissions the settings view has disabled buttons (all besides test connection button)
- Test if a user has proper abilities/permissions the settings buttons are enabled
- Fix a missed only 

- [x] **DONE**

## How was this tested?

Added missing cypress e2e test for settings view. 
Tests were already done in  [previous pr](https://github.com/trento-project/web/pull/2760.)

- [x] **DONE**

No docs update required
- [x] **DONE**

### DIFF
--- a/test/e2e/cypress/e2e/settings.cy.js
+++ b/test/e2e/cypress/e2e/settings.cy.js
@@ -807,7 +807,8 @@ context('Settings page', () => {
       });
     });
   });
-  describe('Forbidden action', () => {
+
+  describe('Forbidden actions', () => {
     const password = 'password';
 
     beforeEach(() => {
--- a/test/e2e/cypress/e2e/settings.cy.js
+++ b/test/e2e/cypress/e2e/settings.cy.js
@@ -838,18 +838,16 @@ context('Settings page', () => {
         .and('be.enabled');
       // SUSE Manager config settings button
       cy.contains('button', 'Test Connection').should('be.enabled');
-      cy.get(
-        '.py-4 > .container > :nth-child(1) > .float-right > .text-jungle-green-500'
-      )
-        .should('be.enabled')
-        .and('contain.text', 'Edit Settings');
+      cy.contains('h2', 'SUSE Manager Config')
+        .next()
+        .contains('button', 'Edit Settings')
+        .should('be.enabled');
       cy.contains('button', 'Clear Settings').should('be.enabled');
       // Activity Logs settings button
-      cy.get(
-        ':nth-child(4) > .container > :nth-child(1) > .float-right > .bg-white'
-      )
-        .should('be.enabled')
-        .and('contain.text', 'Edit Settings');
+      cy.contains('h2', 'Activity Logs')
+        .next()
+        .contains('button', 'Edit Settings')
+        .should('be.enabled');
     });
 
     it('should disable settings buttons if the user has no abilities', () => {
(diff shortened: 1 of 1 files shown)


test/e2e/cypress/e2e/settings.cy.js
```diff
@@ -806,4 +807,73 @@ context('Settings page', () => {
...
+        cy.login(user.username, password);
+      });
+      cy.visit(`/settings`);
+      // API Key settings button
+      cy.contains('button', 'Generate Key')
+        .should('be.visible')
+        .and('be.enabled');
+      // SUSE Manager config settings button
+      cy.contains('button', 'Test Connection').should('be.enabled');
+      cy.get(
```
arbulu89, line 840: Maybe there is a nice way to find the buttons, but well hehe

arbulu89 (approved):
Nice @EMaksy 
LGTM

test/e2e/cypress/e2e/settings.cy.js
EMaksy: Yeah, I know it is ugly, that's why I put some comments. 
As we have two "Edit Settings" buttons, I thought taking the selector directly with cypress to be precise won't harm :) 

Btw just an idea, we could add an aria label prop for the button component, with this we could have a very readable way to find elements like the two "Edit Settings" buttons

arbulu89: Maybe you could find the by the h2 header. Something like: (not tested):
```
cy.contains('h2', 'SUSE Manager Config').next().contains('button', 'Edit Settings').should('be.enabled'):
```
Give a try if you want. Anyway, if the selectors find what they need to find, the tests does what it is supposed to do, so I'm fine as well

test/e2e/cypress/e2e/settings.cy.js
```diff
@@ -806,4 +807,73 @@ context('Settings page', () => {
...
+        cy.createUserWithAbilities(user, userAbilites);
+        cy.login(user.username, password);
+      });
+      cy.visit(`/settings`);
+      // API Key settings button
+      cy.contains('button', 'Generate Key')
+        .should('be.visible')
+        .and('be.enabled');
+      // SUSE Manager config settings button
+      cy.contains('button', 'Test Connection').should('be.enabled');
```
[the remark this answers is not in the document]
EMaksy: I wanted to test just every button on this page :) 
You are right it is not a must, I can remove it, but I would prefer to keep it.

test/e2e/cypress/e2e/settings.cy.js
```diff
@@ -806,4 +807,71 @@ context('Settings page', () => {
...
+        .next()
+        .contains('button', 'Edit Settings')
+        .should('be.enabled');
+      cy.contains('button', 'Clear Settings').should('be.enabled');
+      // Activity Logs settings button
+      cy.contains('h2', 'Activity Logs')
+        .next()
+        .contains('button', 'Edit Settings')
+        .should('be.enabled');
+      cy.contains('button', 'Clear Settings').should('be.enabled');
```
arbulu89, line 851: `cy.contains('button', 'Clear Settings').should('be.enabled');` this line is duplicated

test/e2e/cypress/e2e/settings.cy.js
EMaksy: Yeah, I was playing with cypress [eq](https://docs.cypress.io/api/commands/eq), [first ](https://docs.cypress.io/api/commands/first) and [last ](https://docs.cypress.io/api/commands/last) but it could not find the second edit settings button so meh. 
Your suggestion is much more readable, I would stick with it :+1: 
Thanks for the idea!

PR #2773, opened 2024-07-12
Expose id as part of an activity log entry

nelsonkopliku:
# Description

Exposes the activity log id as part of the api response.

## How was this tested?

automated tests


lib/trento_web/views/v1/activity_log_view.ex
```diff
@@ -2,11 +2,12 @@ defmodule TrentoWeb.V1.ActivityLogView do
...
 
   def render("activity_log.json", %{activity_log: entries}) do
-    render_many(entries, __MODULE__, "activity_log_entry.json", as: :entries)
+    render_many(entries, __MODULE__, "activity_log_entry.json", as: :activity_log_entry)
   end
 
-  def render("activity_log_entry.json", %{entries: entry}) do
+  def render("activity_log_entry.json", %{activity_log_entry: entry}) do
     %{
+      id: entry.id,
```
gagandeepb, line 10: This exposes an internal db identifier to the outside; what are your thoughts about reversible obfuscation/mapping of some sort (hashids, etc.)?

nelsonkopliku: I think we'd be replacing a uuid with its hashed representation.
I am not sure I see the benefit/need for this extra transformation step, currently.

gagandeepb (approved):

PR #2780, opened 2024-07-15
Open api spex object refactor

CDimonaco:
# Description

This PR change the use of OpenApiSpex macro, to include `struct?:false` as parameter.

This allow us to better pattern match into the controller action functions and also allow us to not decode the body_params into a map, and pass them directly to ecto, as phoenix convention.

The validation of the paramters and openapi definition remains untouched.

Other small controller fixes added.

Tested.


lib/trento_web/controllers/v1/suma_credentials_controller.ex
```diff
@@ -116,7 +112,4 @@ defmodule TrentoWeb.V1.SUMACredentialsController do
   end
 
   def get_policy_resource(_), do: Trento.SoftwareUpdates.Settings
-
-  defp decode_body(body) when is_struct(body), do: Map.from_struct(body)
```
arbulu89, line 120: Is this the unique real place where we are forced to decode to map?
As far as I see, we can get the remaining fields exactly the same way without the `struct?` field

arbulu89 (approved):
Nice!
I'm not 100% what we "solve" with this exactly, except one `decode` function, as everything else remains untouched in essence

It's a pity that there is not any way to set `struct?` default value as `false` somehow.
Otherwise, we need to remember always to add the value, and it looks like something we can forget pretty easily. I hope we don't :sweat_smile: 

Maybe we could have our own macro for that

PR #2786, opened 2024-07-16
Activity log pagination and filtering

gagandeepb:
# Description

Implements pagination and filtering for the activity log. 

- [x] **DONE**

## How was this tested?

Unit tests
- [x] **DONE**

- [x] **DONE**
No doc changes needed, yet.


lib/trento_web/controllers/v1/activity_log_controller.ex
```diff
@@ -11,15 +11,110 @@ defmodule TrentoWeb.V1.ActivityLogController do
...
+        pagination: meta
       })
     end
   end
+
+  @default_params %{
+    first: 25,
+    order_by: [:inserted_at],
+    order_direction: [:desc]
+  }
```
balanza, line 79: **question:** What's the need for these defaults, as we already have defaults in the schema? (I tried to remove them and the test are still passing, though)‏

gagandeepb: I see one failing test with these default params set to `%{}`. Let's sync about it.

lib/trento_web/controllers/v1/activity_log_controller.ex
```diff
@@ -11,15 +11,110 @@ defmodule TrentoWeb.V1.ActivityLogController do
...
     ]
 
-  def get_activity_log(conn, _) do
-    with activity_log_entries <- ActivityLog.list_activity_log() do
+  def get_activity_log(conn, params) do
+    parsed_params =
+      case params do
+        [] -> %{}
+        params -> parse_params(params)
+      end
```
balanza, line 65: **issue:** I see some downsides in doing the parsing step in the controller:
* we leak some implementation details (Flop's filters shape) to the integration point (the controller);
* we handle 2 different interfaces for the same problem (the API of the controller and the signature of the context);
* we are _invited_ to use Flop's internals in our test (which we should avoid, our tests should use as plain data as possible).

I think we can have a better result by moving the parsing step into the context: 
* our integration with Flop will be segregated in the core domain without leaking into the integration layer;
* we can test filters and pagination directly against the context (no need to use an intermediate interface in tests).

What do you think?


‏

lib/trento_web/controllers/v1/activity_log_controller.ex
```diff
@@ -11,15 +11,110 @@ defmodule TrentoWeb.V1.ActivityLogController do
...
+        in: :query,
+        schema: %OpenApiSpex.Schema{type: :string},
+        required: false
+      ],
+      to_date: [
+        in: :query,
+        schema: %OpenApiSpex.Schema{type: :string},
+        required: false
+      ],
+      actor: [
```
balanza, line 45: **nitpick:** ‏Maybe `actors`, as it's a list? No strong opinions, though.

lib/trento/activity_logging/activity_log.ex
```diff
@@ -7,6 +7,20 @@ defmodule Trento.ActivityLog.ActivityLog do
 
   @type t() :: %__MODULE__{}
 
+  @derive {
+    Flop.Schema,
+    filterable: [:type, :actor, :inserted_at],
+    sortable: [:type, :actor, :inserted_at],
```
balanza, line 13: **thought:** Being the log a stream of entries, I don't see the value in ordering for some criteria different from time.

It's a concern about the product, we might want to raise the issue in the product design phase. 

Anyway, the implementation itself is fine.


‏

lib/trento/activity_logging/activity_log.ex
```diff
@@ -7,6 +7,20 @@ defmodule Trento.ActivityLog.ActivityLog do
...
+    max_limit: 100,
+    default_limit: 25,
+    default_order: %{
+      order_by: [:inserted_at],
+      order_directions: [:desc]
+    },
+    pagination_types: [:first, :last],
```
balanza, line 20: **praise:** ‏I think first/last pagination is the only viable solution we have. Given the nature of the list, which is frequently populated (logs arrive at any time), limit/offset pagination would give some inconsistencies in the results.

lib/trento/activity_log.ex
```diff
@@ -44,14 +44,24 @@ defmodule Trento.ActivityLog do
...
   end
 
+  @spec list_activity_log(map()) :: {:ok, list(ActivityLog.t()), Flop.Meta.t()}
+  def list_activity_log(params) do
+    case Flop.validate_and_run(ActivityLog, params, for: ActivityLog) do
+      {:ok, {activity_log_entries, meta}} ->
+        {:ok, activity_log_entries, meta}
+
+      error ->
+        log_error(error, "Error while paginating activity log")
```
balanza, line 61: **issue:** ‏Aren't we hiding the error here?

lib/trento_web/openapi/v1/schema/activity_log.ex
```diff
@@ -8,35 +8,44 @@ defmodule TrentoWeb.OpenApi.V1.Schema.ActivityLog do
...
+              },
+              metadata: %Schema{
+                type: :object
+              },
+              occurred_on: %Schema{
+                type: :string,
+                description: "Timestamp upon Activity Log entry insertion."
+              }
+            },
+            required: [:type, :actor, :metadata, :occurred_on]
```
balanza, line 41: **suggestion:** ‏Maybe it's an opportunity to extract this definition?

lib/trento/activity_log.ex
```diff
@@ -44,14 +44,24 @@ defmodule Trento.ActivityLog do
 
   @spec list_activity_log() :: list(ActivityLog.t())
   def list_activity_log do
-    # This will be made filterable/paginatable in a later PR
     query =
       from activity in ActivityLog,
         order_by: [desc: activity.inserted_at]
 
     Repo.all(query)
   end
```
balanza, line 52: **question:** Shouldn't this be removed?‏

lib/trento_web/controllers/v1/activity_log_controller.ex
```diff
@@ -11,15 +11,110 @@ defmodule TrentoWeb.V1.ActivityLogController do
...
+  defp handle_query_params(:to_date, to_date),
+    do: {:filters, %{field: :inserted_at, op: :>=, value: to_date}}
+
+  defp handle_query_params(:actor, actor),
+    do: {:filters, %{field: :actor, op: :ilike_or, value: actor}}
+
+  defp handle_query_params(:type, type),
+    do: {:filters, %{field: :type, op: :ilike_or, value: type}}
+
+  defp handle_query_params(k, v), do: {k, v}
```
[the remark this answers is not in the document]
gagandeepb: Thanks, added.

lib/trento_web/controllers/v1/activity_log_controller.ex
gagandeepb: I am fine with it as it is, since it can be one or more in number.

lib/trento_web/controllers/v1/activity_log_controller.ex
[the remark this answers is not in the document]
gagandeepb: Removed.

lib/trento_web/openapi/v1/schema/activity_log.ex
```diff
@@ -8,6 +8,21 @@ defmodule TrentoWeb.OpenApi.V1.Schema.ActivityLog do
...
+        pagination: %Schema{
+          type: :object
+        }
+      },
+      required: [:data, :pagination]
+    },
+    struct?: false
+  )
+
+  defmodule ActivityLogEntry do
```
nelsonkopliku, line 23: I guess this module needs to be defined before its usage.

test/trento_web/views/v1/activity_log_view_test.exs
```diff
@@ -1,49 +1,11 @@
...
   use ExUnit.Case
 
-  import Phoenix.View
-
   import Trento.Factory
 
   alias Trento.ActivityLog.ActivityLog
   alias TrentoWeb.V1.ActivityLogView
 
-  test "should render activity_log.json" do
```
nelsonkopliku, line 11: Maybe worth keeping this test so that we also cover pagination, or we add a pagination specific test.

lib/trento_web/openapi/v1/schema/activity_log.ex
```diff
@@ -35,9 +49,8 @@ defmodule TrentoWeb.OpenApi.V1.Schema.ActivityLog do
             description: "Timestamp upon Activity Log entry insertion."
           }
         },
-        required: [:id, :type, :actor, :metadata, :occurred_on]
+        required: [:type, :actor, :metadata, :occurred_on]
```
nelsonkopliku, line 52: What made the id become not required in the current schema? :eyes:

lib/trento/activity_log.ex
```diff
@@ -42,14 +40,51 @@ defmodule Trento.ActivityLog do
...
+
+  # ''&& false' is a workaround until we reach OTP 27 that allows doc tag for private functions;
+  # we get a compile warning without this with OTP 26
+  @doc """
+       Parses the query parameters and returns a map with the parsed values as expected by the Flop library.
+       Some parameters are recognized by Flop and are used as is (example: last, first, after, before);
+       some other parameters are used to build filters with custom operator logic (example: from_date, to_date, actor, type).
+       ## Examples
+             iex> parse_params([{:from_date, "2021-01-01"}, {:to_date, "2021-01-31"}, last: 10])
+             %{filters: %{from_date: "2021-01-01", to_date: "2021-01-31"}, last: 10}
```
nelsonkopliku, line 70: Maybe a typo
```suggestion
             %{filters: %{from_date: "2021-01-01", to_date: "2021-01-31"}, last: 10}
             %{filters: %{inserted_at: %{value: "2021-01-31", op: :>=, field: :inserted_at}}, last: 10}
```

nelsonkopliku: @gagandeepb @balanza now that I am looking at this again I noticed that we might be missing something.

if we have both `from_date` and `to_date` as an input, we should expect `inserted_at` to be between those dates

```elixir
# input
# [{:from_date, "2021-01-01"}, {:to_date, "2021-01-31"}]
[
  %{
    filters: [
      %{inserted_at: %{value: "2021-01-01", op: :>=, field: :inserted_at}},
      %{inserted_at: %{value: "2021-01-31", op: :<=, field: :inserted_at}}
    ]
  }
]
```

I guess in the reduction we are not considering multiple filter operations on the same field.

balanza: @nelsonkopliku so you expect an output like this?

```elixir
[
  %{
    filters: [
      %{inserted_at: %{value: ["2021-01-01", "2021-01-31"], op: :between field: :inserted_at}},
    ]
  }
]
```
(I didn't check Flop docs for the exact implementation of the operator)

nelsonkopliku: I did not check Flop's doc either :see_no_evil: but yes: if we provide `from_date` and `to_date` I would expect `inserted_at` to be between those dates.

What we currently generate is a different filter, lacking part of the intent
```elixir
%{
  filters: %{inserted_at: %{value: "2021-01-31", op: :>=, field: :inserted_at}}
}
```

balanza: I see, there is a bug.

Basically because the `filters` map's key is the field name. Good catch.

Let's check Flop docs for how it expects the filter map to be.

lib/trento/activity_log.ex
```diff
@@ -42,14 +40,51 @@ defmodule Trento.ActivityLog do
...
+    |> Enum.map(fn
+      {:from_date, v} -> {:filters, %{field: :inserted_at, op: :<=, value: v}}
+      {:to_date, v} -> {:filters, %{field: :inserted_at, op: :>=, value: v}}
+      {:actor, v} -> {:filters, %{field: :actor, op: :ilike_or, value: v}}
+      {:type, v} -> {:filters, %{field: :type, op: :ilike_or, value: v}}
+      param -> param
+    end)
+    |> Enum.reduce(%{filters: %{}}, fn
+      {:filters, filter}, acc ->
+        Map.put(acc, :filters, Map.put(acc.filters, filter.field, filter))
```
balanza, line 83: **issue:** This introduced a bug. Using filters as a map only allows for one filter per each field. If we have multiple filters for the same field (like dates for examples) they won't be considered. (my bad, the refactor was my suggestion).

Using filters like collection it would solve the issue.

```suggestion
    |> Enum.reduce(%{filters: []}, fn
      {:filters, filter}, acc ->
        Map.put(acc, :filters, [filter | acc.filters])
```

However, by changing the shape of filters I expect tests to fail, but they don't

‏

balanza (changes requested):

lib/trento/activity_log.ex
balanza: Traced the bug in https://github.com/trento-project/web/pull/2786#discussion_r1686329508

We may have some concerns about tests (they should have spoiled the bug)

gagandeepb: I'll add some more tests.

lib/trento_web/openapi/v1/schema/activity_log.ex
```diff
@@ -16,8 +15,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.ActivityLog do
         properties: %{
           id: %Schema{
             type: :string,
-            description: "Identifier of Activity Log entry.",
-            format: :uuid
```
nelsonkopliku, line 20: let's keep the `format : :uuid`

lib/trento_web/openapi/v1/schema/activity_log.ex
```diff
@@ -37,7 +35,26 @@ defmodule TrentoWeb.OpenApi.V1.Schema.ActivityLog do
...
+  defmodule ActivityLog do
+    @moduledoc false
+
+    OpenApiSpex.schema(
+      %{
+        title: "ActivityLog",
+        description: "Activity Log for the current installation.",
+        type: :object,
+        properties: %{
+          data: ActivityLogEntry,
```
nelsonkopliku, line 50: `ActivityLogEntry` should either be `ActivityLogEntries` or just the whole `ActivityLog`

lib/trento_web/views/v1/activity_log_view.ex
```diff
@@ -15,4 +18,29 @@ defmodule TrentoWeb.V1.ActivityLogView do
       occurred_on: entry.inserted_at
     }
   end
+
+  def render("pagination.json", %{pagination: pagination}) when is_map(pagination) do
```
nelsonkopliku, line 22: If the guard is not strictly necessary, let's remove it.
```suggestion
  def render("pagination.json", %{pagination: pagination}) do
```

test/trento/activity_log_test.exs
```diff
@@ -191,14 +191,70 @@ defmodule Trento.ActivityLogTest do
...
+      # default order is by inserted_at timestamp
+      all_logs_sorted =
+        all_logs |> Enum.sort_by(fn entry -> entry.inserted_at end, :desc) |> Enum.take(25)
+
+      assert logs == all_logs_sorted
+    end
+
+    test "should return earliest 5 paginated activity log entries" do
+      all_logs = insert_list(100, :activity_log_entry)
+      params = %{last: 5, order_by: [:inserted_at], order_directions: [:desc]}
```
nelsonkopliku, line 223: What about adding a scenario for the `first` counterpart?

test/trento/activity_log_test.exs
```diff
@@ -191,14 +191,70 @@ defmodule Trento.ActivityLogTest do
...
+    end
+
+    test "should return latest paginated activity log entries number 26 to 50 (desc)" do
+      all_logs = insert_list(100, :activity_log_entry)
+      params = %{}
+
+      {:ok, logs, meta} = ActivityLog.list_activity_log(params)
+
+      assert length(logs) == 25
+      new_params = %{first: 25, after: meta.end_cursor}
```
nelsonkopliku, line 245: what about a scenario for `last+before` combination?

nelsonkopliku (commented):
Noice, great job!

Only thing I can say, besides maybe adding even more test cases, is that we could omit from the `pagination` entry the following:
- `filters`
- `order_by`
- `order_directions`

It's a piece of state that a client (our frontend mainly so far) could keep on its own, as it is responsible to making the requests (in any case we could add them at a later stage, if needed).

We need to work around the pipeline failure on api bc check :see_no_evil:

nelsonkopliku (approved):
LGTM. Thanks!

lib/trento/activity_log.ex
```diff
@@ -62,6 +52,61 @@ defmodule Trento.ActivityLog do
...
+       }
+       """ && false
+  @spec parse_params(map()) :: map()
+  def parse_params(query_params) when query_params == %{} do
+    # Implies
+    # %{first: 25, order_by: [:inserted_at], order_directions: [:desc]}
+    %{}
+  end
+
+  def parse_params(query_params) do
```
balanza, line 92: **suggestion:** ‏It should be private

```suggestion
  defp parse_params(query_params) do
```
‏

lib/trento/activity_log.ex
```diff
@@ -62,6 +52,61 @@ defmodule Trento.ActivityLog do
...
+          last: 10
+       }
+       """ && false
+  @spec parse_params(map()) :: map()
+  def parse_params(query_params) when query_params == %{} do
+    # Implies
+    # %{first: 25, order_by: [:inserted_at], order_directions: [:desc]}
+    %{}
+  end
+
```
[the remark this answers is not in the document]
gagandeepb: Intentional choice, since with your previous suggestion I get a non empty map that is not desirable. The non-empty map is a no-op, but not a priority to add a test for at this time.

lib/trento_web/views/v1/activity_log_view.ex
```diff
@@ -15,4 +18,23 @@ defmodule TrentoWeb.V1.ActivityLogView do
...
+      flop: %{
+        first: first,
+        last: last
+      }
+    } =
+      pagination
+
+    %{
+      start_cursor: start_cursor,
+      end_cursor: end_cursor,
```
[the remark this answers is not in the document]
gagandeepb: It's fine as it is. I prefer the explicit naming with a clearer naming.

lib/trento_web/controllers/fallback_controller.ex
```diff
@@ -3,6 +3,13 @@ defmodule TrentoWeb.FallbackController do
 
   alias TrentoWeb.ErrorView
 
+  def call(conn, {:error, :activity_log_fetch_error}) do
+    conn
+    |> put_status(:bad_request)
+    |> put_view(ErrorView)
+    |> render(:"400", reason: "Activity Log fetch error.")
```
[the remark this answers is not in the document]
gagandeepb: This has been decided elsewhere. It's a client error. Which implies a 4XX.

test/trento_web/views/v1/activity_log_view_test.exs
```diff
@@ -60,7 +22,7 @@ defmodule TrentoWeb.V1.ActivityLogViewTest do
              metadata: ^metadata,
              occurred_on: ^inserted_at
            } =
-             render(ActivityLogView, "activity_log_entry.json", %{
+             ActivityLogView.render("activity_log_entry.json", %{
```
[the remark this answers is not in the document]
gagandeepb: I think it is more idiomatic the way it is. With what you are suggesting, it would require an additional Phoenix View import and using that as a dispatching mechanism adds more indirection than is strictly needed.

lib/trento_web/controllers/fallback_controller.ex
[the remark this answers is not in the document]
nelsonkopliku: AFAIU `Flop.validate_and_run` returns an error if the combination of provided filters is incorrect.
I am not sure whether there are other reasons that might make that function call fail.

Let's keep an eye open on this and improve if we find shortcomings.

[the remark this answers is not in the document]
balanza: Yes, the `validate` part returns errors on invalid parameters provided (client error). The `run` part is a proxy over the `Ecto` query, thus inheriting its failure cases (server error).

As I told you before it's just nitpicking, I'm fine with 400.

lib/trento_web/views/v1/activity_log_view.ex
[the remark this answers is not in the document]
nelsonkopliku: I was initially also thinking about possible different naming.

Then I learned that `start_cursor` is the hash of the first element of the returned list, while - unsurprisingly - `end_cursor` is the hash of the last element in the returned list.
I also learned is that those hashes do not carry any _direction semantics_ (as in back/forth) and a client could freely use any of those hashes as part of an `first+after` or a `last+before` filter to retrieve what it wants.

`prev` and `next` are very nice at a first glance, however I am wondering if they give a too specific direction related meaning to items unbound from direction.

my2cents :sweat_smile: 

(just to give an alternative to `start_cursor` and `end_cursor`, we might consider something like `first_item`/`last_item`, if it helps. I have no strong preferences)

test/trento_web/views/v1/activity_log_view_test.exs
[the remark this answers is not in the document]
balanza: The most important thing is we keep the codebase coherent and consistent. The same instructions must be written with the same shape so that we avoid having a scattered codebase in which each developer feels free to use his/her taste regardless of the others.

If it brings an advantage we can refactor all the occurrences. I advocate we open a PR on that.
‏

balanza (approved):

test/trento_web/views/v1/activity_log_view_test.exs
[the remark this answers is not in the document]
gagandeepb: Consistency is always nice. Not so sure if this introduces any incoherence. Regardless, if you see this as an opportunity for process improvement, please feel free to add a custom Credo consistency check to the CI for this/similar patterns. That would be the way to ensure that a shared state of consistency is evenly followed everywhere.

PR #2787, opened 2024-07-16
Fix navigation form relevant patches to advisory details

janvhs:
This commit fixes the navigation between the pages mentioned above. Previously it would append the path to the end of the current path.

## How was this tested?

~It isn't :D~ I clicked on it lol


dottorblaster (approved):
LGTM

PR #2789, opened 2024-07-16
Fix flickering in `Available Software Updates` panel

jamie-suse:
# Description

This PR fixes an issue in the UI where the _Available Software Updates_ panel was rendering two different _loading_ states at two different times. The result of this was that the users were seeing a flicker as the panel (briefly) moved into a loading state a second time. This created a confusing user experience.

See the image below for the typical sequence of Redux events that would happen during rendering:

software_updates_loading

Note how the panel enters a _loading_ state while retrieving the software updates, the response returns and the software updates are saved. Much later however the network call to retrieve the SUSE Manager settings begins, and sets the second _loading_ state. The settings are then returned, and the panel is set into the _display_ state again. This is a confusing user experience.

This was fixed by removing the superfluous call to retrieve the SUSE Manager settings, as they are not actually relevant to the _Host Details_ page, nor the _Available Software Updates_ panel. Since we know that if we can retrieve the software updates, then obviously the settings are configured correctly. To emphasize; we don't care about the specific SUSE Manager settings, just if we are authorized to access SUSE Manager with our settings.

We differentiate between being unauthorized vs some other network or server error by returning a `422` status code and corresponding error message `"SUSE Manager authentication error."` in the response from Trento's _SUSE Manager Controller_. The frontend can then know that the SUSE Manager settings are not configured correctly and displays the _SUSE Manager is not configured_ state to the user.

## How was this tested?

Units tests updated.

No changes are required for the documentation.

### DIFF
--- a/assets/js/state/sagas/softwareUpdates.js
+++ b/assets/js/state/sagas/softwareUpdates.js
@@ -28,12 +28,15 @@ export function* fetchSoftwareUpdates({ payload: hostID }) {
     yield put(setEmptySoftwareUpdates({ hostID }));
 
     const errorCode = get(error, ['response', 'status']);
+    const { errors } = get(error, ['response', 'data'], []);
+    const suma_unauthorized = errors.some(
+      ({ detail }) => detail === 'SUSE Manager authentication error.'
+    );
 
-    if (errorCode === 403) {
+    if (errorCode === 422 && suma_unauthorized) {
       yield put(setSettingsNotConfigured());
     }
 
-    const errors = get(error, ['response', 'data'], []);
     yield put(setSoftwareUpdatesErrors({ hostID, errors }));
   }
 }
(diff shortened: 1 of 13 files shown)


lib/trento_web/controllers/fallback_controller.ex
```diff
@@ -17,6 +17,13 @@ defmodule TrentoWeb.FallbackController do
     |> render(:"404", reason: "SUSE Manager settings not configured.")
   end
 
+  def call(conn, {:error, :suma_authentication_error}) do
+    conn
+    |> put_status(:forbidden)
+    |> put_view(ErrorView)
+    |> render(:"403", reason: "SUSE Manager authentication error.")
+  end
+
```
dottorblaster, line 26: Since we now rely on 403 for our own permissions system it's better to switch this to 404 or 422, no big deal

dottorblaster (approved):
This is terrific! Thanks for fixing all of this, this is a huge contribution :fire:

assets/js/state/sagas/softwareUpdates.js
```diff
@@ -21,10 +23,20 @@ export function* fetchSoftwareUpdates({ payload: hostID }) {
...
     yield put(setEmptySoftwareUpdates({ hostID }));
 
-    const errors = get(error, ['response', 'data'], []);
+    const errorCode = get(error, ['response', 'status']);
+    const { errors } = get(error, ['response', 'data'], []);
+    const suma_unauthorized = errors.some(
+      ({ detail }) => detail === 'SUSE Manager authentication error.'
+    );
+
+    if (errorCode === 422 && suma_unauthorized) {
```
jamie-suse, line 36: I check the status code here for completeness, though technically we may not need to.

PR #2792, opened 2024-07-17
Schedule activity log cleaning job

balanza:
# Description

Schedule a recurrent job to delete all expired logs, according to the given retention time in settings. 

Fixes # (issue)

Unit test

### DIFF
--- a/lib/trento/activity_log.ex
+++ b/lib/trento/activity_log.ex
@@ -7,6 +7,7 @@ defmodule Trento.ActivityLog do
 
   require Logger
 
+  alias Trento.ActivityLog.RetentionTime
   require Trento.ActivityLog.RetentionPeriodUnit, as: RetentionPeriodUnit
   alias Trento.ActivityLog.ActivityLog
   alias Trento.ActivityLog.Settings
@@ -52,10 +53,42 @@ defmodule Trento.ActivityLog do
     Repo.all(query)
   end
 
+  @spec clear_expired_logs() :: :ok | {:error, any()}
+  def clear_expired_logs do
+    with {:ok, retention_time} <- get_retention_time(),
+         expiration_date <- calculate_expiration_date(retention_time) do
+      delete_logs_before(expiration_date)
+      :ok
+    end
+  end
+
   defp log_error({:error, _} = error, message) do
     Logger.error("#{message}: #{inspect(error)}")
     error
   end
 
   defp log_error(result, _), do: result
+
+  defp get_retention_time do
+    case Trento.ActivityLog.get_settings() do
+      {:ok, settings} -> {:ok, settings.retention_time}
+      {:error, _} = error -> error
+    end
+  end
+
+  defp calculate_expiration_date(%RetentionTime{value: value, unit: :day}),
+    do: DateTime.add(DateTime.utc_now(), -value, :day)
+
+  defp calculate_expiration_date(%RetentionTime{value: value, unit: :week}),
+    do: DateTime.add(DateTime.utc_now(), -value * 7, :day)
+
+  defp calculate_expiration_date(%RetentionTime{value: value, unit: :month}),
+    do: DateTime.add(DateTime.utc_now(), -value * 30, :day)
+
+  defp calculate_expiration_date(%RetentionTime{value: value, unit: :year}),
+    do: DateTime.add(DateTime.utc_now(), -value * 365, :day)
+
+  defp delete_logs_before(%DateTime{} = expiration_date) do
+    Repo.delete_all(from l in ActivityLog, where: l.inserted_at < ^expiration_date)
+  end
 end
--- a/lib/trento/activity_log.ex
+++ b/lib/trento/activity_log.ex
@@ -55,7 +55,7 @@ defmodule Trento.ActivityLog do
 
   @spec clear_expired_logs() :: :ok | {:error, any()}
   def clear_expired_logs do
-    with {:ok, retention_time} <- get_retention_time(),
+    with {:ok, %{retention_time: retention_time}} <- Trento.ActivityLog.get_settings(),
          expiration_date <- calculate_expiration_date(retention_time) do
       delete_logs_before(expiration_date)
       :ok
@@ -68,13 +68,6 @@ defmodule Trento.ActivityLog do
   end
 
   defp log_error(result, _), do: result
-
-  defp get_retention_time do
-    case Trento.ActivityLog.get_settings() do
-      {:ok, settings} -> {:ok, settings.retention_time}
-      {:error, _} = error -> error
-    end
-  end
 
   defp calculate_expiration_date(%RetentionTime{value: value, unit: :day}),
     do: DateTime.add(DateTime.utc_now(), -value, :day)
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -41,7 +41,7 @@
         "@babel/core": "^7.24.9",
         "@babel/plugin-transform-modules-commonjs": "^7.24.7",
         "@babel/preset-env": "^7.24.8",
-        "@babel/preset-react": "^7.24.1",
+        "@babel/preset-react": "^7.24.7",
         "@faker-js/faker": "^8.4.1",
         "@storybook/addon-actions": "^7.6.17",
         "@storybook/addon-essentials": "^7.6.17",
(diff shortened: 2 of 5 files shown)


lib/trento/activity_logging/activity_logger.ex
```diff
@@ -60,4 +74,27 @@ defmodule Trento.ActivityLog.ActivityLogger do
...
   end
+
+  defp get_retention_time do
+    case Repo.one(Settings.base_query()) do
+      %Settings{} = settings -> {:ok, settings.retention_time}
+      nil -> {:error, :activity_log_settings_not_configured}
+    end
+  end
+
+  defp calculate_expiration_date(%RetentionTime{value: value, unit: :day}),
```
nelsonkopliku, line 85: Credo is complaining about pipelining only on function.

Here's an idea of how we could revisit this:
```elixir
  @day 24 * 60 * 60

  defp calculate_expiration_date(retention_time),
    do: DateTime.add(DateTime.utc_now(), -to_seconds(retention_time), :second)

  defp to_seconds(%RetentionTime{value: value, unit: :day}), do: value * @day
  defp to_seconds(%RetentionTime{value: value, unit: :week}), do: value * 7 * @day
  defp to_seconds(%RetentionTime{value: value, unit: :month}), do: value * 30 * @day
  defp to_seconds(%RetentionTime{value: value, unit: :year}), do: value * 365 * @day
```

Side note: we are assuming a month being always `30 days` and a year being always `365 days`.
I am fine with this approximation, just wondering whether worth making it explicit to users.

nelsonkopliku (commented):
Good job!

Besides static code analysis complaints, and what already commented, I would move `clear_expired_logs` (and what comes with it) in the entrypoint context `Trento.ActivityLog`, and leave to `ActivityLogger` the responsibility to deal with actual logging based on the different nature of the activity context.

test/support/factory.ex
```diff
@@ -1036,6 +1038,21 @@ defmodule Trento.Factory do
     }
   end
 
+  def activity_log_entry_factory(attrs) do
```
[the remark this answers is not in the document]
balanza: > What we could do is explicitly set inserted_at to DateTime.utc_now() in current activity_log_entry_factory and let the factories take care of overriding properties for us.

I thought it's better not to assume a default in mocked data, as it may cover the expected default in the schema. If we explicitly set the factory with a default `DateTime.utc_now()`, what if we change the schema not to have defaults for the field?

test/support/factory.ex
```diff
@@ -1021,10 +1021,12 @@ defmodule Trento.Factory do
     })
   end
 
-  def activity_log_settings_factory do
+  def activity_log_settings_factory(attrs) do
+    retention_time = Map.get(attrs, :retention_time, build(:activity_log_retention_time))
```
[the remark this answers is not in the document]
balanza: Thanks, done in https://github.com/trento-project/web/pull/2792/commits/07133341b27b399f01a151b28f60f6dc4ae11555

lib/trento/activity_logging/activity_logger.ex
```diff
@@ -60,4 +74,27 @@ defmodule Trento.ActivityLog.ActivityLogger do
         )
     end
   end
+
+  defp get_retention_time do
```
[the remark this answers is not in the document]
balanza: Great, done in https://github.com/trento-project/web/pull/2792/commits/0850a886e7b2a8f6b0260e8aaaf8f350b5be5219

Thanks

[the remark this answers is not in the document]
nelsonkopliku: I was imagining using `Trento.ActivityLog.get_settings()` directly in the with pipe, but I am fine eitherway.

lib/trento/activity_logging/activity_logger.ex
balanza: The feedback from Credo is pretty straightforward, I fixed it in https://github.com/trento-project/web/pull/2792/commits/a2dc26e7ccaff974e1238595573be50c05d2ad09

gagandeepb: [This function](https://hexdocs.pm/elixir/1.15.7/DateTime.html#add/4) accepts time units as a parameter. I wonder if we can pass the retention time unit to this, and not have to approximate a month as 30 days?

balanza:
@nelsonkopliku I extract your comment because I think it deserves a proper discussion
> Side note: we are assuming a month being always 30 days and a year being always 365 days.
I am fine with this approximation, just wondering whether worth making it explicit to users.

The most unambiguous communication towards our users would be to let them set the retention time in days only so that they do not have to think about any assumptions or internals. That would also remove a bit of complexity from our model. 

Apparently, it's very inconvenient for our users to set _365 days_ instead of _1 year_; this requires us to choose an approximation that brings, by definition, a precision error.

Communication-wise, I see no other solution than a paragraph somewhere in the docs or the UI. I'm of course happy to hear another proposal, though.

Speaking about implementation, we can either approximate by day number or by using [date shifts](https://hexdocs.pm/elixir/1.17.0/Duration.html). In this PR I came out with the easier one, but I'm fine with both approaches (although Durations were introduced in Elixir 1.17)

lib/trento/activity_logging/activity_logger.ex
balanza: Afaik the only unit that intersects our needs is `:day`. We can indeed refactor in this sense, it may be more readable. Good point

balanza: Done in https://github.com/trento-project/web/pull/2792/commits/ab7b9f27a7bd94b926adc60965ecbe0138b9f89e

test/support/factory.ex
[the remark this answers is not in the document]
nelsonkopliku: if we think the approximation of the insertion in the db is not a proper implementation, we can add a different field like `occurred_on` and have the application generate it. I don't have anything against that.

With regards to this specific factory variation, the test passes even if we did not have it. Up to you whether keeping or removing.

nelsonkopliku (approved):
> Speaking about implementation, we can either approximate by day number or by using [date shifts](https://hexdocs.pm/elixir/1.17.0/Duration.html). In this PR I came out with the easier one, but I'm fine with both approaches (although Durations were introduced in Elixir 1.17)

I was indeed about to suggest https://hexdocs.pm/elixir/1.17.2/DateTime.html#shift/3, but we don't have it. Too bad.

For the rest I am fine with the approximation. We can iterate over when needed.

Besides the comment on the factory variation, the only remaining thing is that the test added in `test/trento/activity_logging/activity_log_test.exs` could be part of the already existent test at `test/trento/activity_log_test.exs`

lib/trento/activity_logging/activity_logger.ex
gagandeepb: @balanza : There is also this one that looks promising: https://hexdocs.pm/ecto/Ecto.Query.API.html#datetime_add/3 and has all the units of interest.

test/support/factory.ex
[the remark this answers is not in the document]
balanza: I think that was my misunderstanding of how factories work. I removed the factory, thanks.

lib/trento/activity_log.ex
```diff
@@ -52,10 +53,42 @@ defmodule Trento.ActivityLog do
     Repo.all(query)
   end
 
+  @spec clear_expired_logs() :: :ok | {:error, any()}
+  def clear_expired_logs do
+    with {:ok, retention_time} <- get_retention_time(),
```
nelsonkopliku, line 58: An alternative could be the following
```suggestion
    with {:ok, %{retention_time: retention_time}} <- get_settings(),
```
so we can remove the private `get_retention_time`

nelsonkopliku (approved):
Looking good.

Besides the comment on the opportunity to remove the private `get_retention_time`, the only thing is that the test added in `test/trento/activity_logging/activity_log_test.exs` could be part of the already existent test at `test/trento/activity_log_test.exs`.

Up to you. Good job!

gagandeepb (commented):
Commits need to be signed for merging. Additionally, up to you if you want to address the comments about using Ecto.SQL for time intervals arithmetic.

balanza:
> Commits need to be signed for merging.
Yes, the problem happens when I rebase using the GitHub UI. Now it's fine.

> Additionally, up to you if you want to address the comments about using Ecto.SQL for time intervals arithmetic.
To be honest, for now we are just fine like this. We can refactor it later anyway.

PR #2793, opened 2024-07-17
Add IP addresses netmask discovery and visualization

arbulu89:
# Description
Discover ip addresses netmask and implement visualization.
In order to not overload this small changes in the domain logic (which I have the gut feeling that at some point we will want to upgrade how we deal with network interfaces in general), i have added the logic in the projection itself.

This way, we simply serialize and deserialize ip address+netmask in the policy and projection (basically concat address with netmask).
And the domain and events stay untouched. At the end, changing events and domain is really sensitive, so if i can avoid with this little changes I prefer this way.
Maybe changing `commands/events/domain` is a bit cleaner though.

If you think we should include `netmasks` in the domain, we can of course, but again, I'm not entirely sure if this will be the last time we play with network interfaces...

Based on: https://github.com/trento-project/agent/pull/346

## How was this tested?

Tested with UT and e2e tests updated


test/fixtures/scenarios/single-hana-single-app/e77ab524-c4e4-5381-bb59-5af8639d52a4_host_discovery.json
```diff
@@ -19,6 +19,18 @@
...
+      "128",
+      "16",
+      "24",
+      "24",
+      "64",
+      "24",
+      "24",
+      "64",
+      "64"
+    ],
```
EMaksy, line 33: We have 11 ip_addresses but only 10 netmasks, i guess we need one more :)

EMaksy (approved):
Hey man besides one nitpick everything fine :+1: 

Great job, it just amazes me how much change must happen to add CIDR notation xD

test/fixtures/scenarios/single-hana-single-app/e77ab524-c4e4-5381-bb59-5af8639d52a4_host_discovery.json
arbulu89: damn, really good catch

arbulu89: Fixed!

CDimonaco (approved):

PR #2798, opened 2024-07-18
E2E SUSE Manager integration

dottorblaster:
# Description
This adds the end to end tests for the SUSE Manager integration's overviews

## How was this tested?
It's literally a PR containing tests

### DIFF
--- a/assets/js/state/softwareUpdates.js
+++ b/assets/js/state/softwareUpdates.js
@@ -2,7 +2,7 @@ import { createAction, createSlice } from '@reduxjs/toolkit';
 import { find, get, isEmpty } from 'lodash';
 
 const initialState = {
-  settingsConfigured: false,
+  settingsConfigured: true,
   softwareUpdates: {},
 };
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -204,7 +204,11 @@ config :trento, Trento.SoftwareUpdates.Discovery,
 config :trento, Trento.Infrastructure.SoftwareUpdates.Suma,
   auth: Trento.Infrastructure.SoftwareUpdates.Auth.SumaAuth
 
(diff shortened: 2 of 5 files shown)


assets/js/state/softwareUpdates.js
```diff
@@ -2,7 +2,7 @@ import { createAction, createSlice } from '@reduxjs/toolkit';
 import { find, get, isEmpty } from 'lodash';
 
 const initialState = {
-  settingsConfigured: false,
```
jamie-suse, line 5: Why did this need to be changed? :thinking:

dottorblaster: It's a leftover from a fix I attempted for the same behavior I think you found in #2821, for which you just added additional dispatches in the sagas. I can restore it to `false` if you want

dottorblaster: Yes because if it changes then we need to display a call to action for the user to go configuring the connection details

config/config.exs
```diff
@@ -204,7 +204,11 @@ config :trento, Trento.SoftwareUpdates.Discovery,
 config :trento, Trento.Infrastructure.SoftwareUpdates.Suma,
   auth: Trento.Infrastructure.SoftwareUpdates.Auth.SumaAuth
 
-config :trento, Trento.Infrastructure.SoftwareUpdates.MockSuma, relevant_patches_system_ids: []
+config :trento, Trento.Infrastructure.SoftwareUpdates.MockSuma,
+  relevant_patches_system_ids: [
+    # vmdrbddev01
+    5980
+  ]
```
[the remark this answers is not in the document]
dottorblaster: I think you're right, let's see if we catch any unwanted side effect but I like it

PR #2809, opened 2024-07-22
Flaky test detection & analysis

gagandeepb:
# Description

Provides the `make` targets to check for flaky tests across BE, FE and E2E tests. These `make` targets when invoked with appropriate params, return a mostly complete list of flaky tests, ordered by a computed measure of flakiness.

### How it works

1. **Run phase**: Runs test suite repeatedly (e.g. 100 times), and collecting trace of each run in JUnit format. 
2. **Analyze phase**: Once there are JUnit files collected in one place, running the `analyze-files` target computes measures of flakiness over all executed tests and returns the top N (40 by default) tests.


hack/flaky_tests_analysis/README.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+```
+❯ make PATH-TO-JUNIT-FILES=/tmp analyze-files
+python check_flakes.py --junit-files=/tmp --grouping-option=runs --window-size=5 --window-count=100 --top-n=40
+
+Top 40 flaky tests based on latest window exponential weighted moving average fliprate score
+Elixir.Trento.Infrastructure.Alerting.AlertingTest::test Alerting the configured recipient about crucial facts with email notifications Notify api key will be expired soon --- score: 0.1038
+Elixir.TrentoWeb.V1.ProfileControllerTest::test should not confirm a totp enrollment if totp is already enabled for the user --- score: 0.08546
+Elixir.Trento.Infrastructure.Commanded.EventHandlers.DatabaseDeregistrationEventHandlerTest::test should dispatch DeregisterSapSystem commands when a tenant is removed --- score: 0.08458
```
gagandeepb, line 44: [A recent example from an actual CI run.](https://github.com/trento-project/web/actions/runs/10250239185/job/28355600383?pr=2857#step:8:1822)

hack/flaky_tests_analysis/README.md
```diff
@@ -0,0 +1,84 @@
...
+python check_flakes.py --junit-files=/tmp --grouping-option=runs --window-size=5 --window-count=100 --top-n=40
+
+Top 40 flaky tests based on latest window exponential weighted moving average fliprate score
+SuseManagerSettingsModal component::should try to save all the fields --- score: 0.4945
+SuseManagerSettingsModal component::should attempt saving only what changed --- score: 0.4945
+PatchList::renders the content correctly --- score: 0.1325
+search::should not match not included words --- score: 0.06413
+Table component::should display the header --- score: 0.05145
+Label Component::should display an info tooltip if specified --- score: 0.05001
+UsersPage::should render toast with error message when deleting failed --- score: 0.05001
```
gagandeepb, line 80: [From FE tests on CI.](https://github.com/trento-project/web/actions/runs/10250239185/job/28355537367?pr=2857#step:6:3632)

hack/flaky_tests_analysis/README.md
```diff
@@ -0,0 +1,84 @@
...
+Elixir.Trento.Infrastructure.Alerting.AlertingTest::test Alerting the configured recipient about crucial facts with email notifications Notify api key will be expired soon --- score: 0.1038
+Elixir.TrentoWeb.V1.ProfileControllerTest::test should not confirm a totp enrollment if totp is already enabled for the user --- score: 0.08546
+Elixir.Trento.Infrastructure.Commanded.EventHandlers.DatabaseDeregistrationEventHandlerTest::test should dispatch DeregisterSapSystem commands when a tenant is removed --- score: 0.08458
+Elixir.TrentoWeb.V1.UsersControllerTest::test index lists all users --- score: 0.07423
+Elixir.TrentoWeb.V1.HostControllerTest::test delete should allow the request when the user has cleanup:host ability --- score: 0.07392
+Elixir.TrentoWeb.V1.ProfileControllerTest::test should not confirm a totp enrollment if totp is not valid --- score: 0.05231
+Elixir.Trento.SapSystems.SapSystemTest::test SAP System registration should register and not move an application instance, if the instance number exists but it's in another host and the application is not clustered --- score: 0.05102
```
gagandeepb, line 48: [A more recent failure on main branch.](https://github.com/trento-project/web/actions/runs/10573821516/job/29294084875#step:8:188) cc: @balanza

hack/flaky_tests_analysis/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+###  Setup
+
+`mix test` along with some config that makes the test runner process write to a JUnit file for every run was set up. 
+
+
+### Execution 
+
+```
+❯ make N-TIMES=500 gen-test-data
```
arbulu89, line 13: This is outdated, right?
I mean now we have options in the makefile to run frontend, backend and e2e tests as far as I see,

Maybe this whole file content should be generilzed, as the usage is almost the same in all 3 forms of tests

arbulu89 (approved):
Hey @gagandeepb ,
For me, the approach and the documentation looks great.
I think that the README must be updated, as it is Backend orietend only, and the makefile has been updated since.
Besides that, great job!

PD: As a comment, as we have multiple projects where we might want to have flaky tests deteciton, would it make sense to upload the python script and other utilities to some "tooling" repo (i don't know if we already have one) and make the script download it from there. Just optimization anyway

config/test.exs
```diff
@@ -93,3 +93,14 @@ config :trento, Trento.Infrastructure.SoftwareUpdates.MockSuma,
...
+config :junit_formatter,
+  report_file: "report_file_test_#{DateTime.now!("Etc/UTC")}.xml",
+  report_dir: "/tmp",
+  print_report_file: true,
+  prepend_project_name?: true,
+  include_filename?: true
+
+config :trento,
+       :flaky_tests_detection,
+       enabled?: System.get_env("WRITE_JUNIT") == "1" || false
```
dottorblaster, line 106:
```suggestion
       enabled?: System.get_env("WRITE_JUNIT") == "1"
```

Setting a default flag for a boolean option is redundant as it will be `false` in any case :smile:

hack/flaky_tests_analysis/Makefile
```diff
@@ -0,0 +1,52 @@
...
+			npx cypress run --reporter junit --reporter-options "mochaFile=$(PATH-TO-JUNIT-FILES)/trento-e2e-junit-[hash]-$(ENV_TS).xml" $(RUN_ONLY_E2E_OPTION);\
+		done 
+
+analyze-files:
+	python check_flakes.py --junit-files=$(PATH-TO-JUNIT-FILES) --grouping-option=runs --window-size=$(WINDOW-SIZE) --window-count=$(WINDOW-COUNT) --top-n=$(TOP-N)
+
+install-deps:
+	pip install -r ./requirements.txt
+
+alt-install-deps:
```
dottorblaster, line 50:
```suggestion
nix-install-deps:
```

dottorblaster (commented):
Nice work, as @arbulu89 said in the future I'd move something like this in a separate repo.

Also, I'm not that proficient with Python but you could leverage something like a virtual env to avoid just making `pip` install things all over the place.

gagandeepb:
Thanks @dottorblaster . Yes, I agree, regarding the python script, it is already available as pip package as well. I can also look into using something Poetry to avoid global installation via pip, in the near future. I did briefly look into using poetry for this very purpose, however I ran into some issues with some native python deps (e.g. numpy).

nelsonkopliku (approved):
Not much to add to what already said. Well done!

Let's consider whether we want to move the detection capability to its own place before/after introducing it in the CI for the current repo.

gagandeepb:
@dottorblaster @arbulu89 @nelsonkopliku I have removed the script and am now using it as an external pip package installed inside a venv. A similar approach can potentially be used in other projects/repos that would like similar detection capabilities.

PR #2814, opened 2024-07-23
Fix process manager rename

arbulu89:
# Description

Fix issue happening when trento is upgraded from version 2.2.0 to 2.3.0 as the deregistration process manager name was updated, but the "superseding" not handled, so the old module is not found.

```
> ** (ArgumentError) errors were found at the given arguments:
> 
>   * 1st argument: not an already existing atom
> 
>     :erlang.binary_to_existing_atom("Elixir.Trento.DeregistrationProcessManager", :utf8)
>     (trento 2.3.1) lib/trento/support/jsonb_serializer.ex:35: Trento.Support.JsonbSerializer.deserialize/2
>     (eventstore 1.4.4) lib/event_store/snapshots/snapshot_data.ex:34: EventStore.Snapshots.SnapshotData.deserialize/2
>     (eventstore 1.4.4) lib/event_store/snapshots/snapshotter.ex:14: EventStore.Snapshots.Snapshotter.read_snapshot/3
```

## How was this tested?
Regression test with old `Trento.DeregistrationProcessManager` process manager entry. When new events affecting this PM entry are received, the PM snapshot is loaded and updated by the new one automaticlly.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -19,8 +19,15 @@ env:
 
 jobs:
   elixir-deps:
-    name: Elixir dependencies
+    name: Elixir ${{ matrix.mix_env }} dependencies
     runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - mix_env: dev
+          - mix_env: test
+    env:
+      MIX_ENV: ${{ matrix.mix_env }}
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.12.1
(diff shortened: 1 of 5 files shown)


nelsonkopliku (approved):
Besides being initially confused by `LegacyModule.supersede`, which might not be immediately communicating that the `LegacyModule` is actually _being superseded by a new one_, LGTM. Thanks.

CDimonaco (approved):
LGTM

arbulu89:
> Besides being initially confused by `LegacyModule.supersede`, which might not be immediately communicating that the `LegacyModule` is actually _being superseded by a new one_, LGTM. Thanks.

Ok, I will rename it to `superseded_by`

PR #2820, opened 2024-07-25
 Discover SAP system running on a JAVA stack

EMaksy:
# Description

This PR will enable the discovery of a SAP system running on a JAVA stack. A Java sap system is discovered and displayed in the console in the SAP systems overview.

This pr adds an Type field in the Sap systems overview view so the user can directly see if the discovered system is running on ABAP, JAVA or in a mixed environment.

**Note:** This pr contains a lot of almost duplicated back-end tests, this was needed to cover the new check for a sap system. 
Previously we checked if an sap system runs on abap, now we check for abap and/or java.

## Preview from Storybook
image

- [ ] **DONE**

## How was this tested?

- Added fronted test
- Fixed cypress e2e test (New e2e tests will be added in a second pr as this pr get's a little bit to big)
- Updated existing and added new back-end sap systems tests.

- [ ] **DONE**

No updates needed at this point.
- [ ] **DONE**


assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
```diff
@@ -80,13 +84,93 @@ describe('SapSystemsOverviews component', () => {
...
+        />
+      );
+      const rows = screen.getByRole('table').querySelectorAll('tbody > tr');
+      const firstSapSystem = rows[0];
+      const secondSapSystem = rows[2];
+      const thirdSapSystem = rows[4];
+      const fourthSapSystem = rows[6];
+      const fifthSapSystem = rows[8];
+
+      expect(firstSapSystem.querySelector('td:nth-child(5)')).toHaveTextContent(
```
arbulu89, line 154: Could we have this `expect`s in some sort of loop.
I mean, you are doing the same operation 5 times in a row, just suming the index twice or so

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
```diff
@@ -80,13 +84,93 @@ describe('SapSystemsOverviews component', () => {
...
+      ];
+
+      const expectedSapSystemTypes = [
+        'ABAP',
+        'JAVA',
+        'ABAP/JAVA',
+        'ABAP/JAVA',
+        '',
+      ];
+      const sapSystems = [
```
arbulu89, line 113: Maybe
```
const sapSystems = sapSystemTypes.map((type) => sapSystemFactory.build({ sap_system_type: type }))
```

arbulu89 (changes requested):
Some nitpick comments in the frontend.
About the backend, and its test.
Do you think we need so many new tests for such a simple thing?
I mean, I think that simply with one new test would've been enough. Just a test to see that a j2ee instance emits the sap system discovered event. The rest of the scenarios are already well tested

assets/js/lib/test-utils/factories/sapSystems.js
```diff
@@ -36,10 +36,12 @@ export const sapSystemFactory = Factory.define(({ params }) => {
   const sid = params.sid || generateSid();
   const databaseId = params.database_id || faker.string.uuid();
   const databaseSid = params.database_sid || generateSid();
+  const sapSystemType = params.sap_system_type || 'ABAP';
   return {
     application_instances: sapSystemApplicationInstanceFactory.buildList(2, {
       sap_system_id: id,
       sid,
+      features: sapSystemType,
```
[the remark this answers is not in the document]
EMaksy: Okay, followed the suggestion!

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.jsx
```diff
@@ -120,17 +138,21 @@ function SapSystemsOverview({
...
+      sap_system_id: sapSystem.id,
+    });
+
   const data = sapSystems.map((sapSystem) => ({
     id: sapSystem.id,
     health: sapSystem.health,
     sid: sapSystem.sid,
     attachedRdbms: sapSystem.database_sid,
     tenant: sapSystem.tenant,
+    type: filteredApplicationInstances(sapSystem),
```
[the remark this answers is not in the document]
EMaksy: Yeah, you are right, a leftover from the backend implementation approach :)

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.jsx
```diff
@@ -77,6 +77,24 @@ function SapSystemsOverview({
...
       },
+      {
+        title: 'Type',
+        key: 'applicationInstances',
+        render: (content) => {
+          const instanceTypes = uniq(
+            flatMap(content, (instance) => instance.features.split('|'))
+          )
+            .filter((item) => item === 'J2EE' || item === 'ABAP')
+            .map((item) => (item === 'J2EE' ? 'JAVA' : item));
```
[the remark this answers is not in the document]
EMaksy: Yeah nice!

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
EMaksy: Refactored with a for each!

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
EMaksy: Factory is returned to original state, this is obsolete now, but yeah map would be here the way!

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.jsx
```diff
@@ -77,6 +77,22 @@ function SapSystemsOverview({
         title: 'Tenant',
         key: 'tenant',
       },
+      {
+        title: 'Type',
+        key: 'type',
```
arbulu89, line 82: Here you cannot use `type` as we don't have this key in the data fields.
You need to use `applicationInstances`.
Once you do this, you will have `render: (content)` where content are the `applicationInstances`

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.jsx
```diff
@@ -77,6 +77,22 @@ function SapSystemsOverview({
...
+        render: (_, items) =>
+          uniq(
+            flatMap(
+              filter(items.applicationInstances, { sap_system_id: items.id }),
```
arbulu89, line 86: You shouldn't need this filter, as `applicationInstances` are already coming filtered.

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
```diff
@@ -53,11 +55,19 @@ describe('SapSystemsOverviews component', () => {
         database_sid: attachedRdbms,
       } = sapSystem;
 
+      const sapSystemType = 'ABAP';
+      const modifiedApplicationInstances = applicationInstances.map(
```
arbulu89, line 59: Do we need to modify the variable after creating?
Did you try something like this above:
```
      const sapSystem = sapSystemFactory.build({
        ensa_version: 'ensa1',
        application_instances: applicationInstanceFactory.buildList(2, { features: 'ABAP' }),
      });
```

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
```diff
@@ -80,13 +90,69 @@ describe('SapSystemsOverviews component', () => {
...
+      expect(mainRow.querySelector('td:nth-child(7)')).toHaveTextContent(
         'ENSA1'
       );
     });
 
+    it('should display the correct SAP system type', () => {
+      const sapSystemTypes = [
+        'ABAP',
+        'J2EE',
+        'ABAP|J2EE',
```
arbulu89, line 107: I understand what you are trying to test, but this entry you are putting is not a real scenario.
Here, if you want to test ABAP and JAVA together, you should have one instance with `ABAP` features and other instance with `J2EE` features. Not 1 instance with both, which doesn't exist

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
```diff
@@ -80,13 +90,69 @@ describe('SapSystemsOverviews component', () => {
...
+
+      const expectedSapSystemTypes = [
+        'ABAP',
+        'JAVA',
+        'ABAP/JAVA',
+        'ABAP/JAVA',
+        '',
+      ];
+
+      const sapSystems = sapSystemFactory.buildList(5);
```
arbulu89, line 120: Creating a factory to modify just after that looks like an anti-pattern. 
Why don't you create the value using the factory with the wanted entries?
Something maybe like:
```
const sapSystems = expectedSapSystemTypes
.map((type) => sapSystemFactory.build({ application_instances : applicationInstanceFactory.buildList(2, { features: type }))
```

arbulu89 (commented):
Hey @EMaksy 
Yet other set of comments about the frontend changes :stuck_out_tongue_closed_eyes: 

As a comment, now in Storybook, in `SapSystemsOverview` entry, all the types, always are `ABAP/JAVA` which is a totally exceptional deployment.
Could we have only `ABAP` or `JAVA` and the combo of both only in the custom story?

assets/js/pages/SapSystemsOverviewPage/SapSystemsOverview.test.jsx
EMaksy: No not really, overwriting makes more sense :)

arbulu89 (approved):
Thank you buddy!
Looks great

PR #2823, opened 2024-07-25
Refactor: move `<Filter />` component

balanza:
# Description

The  `<Filter />` component has been moved out from the `<Table />` component so that it can be used separately. 
This is a preparatory work in order to enhance the usage of tables in more diversified scenarios.

No changes in the behavior of the component have been introduced (except for the default value).

All references to the component has been updated.


assets/js/common/Filter/Filter.stories.jsx
```diff
@@ -0,0 +1,79 @@
+import React, { useState } from 'react';
+import { action } from '@storybook/addon-actions';
+
+import Filter from '.';
+
+export default {
```
nelsonkopliku, line 6: **suggestion:** we might want to follow storybook's [guidelines about writing stories](https://storybook.js.org/docs/writing-stories).
I think it maximizes expressiveness and minimizes repetition.

```js
export default {
  title: 'Components/Filter',
  component: Filter,
  argTypes: {
    options: {/** type, description, control... */},
    title: {/** type, description, control... */},
    value: {/** type, description, control... */},
    onChange: {/** type, description, control... */}
  },
  render: (args) => {
    const [value, setValue] = useState(args.value);

    return (
      <Filter
        title={args.title}
        options={args.options}
        value={value}
        onChange={(newValue) => {
          setValue(newValue);
          action('onChange')(newValue);
        }}
      />
    );
  }
};

export const Default = {
  args: {
    title: 'Title',
    options: [
      'Tony Kekw',
      'Chad Carbonara',
      'Chuck Amatriciana',
      'K.C.O. Pepe',
      'Virgin Gricia',
    ],
    value: [],
  },
};

export const WithValue = {
  args: {
    ...Default.args,
    value: ['Tony Kekw'],
  },
};

export const WithMultipleValues = {
  args: {
    ...Default.args,
    value: ['Tony Kekw', 'Chad Carbonara'],
  },
};
```

nelsonkopliku (commented):
Cool, thanks!
I shared a suggestion we might want to consider, although not blocking.

assets/js/common/Filter/Filter.stories.jsx
balanza: Good point, it's indeed better to stick with guidelines. Done in https://github.com/trento-project/web/pull/2823/commits/c37218fcf8ac3fe74e9066d42e1d7e05e18318e1

nelsonkopliku (approved):
LGTM

PR #2824, opened 2024-07-25
Remove unused prop in `<Table />` component

balanza:
# Description

The `withPadding` property is completely overlapped by the `config.usePadding` property. We removed the first and used the latter when needed.


nelsonkopliku (approved):
LGTM. The visual result seems to be unchaged.

EMaksy (commented):
Thanks for taking care LGTM!

PR #2827, opened 2024-07-26
Handle unexpected successful api response

nelsonkopliku:
# Description

Recent changes to the shape of the activity log api response broke the frontend.
This PR makes the client more reliable on unexpected successful response shapes from the api.

## How was this tested?
Automated tests


assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
```diff
@@ -14,7 +14,7 @@ function ActivityLogPage() {
   useEffect(() => {
     getActivityLog()
       .then((response) => {
-        setActivityLog(response.data);
+        setActivityLog(response.data?.data ?? []);
```
EMaksy, line 17: Nullish coalescing operator, nice!

PR #2828, opened 2024-07-26
Default to activity type as a message for unrecognized activities

nelsonkopliku:
# Description

With this PR the frontend uses the activity type as the message for an activity which is not yet mapped to be nicely shown in the UI.

In this way we avoid a perhaps too generic `Unrecognized activity` as message, and we show `the_new_unmapped_activity_log_type`, which provides at least some value.

## How was this tested?
Automated test.


nelsonkopliku:
@balanza @gagandeepb any input here?

gagandeepb (approved):
Thanks for this! LGTM.

PR #2831, opened 2024-07-29
Upgrade to Phoenix 1.7

CDimonaco:
# Description

This pr upgrade the `web` codebase to Phoenix 1.7

All the major critical libraries are updated, some nitpicks in the codebase/tests changed to reflect the apis of the new version.

Removed eventstore_dashboard library, the project is no longer updated, and it's not compatible with new liveview/live_dashboard versions.

## How was this tested?

Automated/manual testing


lib/trento_web/templates/layout/live.html.heex
```diff
@@ -1,11 +1,11 @@
 <main class="container">
-  <p class="alert alert-info" role="alert"
-    phx-click="lv:clear-flash"
-    phx-value-key="info"><%= live_flash(@flash, :info) %></p>
+  <p class="alert alert-info" role="alert" phx-click="lv:clear-flash" phx-value-key="info">
+    <%= Phoenix.Flash.get(@flash, :info) %>
```
arbulu89, line 3: I think simply `get` instead of `Phoenix.Flash.get` would work as well, as we are importing `Phoenix.Component` in the router.
Anyway, nothing bad having explicit :+1:

arbulu89 (approved):
Perfect @CDimonaco 
Really appreciated.
It is a pity that we need to remove the eventstore dashboard temporarily at least. I loved that view in the dashboard.
Anyway, it is what it is...

PD:
If anyone wants to some references about this upgrade:
https://elixircasts.io/upgrading-to-phoenix-1.7

dottorblaster (approved):
Nothing wrong from my point of view :+1:

PR #2833, opened 2024-07-29
E2E Tests parallel execution

CDimonaco:
# Description

This pr splits the cypress tests execution across multiple CI jobs, we use the `cypress-split` plugin.
This PR does not fix any issue on flakiness or test logic, it just speeds up the execution of test pipeline.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -353,10 +353,26 @@ jobs:
         if: steps.npm-e2e-cache.outputs.cache-hit != 'true'
         run: cd test/e2e && npm install
 
+  prepare:
+    runs-on: ubuntu-20.04
+    outputs:
+      matrix: ${{ steps.prepare.outputs.matrix }}
+    steps:
+      - name: Create matrix ⊹
+        id: prepare
+        uses: bahmutov/gh-build-matrix@main
+        with:
+          n: 4
+      - name: Print result 🖨
+        run: echo '${{ steps.prepare.outputs.matrix }}'
+
   test-e2e:
     name: End to end tests
-    needs: [elixir-deps, npm-deps, npm-e2e-deps]
+    needs: [elixir-deps, npm-deps, npm-e2e-deps, prepare]
     runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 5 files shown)


gagandeepb (approved):
Thanks for this! I wonder what is the expected/observed speed-up with these changes ?

CDimonaco:
> Thanks for this! I wonder what is the expected/observed speed-up with these changes ?

That's a good question! I observed an increase of speed of about 5 to 7 minutes, but this is just the first step, when we refactor the e2e tests, in order to make them less flaky and without some useless costly operations (mostly photofinish dump when not needed), this pr should make a lot more sense

.github/workflows/ci.yaml
```diff
@@ -344,10 +344,26 @@ jobs:
...
+  prepare:
+    runs-on: ubuntu-20.04
+    outputs:
+      matrix: ${{ steps.prepare.outputs.matrix }}
+    steps:
+      - name: Create matrix ⊹
+        id: prepare
+        uses: bahmutov/gh-build-matrix@main
+        with:
+          n: 4
```
arbulu89, line 356: Can we have this `4` in some variable on top?
So we can change if needed, as it is used in 2 places

arbulu89 (approved):
Perfect! Really really appreciated!
Just a comment about putting the `4` as a variable

.github/workflows/ci.yaml
```diff
@@ -415,14 +431,18 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Give executable permissions to photofinish
         run: chmod +x $(whereis photofinish | cut -d" " -f2)
-      - name: Cypress run
+      - name: Parallel cypress
         uses: cypress-io/github-action@v6
         env:
+          SPLIT: ${{ strategy.job-total }}
+          SPLIT_INDEX: ${{ strategy.job-index }}
```
balanza, line 438: **question:** ‏Where do those variables come out? Shouldn't be something like `strategy.matrix...` or `prepare.output.matrix...`?

.github/workflows/ci.yaml
balanza: Also: where does this number come from? Should it be the number of cores?

PR #2835, opened 2024-07-29
Fix schedule of `activity_log_cleaning` job

balanza:
# Description

In #2792 we erroneously configured the `activity_log_cleaning` job.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -129,7 +129,7 @@ config :trento, Trento.Scheduler,
     ],
     activity_log_cleaning: [
       schedule: "@daily",
-      task: {Trento.ActivityLog.ActivityLogger, :clear_expired_logs, []},
+      task: {Trento.ActivityLog, :clear_expired_logs, []},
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):

balanza:
**thought:** ‏None of our checks spotted the problem. Should we think about a check dedicated to scheduled jobs? It seems they are not covered by our test suites.

nelsonkopliku:
> **thought:** ‏None of our checks spotted the problem. Should we think about a check dedicated to scheduled jobs? It seems they are not covered by our test suites.

It might indeed help for these cases. Shall we track a tech debt?

gagandeepb (commented):
Thanks. Oban allows for testing of workers and assert that jobs get enqueued, and is on the backlog already, as something to evaluate. Alternatively, a short-lived solution is to implement a custom Credo check to verify if the specific module even exports a function when put in Quantum config.

PR #2840, opened 2024-07-30
Default admin check with username

CDimonaco:
# Description

This PR changes the way we detect if a user is the default admin.
We used to check against the user id, checking from id == 1

We change the strategy in order to check against the username, the guarantees are the same but we don't assume the user id, this will be useful also in future scenarios when the admin will come from an external IDP and we can't be sure if it's the user with id 1 in our database.

## How was this tested?

Automated tests


config/config.exs
```diff
@@ -215,6 +215,9 @@ config :bodyguard,
 
 config :flop, repo: Trento.Repo
 
+config :trento,
+  admin_username: "admin"
```
arbulu89, line 219: nitpick: Maybe just to match with the env variable, could we name this `admin_user`?
As the env variable is `ADMIN_USER`

lib/trento/users.ex
```diff
@@ -34,6 +34,8 @@ defmodule Trento.Users do
     |> Repo.all()
   end
 
+  def admin_username, do: Application.fetch_env!(:trento, :admin_username)
```
arbulu89, line 37: question: Must this be public?
I mean, if it is only for tests, you know that the value is `admin` coming from the config file

lib/trento/users.ex
```diff
@@ -76,18 +78,18 @@ defmodule Trento.Users do
     |> Repo.insert()
   end
 
-  def update_user_profile(%User{id: 1}, _), do: {:error, :forbidden}
-
-  def update_user_profile(%User{} = user, attrs) do
-    updated_attrs = maybe_set_password_change_requested_at(attrs, true)
+  def update_user_profile(%User{username: username} = user, attrs) do
```
arbulu89, line 81: This change is so painful. It was so nice with that readable pattern matching :sweat_smile: 
Could we maybe try to experiment with some macro magic?
Or assign the value to a constant? 
Or something like that.
I don't know.
This change makes me sad :stuck_out_tongue_closed_eyes:

arbulu89 (commented):
The idea is needed, i understand that, but the changes in the users file makes me so sad.
If there is not any other alternative, we will need to go this way, but it looks so ugly codewise :sweat_smile:

arbulu89 (approved):
Hey @CDimonaco 
It looks good to me.
Simply a comment about making the `admin_username` function maybe private.

About the ugliness, i understand that there must not be any other option.
If you researched it, and there isn't, go ahead. It is what it is

lib/trento/users.ex
CDimonaco: done!

config/config.exs
CDimonaco: done!

lib/trento/users.ex
CDimonaco: It's ugly has hell, but the value is set at runtime, so we can't use compile_env, or some costant in the pattern matching, even with macros the value is not known at compile time

PR #2841, opened 2024-07-31
Domain events activity frontend

nelsonkopliku:
# Description

Mapping messages/activity types/resources to the domain events activities in the UI.
I found it useful consolidating the mapping in one place.

As we iterate we'll figure out what kind of mapping should be moved from the frontend to the backend. 

## How was this tested?

Automated tests.

### DIFF
--- a/assets/js/common/ActivityLogDetailsModal/ActivityLogDetailModal.jsx
+++ b/assets/js/common/ActivityLogDetailsModal/ActivityLogDetailModal.jsx
@@ -7,43 +7,8 @@ import Button from '@common/Button';
 import Modal from '@common/Modal';
 import ListView from '@common/ListView';
 
-import {
-  API_KEY_GENERATION,
-  CHANGING_SUMA_SETTINGS,
-  CLEARING_SUMA_SETTINGS,
-  CLUSTER_CHECKS_EXECUTION_REQUEST,
-  LOGIN_ATTEMPT,
-  PROFILE_UPDATE,
-  RESOURCE_TAGGING,
-  RESOURCE_UNTAGGING,
-  SAVING_SUMA_SETTINGS,
-  USER_CREATION,
-  USER_DELETION,
-  USER_MODIFICATION,
-} from '@lib/model/activityLog';
+import { toResource, toLabel } from '@lib/model/activityLog';
 import classNames from 'classnames';
-
-const activityTypesLabels = {
-  [LOGIN_ATTEMPT]: 'Login Attempt',
-  [RESOURCE_TAGGING]: 'Tag Added',
-  [RESOURCE_UNTAGGING]: 'Tag Removed',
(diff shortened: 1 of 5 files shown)


assets/js/common/ActivityLogDetailsModal/ActivityLogDetailModal.jsx
```diff
@@ -20,6 +20,72 @@ import {
...
   USER_DELETION,
   USER_MODIFICATION,
+  // Host events
+  HEARTBEAT_FAILED,
+  HEARTBEAT_SUCCEEDED,
+  HOST_CHECKS_HEALTH_CHANGED,
+  HOST_CHECKS_SELECTED,
+  HOST_DEREGISTERED,
+  HOST_DEREGISTRATION_REQUESTED,
+  HOST_DETAILS_UPDATED,
```
gagandeepb, line 30: Possible opportunity to evaluate openapi front-end code generation here?

assets/js/common/ActivityLogDetailsModal/ActivityLogDetailModal.jsx
```diff
@@ -36,6 +102,71 @@ const activityTypesLabels = {
   [USER_DELETION]: 'User Deleted',
   [PROFILE_UPDATE]: 'Profile Updated',
   [CLUSTER_CHECKS_EXECUTION_REQUEST]: 'Checks Execution Requested',
+  [HOST_ADDED_TO_CLUSTER]: 'Host Added to Cluster',
+  [HEARTBEAT_FAILED]: 'Heartbeat Failed',
+  [HEARTBEAT_SUCCEEDED]: 'Heartbeat Succeeded',
+  [HOST_CHECKS_HEALTH_CHANGED]: 'Checks Health Changed',
```
gagandeepb, line 108: This could potentially be a string manipulating function, I guess? Turn the event names into a string is a bit like writing a custom serializer, for the most part.

assets/js/common/ActivityLogDetailsModal/ActivityLogDetailModal.jsx
nelsonkopliku: Could be an option indeed. Let's see how it evolves.

assets/js/common/ActivityLogDetailsModal/ActivityLogDetailModal.jsx
nelsonkopliku: Yes, it could be something like
```js
const activityTypesLabels = _.fromPairs(
  ACTIVITY_TYPES.map((type) => [
    type,
    type.split('_').map(_.capitalize).join(' '),
  ])
);
```

However I'd like to get some more feedback on the rendered entries in order to figure out the transformation rule.

Additionally let's take into account that there is not yet a consistent pattern throughout all the activity types. But we can improve as we proceed.

PR #2842, opened 2024-07-31
Add a global config variable to Storybook

dottorblaster:
Since literally now our Storybook build began failing because of a missing `config` global variable. This PR fixes it.


CDimonaco (approved):
LGTM 🚀

PR #2849, opened 2024-08-02
Add `<DateFilter >` component

balanza:
# Description

Add <DateFilter /> component to filter some dates in the past. It can be configured with the options to show, although it has some defaults. Please take a look at the storybook for all the configurable options.

Every option is composed of a triple:
* ID of the option
* value function, to calculate the actual date at selection time
* rendering function, to show a custom label for the option; this function is optional and falls back to the id.

As a shortcut, an option can be specified as just its ID if it's available amongst the pre-configured ones/

The exposed interface comes with some assumptions, I'll leave comments in this PR pointing to the relevant code parts.

----

The component has also been added as a configurable filter in `<ComposedFilter />`.

### DIFF
--- a/assets/js/common/ComposedFilter/ComposedFilter.jsx
+++ b/assets/js/common/ComposedFilter/ComposedFilter.jsx
@@ -3,12 +3,21 @@ import Button from '@common/Button';
 import Filter from '@common/Filter';
 import DateFilter from '@common/DateFilter';
 
-const renderFilter = ({ type, ...filterProps }, value, onChange) => {
+const renderFilter = (key, { type, ...filterProps }, value, onChange) => {
   switch (type) {
     case 'select':
-      return <Filter {...filterProps} value={value} onChange={onChange} />;
+      return (
+        <Filter key={key} {...filterProps} value={value} onChange={onChange} />
+      );
     case 'date':
-      return <DateFilter {...filterProps} value={value} onChange={onChange} />;
+      return (
+        <DateFilter
+          key={key}
+          {...filterProps}
+          value={value}
+          onChange={onChange}
+        />
+      );
     default:
       return null;
   }
--- a/assets/js/common/ComposedFilter/ComposedFilter.jsx
+++ b/assets/js/common/ComposedFilter/ComposedFilter.jsx
@@ -1,11 +1,18 @@
 import React, { useState } from 'react';
 import Button from '@common/Button';
 import Filter from '@common/Filter';
+import DateFilter from '@common/DateFilter';
 
-const renderFilter = (key, { type, ...filterProps }, value, onChange) =>
-  type === 'select' ? (
-    <Filter key={key} {...filterProps} value={value} onChange={onChange} />
-  ) : null;
+const renderFilter = ({ type, ...filterProps }, value, onChange) => {
+  switch (type) {
+    case 'select':
+      return <Filter {...filterProps} value={value} onChange={onChange} />;
+    case 'date':
+      return <DateFilter {...filterProps} value={value} onChange={onChange} />;
+    default:
+      return null;
+  }
+};
 
 /**
  * Define a filter which is the composition of several filters.
--- a/assets/js/common/DateFilter/DateFilter.jsx
+++ b/assets/js/common/DateFilter/DateFilter.jsx
@@ -1,0 +1,211 @@
+import React, { Fragment, useState, useRef } from 'react';
+import classNames from 'classnames';
+import { Transition } from '@headlessui/react';
+
+import useOnClickOutside from '@hooks/useOnClickOutside';
+import { EOS_CLOSE } from 'eos-icons-react';
+
+const oneHour = 60 * 60 * 1000;
+const preconfiguredOptions = {
+  '1h ago': () => new Date(Date.now() - oneHour),
+  '24h ago': () => new Date(Date.now() - 24 * oneHour),
+  '7d ago': () => new Date(Date.now() - 7 * 24 * oneHour),
+  '30d ago': () => new Date(Date.now() - 30 * 24 * oneHour),
+};
+
+const renderOptionItem = (option, placeholder) => {
+  if (!option || !Array.isArray(option)) {
+    return placeholder;
+  }
+  if (typeof option[2] === 'function') {
+    return option[2]();
+  }
+  return option[0];
+};
+
+const parseInputOptions = (options) =>
+  options
+    .map((option) => {
+      if (typeof option === 'string' && option in preconfiguredOptions) {
+        return [option, preconfiguredOptions[option]];
+      }
+      if (
+        Array.isArray(option) &&
+        typeof option[1] === 'function' &&
+        option[1]() instanceof Date
+      ) {
+        return option;
+      }
+      return undefined;
+    })
+    .filter(Boolean)
+    .reduceRight(
+      (acc, el) =>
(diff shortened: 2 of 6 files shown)


assets/js/common/DateFilter/DateFilter.jsx
```diff
@@ -0,0 +1,213 @@
...
+ * In case the render function is not provided, the id will be used as the label.
+ * An option can also be a string, in which case it will be considered as a pre-configured option.
+ * In case more options with the same id are provided, only the last one will be considered.
+ * Options will be displayed sorted by date in descending order.
+ * @param {string} props.title - The title of the date filter, to be shown as placeholder when no value is selected.
+ * @param {string} props.value - The selected id of the selected option. It accepted either a string or an array with the id as the first element.
+ * @param {boolean} props.prefill - Whether to include pre-configured options in the options list. Default is false.
+ * @param {function} props.onChange - The callback function to be called when the value of the date filter changes. It will provide a couple with the selected id and the actual date.
+ */
+function DateFilter({ options = [], title, value, prefill, onChange }) {
```
balanza, line 84: **naming (value)** I'm using `value` for the actual selection to be set from the using component, to be consistent with other filters. Maybe `selection` is more appropriate?

balanza, line 84: **naming (prefill)** `prefill` or `prefilled`?

assets/js/common/DateFilter/DateFilter.jsx
```diff
@@ -0,0 +1,213 @@
...
+    .filter(Boolean)
+    .reduceRight(
+      (acc, el) =>
+        acc.find(([label]) => label === el[0]) ? acc : [...acc, el],
+      []
+    )
+    .sort((a, b) => b[1]().getTime() - a[1]().getTime());
+
+const getSelectedOption = (options, value) => {
+  const selectedId = Array.isArray(value) ? value[0] : value;
```
balanza, line 52: **decision** Ideally, the option ID is enough to set the selected item. However, the value exposed on change is a couple `[ID, date]` and I thought it would be convenient to accept it as value as well otherwise it would require us to handle the difference in format when storing it in the local/global state.

assets/js/common/DateFilter/DateFilter.jsx
```diff
@@ -0,0 +1,213 @@
...
+const parseInputOptions = (options) =>
+  options
+    .map((option) => {
+      if (typeof option === 'string' && option in preconfiguredOptions) {
+        return [option, preconfiguredOptions[option]];
+      }
+      if (
+        Array.isArray(option) &&
+        typeof option[1] === 'function' &&
+        option[1]() instanceof Date
```
balanza, line 37: **decision** Options come with a value function so that the actual reference date is calculated at selection time. That's simpler for the consumer as it already has a date, as expected from a date filter, which can just store the value and use it at its convenience.

That comes with a little drawback: if the date reflects a period (for example, 10 minutes ago), that value is never updated unless a new selection is made.

Another viable solution would be to return a number instead of a Date object; such a number can represent the milliseconds back from now. Being a relative value it does not have to update. However, the using component must handle the calculation itself.

nelsonkopliku (commented):
I don't see any major blocker. Eager to start using it in the activity log page :smile: 

A thought that came to my mind is about what gets send to the API call.
I believe the following format is what is going to be send and what the API accepts `2024-08-05T08:00:16.549046Z`, @gagandeepb could you confirm, please?

**question:** should we be concerned about browser timezone vs server timezone? :thinking:

gagandeepb (commented):
If server assumes system/db time is UTC, the browser would/should do the same, I guess. All timestamps are assumed to be UTC, in ISO8601 format serialized as a string and expected in this format by the BE API. cc: @nelsonkopliku

assets/js/common/DateFilter/DateFilter.jsx
```diff
@@ -0,0 +1,211 @@
...
+  }
+  return undefined;
+};
+
+/**
+ * A component for filtering dates.
+ *
+ * @component
+ * @example
+ * // <DateFilter
```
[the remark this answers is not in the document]
balanza: > might lead to thinking it is about selecting a specific date

It's exactly what happens. There are shortcuts, but also a date picker (which will be implemented in the next iteration)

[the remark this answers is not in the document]
nelsonkopliku: Ok, feel free to keep current name :+1:

balanza:
> question: should we be concerned about browser timezone vs server timezone? 

@nelsonkopliku I think we shouldn't, the filter should refer to values as they come from the server. It will be up to the user to consider that he/she is looking at something generated into a server with its timezone.

nelsonkopliku (commented):
At the end, what did you decide to do or not do with regards to the following :sweat_smile::
- [comment](https://github.com/trento-project/web/pull/2849/files#r1703669817) on `prefill` vs `prefilled` and the question about it defaulting to `true`
- [comment](https://github.com/trento-project/web/pull/2849/files#r1703658264) about a story showing `Filter undefined` because of a missing `title`

assets/js/common/ComposedFilter/ComposedFilter.jsx
```diff
@@ -1,11 +1,18 @@
 import React, { useState } from 'react';
 import Button from '@common/Button';
 import Filter from '@common/Filter';
+import DateFilter from '@common/DateFilter';
 
-const renderFilter = (key, { type, ...filterProps }, value, onChange) =>
-  type === 'select' ? (
-    <Filter key={key} {...filterProps} value={value} onChange={onChange} />
-  ) : null;
+const renderFilter = ({ type, ...filterProps }, value, onChange) => {
```
nelsonkopliku, line 6: **issue:** I guess we lost `key` during a rebase. This is causing test failures.

balanza: Good catch, thanks! Fixed in https://github.com/trento-project/web/pull/2849/commits/57e72fdff71b304074a0503de8950cc50217c776

balanza:
> At the end, what did you decide to do or not do with regards to the following 😅:
> 
> * [comment](https://github.com/trento-project/web/pull/2849/files#r1703669817) on `prefill` vs `prefilled` and the question about it defaulting to `true`
> * [comment](https://github.com/trento-project/web/pull/2849/files#r1703658264) about a story showing `Filter undefined` because of a missing `title`

Yes, I missed that. Done in https://github.com/trento-project/web/pull/2849/commits/f5b9d3d151b9bd1029b59b37d77c17eb6db633fa

assets/js/common/DateFilter/DateFilter.jsx
balanza: Done in https://github.com/trento-project/web/pull/2849/commits/f5b9d3d151b9bd1029b59b37d77c17eb6db633fa

assets/js/common/DateFilter/DateFilter.stories.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+export const WithSelectedValue = {
+  args: {
+    ...Default.args,
+    value: '1h ago',
+  },
+};
+
+export const WithCustomOptions = {
+  args: {
```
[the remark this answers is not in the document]
nelsonkopliku: `title` still missing :see_no_evil:

nelsonkopliku (approved):
LGTM, just the story needs to be adjusted.

PR #2853, opened 2024-08-05
Adjust filtering

nelsonkopliku:
# Description

Tiny fixes:
- adds a `key` prop to composed filters rendered in a loop, so that we make react stop complain in the console
- adds a missing bottom padding for table filters


arbulu89 (approved):
Thank you!

balanza (approved):
Thanks for fixing it!

PR #2859, opened 2024-08-05
Disable and update user forms if OIDC is enabled

arbulu89:
# Description

Disable certain fields and options in user management views if SSO (OIDC in this case) is enabled.
The `/users/new` url is redirected to not found.

image

image

## How was this tested?

Tested with jest and storybook

### DIFF
--- a/assets/js/lib/model/users.js
+++ b/assets/js/lib/model/users.js
@@ -1,5 +1,7 @@
 import { getFromConfig } from '@lib/config';
 
 const TRENTO_ADMIN_USERNAME = getFromConfig('adminUsername') || 'admin';
+const OIDC_ENABLED = getFromConfig('oidcEnabled') || false;
 
 export const isAdmin = (user) => user.username === TRENTO_ADMIN_USERNAME;
+export const isSingleSignOnEnabled = () => OIDC_ENABLED;
--- a/assets/js/lib/model/users.js
+++ b/assets/js/lib/model/users.js
@@ -1,7 +1,5 @@
 import { getFromConfig } from '@lib/config';
 
 const TRENTO_ADMIN_USERNAME = getFromConfig('adminUsername') || 'admin';
-const OIDC_ENABLED = getFromConfig('oidcEnabled') || false;
 
 export const isAdmin = (user) => user.username === TRENTO_ADMIN_USERNAME;
-export const isSingleSignOnEnabled = () => OIDC_ENABLED;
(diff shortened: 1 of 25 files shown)


assets/js/pages/Profile/ProfileForm.jsx
```diff
@@ -25,6 +25,7 @@ function ProfileForm({
   disableForm,
   passwordModalOpen = false,
   totpBoxOpen = false,
+  singleSignOnEnabled = false,
```
balanza, line 28: **suggestion:** No need for default value when is `false`‏
```suggestion
  singleSignOnEnabled,
```

assets/js/lib/model/users.js
```diff
@@ -1,5 +1,7 @@
 import { getFromConfig } from '@lib/config';
 
 const TRENTO_ADMIN_USERNAME = getFromConfig('adminUsername') || 'admin';
+const OIDC_ENABLED = getFromConfig('oidcEnabled') || false;
 
 export const isAdmin = (user) => user.username === TRENTO_ADMIN_USERNAME;
+export const isSingleSignOnEnabled = () => OIDC_ENABLED;
```
CDimonaco, line 7: TBH I don't know if lib/model/users is the right place to put SSO stuff, it seems also strange in tests checking and mocking for an app configuration in the users model

Maybe a file in @lib/auth called config or something like that is more appropriate?

Wdyt?

CDimonaco (approved):
Perfect 🚀 

Just one nitpick but really up to you, one I commented on the code, about moving the oidc app configuration stuff from model/users to @lib/auth in a dedicate file

assets/js/lib/model/users.js
arbulu89: Yes, I prefer it in fact.
Maybe `sso.js` of `singlesignon.js`.
I will think about it.
I will move things around

assets/js/pages/Profile/ProfileForm.jsx
arbulu89: Thank you @balanza 
Yes, good tip. Let me send my own commit, as there are other `false` defaults on top, and I will change them all, otherwise it will be strange

assets/js/lib/model/users.js
arbulu89: Done

assets/js/pages/Profile/ProfileForm.jsx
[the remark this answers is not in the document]
arbulu89: I think we need it to make the difference between actual editing enabled (which now is nambed `disableForm`, but still shows the password and topt button) and sso option, which doesn't show the fields.

arbulu89: @balanza At the end, I think I'm going to revert this change...

PR #2865, opened 2024-08-06
View field for idp users

CDimonaco:
# Description

This PR includes a new field called `idp_user` in the api response of profile and users controller.
This field will be true when the user has one or more user_identities associated, so it comes from an external IDP

## How was this tested?

Automated tests


arbulu89 (approved):
Thank you!
Simply remind me why we need this field in the frontend :sweat_smile: 
I don't remember any mockup where we identify this variable

CDimonaco:
Force merging @arbulu89 is working on eslint errors in another branch

PR #2866, opened 2024-08-06
Single sign on login view

arbulu89:
# Description

Style login view for SSO. It is used if `oidc_enabled` is set to true in the configuration.

image

## How was this tested?

Jest

### DIFF
--- a/assets/js/pages/Login/LoginSSO.jsx
+++ b/assets/js/pages/Login/LoginSSO.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+
+import Button from '@common/Button';
+
+export default function LoginSSO({ singleSignOnUrl }) {
+  return (
+    <Button
+      onClick={() => {
+        window.location.href = singleSignOnUrl;
+      }}
+    >
+      Login with Single Sign-on
+    </Button>
+  );
+}
--- a/assets/jest.config.js
+++ b/assets/jest.config.js
@@ -69,6 +69,7 @@ module.exports = {
       suseManagerEnabled: true,
       adminUsername: 'admin',
       oidcEnabled: false,
+      oidcLoginUrl: 'http://localhost:4000/auth/oidc_callback',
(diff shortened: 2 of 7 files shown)


assets/js/pages/Login/LoginSSO.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import Button from '@common/Button';
+
+export default function LoginSSO({ singleSignOnUrl }) {
+  return (
+    <Button
+      onClick={() => {
+        window.location.href = singleSignOnUrl;
```
CDimonaco, line 9: We are not using a child href for styling purposes?

CDimonaco (approved):
LGTM 🚀

jagabomb (approved):
LGTM, (reviewed using the screenshot)

assets/js/pages/Login/LoginSSO.jsx
arbulu89: You mean to have `<a href={ singleSignOnUrl }>Login with Single Sign-on</a>` as child, right?
This has the difference that the url is visible in the browser, when you put the mouse on the button.
I don't have any preference. I have seen the `window.location.href` in other login/session react component in our code. That's why I used it

EMaksy (approved):
Hey the code lgtm!
I guess you took a wip from @CDimonaco and the unused imports variables and console logs will go away inn a follow up pr right ?

arbulu89:
> Hey the code lgtm! I guess you took a wip from @CDimonaco and the unused imports variables and console logs will go away inn a follow up pr right ?

Yes, fix coming in next PR

assets/js/pages/Login/LoginSSO.jsx
arbulu89: And yes, the button is what we need related to styling, so it was an "obvious" option

PR #2867, opened 2024-08-06
Add date and time picker to `<DateFilter />`

balanza:
# Description

Add a DateTime for selecting a specific date and time as a filter value.

### DIFF
--- a/assets/js/common/DateFilter/DateFilter.jsx
+++ b/assets/js/common/DateFilter/DateFilter.jsx
@@ -5,6 +5,8 @@ import { Transition } from '@headlessui/react';
 import useOnClickOutside from '@hooks/useOnClickOutside';
 import { EOS_CLOSE } from 'eos-icons-react';
 
+import Input from '@common/Input';
+
 const oneHour = 60 * 60 * 1000;
 const preconfiguredOptions = {
   '1h ago': () => new Date(Date.now() - oneHour),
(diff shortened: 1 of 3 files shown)


assets/js/common/DateFilter/DateFilter.jsx
```diff
@@ -48,12 +55,58 @@ const parseInputOptions = (options) =>
...
   if (typeof selectedId === 'string') {
     return options.find((option) => option[0] === selectedId);
   }
   return undefined;
 };
 
+function Tick() {
+  return (
+    <span className="absolute inset-y-0 right-0 flex items-center pr-4">
+      <svg
```
nelsonkopliku, line 70: **suggestion:** we might use `EOS_CHECK` icon from eos icon-set.

assets/js/common/Input/Input.stories.jsx
```diff
@@ -160,6 +160,12 @@ export const Date = {
   },
 };
 
+export const DateTime = {
+  args: {
+    type: 'datetime-local',
```
nelsonkopliku, line 165: **praise:** well done in leveraging browser's calendar.

nelsonkopliku (commented):
Cool!

**question:** I was looking for 24h formatting, but I understood it being a bit of a pain. Same conclusion from you, I guess :sweat_smile:

assets/js/common/DateFilter/DateFilter.jsx
balanza: Good point. I'd rather do it in another PR, as we might also refactor the `Filter` component. I prepared it already: https://github.com/trento-project/web/pull/2870

nelsonkopliku (approved):

PR #2868, opened 2024-08-07
User identities context assigns global abilities to a oidc user

CDimonaco:
# Description

This PR enrich the user identities context to assign the global abilities to a oidc user, when the user is recognized as trento default admin.

We have a dedicated method in the user identities context, we cannot and we should not rely on the Users context directly because the Users context by design cannot update a user that is recognized as admin.

## How was this tested?

Automated tests


arbulu89 (approved):
Perfection!

EMaksy (commented):
Nice job!

PR #2874, opened 2024-08-07
Add option to load oidc variables in runtime

arbulu89:
# Description

Load OIDC integration required variables in runtime.
Nothing special. Only the `OIDC_CALLBACK_URL` is built using with origin and port if a specific url is not given.

We might want to update this when we have more providers, but by now it looks simple enough

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -174,4 +174,31 @@ if config_env() in [:prod, :demo] do
         iv_length: 12
       }
     ]
+
+  enable_oidc = System.get_env("ENABLE_OIDC", "false") == "true"
+
+  config :trento, :oidc,
+    enabled: enable_oidc,
+    callback_url:
+      System.get_env(
+        "OIDC_CALLBACK_URL",
+        "https://#{System.get_env("TRENTO_WEB_ORIGIN")}:#{System.get_env("PORT", "4000")}/auth/oidc_callback"
+      )
+
+  if enable_oidc do
+    config :trento, :pow_assent,
+      providers: [
+        oidc_local: [
+          client_id:
+            System.get_env("OIDC_CLIENT_ID") ||
+              raise("environment variable OIDC_CLIENT_ID is missing"),
+          client_secret:
+            System.get_env("OIDC_CLIENT_SECRET") ||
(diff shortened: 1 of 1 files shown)


config/runtime.exs
```diff
@@ -174,4 +174,31 @@ if config_env() in [:prod, :demo] do
...
+        "https://#{System.get_env("TRENTO_WEB_ORIGIN")}:#{System.get_env("PORT", "4000")}/auth/oidc_callback"
+      )
+
+  if enable_oidc do
+    config :trento, :pow_assent,
+      providers: [
+        oidc_local: [
+          client_id:
+            System.get_env("OIDC_CLIENT_ID") ||
+              raise("environment variable OIDC_CLIENT_ID is missing"),
```
EMaksy, line 194: What about adding a more explicit error message --> "environment variable OIDC_CLIENT_ID is missing in trento-web configuration" or something like this?

config/runtime.exs
```diff
@@ -174,4 +174,31 @@ if config_env() in [:prod, :demo] do
...
+          client_secret:
+            System.get_env("OIDC_CLIENT_SECRET") ||
+              raise("environment variable OIDC_CLIENT_SECRET is missing"),
+          base_url:
+            System.get_env("OIDC_BASE_URL") ||
```
EMaksy, line 199: **OIDC_CLIENT_ID**, **OIDC_CLIENT_SECRET** and **OIDC_BASE_URL** are the only new variables that needs to be added to the documentation right?

EMaksy (approved):
LGTM --> Just left a question and a small suggestion.

config/runtime.exs
arbulu89: No, `ENABLE_OIDC` and `OIDC_CALLBACK_URL` (this 2nd optional) are required as well

config/runtime.exs
arbulu89: This message is what we are using for other env variables. atmost, I could put examples, but I would assume that whoever is using OIDC knows what these variable are for

config/runtime.exs
EMaksy: Ah yeah you are right okay okay!

PR #2879, opened 2024-08-08
Allow only abilities update with OIDC

CDimonaco:
# Description

This PR allows only the abilities to be updated when oidc is enabled and the users update endpoint is used.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento_web/controllers/v1/users_controller.ex
+++ b/lib/trento_web/controllers/v1/users_controller.ex
@@ -182,7 +182,7 @@ defmodule TrentoWeb.V1.UsersController do
   end
 
   # when oidc is enabled, we only allow abilities as parameter
-  defp clean_params_for_oidc_integration(attrs, true), do: Map.take(attrs, [:abilities])
+  defp clean_params_for_oidc_integration(attrs, true), do: Map.take(attrs, [:abilities, :enabled])
   defp clean_params_for_oidc_integration(attrs, _), do: attrs
 
   defp oidc_enabled?, do: Application.fetch_env!(:trento, :oidc)[:enabled]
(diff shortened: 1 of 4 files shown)


lib/trento_web/controllers/v1/users_controller.ex
```diff
@@ -179,4 +180,10 @@ defmodule TrentoWeb.V1.UsersController do
   defp attach_user_version_as_etag_header(conn, %User{lock_version: lock_version}) do
     put_resp_header(conn, "etag", Integer.to_string(lock_version))
   end
+
+  # when oidc is enabled, we only allow abilities as parameter
+  defp clean_params_for_oidc_integration(attrs, true), do: Map.take(attrs, [:abilities])
```
arbulu89, line 185: We would need to take `:locked_at` as well, right?
Otherwise we cannot lock the user internally in trento

arbulu89 (commented):
@CDimonaco 
Looks good. But I think we need to include the `locked_at` field as option. Otherwise we cannot lock the user in trento

lib/trento_web/controllers/v1/users_controller.ex
CDimonaco: Yeah we should whitelist the enabled field, nice catch!

arbulu89: oh yes, `enabled` field, you are right

CDimonaco:
@arbulu89 done!

arbulu89 (approved):
:green_heart:

EMaksy (approved):
Nice job

PR #2880, opened 2024-08-08
Existing user is recovered when login with oidc

CDimonaco:
# Description

This pr, recover an existing user when the login with the idp is performed, the user has not a idp identity (could happen when customers have an already existing trento installation with local users, configure oidc and wants to login with the same users using oidc) but is present on trento database.

## How was this tested?

Automated tests

### DIFF
--- a/lib/trento/user_identities.ex
+++ b/lib/trento/user_identities.ex
@@ -17,17 +17,15 @@ defmodule Trento.UserIdentities do
 
   @impl true
   @doc """
-  redefining the PowAssent create user method, this is called when the user login through idp but
-  does not exist in our user database
+  redefining the PowAssent create user method, this is called when the user login through idp and a user identity
+  does not exists on our database.
+
+  If a user with the same username exists on our database, the user will be recovered and associated with the idp identity,
+  otherwise the user will be created.
   """
   def create_user(user_identity_params, user_params, user_id_params) do
-    case pow_assent_create_user(user_identity_params, user_params, user_id_params) do
-      {:ok, %User{} = user} ->
-        {:ok, maybe_assign_global_abilities(user)}
-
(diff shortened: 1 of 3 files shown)


lib/trento/user_identities.ex
```diff
@@ -17,17 +17,15 @@ defmodule Trento.UserIdentities do
...
   """
   def create_user(user_identity_params, user_params, user_id_params) do
-    case pow_assent_create_user(user_identity_params, user_params, user_id_params) do
-      {:ok, %User{} = user} ->
-        {:ok, maybe_assign_global_abilities(user)}
-
-      error ->
-        error
-    end
+    existing_user = Users.get_by(username: user_params["username"])
```
arbulu89, line 27: Maybe you can extract this `username` in function signature pattern matching.
`create_user(user_identity_params, %{"username" => usernam} = user_params...`

lib/trento/user_identities.ex
```diff
@@ -53,6 +51,20 @@ defmodule Trento.UserIdentities do
     end
   end
 
+  defp maybe_create_user(nil, user_identity_params, user_params, user_id_params) do
```
arbulu89, line 54: Most probably this function needs to go before `maybe_assign_global_abilities`, right?

arbulu89 (approved):
Good idea!

EMaksy (approved):
nothing to add besides @arbulu89 comments

lib/trento/user_identities.ex
CDimonaco: done! thx

lib/trento/user_identities.ex
CDimonaco: done!

PR #2882, opened 2024-08-08
Expose info about existence of prev/next page

nelsonkopliku:
# Description

Exposes `has_previous_page` and `has_next_page` boolean info as part of the pagination response.

It comes handy to allow/disallow back/forth navigation actions when we know there are no items.

## How was this tested?

Automated tests.


gagandeepb (approved):
Looks good. Thanks, especially, for adding the pagination schema object.

PR #2886, opened 2024-08-09
Add support for prev/next pagination

nelsonkopliku:
# Description

This PR introduces support for a cursor (ie prev/next) based pagination.
It can be used purely on the client or it can be hooked with callbacks interacting with a backend paginated API (what would happen for activity log page)

## How was this tested?

Automated tests and storybook.


balanza:
This PR is obsolete due to the refactoring of the Paginator component in #2915, #2918 and #2926.

PR #2892, opened 2024-08-12
Add labeling opportunity to `<Filter>` component

balanza:
# Description

The `<Filter>` component now allows options as key/value pairs. This is to cope with cases in which the value of the element must be different than the one shown.
Also, it enables a single value to be selected using a simple string, being more permissive on the allowed input.

This PR introduces no breaking changes.

## Usage
```ts
// normal usage
<Filter 
     value={["Mario Rossi"]}
     options={["Mario Rossi", "Giuseppe Verdi"]} />

// using string as value
<Filter 
     value="Mario Rossi"
     options={["Mario Rossi", "Giuseppe Verdi"]} />

// with key/value
<Filter 
     value={["mario_rossi"}
     options={[
       ["mario_rossi", "Mario Rossi"], 
       ["giuseppe_verdi", "Giuseppe Verdi"]
     ]} />
```

### DIFF
--- a/assets/js/common/Filter/Filter.jsx
+++ b/assets/js/common/Filter/Filter.jsx
@@ -16,7 +16,7 @@ const getLabel = (value, placeholder) =>
  *
  * @param {string[]|[string, string][]} props.options List of options to filter, e.g. ['Option 1', 'Option 2']
  * @param {string} props.title Title of the filter. It will be displayed in the button when the filter is empty
- * @param {string[]} props.value Selected options. Default is an empty array
+ * @param {string|string[]} props.value Selected options. Default is an empty array
  * @param {function} props.onChange Function to call when the selected options change
  */
 function Filter({ options, title, value = [], onChange }) {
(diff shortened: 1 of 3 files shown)


assets/js/common/Filter/Filter.jsx
```diff
@@ -14,26 +14,37 @@ const getLabel = (value, placeholder) =>
...
+    .map((option) => (typeof option === 'string' ? [option, option] : option));
+
+  const filteredOptions = labeledOptions.filter((option) =>
+    option[0].toLowerCase().includes(query.toLowerCase())
+  );
+
+  const parsedValue = typeof value === 'string' ? [value] : value;
+
+  const selectedLabels = parsedValue.reduce((acc, key) => {
+    const e = labeledOptions.find(([optionKey]) => optionKey === key);
```
dottorblaster, line 38: immagine

dottorblaster, line 38:
```suggestion
    const element = labeledOptions.find(([optionKey]) => optionKey === key);
```

dottorblaster (changes requested):
Just one tiny change but overall:

immagine

assets/js/common/Filter/Filter.jsx
balanza: Done in https://github.com/trento-project/web/pull/2892/commits/d039a74a8a8575e808a6b48f36b179d6aaae46e4

dottorblaster (approved):
d039a74a8a8575e808a6b48f36b179d6aaae46e4: `--pedantic mode`

Yes I am

PR #2893, opened 2024-08-13
Fetch patches for packages only if list is populated

dottorblaster:
As above, just fetch patches for packages only if the list is actually populated. Otherwise save up on the HTTP request since it would return an error.

## How was this tested?
Existing unit tests in place


jamie-suse (approved):
nice! :rocket:

PR #2898, opened 2024-08-14
Add filters to Activity log page

balanza:
# Description

Add filtering to the Activity log page

## Changes
* Allow the page to call the underlying API with needed params
* Mount filter by resource type
* Mount filter by date from
* Allow filters to be loaded from querystring
* refine `[Apply]` and `[Reset]` button sizes
* refine `<DateFilter>` input

## How was this tested?
E2E tests

### DIFF
--- a/assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
+++ b/assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
@@ -7,6 +7,11 @@ import ComposedFilter from '@common/ComposedFilter';
 
 import { getActivityLog } from '@lib/api/activityLogs';
 import { ACTIVITY_TYPES_CONFIG } from '@lib/model/activityLog';
+import {
+  filterValueToSearchParams,
+  searchParamsToAPIParams,
+  searchParamsToFilterValue,
+} from './searchParams';
 
 const filters = [
   {
@@ -26,27 +31,6 @@ const filters = [
   },
 ];
 
-const removeUndefinedFields = (obj) =>
-  Object.fromEntries(
-    Object.entries(obj).filter(([_, v]) => typeof v !== 'undefined')
-  );
-
-const searchParamsToFilters = (searchParams) =>
-  [...searchParams.entries()].reduce(
-    (acc, [key, value]) => ({ ...acc, [key]: [...(acc[key] || []), value] }),
-    {}
-  );
-
-const filtersToParams = (selectedFilters) =>
-  Object.entries(selectedFilters).reduce((acc, [key, value]) => {
-    switch (key) {
-      case 'from_date':
-        return { ...acc, from_date: new Date(value[1]).toISOString() };
-      default:
-        return { ...acc, [key]: value };
-    }
-  }, {});
-
 function ActivityLogPage() {
   const [searchParams, setSearchParams] = useSearchParams();
   const [activityLog, setActivityLog] = useState([]);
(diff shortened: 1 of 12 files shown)


balanza:
Depends on #2892 

Draft until it's merged

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
```diff
@@ -1,39 +1,98 @@
...
+  },
+];
+
+const removeUndefinedFields = (obj) =>
+  Object.fromEntries(
+    Object.entries(obj).filter(([_, v]) => typeof v !== 'undefined')
+  );
+
+const searchParamsToFilters = (searchParams) =>
+  [...searchParams.entries()].reduce(
```
balanza, line 35: **TIL** being `searchParams` an instance of `URLSearchParams`, `searchParams.entries()` returns an `Iterator` object. Or better: an object that implements the `Iterator` interface.

When running on the browser (Chrome), `searchParams.entries().reduce` is fine as it returns an `Array` object (which implements `Iterator`), to mimic the behavior of `Object.entries`.

However, when running tests `searchParams.entries().reduce` is not defined, as Node implements [its own `Iterator` class](https://github.com/nodejs/node/blob/123693ccbf97df37f928a8ec3e39b85d5e719a7c/lib/internal/url.js#L217).

Thus the need to convert the iterator into an Array :shrug:

gagandeepb (commented):
Thanks, looks great. I wonder if you want to address the below behavior in this PR ?

image

balanza:
> Thanks, looks great. I wonder if you want to address the below behavior in this PR ?
> 
> image

Good point. The current pagination is fake, I think we should not care now as we will implement it soon. 
Without proper pagination the feature won't be usable anyway, so I wouldn't bother for now.

gagandeepb (changes requested):
When I want to see logs from 30 days ago, the cutoff date would be say `2024-07-15T...`, then this date should be passed to the `to_date` parameter, with `from_date` being `utc_now()`. IIUC, the usage here is flipped.

balanza:
> When I want to see logs from 30 days ago, the cutoff date would be say `2024-07-15T...`, then this date should be passed to the `to_date` parameter, with `from_date` being `utc_now()`. IIUC, the usage here is flipped.

Probably I misunderstood the meaning of `from_date` and `to_date`. The way I get it, is the chart below:
image

I created a branch with that use the `to_date` as you suggested: https://github.com/trento-project/web/tree/ui-activity-log-filter-alternate, we can compare the behaviour.

balanza:
There were some issues when converting dates back and forth from UTC to local time zones. They have been addressed in https://github.com/trento-project/web/pull/2898/commits/54e8be5031f86e7256a4957fac78d696c225214e.

Some notes about the modifications:
* the `data-fns-tz` dependency has been added
* helpers `fromZonedTime` and `toZonedTime` are used when needed.
* better formatting has been applied to rendered data items

We might want to move all time zone conversions to the `<DateFilter>` component to segregate the responsibility. That will require a review of the component API, I'll leave it for another PR.

balanza:
@gagandeepb in https://github.com/trento-project/web/pull/2898/commits/8801a4d55bb4ce778e2e086af136de3945d6efd4 I also added the `to_date` filter for cutting off entries older than a given date. I also change the labels as they may be more clear to users:
* _Filter older than_ for `from_date`
* _Filter newer than_ for `to_date`

We might consider adding some semantic validation (valid date range) in the future, but I think it is enough to have users use the feature now.

Tbh, I'm still of the idea that we won't need a _newer than_  filter, as the collection is sequential and the user will have to scroll and scan anyway. But I think we will understand better as soon as we start using the feature with relevant data.

gagandeepb (approved):
Thanks. It already looks much better! :smile:

PR #2903, opened 2024-08-16
User management logs segregation

gagandeepb:
# Description

Only users with `all:all` or `all:users` can see user management logs.

- [x] **DONE**

## How was this tested?

- [x] **DONE**
Unit tests

No docs changes required


lib/trento/activity_log.ex
```diff
@@ -54,10 +62,12 @@ defmodule Trento.ActivityLog do
 
   @spec list_activity_log(map()) ::
           {:ok, list(ActivityLog.t()), Flop.Meta.t()} | {:error, :activity_log_fetch_error}
-  def list_activity_log(params) do
+  def list_activity_log(params, user_has_ability? \\ false) do
```
CDimonaco, line 65: We left on purpose the concept of abilities and "what a user can do with abilities", outside the contexts, please rename this parameter to something else, like `include_all_log_types` or `with_user_management_log_types`.

lib/trento/activity_log.ex
```diff
@@ -67,6 +77,17 @@ defmodule Trento.ActivityLog do
     end
   end
 
+  # for usage in tests only
+  def user_management_log_types do
```
CDimonaco, line 81: Please remove this and leave them in a test file

lib/trento/activity_log.ex
```diff
@@ -67,6 +77,17 @@ defmodule Trento.ActivityLog do
...
+    @user_management_log_types
+  end
+
+  defp scope(ActivityLog = q, true = _user_has_ability?), do: q
```
CDimonaco, line 85: Please rename this function to something more meaningful, like `maybe_with_user_logs bla bla` but not scope

lib/trento/activity_logging/policy.ex
```diff
@@ -0,0 +1,32 @@
...
+  Policy for the activity log.
+
+  User with the ability all:all/all:user can see user management activity log entries.
+  User with all other abilities can not see any user management activity log entries.
+  """
+
+  @behaviour Bodyguard.Policy
+  alias Trento.Support.PolicyHelper, as: Helper
+
+  @all_user_ability %{
```
CDimonaco, line 12: We can use them as parameter directly into the function call when we ask for abilities, no need for a module attribute

lib/trento/activity_logging/policy.ex
```diff
@@ -0,0 +1,32 @@
...
+    name: "all",
+    resource: "users"
+  }
+  def authorize(:get_activity_log, user, _params) do
+    case {Helper.has_global_ability?(user), Helper.user_has_ability?(user, @all_user_ability)} do
```
CDimonaco, line 17: Please use a simple or statement with a one line function

Global ability or user ability.

lib/trento/activity_logging/policy.ex
```diff
@@ -0,0 +1,32 @@
...
+
+      {_, true} ->
+        true
+
+      _ ->
+        false
+    end
+  end
+
+  def authorize(_, _, _) do
```
CDimonaco, line 29: Please one line function

lib/trento_web/controllers/v1/activity_log_controller.ex
```diff
@@ -60,7 +62,12 @@ defmodule TrentoWeb.V1.ActivityLogController do
     ]
 
   def get_activity_log(conn, params) do
-    with {:ok, activity_log_entries, meta} <- ActivityLog.list_activity_log(params) do
+    {current_function, _} = __ENV__.function
```
CDimonaco, line 65: Please use the `:get_activity_log` as an atom and use it as parameter to the bodyguard policy.

If you want to be more `phoenixish` you can use `https://hexdocs.pm/phoenix/Phoenix.Controller.html#action_name/1` to retrieve the action name.

lib/trento_web/controllers/v1/activity_log_controller.ex
```diff
@@ -60,7 +62,12 @@ defmodule TrentoWeb.V1.ActivityLogController do
...
+    user = conn.assigns.current_user
```
CDimonaco, line 66: Current user should be retrieved using Pow primities and pow context.
This is useful because we can change the pow configuration or update the pow library and be able to retrieve the current user, the fact that is hardwired into the assigns with `current_user` as key is a pow config implementation detail.

Use

` %User{} = user = Pow.Plug.current_user(conn)`

lib/trento/activity_logging/policy.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Trento.ActivityLog.Policy do
+  @moduledoc """
+  Policy for the activity log.
+
+  User with the ability all:all/all:user can see user management activity log entries.
+  User with all other abilities can not see any user management activity log entries.
+  """
+
+  @behaviour Bodyguard.Policy
+  alias Trento.Support.PolicyHelper, as: Helper
```
CDimonaco, line 10: Please don't alias as Helper, PolicyHelper is more readable and in context with the file

arbulu89 (commented):
@gagandeepb 
Have you tried the `Bodyguard` scope pattern?
https://hexdocs.pm/bodyguard/readme.html#schema-scopes

It looks pretty much aligned on what you want.
You add the `scope` function in the `ActivityLog` schema file (not the context) and there you can have the `from` implementation that you have now in the context.
You can add arbitrary params to improve pattern matching if you wish.

It looks the more appropriate way to do it, as in fact, you implemented already have the scope function. I think you already had something similar. It looks more accurate than having a `Policy` fail that it looks a bit out of place, as you are not really authorazing, you are scoping.

CDimonaco:
I personally don't like the idea of bodyguard scopes because it forces the pass of the user in the context, something we wanted to avoid from the beginning because it defeats the context segregation we are doing.

For me the feedbacks you addressed are ok, doing other things open the pandora box and can make things much difficult to mantain, so having the context isolated, testable and without any interaction or invariance based on the user it's the correct way

Maybe we could change in the future idk

gagandeepb:
After a conversation with @CDimonaco, I understand that a separate endpoint that returns all logs (including user-management ones) is the preferred way to go. To summarize there will be two endpoints: one that returns all logs and another that returns the subset of all logs without the user-management entries.

balanza:
**question:** Can you tell me more about the rationale behind having to separate endpoints? I read the purpose is to not leak permission concepts into the context, which I can understand. 
 At the same time, would it be bad to check permission at the controller level? The controller's `conn` ships the current user already. I imagine something like this:
 
 ```elixir
   def get_activity_log(conn, params) do
     %User{} = user = Pow.Plug.current_user(conn)
     include_all_log_types = has_all_users_abilities(user)
 
     with {:ok, activity_log_entries, meta} <-
            ActivityLog.list_activity_log(params, include_all_log_types) do
       render(conn, "activity_log.json", %{
         activity_log: activity_log_entries,
         pagination: meta
       })
     end
   end
 
   defp has_all_users_abilities(%User{} = user),
     do:
       PolicyHelper.has_global_ability?(user) or
         PolicyHelper.user_has_ability?(user, %{
           name: "all",
           resource: "users"
         })
 ```
 
 I'm asking because I want to understand more about the _dos and don'ts_ of the permissions.‏

arbulu89:
> **question:** Can you tell me more about the rationale behind having to separate endpoints? I read the purpose is to not leak permission concepts into the context, which I can understand. At the same time, would it be bad to check permission at the controller level? The controller's `conn` ships the current user already. I imagine something like this:
> ...

In fact, I really like this solution. Maybe it could be a temporary solution ( as we will face scoping requirements for other endpoints in the future), but it does the work, and it doesn't add any real user facing differences, it is just code (maybe tech-debt, but internal tech debt).
We don't use bodyguard strangely (which is what it was smelling bad to me) and the code is really simple.
We can have further improvement and generic solutions afterwards

CDimonaco:
Ok guys, you convinced me, let's do @balanza proposal,

Just a little sidenote, rename the PolicyHelper module to AbilitiesHelper :D

arbulu89 (approved):
Thank you @gagandeepb 
Let's just rename `PolicyHelper` to `AbilitiesHelper` (you will need to rename it in multipe occasions), and ready to merge

PD: Or you can do it in an additional PR to keep this one clean

CDimonaco (approved):
LGTM

PR #2905, opened 2024-08-19
OIDC integration

arbulu89:
# Description
OIDC integration enabling the user to use an existing OIDC identity provider to login in Trento.
The user permissions are still managed by trento, so OIDC is only used for pure authentication

## How was this tested?
Tested with UT and E2E

Documentation is being worked in other PRs


CDimonaco (approved):
OIDC ALL THE THINGS!

PR #2906, opened 2024-08-19
Fix OIDC Runtime config

CDimonaco:
# Description

This pr fix the oidc callback url parameter in runtime configuration.
We don't need to inject the port, by default the trento origin will include the port if present


arbulu89 (approved):
Yep, the port is only needed if you don't really use the origin, so it doesn't really make sense

PR #2908, opened 2024-08-19
OIDC integration E2E  tests

arbulu89:
# Description

Add e2e tests for the OIDC feature integration.
In order to run them, we need to start our docker compose with the `--profile idp` flag, as it will use our keycloack deployment in the testing.

In the same case, to run the tests in the github CI, we need to add the `integration` label to the PR.

## How was this tested?

E2e tests


CDimonaco (approved):
Perfect

EMaksy (approved):
Nice trhanks!

PR #2914, opened 2024-08-21
Fix flacky UserPage test

EMaksy:
# Description

Long Ci runs may cause some timing issues in the UserPage test. Adding [waitFor](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) should help.

## How was this tested?

Run locally multiple times, let's rerun it some times in the whole ci.
After running the ci 5 times, no more errors.


arbulu89 (approved):
Perfect!
Appreciated

PR #2915, opened 2024-08-22
Move `<Pagination>` component to dedicated folder

balanza:
# Description

Move the component to a dedicated folder so it has a dedicated test and storybook. This will help for future improvements on the component.


assets/js/common/Pagination/Pagination.test.jsx
```diff
@@ -0,0 +1,74 @@
...
+        itemsPerPageOptions={[10, 20, 50]}
+        onSelect={noop}
+        onChangeItemsPerPage={onChangeItemsPerPage}
+      />
+    );
+
+    await act(() => user.click(screen.getByText('10')));
+    await act(() => user.click(screen.getByText('20')));
+
+    expect(onChangeItemsPerPage).toHaveBeenCalledWith(20);
```
jamie-suse, line 72: I was just reading the earlier tests and I would think that keeping with that style we would loop through selecting all `itemsPerPageOptions` options? What do you think?

jamie-suse (approved):

assets/js/common/Pagination/Pagination.test.jsx
balanza: Sure, done in https://github.com/trento-project/web/pull/2915/commits/8547d4bc949cdc3a274df7125d0f5850e26825fe

PR #2916, opened 2024-08-22
Fix patches for packages' request too long

dottorblaster:
This PR orthogonally applies two optimization in order to:

- Make things actually work;
- Reduce the load on the SUSE Manager instance linked to Trento.

The first optimization is about chunking the list of packages to request the current patches for, since otherwise the URI for the corresponding GET request becomes really massive and in the end the HTTP client just throws an error.

The second optimization is about just requesting patches for packages that a user is actually viewing, dramatically reducing the number of HTTP requests. A new callback called `onPageChange` was added to the `Table` component to achieve this.

## Testing
Jest tests added.

### DIFF
--- a/assets/js/common/Table/Table.jsx
+++ b/assets/js/common/Table/Table.jsx
@@ -2,6 +2,8 @@
 
 import React, { useState, useEffect } from 'react';
 import classNames from 'classnames';
+import { noop } from 'lodash';
+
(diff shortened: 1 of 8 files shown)


dottorblaster:
@janvhs I thought the same and I hit this the hard way instead

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414

PR #2887, opened 2024-08-09
Activity log pagination

nelsonkopliku:
# Description

Wire up activity log pagination ui to the API.

## How was this tested?

Automated tests.

PR #2930, opened 2024-08-29
Activity log add pagination

balanza:
This PR allows browsing the entire activity log by scrolling different pages.

Key features:
* The default page size is 20 items, which can be changed due to the options in the dropdown
* the list can be navigated sequentially forward and backward
* if there are no previous entries, the `[<]` button is disabled. The same goes for the `[>]` if no entries are next.
* when filters change, the pagination is reset to the first page

### DIFF
--- a/assets/js/common/ActivityLogOverview/ActivityLogOverview.jsx
+++ b/assets/js/common/ActivityLogOverview/ActivityLogOverview.jsx
@@ -56,7 +56,7 @@ function ActivityLogOverview({
   const [selectedEntry, setEntry] = useState({});
 
   const activityLogTableConfig = {
-    pagination: true,
+    pagination: false,
     usePadding: false,
     columns: [
       {
--- a/assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
+++ b/assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
@@ -7,10 +7,15 @@ import ComposedFilter from '@common/ComposedFilter';
 
 import { getActivityLog } from '@lib/api/activityLogs';
 import { ACTIVITY_TYPES_CONFIG } from '@lib/model/activityLog';
+import { PaginationPrevNext } from '@common/Pagination/Pagination';
+import { pipe } from 'lodash/fp';
 import {
-  filterValueToSearchParams,
+  getItemsPerPageFromSearchParams,
+  resetPaginationToSearchParams,
   searchParamsToAPIParams,
   searchParamsToFilterValue,
+  setFilterValueToSearchParams,
+  setPaginationToSearchParams,
 } from './searchParams';
 
 const filters = [
@@ -37,19 +42,47 @@ const filters = [
   },
 ];
 
+const defaultItemsPerPage = 20;
+const itemsPerPageOptions = [10, 20, 50, 75, 100];
+
+const changeItemsPerPage = (searchParams) => (items) => {
+  if (searchParams.has('after')) {
+    return {
+      first: items,
+      after: searchParams.get('after'),
+    };
+  }
+  if (searchParams.has('before')) {
+    return {
+      last: items,
+      before: searchParams.get('before'),
+    };
+  }
+  return { first: items };
+};
+
 function ActivityLogPage() {
-  const [searchParams, setSearchParams] = useSearchParams();
+  const [searchParams, setSearchParams] = useSearchParams(
+    resetPaginationToSearchParams(defaultItemsPerPage)()
+  );
   const [activityLog, setActivityLog] = useState([]);
+  const [activityLogMetadata, setActivityLogMetadata] = useState({});
+
   const [isLoading, setLoading] = useState(true);
   const [activityLogDetailModalOpen, setActivityLogDetailModalOpen] =
     useState(false);
+
+  const itemsPerPage = pipe(getItemsPerPageFromSearchParams, (n) =>
+    itemsPerPageOptions.includes(n) ? n : itemsPerPageOptions[0]
+  )(searchParams);
 
   const fetchActivityLog = () => {
     setLoading(true);
     const params = searchParamsToAPIParams(searchParams);
     getActivityLog(params)
-      .then((response) => {
-        setActivityLog(response.data?.data ?? []);
+      .then(({ data: { data = [], ...metadata } = {} }) => {
+        setActivityLog(data);
+        setActivityLogMetadata(metadata);
       })
       .catch(() => setActivityLog([]))
       .finally(() => {
@@ -71,7 +104,11 @@ function ActivityLogPage() {
             filters={filters}
             autoApply={false}
             value={searchParamsToFilterValue(searchParams)}
-            onChange={(p) => setSearchParams(filterValueToSearchParams(p))}
+            onChange={pipe(
+              setFilterValueToSearchParams,
+              resetPaginationToSearchParams(itemsPerPage),
+              setSearchParams
+            )}
           />
         </div>
         <ActivityLogOverview
@@ -82,6 +119,31 @@ function ActivityLogPage() {
           onCloseActivityLogEntryDetails={() =>
             setActivityLogDetailModalOpen(false)
           }
+        />
+        <PaginationPrevNext
+          hasPrev={activityLogMetadata.pagination?.has_previous_page}
+          hasNext={activityLogMetadata.pagination?.has_next_page}
+          currentItemsPerPage={itemsPerPage}
+          itemsPerPageOptions={itemsPerPageOptions}
+          onSelect={pipe(
+            (selection) =>
+              selection === 'prev'
+                ? {
+                    last: itemsPerPage,
+                    before: activityLogMetadata.pagination.start_cursor,
+                  }
+                : {
+                    first: itemsPerPage,
+                    after: activityLogMetadata.pagination.end_cursor,
+                  },
+            setPaginationToSearchParams,
+            setSearchParams
+          )}
+          onChangeItemsPerPage={pipe(
+            changeItemsPerPage(searchParams),
+            setPaginationToSearchParams,
+            setSearchParams
+          )}
         />
       </div>
     </>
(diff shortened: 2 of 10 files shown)


assets/js/common/ActivityLogOverview/ActivityLogOverview.jsx
```diff
@@ -56,7 +56,7 @@ function ActivityLogOverview({
   const [selectedEntry, setEntry] = useState({});
 
   const activityLogTableConfig = {
-    pagination: true,
+    pagination: false,
```
nelsonkopliku, line 59: **nitpick:** In order to disable a table's pagination it is sufficient not opting in for it
```suggestion
```
---
**thought:** as a developer I would find it useful being able to instrument the table component to use either a page based or cursor based pagination.‏
ie:
```js
const tableConfig: {
    pagination: true, // <-- If I want the page based pagination
    cursorPagination: true // <-- If I want the cursor based pagination
}
```
This might be overly simplistic as we might need further configuration based on the chosen pagination (ie what to do on page selection or what to do on prev/next click) so we could consider options like:
```js
const tableConfig: {
    pagination: {
        onThis: () => {},
        onThat: () => {},
    },
    cursorPagination: {
        onThis: () => {},
        onThat: () => {},
    },
}

// or even
const tableConfig: {
    pagination: {
        type: 'page_based' // or 'cursor_based' 
        onSomethingValidOnlyForPageBasedPagination: () => {},
        onSomethingValidOnlyForCursorBasedPagination: () => {}
    }
}
```
Currently if we want a table with page based pagination we can configure the table, otherwise if we want a table with cursor based pagination we need a different approach and I see an opportunity to improve semantics and consistency. In whatever direction.

I am aware that _currently_ tables do not strictly necessarily need cursor based pagination, except for the activity log one, so I'd be fine with this. Let's see how it evolves.

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
```diff
@@ -37,19 +42,47 @@ const filters = [
...
+    itemsPerPageOptions.includes(n) ? n : itemsPerPageOptions[0]
+  )(searchParams);
+
   const fetchActivityLog = () => {
     setLoading(true);
     const params = searchParamsToAPIParams(searchParams);
     getActivityLog(params)
-      .then((response) => {
-        setActivityLog(response.data?.data ?? []);
+      .then(({ data: { data = [], ...metadata } = {} }) => {
```
nelsonkopliku, line 83: **suggestion:** what about destructuring `pagination` or `paginationMetadata` to not create confusion with a log entry metadata? 
```suggestion
      .then(({ data: { data = [], ...pagination } = {} }) => {
```
Also related state and state setter `activityLogMetadata` and `setActivityLogMetadata`

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
```diff
@@ -83,6 +120,31 @@ function ActivityLogPage() {
             setActivityLogDetailModalOpen(false)
           }
         />
+        <PaginationPrevNext
```
nelsonkopliku, line 123: **question:** what made you opt for placing this component here and not in `ActivityLogOverview` after the table?‏
I see it fitting there better.

nelsonkopliku (commented):
Thanks for this @balanza!

Besides what already commented here's a few extra feedback raised also during the review:
**Filter/pagination retention:**
- changing filters should keep previously applied pagination
- changing results per page should keep previously applied filters
- navigating backwards or forward in the activity log should retain previously applied filters

**Visual** (we can address as a followup): 
`PREV_LABEL` and `NEXT_LABEL` should be respectively `EOS_KEYBOARD_DOUBLE_ARROW_LEFT` and `EOS_KEYBOARD_DOUBLE_ARROW_RIGHT`

**praise:** ‏well done in keeping filters in the query string. It enables for fast sharing of filtered activity log :+1:

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
balanza: `metadata` is an object that contains all the sub-objects including the pagination. If we rename like this we will have `pagination.pagination`. Maybe we can rename it like

```js
      .then(({ data: { data = [], pagination } = {} }) => {
```

It's what you intended?

nelsonkopliku: otherwise, if we need/want to keep access to the whole rest of the response, we could actually go after what it is: `activityLogResponse` or the like.

assets/js/common/ActivityLogOverview/ActivityLogOverview.jsx
balanza: The difference in pagination is not numbered/cursor but controlled/uncontrolled.

The pagination in `<Table>` uses the local state of the component. The kind of pagination we need in activity log, instead, needs to be controlled by the page component, as every changes involves a HTTP request. 

Given that 99% of our tables use local data, I opted for the following architecture:
```
IFyou need numbered&uncontrolled pagination
THEN you can use build-in pagination
ELSE you must override it with a custom implementation
```

Should we aim for a _do-it-all_ `<Table>` component?

I think providing the `<Table>` component with both controlled and uncontrolled pagination (combined with numbered/cursor) will lead to an overengineered `<Table>` component. We may prefer composition to inheritance/parametrization, hence if we don't like the logic above I praise for Separating the pagination from the table rendering instead of stuffing everything into one component.

We may want to refactor the component, though. IMHO we should do it in a dedicated PR. We can:
1. block this PR until such refactor
2. merge this PR, then refactor
3. decide we don't need such refactor

Thoughts?

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
balanza: I thought the same before I started to develop. Then I realized that actions on the pagination trigger HTTP requests that must be intercepted and executed at the page level. If we put pagination inside `ActivityLogOverview` we need to bubble up events and have a _prop-drilling_ situation.

assets/js/common/ActivityLogOverview/ActivityLogOverview.jsx
nelsonkopliku: I understand the rationale about 99% of our tables use local data, and as mentioned I am fine with the proposal :smile: 

No, let's not block this PR and let's merge it.
I am coming from a consistency/semantic stance, but I understand the reasons and I can live with this currently and see how it evolves.

balanza:
> changing filters should keep previously applied pagination

I added such a requirement to prevent unclear experiences for users, but I guess they have been considered minor issues by the team during the review. Done in https://github.com/trento-project/web/pull/2930/commits/4359d5ca2de8b733150117e5f26b30de737e3771


> changing results per page should keep previously applied filters
> navigating backwards or forward in the activity log should retain previously applied filters

It should already be like this, have you found something different while trying? if so it's not intended, it's a bug.

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
nelsonkopliku: Yes, indeed we'd be passing props to `ActivityLogOverview` that will be just then passed down to `PaginationPrevNext`.

I am not sure what is _less bad_: prop-drilling or having `PaginationPrevNext` here :smile: 
Up to you, no strong opinions.

What would you do to mitigate prop-drilling and still have the `PaginationPrevNext` next to the table?
React contexts? HOCs? Other tecniques?

balanza: I see no evil in leaving the pagination at page level, to be honest. The purpose of having `ActivityLogOverview` is to have unit tests on that, and adding pagination won't allow us to add any meaningful test imho - we should just test "the prop is assigned to the component" or replicate things already tested in the pagination component itself.

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
balanza: Good, I think it's more neat. Done in https://github.com/trento-project/web/pull/2930/commits/c8d3e60c918dc593fe8333de36ca3cb6b503ddc6

nelsonkopliku:
> It should already be like this, have you found something different while trying? if so it's not intended, it's a bug.
>
I tried the most obvious things:
Scenario 1:
- filter by an activity type (which returns a long enough list to be paginated)
- navigate on next page

The first request has the `type`, the second one has `first` and `after`, but not the previously selected type (also date filters are lost)

Scenario 2:
- filter by an activity type (which returns a long enough list to be paginated)
- change the `Results per page`

The first request has the `type`, the second one has only `first`, but not the previously selected type (also date filters are lost)

> I added such a requirement to prevent unclear experiences for users
>
Maybe I am missing some cases where things go weird :eyes: I also got some edge cases while implementing this https://github.com/trento-project/web/pull/2887/files maybe there is something useful there?

nelsonkopliku:
> Maybe I am missing some cases where things go weird 👀 I also got some edge cases while implementing this https://github.com/trento-project/web/pull/2887/files maybe there is something useful there?
> 
> EDIT: anyway I guess that resetting pagination on filter change could be acceptable, as long as we keep filters while paginating.

@balanza I guess there might be some misunderstanding about _keep previously applied pagination_.
That means keeping the previously applied `Results per page`, which does not conflict with what you mentioned in the PR's description
> when filters change, the pagination is reset to the first page
>
which is still valid and the way to go.

Anyway it's a minor thing, I don't mind much if when applying new filters we reset `Results per page` to the default of 20,  as long as we keep filters while navigating pages.

nelsonkopliku:
Hey @balanza, in your absence I tried addressing what we were discussing in the last two commits.

However, let's check in together when you're back. I am sure that together we can improve :smile:

balanza:
Closing this as the suggestion will be addressed in #2960, #2961 and #2967 for more-scoped activities.

PR #2949, opened 2024-09-06
Table Header Alignment

jagabomb:
# Description

Tables displaying the text as left aligned for both the table header and rows had a slight misalignment of about 4px. This PR attempts to make the horizontal padding the same value for consistency purposes. There is also the inclusion of test to prevent values being changed independently and to prevent this issue resurfacing.

**Before** (Misalignment highlighted in red)
Frame 56

**Changes**
Screenshot 2024-09-06 at 12 09 42

## How was this tested?

- [x] **DONE** Check for the same 'px-?' value between the table header and rows.

### DIFF
--- a/assets/js/common/Table/Table.jsx
+++ b/assets/js/common/Table/Table.jsx
@@ -206,7 +206,7 @@ function Table({
                             sortable
                               ? 'cursor-pointer hover:text-gray-700 '
                               : ''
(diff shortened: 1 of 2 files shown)


dottorblaster (approved):
LGTM with a small nitpick :+1:

jamie-suse (approved):
Let's go :rocket:

PR #2951, opened 2024-09-09
WIP injecting component names into activity log

nelsonkopliku:
# Description

This is an exploration about the possibility to enrich Activity log metadata.

The rationale behind is improving human-friendly identification of the resource the activity log entry is related to and improving log searchability.

The thing could be summarized as: adding a human friendly resource name to the activity log metadata when it does not already have one

- host -> hostname
- cluster -> cluster name
- database -> sid
- sap system -> sid
- application instance -> instance number
- database instance -> instance number

Alternative to the approach tried in this PR could be upcasting all necessary events to also carry the extra information needed. However this would not be sufficient for non event sourced related activities: ie when tagging/untagging we still need some sort of metadata enrichment process to get the resource name being tagged/untagged.

### DIFF
--- a/assets/js/lib/model/activityLog.js
+++ b/assets/js/lib/model/activityLog.js
@@ -131,13 +131,13 @@ export const ACTIVITY_TYPES_CONFIG = {
   [RESOURCE_TAGGING]: {
     label: 'Tag Added',
     message: ({ metadata }) =>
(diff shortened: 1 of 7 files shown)


nelsonkopliku:
Followed up by https://github.com/trento-project/web/pull/2992, https://github.com/trento-project/web/pull/2996 and https://github.com/trento-project/web/pull/2997

PR #2959, opened 2024-09-10
General Form Label Alignment Fixes

jagabomb:
# Description

This PR makes the following improvements to the **UserForm** and **ProfileForm** components.

1. Improves the alignment of the **Label** so that it is more vertically centred with each corresponding input field.
2. Increases the width of the **Label** (column) so that slightly longer labels (e.g: Confirm Password and Authenticator App) don't wrap over two lines.

Below are the comparisions:
UserForm
ProfileForm

## How was this tested?

- [x] **DONE** Visually


dottorblaster (approved):
Thanks our god and savior of good UI :grimacing:

PR #2960, opened 2024-09-10
Refactor `<DateFilter>`

balanza:
# Description

Being it uses the browser native `<input type="datetime-local">` HTML element, the component needs to convert dates in the local timezone from/to UTC. 
This PR uses formatting instead of timezone offset, so the conversion is idempotent if applied multiple times. 

This avoid the need to control balance the number of conversion from and to the target timezone (as it's the case in a controlled component scenario)


nelsonkopliku (approved):
LGTM :eyes:

PR #2961, opened 2024-09-10
Add default pagination values

balanza:
# Description

Expose pagination defaults as discussed in https://github.com/trento-project/web/pull/2930


nelsonkopliku (approved):
:eyes:

PR #2968, opened 2024-09-11
Adust length of "Results per page"-field

janvhs:
This commit adjusts the length of the "Results per page"-field in the views containing pagination.  The field now adjusts its size based on its content.  For this, I use a margin on the chevron element, but utilising CSS' after pseudo class would be better

- [x] **DONE**

## How was this tested?

Visually

### DIFF
--- a/assets/js/common/Select/Select.jsx
+++ b/assets/js/common/Select/Select.jsx
@@ -31,7 +31,7 @@ function Select({
   )}-selection-dropdown`;
 
   return (
-    <div className={classNames('w-64 flex-2', className)}>
+    <div className={classNames('flex-2', className)}>
       <Listbox disabled={disabled} value={value} onChange={onChange}>
         <div className="relative mt-1">
           <Listbox.Button
--- a/assets/js/common/ActivityLogsSettingsModal/ActivityLogsSettingsModal.jsx
+++ b/assets/js/common/ActivityLogsSettingsModal/ActivityLogsSettingsModal.jsx
@@ -45,6 +45,7 @@ function TimeSpan({ time: initialTime, error = false, onChange = noop }) {
       </div>
       <div className="flex w-1/4">
         <Select
+          className="w-64"
           optionsName=""
           options={timeUnitOptions}
           value={time.unit || defaultTimeUnit}
--- a/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
+++ b/assets/js/common/ApiKeySettingsModal/ApiKeySettingsModal.jsx
@@ -152,7 +152,7 @@ function ApiKeySettingsModal({
               </div>
               <div className="w-2/4 pt-4">
                 <Select
-                  className="pb-4"
+                  className="pb-4 w-64"
                   optionsName=""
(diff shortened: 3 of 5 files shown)


assets/js/common/Pagination/Pagination.jsx
```diff
@@ -38,6 +38,23 @@ const containerClassNames = 'flex items-center';
...
+  const lowerBound = itemsBase + 1;
+  const upperBound = itemsBase + itemsPresent;
+
+  return (
+    <span className="mr-4 text-gray-600">
+      {lowerBound}–{upperBound} of {itemsTotal}
+    </span>
+  );
+}
+
```
dottorblaster, line 57: Could you move this to another file?

dottorblaster: Also I think you wanna give this a `text-sm` class in order to make the text a little bit smaller

assets/js/common/Select/Select.jsx
```diff
@@ -31,13 +31,13 @@ function Select({
...
   return (
-    <div className={classNames('flex-2 w-64', className)}>
+    <div className={classNames('flex-2', className)}>
       <Listbox disabled={disabled} value={value} onChange={onChange}>
         <div className="relative mt-1">
           <Listbox.Button
             className={classNames(
               dropdownSelector,
-              'relative w-full py-2 px-3 text-left bg-white rounded-lg cursor-default border border-gray-300 focus:outline-none focus-visible:ring-2 focus-visible:ring-opacity-75 focus-visible:ring-white focus-visible:ring-offset-orange-300 focus-visible:ring-offset-2 focus-visible:border-indigo-500 sm:text-sm disabled:bg-gray-50 disabled:text-gray-400'
+              'relative w-full ml-7 py-2 px-3 text-left bg-white rounded-lg cursor-default border border-gray-300 focus:outline-none focus-visible:ring-2 focus-visible:ring-opacity-75 focus-visible:ring-white focus-visible:ring-offset-orange-300 focus-visible:ring-offset-2 focus-visible:border-indigo-500 sm:text-sm disabled:bg-gray-50 disabled:text-gray-400'
```
dottorblaster, line 40: Do we need this `ml-7`? Is it something we are making just for the footer of the table and we're letting it percolate into a generic component?

dottorblaster (changes requested):
Just some minor changes. Also, could you write a Jest test for this?

assets/js/common/Pagination/Pagination.jsx
jagabomb: I agree with @dottorblaster here. The text size need to be smaller so that it not disproportional in the table footer section and that it also matches the size as the "**Results per page**" text.

jagabomb (commented):
A small note from my side... I think component should have been aligned next to the **Results per page dropdown** component. 
Screenshot 2024-09-13 at 13 37 28

jagabomb (approved):
LGTM

dottorblaster (approved):
LGTM

dottorblaster:
@janvhs all good here, smash the green button :+1:

PR #2967, opened 2024-09-11
Activity log add pagination (rework)

balanza:
# Description

Retaking the work from https://github.com/trento-project/web/pull/2930 as the original PR got too messy and required more work for rebasing on the main branch. Also, some of the suggestions have been addressed in https://github.com/trento-project/web/pull/2961 and https://github.com/trento-project/web/pull/2960.

In respect of the original PR, this one keeps the proposed key features:

* The default page size is 20 items, which can be changed due to the options in the dropdown
* the list can be navigated sequentially forward and backward
* if there are no previous entries, the [<] button is disabled. The same goes for the [>] if no entries are next.
* when filters change, the pagination is reset to the first page

It also addressed the found issues:
* items per page to be maintained when scrolling pages
* filters to be maintained when scrolling pages
* default of 20 elements per page
* no duplicate requests when clicking multiple times on the navigation link

-----

Depends on #2960
Depends on #2961

PR #2977, opened 2024-09-13
Fix `<ComposedFilter>` behavior as controlled component

balanza:
# Description

The  `<ComposedFilter>` can now be used as a controlled component. The use case is when the `value` changes in parent component and thus it needs to be re-rendered.

Before this PR the `<ComposedFilter>` wouldn't show the changed value as the `useState` hook prevents it. The component re-renders as expected by forcing `setValue` on the initial value change. 

The bug has been spotted working on https://github.com/trento-project/web/pull/2967#pullrequestreview-2302325045, an e2e test has been added for that.


nelsonkopliku (approved):
Thank you!

gagandeepb (approved):

PR #2983, opened 2024-09-16
Reset pagination when filter changes

balanza:
# Description

`after` and `before` must be removed when changing filters, as the pagination only applies to a given query.


assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
```diff
@@ -133,6 +134,7 @@ function ActivityLogPage() {
             value={searchParamsToFilterValue(searchParams)}
             onChange={pipe(
               setFilterValueToSearchParams,
+              resetPaginationToSearchParams(),
```
nelsonkopliku, line 137: By adding back `resetPaginationToSearchParams` I guess we could `resetPaginationToSearchParams(itemsPerPage)` and remove the following `applyItemsPerPage(itemsPerPage)` and the function altogether since `resetPaginationToSearchParams` already covers the need.

nelsonkopliku (approved):
Awesome, thanks!

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
balanza: Good catch. Actually, I prefer to keep them separated. As there is no need for the items number, I simplified the code like this https://github.com/trento-project/web/pull/2983/commits/87721bbb6792122c9e82e494292d370980861bd5

PR #2986, opened 2024-09-16
Adjust activity log columns width

nelsonkopliku:
# Description

Tiny colums width fix as per collected feedback.


jagabomb (approved):
LGTM

PR #2995, opened 2024-09-18
Enable SAML messages signing

arbulu89:
# Description

Enable SAML messages signing with SSL for dev environment.
Just it mimics better a real world application

I just uploaded the dev certs to the `container_fixtures` folder as they are related to keycloak. 
I didn't think putting them in `priv` was the best.
I have uploaded the `metadata.xml` file as well, less things to worry about for developers (if the realm is changed, it must be changed of course)

But I accept proposals (maybe just `priv/cert` is the correct folder...)


EMaksy (approved):
LGTM!

CDimonaco (approved):

PR #2996, opened 2024-09-18
Activity log metadata enrichment

nelsonkopliku:
# Description

This PR introduces a `MetadataEnricher` enriching metadata extracted from activity parsing.

Depends on https://github.com/trento-project/web/pull/2992, see also https://github.com/trento-project/web/pull/2951

The rationale behind is improving human-friendly identification of the resource the activity log entry is related to and improving log searchability.

The thing could be summarized as: adding a human friendly resource name to the activity log metadata when it does not already have one

- host -> hostname - field named `hostname`
- cluster -> cluster name - field named `name` 
- database -> sid - field named `sid`
- sap system -> sid - field named `sid`
- application instance -> instance number (always present when needed, no enrichment required)
- database instance -> instance number (always present when needed, no enrichment required)

I just kept field names as they come from the read model (that is why we have `hostname` for a host and `name` for a cluster).
Had thought of options where:
- we always enrich with `component_name`|`resource_name`|`name`  or the like
- making `name` for clusters `clustername` to be consistent with `hostname`

Open for feedback.

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/activity_logging/parser/metadata_enricher.ex
+++ b/lib/trento/activity_logging/parser/metadata_enricher.ex
@@ -1,0 +1,125 @@
+defmodule Trento.ActivityLog.Logger.Parser.MetadataEnricher do
+  @moduledoc """
+  Metadata enricher enriches metadata extracted by activity parser.
+  """
+
+  alias Trento.ActivityLog.ActivityCatalog
+
+  alias Trento.Clusters.Projections.ClusterReadModel
+  alias Trento.Databases.Projections.DatabaseReadModel
+  alias Trento.Hosts.Projections.HostReadModel
+  alias Trento.SapSystems.Projections.SapSystemReadModel
+
+  alias Trento.{Clusters, Databases, Hosts, SapSystems}
+
+  @spec enrich(activity :: ActivityCatalog.activity_type(), metadata :: map()) ::
+          {:ok, maybe_enriched_metadata :: map()}
+  def enrich(activity, metadata) do
+    case ActivityCatalog.detect_activity_category(activity) do
+      supported_activities
+      when supported_activities in [:connection_activity, :domain_event_activity] ->
+        {:ok, enrich_metadata(activity, metadata)}
+
+      _ ->
+        {:ok, metadata}
+    end
+  end
+
+  defp enrich_metadata(activity, metadata),
+    do:
+      {activity, metadata}
+      |> maybe_enrich_with_hostname()
+      |> maybe_enrich_with_cluster_name()
+      |> maybe_enrich_with_database_sid()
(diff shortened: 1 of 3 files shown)


lib/trento/activity_logging/parser/metadata_enricher.ex
```diff
@@ -0,0 +1,125 @@
...
+    enriched_metadata =
+      with {:ok, sap_system_id} <- detect_enrichment(:sap_system, {activity, metadata}),
+           {:ok, %SapSystemReadModel{sid: sid}} <- SapSystems.by_id(sap_system_id) do
+        Map.put(metadata, :sid, sid)
+      else
+        _ -> metadata
+      end
+
+    {activity, enriched_metadata}
+  end
```
gagandeepb, line 83: These functions are identical except for a parameter (i.e. one of host, database, sap system etc.). What are your thoughts about re-parameterizing/genericizing this function into something like `maybe_enrich({activity, metadata}, CtxModule, field_to_extract, entity_atom)` where CtxModule is `SapSystems | Databases`, field to extract is `:name | :sid ...`, entity_atom is `:database |  :sap_system ...`.

lib/trento/activity_logging/parser/metadata_enricher.ex
```diff
@@ -0,0 +1,125 @@
...
+  @moduledoc """
+  Metadata enricher enriches metadata extracted by activity parser.
+  """
+
+  alias Trento.ActivityLog.ActivityCatalog
+
+  alias Trento.Clusters.Projections.ClusterReadModel
+  alias Trento.Databases.Projections.DatabaseReadModel
+  alias Trento.Hosts.Projections.HostReadModel
+  alias Trento.SapSystems.Projections.SapSystemReadModel
```
gagandeepb, line 11: These are deep dependencies, not top level contexts. What do we lose if we do not assert on the type of the return (e.g. `Hosts.by_id()`) ?

nelsonkopliku: We wouldn't be losing much. Here we go a56f2fadf47014161506788a4defb2e62c9a47cc

lib/trento/activity_logging/parser/metadata_enricher.ex
nelsonkopliku: I thought about that, thanks for helping me fighting my laziness :smile:  :see_no_evil: 

Here's the change 03f6d8c9dfa3001fede760fc8601c96df0ee02d6: if we're gonna be needing to let's say enrich with multiple values, we'll revisit what needed :+1:

gagandeepb (approved):

balanza (approved):

PR #2998, opened 2024-09-19
Improve domain event activity detection

nelsonkopliku:
# Description

Fixing the way we detect domain events related activities and how we ignore legacy events.

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/activity_logging/activity_catalog.ex
+++ b/lib/trento/activity_logging/activity_catalog.ex
@@ -107,11 +107,7 @@ defmodule Trento.ActivityLog.ActivityCatalog do
             false
         end)
         |> Enum.map(&Module.concat/1)
-        |> Enum.filter(fn module ->
-          Code.ensure_loaded?(module)
-
-          not function_exported?(module, :legacy?, 0)
-        end)
+        |> Enum.filter(&(not &1.legacy?()))
         |> Map.new(fn event_module ->
           {event_module,
            {event_module
--- a/lib/trento/activity_logging/activity_catalog.ex
+++ b/lib/trento/activity_logging/activity_catalog.ex
@@ -107,7 +107,11 @@ defmodule Trento.ActivityLog.ActivityCatalog do
             false
         end)
         |> Enum.map(&Module.concat/1)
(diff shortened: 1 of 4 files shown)


lib/trento/activity_logging/activity_catalog.ex
```diff
@@ -107,7 +107,11 @@ defmodule Trento.ActivityLog.ActivityCatalog do
             false
         end)
         |> Enum.map(&Module.concat/1)
-        |> Enum.filter(&(not function_exported?(&1, :legacy, 0)))
+        |> Enum.filter(fn module ->
+          Code.ensure_loaded?(module)
```
gagandeepb, line 111: I wonder what is the expected/desired behavior here in case of successful or unsuccessful response ? Additionally, what are the scenarios at run time when this will return an error response ?

nelsonkopliku: We're calling `Code.ensure_loaded?(module)` to make sure that the following [`function_exported?`](https://hexdocs.pm/elixir/Kernel.html#function_exported?/3) works as expected.

> what is the expected/desired behavior here in case of successful or unsuccessful response
>
Unless something very wrong with beam, I wouldn't expect `Code.ensure_loaded?(module)` to fail in this context as `module` is always an actual module (probably not loaded yet) that we get from `:application.get_key(:trento, :modules)`.

> what are the scenarios at run time when this will return an error response
>
None that I am aware of, unless something very wrong with beam is going on.

---

Alternatively to this we could just always have a `legacy?/0` function in events returning a boolean: like here e9be79658f0d124346c7216c017ba5ce74aac182

gagandeepb (approved):
I see, now. As I understand it, using `function_exported?` on event modules has a runtime implication about module load order at application startup. Executing this before the other dependent modules are loaded (as implied by the Application ordering) can be problematic. This load order problem gets resolved by the early resolution provided by having the `legacy?` declaration so now it is enough for the necessary code to be available as compiled at run time (usually/always true) but not necessarily loaded. I prefer the latter, i.e. early resolution. Thanks for looking into this.

PR #2999, opened 2024-09-19
SAML runtime operations

arbulu89:
# Description

Add SAML release task initialization and runtime options.

A new `/api/public_keys` route has been added in order to show the user the uploaded keys, so they can copy the certificate in their IDP configuration

A simple usage of the env variables consists of:
```
ENABLE_SAML=true SAML_IDP_ID=saml SAML_SP_DIR=/tmp/saml SAML_SP_ID=trento-saml TRENTO_WEB_ORIGIN=localhost:4000 SAML_METDATA_URL=http://localhost:8081/realms/trento/protocol/saml/descriptor SAML_SP_ENTITY_ID=trento-web-saml ........
```

## How was this tested?
Manual tests with created releases + UT

No, we will do this at the end. Specially for this case, where the setup and configuration is slightly more difficult than other integrations


CDimonaco (approved):
:chef_kiss:

EMaksy (approved):
NICE!

PR #3005, opened 2024-09-20
Patches for packages alternative approach

dottorblaster:
# Description
This PR leverages the existing list of vulnerabilities for the host in order to achieve a more comprehensive list of patches available for the upgradable packages view coming from SUSE Manager.

## How was this tested?
Automated tests updated


dottorblaster:
Just sending this on main despite the BC check because we didn't have the chance to deploy it to actual user, so in the end nothing would break.

PR #3011, opened 2024-09-23
SAML configurable attr fields

arbulu89:
# Description

As the title says, make SAML user profile attributes configurable.
The username, email, first name and last name attributes must continue existing, even though the user can have different naming for them in their SAML configuration.

## How was this tested?

UT + manual testing


EMaksy (approved):
LGTM!

CDimonaco (approved):

PR #3016, opened 2024-09-24
Improve Activity Log page request lifecycle

balanza:
# Description

Refine how the Activity log page handles the API request
* define the HTTP request lifecycle in `lib/api/request`
* show an empty view initially
* show an error view in case an error occurred
* add a spinner when loading data

### DIFF
--- a/assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
+++ b/assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
@@ -6,6 +6,7 @@ import { EOS_REFRESH } from 'eos-icons-react';
 import { map, pipe } from 'lodash/fp';
 
 import { getActivityLog } from '@lib/api/activityLogs';
+import * as request from '@lib/api/request';
 import { allowedActivities } from '@lib/model/activityLog';
 import { getActivityLogUsers } from '@state/selectors/activityLog';
 import { getUserProfile } from '@state/selectors/user';
@@ -15,7 +16,11 @@ import PageHeader from '@common/PageHeader';
 import ActivityLogOverview from '@common/ActivityLogOverview';
 import ComposedFilter from '@common/ComposedFilter';
 import Pagination, { defaultItemsPerPageOptions } from '@common/Pagination';
+import Spinner from '@common/Spinner';
 
+import ConnectionErrorAntenna from '@static/connection-error-antenna.svg';
+
+import NotificationBox from '@common/NotificationBox';
 import {
   applyItemsPerPage,
   setFilterValueToSearchParams,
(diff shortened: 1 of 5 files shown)


assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
```diff
@@ -103,19 +156,21 @@ function ActivityLogPage() {
...
-          data: data?.data || [],
-          pagination: data?.pagination,
-        });
-      })
-      .catch(() => setActivityLogResponse(emptyResponse))
-      .finally(() => {
-        setLoading(false);
-      });
+    // defer the loading state to avoid flickering
+    const tid = setTimeout(() => setActivityLogRequest(request.loading()), 500);
```
nelsonkopliku, line 160: **question:** how can the flickering be reproduced?

nelsonkopliku (commented):
Interesting the introduction of the `MainView` and the modeling of the lifecycle of an API request. Curious about the possibility to be reused elsewhere.

However tests for `ActivityLogOverview` and `ActivityLogPage` are failing and need to be adjusted.

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
```diff
@@ -52,19 +52,72 @@ const applyDefaultItemsPerPage = (params) =>
...
+  }
+  if (status === 'loading') {
+    return (
+      <div className="flex flex-row min-h-screen justify-center items-center">
+        <Spinner size="xl" />
+      </div>
+    );
+  }
+  if (status === 'failure') {
+    return <div>An error occurred.</div>;
```
nelsonkopliku, line 78: **suggestion:** we might want to style a tiny bit this case.

<details>
  <summary>See a screenshot here</summary>
  
image
</details>

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
balanza: Good question, it is something visually annoying that has no functional effect. Maybe it only applies to local executions, when the latency is near zero.

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
balanza: I reused the component from the other pages

image

nelsonkopliku: Awesome, thanks!

assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
nelsonkopliku: Alright thanks for clarifying.

nelsonkopliku (approved):
Besides the conflict to be resolved (my bad :smile: ) LGTM! Thanks!

PR #3026, opened 2024-09-26
Streamline Pagination

nelsonkopliku:
# Description

This PR addresses following Pagination aspects:
- icons are used as per design
- pagination streamlined application-wide

I kept the legacy Pagination as a `PageNumberPagination` component. If no one against it I will remove in a follow up PR.

## How was this tested?

Existent tests.


gagandeepb (approved):

balanza (approved):
LGTM

Thanks, I thought it would take more hassle to switch to vector icons.

About the `PageNumberPagination`: I'm keen to remove it as we are not using it now. We can always resume the code from the history. If you do, remember to also uninstall the `react-paginate` packages as it's used nowhere else.

PR #3040, opened 2024-10-01
Discarded discovery events to be recorded 

balanza:
# Description

Discovery Events that fail to parse into commands are discarded and stored in the `discarded_discovery_events` table.

However, if a Discovery Event requires a step in the `before_dispatch` handler of a Commanded Middleware, and the step fails, the Discovery Event is still discarded, but it's not saved.

So far, this only applies to application instance registration events in the context of a sap system discovery.

### DIFF
--- a/test/trento_web/controllers/v1/discovery_controller_test.exs
+++ b/test/trento_web/controllers/v1/discovery_controller_test.exs
@@ -3,10 +3,11 @@ defmodule TrentoWeb.V1.DiscoveryControllerTest do
 
   import Trento.DiscoveryFixturesHelper
 
-  alias Trento.Discovery.{
-    DiscardedDiscoveryEvent,
-    DiscoveryEvent
-  }
+  import Mox
+
+  alias Trento.Discovery
+
+  alias Trento.Discovery.DiscardedDiscoveryEvent
 
   describe "discovery" do
     test "collect action should return an unprocessable entity error when the event is unknown",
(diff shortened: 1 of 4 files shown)


arbulu89 (approved):
Looks good!
One step forward to improve the mighty 404  "bug" XD

test/trento_web/controllers/v1/discovery_controller_test.exs
```diff
@@ -32,5 +40,27 @@ defmodule TrentoWeb.V1.DiscoveryControllerTest do
...
+      expect(Trento.Commanded.Mock, :dispatch, fn _ ->
+        {:error, :any_error}
+      end)
+
+      %{status: status} =
+        conn
+        |> put_req_header("content-type", "application/json")
+        |> post("/api/v1/collect", body)
+
+      assert status != 202
```
nelsonkopliku, line 58: **question:** the endpoint returns 500. Is it intended?

nelsonkopliku (approved):
Just a question about the 500 being returned by the endpoint. Otherwise LGTM.

test/trento_web/controllers/v1/discovery_controller_test.exs
balanza: TL;DR We shouldn't care

The response code is inherited by the error returned by the dispatcher, which the fallback controller then handles. AS `:any_error` is not handled, `500` is returned.

balanza: Anyway, thanks to your comment I realized that the test might be weak, because the controller may fail for other reasons that the declared in the test definition.

I narrowed the case in https://github.com/trento-project/web/pull/3040/commits/d55d837204c1dc8199c0783a8e83c4cb85751f91

nelsonkopliku: Correct, thanks. In reality it returns the 404 :+1:

test/trento_web/controllers/v1/discovery_controller_test.exs
```diff
@@ -32,5 +40,27 @@ defmodule TrentoWeb.V1.DiscoveryControllerTest do
...
     end
+
+    test "collect action discards application instance registrations when the associated database does not exists",
+         %{conn: conn} do
+      body =
+        load_discovery_event_fixture("sap_system_discovery_application")
+
+      expect(Trento.Commanded.Mock, :dispatch, fn _ ->
+        {:error, :any_error}
+      end)
```
dottorblaster, line 51: In the end you went the mock way, but I see :smile: no other option available in this case.

dottorblaster (approved):
:rocket:

PR #3052, opened 2024-10-07
Improve usability of select search on large options content

nelsonkopliku:
# Description

This PR hopefully improves usability of Activity Log's search filters, specifically the filter by type.

Currently the activity types options are truncated in the selection, and it might be difficult identifying an exact type.

https://github.com/user-attachments/assets/7d425fae-dcc7-429a-83ad-4017bf190edb

With the proposed change we show the whole options content.

https://github.com/user-attachments/assets/59bf95ab-4048-43ff-adbc-1ed3b331a8b0

## How was this tested?

Storybook.

### DIFF
--- a/assets/js/common/Filter/Filter.jsx
+++ b/assets/js/common/Filter/Filter.jsx
@@ -18,15 +18,8 @@ const getLabel = (value, placeholder) =>
  * @param {string} props.title Title of the filter. It will be displayed in the button when the filter is empty
  * @param {string|string[]} props.value Selected options. Default is an empty array
  * @param {function} props.onChange Function to call when the selected options change
- * @param {boolean} props.truncateOptionsContent If true, the options will be truncated to fit the width of the filter select, otherwise the full value will be displayed. Default is true
  */
-function Filter({
-  options,
-  title,
-  value = [],
-  onChange,
-  truncateOptionsContent = true,
-}) {
+function Filter({ options, title, value = [], onChange }) {
   const ref = useRef();
   const [open, setOpen] = useState(false);
   const [query, setQuery] = useState('');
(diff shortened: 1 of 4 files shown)


assets/js/common/Filter/Filter.jsx
```diff
@@ -18,8 +18,15 @@ const getLabel = (value, placeholder) =>
  * @param {string} props.title Title of the filter. It will be displayed in the button when the filter is empty
  * @param {string|string[]} props.value Selected options. Default is an empty array
  * @param {function} props.onChange Function to call when the selected options change
+ * @param {boolean} props.truncateOptionsContent If true, the options will be truncated to fit the width of the filter select, otherwise the full value will be displayed. Default is true
```
balanza, line 21: Bold opinion: this should be the behavior of the component every time. I see no harm in doing that everywhere. Let's take a trip on the app to see id something breaks, but I'm quite confident it doesn't.

balanza (commented):
Thanks @nelsonkopliku :pray: good catch

jagabomb (approved):
LGTM

gagandeepb (approved):

assets/js/common/Filter/Filter.jsx
nelsonkopliku: I tend to agree, however there might be edge cases where the select is the last element on the right side of the page and in that case a non truncating behaviour might be suboptimal (ie think of the overviews where we have tags. A very long tag plus a non truncating option would break the UI).

What about deferring further improvements to this aspect?

balanza: Are you referring to the case when:
* the filter is the right-most one
* the browser window is narrow

In that case, the content will be truncated anyway, so no practical difference for users :upside_down_face: 

image

nelsonkopliku: the last commit might address it https://github.com/trento-project/web/pull/3052/commits/8d1b857c397e18a285c0f7bce14a322f1b1a23cf. The default behavior becomes not truncating the options of a select when the content is large.

balanza (approved):

PR #3055, opened 2024-10-08
Remove Phoenix.View and use the new Phoenix.Component structure

EMaksy:
# Description

This PR has the main goal to update Trento Web console to [Phoenix 1.7 standards](https://hexdocs.pm/phoenix/changelog.html)

## What has been done ? 

- Removed Phoenix.View and replaced it with [Phoenix.Component/HTML](https://hexdocs.pm/phoenix_view/Phoenix.View.html#module-replaced-by-phoenix-component)

- Updated [configs for new html and json rendering](https://gist.github.com/chrismccord/00a6ea2a96bc57df0cce526bd20af8a7)

- Replaced [old view modules with json modules with new style of json rendering](https://hexdocs.pm/phoenix/json_and_apis.html#rendering-json)

- Updated Projectors when required for JSON rendering

- Added components/core component and layouts to fit new Phoenix.Component style

- Updated heex templates

- Added [verified routes](https://hexdocs.pm/phoenix/Phoenix.VerifiedRoutes.html)

- Updated existing controllers for json rendering

- Switched some functions to private when not used elsewhere

  

## What has been not done? 

Phoenix View was not completely removed as currently Swoosh still uses it, there are ways to remove Phoenix.View, but i would prefer to do it when Swoosh officially removes Phoenix.View.

Check out the [discussion](https://github.com/swoosh/phoenix_swoosh/issues/287) and [official docs](https://hexdocs.pm/phoenix_swoosh/readme.html)

This PR relates to wanda  https://github.com/trento-project/wanda/pull/513

## How was this tested?

- Updated all existing tests

- Added test for about json view


lib/trento_web.ex
```diff
@@ -20,51 +20,53 @@ defmodule TrentoWeb do
...
       import Plug.Conn
       import TrentoWeb.Gettext
 
       unquote(verified_routes())
-      alias TrentoWeb.Router.Helpers, as: Routes
     end
   end
 
-  def view do
+  def html do
```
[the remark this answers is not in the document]
EMaksy: It is not really documented, you can generate a new phoenix app to compare as suggested in the [Phoenix.View migration ](https://hexdocs.pm/phoenix_view/Phoenix.View.html#module-migrating-to-phoenix-component)

EMaksy:
Thanks for the review, as mentioned by @arbulu89 opened a follow up pr on wanda

arbulu89 (approved):
Nice!
Thank you

lib/trento_web/controllers/v1/database_json.ex
```diff
@@ -1,48 +1,45 @@
...
+        |> Map.delete(:sap_systems)
+        |> Map.delete(:tenants)
+        |> Map.put(
+          :database_instances,
+          Enum.map(database_instances, &database_instance(%{instance: &1}))
+        )
+        |> add_system_replication_status_to_secondary_instance
+
+  def database_instance(%{instance: %{database_id: database_id} = instance}),
+    do:
```
dottorblaster, line 24: Props for the `, do:`

dottorblaster (approved):
LGTM with some minor nitpicks

lib/trento_web/controllers/v1/database_json.ex
arbulu89: For me personally, it looks worst this way :see_no_evil: 
Not for being one liner we should go this way. I would use the `, do:` in really easy function declarations. (More if you even have comments in the code)
Anyway, just my opinion

EMaksy: Well the question is, what  looks better for the eye. 

```
  def database_instance(%{instance: %{database_id: database_id} = instance}) do
    instance
    |> Map.from_struct()
    # keep backward compatibility
    |> Map.put(:sap_system_id, database_id)
    |> Map.delete(:__meta__)
    |> Map.delete(:host)
  end
```

vs

```
  def database_instance(%{instance: %{database_id: database_id} = instance}),
    do:
      instance
      |> Map.from_struct()
      # keep backward compatibility
      |> Map.put(:sap_system_id, database_id)
      |> Map.delete(:__meta__)
      |> Map.delete(:host)
```

It is almost the same with the pipe operator, just one line less without the end block. 
For me i prefer the shorter version, both works perfectly,  but i think we should just decide and have the same way in the whole code base :)

arbulu89: I would say this is subjective (as elixir allow both options), but for me, one liners are for one liners, like having something like:
```
def function(), do: do_one_line_things
# or
def some_long_function(), do:
  do_one_line_things
``` 

Otherwise, it is not obvious to your eye where a function start and ends.
Again, just me opinion.
And I don't think we should extend on this and have a "strict code rule"

PR #3060, opened 2024-10-10
Add junit.xml to gitignore

dottorblaster:
# Description
Right now a `junit.xml` file gets generated upon `npm test`, I'm just adding it to the gitignore in order to avoid committing it by mistake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,8 @@ npm-debug.log
 /assets/node_modules/
 /assets/.env
 
+/assets/junit.xml
+
 .pgdata/
 
 # Local dev.local.exs file


gagandeepb (commented):
In case you want to consider this alternative: There is also an option that can be set under `jest-junit` in `jest.config.js`:

```js
outputDirectory: "/tmp"
```

Thanks for looking into this. It escaped my attention and should have been handled in a way similar to the mix tests, when I introduced these junit deps.

dottorblaster:
No worries @gagandeepb, I just stumbled upon this right now. Can you do the same for the other tests?

gagandeepb:
For the other categories of tests: `mix` tests already have an opt-in flag (`WRITE_JUNIT=1`) to generate the junit so they do/should not generate junit files by default. Even when opted-in, the junit files for `mix` tests go to `/tmp` by default.

gagandeepb:
For jest tests, the write destination of the junit files can also be customized/overridden, if needed, like: `JEST_JUNIT_OUTPUT_DIR=$DESIRED_DIRECTORY npm test`

EMaksy (approved):

PR #3073, opened 2024-10-16
Load cacerts in sso

arbulu89:
**Update**: Apparently, the initial certificates we created for our demo IDP, they were created wrong, and that's why we were required to upload them to the machine. This means, that the requirement of uploading custom CA certs shouldn't be that important (actually, It would become a small corner case), so we can postpone this by now

# Description

Load CA certificates for Assent to enable https connection in SSO.
Otherwise, we get the imfamous `{:tls_alert, {:unknown_ca, ~c"TLS client: In state wait_cert_cr at ssl_handshake.erl:2133 generated CLIENT ALERT: Fatal - Unknown CA\n"}}` error.

Besides that, I have improved a bit the SSO login page to prompt a basic error message if the configuration fails.
We could've gone with the standard 500 error, but it looks less obvious. Either way, you need to access the server logs to see the exact error (which can be of different reason).

This is how it looks like:
image

PD: To test it locally, add the next lines to your local `dev.local.exs` file:
```
:ok = :public_key.cacerts_load()
[_ | _] = cacerts = :public_key.cacerts_get()

config :assent, http_adapter: {Assent.HTTPAdapter.Httpc, [ssl: [cacerts: cacerts]]}
```

## How was this tested?

Tested with some test and specially manually, using our demo IDP environment.

No, but we should consider adding maybe some troubleshooting section. 
Besides, once we decide how to load the CA certificates, specially in helm, we will need to document this is it requires user intervention (which will be the case almost certainly)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 # tar is required by kubectl cp
-RUN zypper -n in tar
+RUN zypper -n in tar ca-certificates ca-certificates-cacert
 WORKDIR /app
 COPY --from=release /build/_build/$MIX_ENV/rel/trento .
 EXPOSE 4000/tcp
--- a/assets/js/pages/Login/Login.test.jsx
+++ b/assets/js/pages/Login/Login.test.jsx
@@ -206,5 +206,26 @@ describe('Login component', () => {
       });
       expect(loginButton).toBeVisible();
     });
+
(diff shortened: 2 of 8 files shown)


arbulu89:
> General comment for later/sync consideration/conversation: What are your thoughts about having some "smoke" tests (e.g. on CI) for config hardening by running existing ssl config against something like badssl.com ? It provides a lot of positive as well as negative test cases that the config should account for, that can possibly be asserted for/against.

It looks interesting. Let's chat about it offline

dottorblaster (approved):
LGTM

PR #3074, opened 2024-10-16
Handle `associated_database_not_found` as accepted event.

balanza:
# Description

This PR comes as a rework of https://github.com/trento-project/web/pull/3069 after the review. The purpose is the same:
A valid discovery event that is discarded for context reasons should be considered as accepted.

It does it by handling the `:associated_database_not_found` rejection as an exceptional case in the fallback controller.

### DIFF
--- a/lib/trento_web/controllers/fallback_controller.ex
+++ b/lib/trento_web/controllers/fallback_controller.ex
@@ -75,6 +75,12 @@ defmodule TrentoWeb.FallbackController do
     |> render(:"422", changeset: changeset)
   end
 
+  def call(conn, {:error, :associated_database_not_found}) do
+    conn
+    |> put_status(:accepted)
+    |> json(%{})
+  end
+
   def call(conn, {:error, {:validation, _} = reason}) do
     conn
     |> put_status(:unprocessable_entity)
--- a/test/trento_web/controllers/v1/discovery_controller_test.exs
+++ b/test/trento_web/controllers/v1/discovery_controller_test.exs
@@ -41,7 +41,7 @@ defmodule TrentoWeb.V1.DiscoveryControllerTest do
       |> json_response(202)
     end
 
-    test "collect action discards application instance registrations when the associated database does not exists",
+    test "collect action discards application instance registrations when the dispatch fails",
          %{conn: conn} do
       body =
         load_discovery_event_fixture("sap_system_discovery_application")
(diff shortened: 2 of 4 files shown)


test/trento_web/controllers/v1/discovery_controller_test.exs
```diff
@@ -62,5 +62,27 @@ defmodule TrentoWeb.V1.DiscoveryControllerTest do
...
+      expect(Trento.Commanded.Mock, :dispatch, fn _ ->
+        {:error, :associated_database_not_found}
+      end)
+
+      %{status: status} =
+        conn
+        |> put_req_header("content-type", "application/json")
+        |> post("/api/v1/collect", body)
+
+      assert status == 202
```
arbulu89, line 80: nitpick: You can pipe this assert in the previous function adding `|> json_response(202)` as well

lib/trento_web/controllers/fallback_controller.ex
```diff
@@ -75,6 +75,13 @@ defmodule TrentoWeb.FallbackController do
     |> render(:"422", changeset: changeset)
   end
 
+  def call(conn, {:error, :associated_database_not_found}) do
+    conn
+    |> put_status(:accepted)
+    |> put_view(json: ErrorJSON)
```
arbulu89, line 81: `ErrorJSON`?
I think we can return an empty json with `json(%{})` or `even just try to `send_resp(conn, 202, %{})`.
Because I don't think we have a 202 view to render (unless phoenix has it by default)

arbulu89 (approved):
Good to merge once the tests are fixed and nitpick considered!
But I'm afraid e2e tests might start after doing that, as database not found errors won't be replied by photofinish

test/trento_web/controllers/v1/discovery_controller_test.exs
balanza: I think leaving this separated makes it clear to the reader that this is an expectation for the test.

lib/trento_web/controllers/fallback_controller.ex
balanza: FIxed in https://github.com/trento-project/web/pull/3074/commits/b4b35325eb66fa446bfe6d312c82152e5bca72ac

nelsonkopliku (commented):
The only missing bit is fixing `enrich_register_application_instance_test.exs`. For the rest LGTM.

PR #3084, opened 2024-10-21
Display active users only in Activity Log filter

gagandeepb:
Now only those usernames that correspond to active users are displayed in the users filter in the Activity Log page.


nelsonkopliku (approved):
LGTM

I labeled the PR as `env`, just to test it in the pr environment.

balanza (approved):

PR #3085, opened 2024-10-21
Reduce ClusterDetailsUpdated events on irrelevant node attribute change

nelsonkopliku:
# Description

This PR sptrip `lpa_<sid>_lpt` cluster node attributes during a cluster discovery, specifically when a `RegisterClusterHost` command is dispatched.

`lpa_<sid>_lpt` is a timestamp attribute attached to a cluster node that possibly changes at every discovery tick, resulting in many `ClusterDetailsUpdated` being emitted.

We want those possibly frequent changes to that attribute not be issuing the mentioned domain event.

Considerations:
- change is currently implemented for hana scale up clusters only. We will extend to other cluster types when/if detected it being necessary
- `lpa_<sid>_lpt` might be present for every cluster node, and it is being ignored for every cluster node
- cluster enrichment data is extended with a `nodes_attributes` jsonb field in order to keep an up to date representation of that piece of information and its content will be merged into clusters read model, similarly to what happens for `cib_last_written`
- enrichment data write is kept in one query

**Note:** besides a silly mistake of mine, the initial aggregate based implementation was working fine. However if there is preference in dropping the information at middleware level I am fine with it.

Latest changes in the UI will follow up when this PR gets through.

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -95,8 +95,12 @@ defmodule Trento.Clusters do
       nil ->
         cluster
 
-      enriched_data ->
-        %ClusterReadModel{cluster | cib_last_written: enriched_data.cib_last_written}
+      %{cib_last_written: cib_last_written, details: enriching_details} ->
+        enrich_cluster_details(%ClusterReadModel{
+          cluster
+          | cib_last_written: cib_last_written,
+            enriching_details: enriching_details
+        })
     end
   end
 
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -11,6 +11,8 @@ defmodule Trento.Clusters do
   require Trento.Clusters.Enums.ClusterEnsaVersion, as: ClusterEnsaVersion
   require Trento.Clusters.Enums.HanaArchitectureType, as: HanaArchitectureType
 
+  alias Trento.Support.MapHelper
+
   alias Trento.Hosts.Projections.HostReadModel
 
   alias Trento.SapSystems.Projections.ApplicationInstanceReadModel
@@ -74,6 +76,16 @@ defmodule Trento.Clusters do
     )
     |> enrich_cluster_model_query()
     |> Repo.all()
+    |> Enum.map(fn %ClusterReadModel{
+                     details: details,
+                     enriching_details: enriching_details
+                   } = cluster ->
+      %ClusterReadModel{
+        cluster
+        | details: MapHelper.atomize_keys(details),
+          enriching_details: MapHelper.atomize_keys(enriching_details)
+      }
+    end)
     |> Enum.map(&enrich_cluster_details/1)
   end
 
@@ -90,7 +102,7 @@ defmodule Trento.Clusters do
   end
 
   @spec enrich_cluster_model(ClusterReadModel.t()) :: ClusterReadModel.t()
-  def enrich_cluster_model(%ClusterReadModel{id: id} = cluster) do
+  def enrich_cluster_model(%ClusterReadModel{id: id, details: details} = cluster) do
     case Repo.get(ClusterEnrichmentData, id) do
       nil ->
         cluster
@@ -99,7 +111,8 @@ defmodule Trento.Clusters do
         enrich_cluster_details(%ClusterReadModel{
           cluster
           | cib_last_written: cib_last_written,
-            enriching_details: enriching_details
+            details: MapHelper.atomize_keys(details),
+            enriching_details: MapHelper.atomize_keys(enriching_details)
         })
     end
   end
@@ -152,8 +165,8 @@ defmodule Trento.Clusters do
   defp enrich_cluster_details(
          %ClusterReadModel{
            type: ClusterType.hana_scale_up(),
-           details: %{"nodes" => initial_nodes} = initial_details,
-           enriching_details: %{"nodes" => enriching_nodes} = enriching_details
+           details: %{nodes: initial_nodes} = initial_details,
+           enriching_details: %{nodes: enriching_nodes} = enriching_details
          } = cluster
        )
        when is_map(initial_details) and is_map(enriching_details) do
@@ -161,7 +174,7 @@ defmodule Trento.Clusters do
       cluster
       | details: %{
           initial_details
-          | "nodes" => enrich_hana_scale_up_nodes(initial_nodes, enriching_nodes)
+          | nodes: enrich_hana_scale_up_nodes(initial_nodes, enriching_nodes)
         }
     }
   end
@@ -169,16 +182,16 @@ defmodule Trento.Clusters do
   defp enrich_cluster_details(cluster), do: cluster
 
   defp enrich_hana_scale_up_nodes(initial_nodes, enriching_nodes) do
-    Enum.map(initial_nodes, fn %{"name" => node_name, "attributes" => initial_attributes} = node ->
+    Enum.map(initial_nodes, fn %{name: node_name, attributes: initial_attributes} = node ->
       enriching_attributes =
         Enum.find_value(enriching_nodes, %{}, fn
-          %{"name" => ^node_name, "attributes" => enriching_attributes} -> enriching_attributes
+          %{name: ^node_name, attributes: enriching_attributes} -> enriching_attributes
           _ -> false
         end)
 
       %{
         node
-        | "attributes" =>
+        | attributes:
             Map.merge(
               initial_attributes,
               enriching_attributes
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -124,13 +124,13 @@ defmodule Trento.Clusters do
     end)
   end
 
-  @spec update_cib_last_written(String.t(), String.t()) :: {:ok, Ecto.Schema.t()} | {:error, any}
-  def update_cib_last_written(cluster_id, cib_last_written) do
+  @spec update_enrichment_data(String.t(), map()) :: {:ok, Ecto.Schema.t()} | {:error, any}
+  def update_enrichment_data(cluster_id, enriching_changeset) do
     case Repo.get(ClusterEnrichmentData, cluster_id) do
       nil -> %ClusterEnrichmentData{cluster_id: cluster_id}
       enriched_cluster -> enriched_cluster
     end
-    |> ClusterEnrichmentData.changeset(%{cib_last_written: cib_last_written})
+    |> ClusterEnrichmentData.changeset(enriching_changeset)
     |> Repo.insert_or_update()
   end
 
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -169,14 +169,12 @@ defmodule Trento.Clusters do
   defp enrich_cluster_details(cluster), do: cluster
 
   defp enrich_hana_scale_up_nodes(initial_nodes, enriching_nodes) do
-    Enum.map(initial_nodes, fn %{
-                                 "name" => node_name,
-                                 "attributes" => initial_attributes
-                               } = node ->
+    Enum.map(initial_nodes, fn %{"name" => node_name, "attributes" => initial_attributes} = node ->
       enriching_attributes =
-        enriching_nodes
-        |> Enum.find(%{}, &(Map.get(&1, "name") == node_name))
-        |> Map.get("attributes", %{})
+        Enum.find_value(enriching_nodes, %{}, fn
+          %{"name" => ^node_name, "attributes" => enriching_attributes} -> enriching_attributes
+          _ -> false
+        end)
 
       %{
         node
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -74,6 +74,7 @@ defmodule Trento.Clusters do
     )
     |> enrich_cluster_model_query()
     |> Repo.all()
+    |> Enum.map(&enrich_cluster_details/1)
   end
 
   @spec get_cluster_id_by_host_id(String.t()) :: String.t() | nil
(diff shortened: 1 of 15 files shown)


arbulu89:
@nelsonkopliku I said that I would prefer to see the complete solution in place to review because depending how we do other things, we might decide changing things done in this PR.
For example, if you extract these values in the enrichment middleware and there are functions to just pop the value modifying the object, we wouldn't even need to care what's going in the in aggregate.

And at the end, for my point of view, this value is simply a nuisance in the aggregate, as it simply complicates things without adding any real value. For example, in the current code, if I'm right, to decide that details differ, you are simply checking that the nodes are different excluding that value, but the details contain more fields than only the nodes. What happens if the `maintenance_mode` changes?
I mean, the comparison becomes not that simple. Maybe you just can add the updated nodes to the details again, and compare the whole details content. But you need to add extra tests, etc

And we don't really use it in the aggregate nor in the future projection. The same reason why we don't even use the `cib_last_written` there I guess (it is more obvious there, as not comparing things is just not pattern matching, and we don't need to store the value in the aggregate neither).

That's why if I had the chance to review once the other pieces are, I think we have more information to take decisions.
Anyway, after my text wall :sweat: if you think that this is the correct solution, let's move on once the `details_differs?` is fixed.

nelsonkopliku:
hey @arbulu89 this is exactly the feedback I was looking for. Thanks!

Catching the difference in the aggregate does indeed require more juggling with state, so I am going to revisit and move the attribute exclusion logic in the middleware.

balanza:
What if we just don't do it? What are the downsides for both Trento's regular users and administrators?

arbulu89:
> What if we just don't do it? What are the downsides for both Trento's regular users and administrators?

I think we need yes or yes, otherwise it breaks the "events are meaningful" things rule.

nelsonkopliku:
@arbulu89 @balanza @gagandeepb ready for you

lib/trento/clusters/cluster_enrichment_data.ex
```diff
@@ -13,6 +13,7 @@ defmodule Trento.Clusters.ClusterEnrichmentData do
   @primary_key {:cluster_id, :binary_id, autogenerate: false}
   schema "clusters_enrichment_data" do
     field :cib_last_written, :string
+    field :nodes_attributes, :map, default: %{}
```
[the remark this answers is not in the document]
nelsonkopliku: do you mean just storing the whole cluster details?
It would be simple for the final merging, however the question then is: does it even make sense projecting the details then? 🤔

lib/trento/infrastructure/commanded/middleware/enrich_register_cluster_host.ex
```diff
@@ -22,6 +38,46 @@ defimpl Trento.Infrastructure.Commanded.Middleware.Enrichable,
...
 
-    {:ok, command}
+    {:ok,
+     %RegisterClusterHost{
+       command
+       | details: new_details
+     }}
+  end
+
+  defp strip_irrelevant_details(
```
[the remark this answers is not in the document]
nelsonkopliku: I tried the code and it took me a while to understand why it was not stripping properly. There is a typo `_ltp` should be `_lpt`.

However, thanks for the suggestion I embedded many of them here c92c6be37df36e375aa734e52b134f6fa3738278
Note:
- the node name kept in the resulting enrichment data is useful (in my idea) when enriching the read model, and matching which attributes go in which node
- Map.pop vs Map.split_with: not much difference, I just found`split_with` handier to have already the correct key => value at hand

lib/trento/clusters/cluster_enrichment_data.ex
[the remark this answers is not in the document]
nelsonkopliku: For the record, currently what gets store is something like this
```
{
  "vmhdbqas01": {
    "lpa_hdq_lpt": "1643125033"
  },
  "vmhdbqas02": {
    "lpa_hdq_lpt": "30"
  }
}
```
which is basically node_name => attributes_enriching_the_attributes_in_the_read_model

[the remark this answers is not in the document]
arbulu89: Not the whole details data, just the same data "struct" format. Like details being:
`%{nodes: [%{attributes: %{"lpa_hdq_lpt" => "1643125033"}}, %{attributes: %{"lpa_hdq_lpt" => "30}}]}`, so it resembles a partial details struct. Something that you could just maybe merge with some basic `Map.merge` usage (maybe combining with some other Enum functions)

lib/trento/infrastructure/commanded/middleware/enrich_register_cluster_host.ex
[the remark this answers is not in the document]
arbulu89: About the name, I cannot say, as I don't know how the final merge looks like, but in essence, as they nodes are a ordered list, if you keep the same order, you should be applying the correct attributes at the end as well

lib/trento/clusters/cluster_enrichment_data.ex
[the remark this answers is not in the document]
nelsonkopliku: as you suggested I kept the same structure. Could indeed be helpful if we need to do something similar for other parts of the detail.

nelsonkopliku:
Hey @arbulu89 I addressed the merging by adding a virtual field on top of which we do the merge on the real field. Then the virtual field gets stripped in the view. https://github.com/trento-project/web/pull/3085/commits/00ef2cb7477e38aacdcc6f7c0e11cc08a5c35bd0 and https://github.com/trento-project/web/pull/3085/commits/cc7efaba5bb3aab8e33f603315be339d1c95eb07

I could not find anything different allowing something like the following 😞 
```elixir
select_merge([c, e], %{cib_last_written: e.cib_last_written, details: merge_details(c.details, e.details)})
```

lib/trento/clusters.ex
```diff
@@ -141,7 +142,50 @@ defmodule Trento.Clusters do
...
+  end
+
+  defp enrich_cluster_details(cluster), do: cluster
+
+  defp enrich_hana_scale_up_nodes(initial_nodes, enriching_nodes) do
+    Enum.map(initial_nodes, fn %{
+                                 "name" => node_name,
+                                 "attributes" => initial_attributes
+                               } = node ->
+      enriching_attributes =
```
arbulu89, line 179:
```
enriching_attributes = 
    Enum.find_value(enriching_nodes, %{}, fn 
        %{"name" => ^node_name, "attributes" => attributes} -> attributes
       _ -> false
    end)
```
(of with `if enriching_node_name == node_name, do: attributes`)

lib/trento/clusters.ex
```diff
@@ -141,7 +142,50 @@ defmodule Trento.Clusters do
...
+        enriching_nodes
+        |> Enum.find(%{}, &(Map.get(&1, "name") == node_name))
+        |> Map.get("attributes", %{})
+
+      %{
+        node
+        | "attributes" =>
```
arbulu89, line 186: This should be `attributes` instead of `"attributes"` otherwise we are returning some random map and not `HanaClusterNode`

lib/trento/clusters.ex
```diff
@@ -141,7 +142,50 @@ defmodule Trento.Clusters do
   defp enrich_cluster_model_query(query) do
     query
     |> join(:left, [c], e in ClusterEnrichmentData, on: c.id == e.cluster_id)
-    |> select_merge([c, e], %{cib_last_written: e.cib_last_written})
+    |> select_merge([c, e], %{cib_last_written: e.cib_last_written, enriching_details: e.details})
```
arbulu89, line 149: If you want to remove the `virtual` field, you could skip this join, and simply query the details in `enrich_cluster_details`. We actually do the same in `enrich_cluster_model`. 
This way we don't have the risk to slip away this field anywhere.
In fact, we could reuse the function in other parts (I think we need to do it in `ClusterRegistered` projection, where we use `enrich_cluster_model`

It adds an extra query, but it removes the virtual field and its inconveniences.

test/trento/clusters_test.exs
```diff
@@ -246,6 +246,55 @@ defmodule Trento.ClustersTest do
...
+                node
+                | attributes: %{
+                    "foo_attribute" => "foo_value",
+                    "bar_attribute" => "bar_value"
+                  }
+              }
+            ]
+        }
+        |> Jason.encode!()
+        |> Jason.decode!()
```
arbulu89, line 289: duped entry.
in any case, i don't know why we need to encode to json, as the details are a struct

test/trento/clusters_test.exs
```diff
@@ -264,21 +313,157 @@ defmodule Trento.ClustersTest do
...
+              %ClusterEnrichmentData{
+                cib_last_written: ^new_cib_last_written,
+                details: ^unchanged_details
+              }} =
+               Clusters.update_enrichment_data(cluster_id, %{
+                 cib_last_written: new_cib_last_written
+               })
+    end
+
+    test "shuld enrich a cluster read model with enrichment data" do
```
arbulu89, line 416: typo `shuld`

test/trento/clusters_test.exs
```diff
@@ -264,21 +313,157 @@ defmodule Trento.ClustersTest do
...
+                node
+                | attributes: %{
+                    "foo_attribute" => "foo_value",
+                    "bar_attribute" => "bar_value"
+                  }
+              }
+            ]
+        }
+        |> Jason.encode!()
+        |> Jason.decode!()
```
arbulu89, line 457: duped

arbulu89 (commented):
Hey @nelsonkopliku 
I wrote some other comments, many of them nitpicks.
About using the `virtual` field, other option is just to replicate what we do with `enrich_cluster_model` and query again the enrichment data (without a `join`) and do the same operation afterwards. This has the pro that we don't have the virtual field, and the con that we have an extra query.

One missing thing is that I think we need to add something in the `ClusterRegistered` projection `after_update`, so we enrich this data as well. Otherwise, we might not have this attribute in the first registration until the next HA payload coming from the agent as it is not coming in the event.

test/trento/clusters_test.exs
nelsonkopliku: unfortunately necessary to match what we get from the db with what we have at hand.
when we do insert of factories we have shiny Structs, when we query back from the db the inner detail is just a plain map with "key" => "value" instead of key: "value" :shrug:

lib/trento/clusters.ex
nelsonkopliku: related to https://github.com/trento-project/web/pull/3085#discussion_r1815132706

lib/trento/clusters.ex
```diff
@@ -141,7 +142,50 @@ defmodule Trento.Clusters do
...
+       )
+       when is_map(initial_details) and is_map(enriching_details) do
+    initial_nodes = Map.get(initial_details, "nodes", [])
+    enriching_nodes = Map.get(enriching_details, "nodes", [])
+
+    %ClusterReadModel{
+      cluster
+      | details: %{
+          initial_details
+          | "nodes" => enrich_hana_scale_up_nodes(initial_nodes, enriching_nodes)
```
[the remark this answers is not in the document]
nelsonkopliku: related to https://github.com/trento-project/web/pull/3085#discussion_r1815132706

arbulu89 (approved):
@nelsonkopliku 
Green light from my side.
If you decide to change the virtual field, feel free to ask me other review.
Everything else are just nitpick comments

priv/repo/migrations/20241022133945_add_cluster_details_enrichment.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Trento.Repo.Migrations.AddClusterDetailsEnrichment do
+  use Ecto.Migration
+
+  def change do
+    alter table(:clusters_enrichment_data) do
+      add :details, :map
```
gagandeepb, line 6: I wonder what are the arguments for/against setting the default value at the schema vs at the migration level ?

lib/trento/infrastructure/commanded/middleware/enrich_register_cluster_host.ex
```diff
@@ -22,6 +39,45 @@ defimpl Trento.Infrastructure.Commanded.Middleware.Enrichable,
...
+         ClusterType.hana_scale_up(),
+         sid,
+         %HanaClusterDetails{nodes: cluster_nodes} = details
+       ) do
+    {stripped_nodes, ignored_nodes_attributes} =
```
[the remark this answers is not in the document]
nelsonkopliku: thanks, addressed here 61c042be08bddf9b08599098cfe758c1522a7216

priv/repo/migrations/20241022133945_add_cluster_details_enrichment.exs
nelsonkopliku: No arguments for/against it. Added also in the migration so we're 100% sure f8627b7b4ee0cde8cca8d95e33db9189254120ed

lib/trento/clusters.ex
```diff
@@ -141,7 +142,50 @@ defmodule Trento.Clusters do
...
+
+  defp enrich_cluster_details(
+         %ClusterReadModel{
+           type: ClusterType.hana_scale_up(),
+           details: initial_details,
+           enriching_details: enriching_details
+         } = cluster
+       )
+       when is_map(initial_details) and is_map(enriching_details) do
+    initial_nodes = Map.get(initial_details, "nodes", [])
```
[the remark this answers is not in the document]
nelsonkopliku: true, I also added a test here 62d9f90f2f2f2ca4dd4b6d7e2ef9528b653100e8

lib/trento/clusters.ex
nelsonkopliku: here we go c5a114fab4573f77c4dd6fda2669ae8b0b0d526c

nelsonkopliku:
Superseded by https://github.com/trento-project/web/pull/3112.

PR #3091, opened 2024-10-22
Add username to activity log metadata on user deletion

balanza:
# Description
Add username to activity log metadata on user deletion.

Being the user have been soft deleted, its username must be polished before presenting.


lib/trento/activity_logging/parser/metadata_enricher.ex
```diff
@@ -79,7 +80,15 @@ defmodule Trento.ActivityLog.Logger.Parser.MetadataEnricher do
...
 
   defp detect_enrichment(_target_entity, {_activity, _metadata}),
     do: {:error, :no_enrichment_needed}
+
+  defp clean_metadata(%{username: username} = metadata) do
+    clean_username = username |> String.split("__") |> List.first()
+    Map.put(metadata, :username, clean_username)
+  end
+
+  defp clean_metadata(metadata), do: metadata
```
gagandeepb, line 93: What are your thoughts about handling the case with usernames like `foo__bar__2024...`?

lib/trento/users.ex
```diff
@@ -15,6 +15,14 @@ defmodule Trento.Users do
 
   alias Trento.Users.User
 
+  @spec by_id(Number.t()) :: {:ok, User.t()} | {:error, :not_found}
```
gagandeepb: I would suggest `non_neg_integer()` as a more precise typespec. I am not sure/fully aware if there is a `Number.t()`  type. :thinking:

lib/trento/users.ex
```diff
@@ -15,6 +15,14 @@ defmodule Trento.Users do
...
+  def by_id(id) do
```
gagandeepb, line 19: Naming suggestion: there is a convention (in this and in many other context modules) to have a verb followed by descriptive qualifier (or something along these lines; it's somewhat informal). This would mean that something like `get_by_id` or something similar might be better suited.

balanza: I agree. The problem is that other contexts declare the `by_id` method and it's also expected by this (loosy) abstraction: https://github.com/trento-project/web/blob/b522d734e62eaedfcd598e8ce81b268296506262/lib/trento/activity_logging/parser/metadata_enricher.ex#L36

Maybe we can do a refactor afterward?

lib/trento/activity_logging/parser/metadata_enricher.ex
balanza: I didn't get your point sorry :upside_down_face: 

What's the question?

gagandeepb: With the current implementation, a user with username `foo__bar`, upon deletion, will have the username just `foo` persisted, IIUC. However we want the actual user name `foo__bar` to be used, not the truncation that would occur due to usage of two successive `_`s. The timestamp also uses the same `__` as separator.

test/trento/activity_logging/activity_logger_test.exs
```diff
@@ -95,6 +95,29 @@ defmodule Trento.ActivityLog.ActivityLoggerTest do
...
+      |> with_token(admin_id)
+      |> delete("/api/v1/users/#{user_id}")
+
+      wait_for_tasks_completion()
+
+      assert [
+               %ActivityLog{
+                 type: "user_deletion",
+                 actor: ^admin_username,
+                 metadata: %{"username" => ^username, "user_id" => ^user_id}
```
nelsonkopliku, line 115: after deletion the username becomes `<real-username>__<deletion-timestamp>` so this is not going to match, unfortunately.

nelsonkopliku (commented):
Hey @balanza thanks for this.

The main thing we need to address is making sure the tagging enrichment detection does not conflict with username enrichment detection.

Then there is the point about the deleted user's username.
I personally believe it'd be more friendly tracking without the leading deletion timestamp. I know, it's ugly, but it is what it is.

lib/trento/activity_logging/parser/metadata_enricher.ex
```diff
@@ -79,6 +80,7 @@ defmodule Trento.ActivityLog.Logger.Parser.MetadataEnricher do
   defp detect_enrichment(:cluster, {_, %{cluster_id: id}}), do: {:ok, id}
   defp detect_enrichment(:database, {_, %{database_id: id}}), do: {:ok, id}
   defp detect_enrichment(:sap_system, {_, %{sap_system_id: id}}), do: {:ok, id}
+  defp detect_enrichment(:user, {_, %{user_id: id}}), do: {:ok, id}
```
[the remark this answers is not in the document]
balanza: In the end, I found out this works:
```elixir
 defp detect_enrichment(
         target_entity,
         {_,
          %{
            resource_id: resource_id,
            resource_type: target_entity
          }}
       ),
       do: {:ok, resource_id}
```

Also, I moved it at the bottom being less specific than the others

test/trento/activity_logging/activity_logger_test.exs
balanza: I couldn't come up with a better option than checking string starts with :shrug:

nelsonkopliku: I see. We could consider making sure that when the `Users` context returns possibly deleted users we clean up the username.

balanza: I also think that it's the `Users` context duty to come up with the cleanup logic

lib/trento/activity_logging/parser/metadata_enricher.ex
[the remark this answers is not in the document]
balanza: Do you think it's necessary? What's wrong with having (for example) a cluster enriched regardless of the activity type?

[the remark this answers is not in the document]
nelsonkopliku: Not strictly necessary perhaps and there is nothing wrong in enriching a cluster (or anything else) for different activities.

For example it happens for cluster checks execution being requested, cluster tagging/untagging and for any domain event carrying a cluster id but not the cluster name.

There is a slightly difference in the semantics, imho.

For instance the following two clauses could be merged in one, however we'd miss a bit of context about connection vs domain event activities.
```elixir
# connection related
defp detect_enrichment(
         :cluster,
         {:cluster_checks_execution_request, %{cluster_id: cluster_id}}
       ),
       do: {:ok, cluster_id}

# domain event related
defp detect_enrichment(:cluster, {_, %{cluster_id: id}}), do: {:ok, id}
```
Now that you make me think a couple of comments would be helpful 😄

test/trento/activity_logging/activity_logger_test.exs
nelsonkopliku: Sweet, so are we going to address this here or in a follow up?

balanza: I don't know :shrug:  

The only I know is it doesn't bother the activity log as it's something that refers to the user management logic.

lib/trento/activity_logging/parser/metadata_enricher.ex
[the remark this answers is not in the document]
balanza: Same goes for 
```elixir
  defp detect_enrichment(:host, {:host_checks_execution_request, %{host_id: host_id}}),
    do: {:ok, host_id}
```
it can be avoided.

lib/trento/activity_logging/parser/phoenix_conn_parser.ex
```diff
@@ -124,6 +124,11 @@ defmodule Trento.ActivityLog.Logger.Parser.PhoenixConnParser do
       ),
       do: %{host_id: Map.get(params, :id)}
 
+  def get_activity_metadata(:user_deletion, %Plug.Conn{
```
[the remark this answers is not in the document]
nelsonkopliku: Well' I guess it is covered already indirectly.

lib/trento/activity_logging/parser/metadata_enricher.ex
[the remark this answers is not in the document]
nelsonkopliku: I am fine either way, as long as we are clear on the two natures of activities (connection related and domain event related) and nothing breaks 😄

nelsonkopliku (approved):
Overall LGTM.

The only reserve I have is that I'd find it more consistent if in the log we track the deleted user's username stripped from the deletion timestamp.

Let's proceed anyways. If needed we can follow up on that.

lib/trento/activity_logging/parser/metadata_enricher.ex
```diff
@@ -79,7 +69,31 @@ defmodule Trento.ActivityLog.Logger.Parser.MetadataEnricher do
...
+            resource_type: target_entity
+          }}
+       )
+       when activity in [:resource_tagging, :resource_untagging],
+       do: {:ok, resource_id}
 
   defp detect_enrichment(_target_entity, {_activity, _metadata}),
     do: {:error, :no_enrichment_needed}
+
+  defp polish_entity(%{username: username, deleted_at: deleted_at} = entity) do
```
nelsonkopliku, line 88: I thought you would add this to the `Users.by_id/1` 👀 However, as you wish.

balanza: Well, huge no. I won't add a custom patch to a context. I'm already ashamed to add a patch at this level, but I can swallow it.

Whether the data is "dirty" or not must be decided by the context, in a dedicated discussion hopefully

nelsonkopliku: Ok, I see. Fine with this, thanks 👍

test/trento/activity_logging/metadata_enricher_test.exs
```diff
@@ -203,6 +203,19 @@ defmodule Trento.ActivityLog.MetadataEnricherTest do
               }} =
                MetadataEnricher.enrich(:cluster_checks_execution_request, initial_metadata)
     end
+
+    test "should enrich user deletion request" do
+      %{id: user_id, username: username} = insert(:user)
```
nelsonkopliku, line 208: We're just missing a deletion date for a deleted user.
```suggestion
     %{id: user_id, username: username} = insert(:user, deleted_at: Faker.DateTime.backward(1))
```

nelsonkopliku (approved):
We might want to rebase and have a clean history of the PR. 
Besides this LGTM :shipit:

lib/trento/activity_logging/parser/metadata_enricher.ex
```diff
@@ -79,7 +69,31 @@ defmodule Trento.ActivityLog.Logger.Parser.MetadataEnricher do
...
+    # If the user is deleted, we append the deletion date to the username
+    # It's a implementation detail that we don't want to expose to the user
+    # See Trento.Users context for more information
+    clean_username =
+      String.trim_trailing(username, "__" <> DateTime.to_string(deleted_at))
+
```
gagandeepb, line 94: I wonder: do we need to guard against the `deleted_at` value being `nil` ?

balanza: It should never happen but I see no harm in doing it.

PR #3092, opened 2024-10-22
Fix SUSE Manager errors label computation

dottorblaster:
The computation for the SUSE Manager status of the host was wrong. Even when the errors weren't critical it reported `SUMA connection not working`. This PR fixes it making the unknown state reachable again.

## Testing
Jest tests both updated and added


EMaksy (approved):
something went wrong... but everything went right with this pr !

PR #3098, opened 2024-10-23
Add a button to download upgradable packages as CSV

janvhs:
# Description
This commit adds a CSV download button to include all patches to the UpgradablePackages view.

## Preview
image

## How was this tested

- Added test for UpgradablePackages
- Added UpgradablePackages empty story

## What is open ? 
- Missing e2e test, maybe in follow up pr

### DIFF
--- a/assets/js/common/UpgradablePackagesList/UpgradablePackagesList.test.jsx
+++ b/assets/js/common/UpgradablePackagesList/UpgradablePackagesList.test.jsx
@@ -10,7 +10,9 @@ describe('UpgradablePackagesList component', () => {
     const { patches } = upgradablePackage;
 
     const expectedInstalledPackage = `${upgradablePackage.name}-${upgradablePackage.from_version}-${upgradablePackage.from_release}.${upgradablePackage.arch}`;
+    upgradablePackage.installed_package = expectedInstalledPackage;
     const expectedLatestPackage = `${upgradablePackage.name}-${upgradablePackage.to_version}-${upgradablePackage.to_release}.${upgradablePackage.arch}`;
+    upgradablePackage.latest_package = expectedLatestPackage;
 
     render(<UpgradablePackagesList upgradablePackages={[upgradablePackage]} />);
 
--- a/assets/js/common/UpgradablePackagesList/UpgradablePackagesList.test.jsx
+++ b/assets/js/common/UpgradablePackagesList/UpgradablePackagesList.test.jsx
@@ -7,18 +7,18 @@ import UpgradablePackagesList from './UpgradablePackagesList';
 describe('UpgradablePackagesList component', () => {
   it('should render the upgradable packages list', () => {
     const upgradablePackage = upgradablePackageFactory.build();
-    const { original_patches } = upgradablePackage;
+    const { patches } = upgradablePackage;
 
     const expectedInstalledPackage = `${upgradablePackage.name}-${upgradablePackage.from_version}-${upgradablePackage.from_release}.${upgradablePackage.arch}`;
-    upgradablePackage.installed_package = expectedInstalledPackage;
+    upgradablePackage.installedPackage = expectedInstalledPackage;
     const expectedLatestPackage = `${upgradablePackage.name}-${upgradablePackage.to_version}-${upgradablePackage.to_release}.${upgradablePackage.arch}`;
-    upgradablePackage.latest_package = expectedLatestPackage;
+    upgradablePackage.latestPackage = expectedLatestPackage;
 
     render(<UpgradablePackagesList upgradablePackages={[upgradablePackage]} />);
 
     expect(screen.getByText(expectedInstalledPackage)).toBeVisible();
     expect(screen.getByText(expectedLatestPackage)).toBeVisible();
-    original_patches.forEach(({ advisory }) => {
+    patches.forEach(({ advisory }) => {
       expect(screen.getByText(advisory)).toBeVisible();
     });
   });
--- a/assets/js/common/UpgradablePackagesList/UpgradablePackagesList.test.jsx
+++ b/assets/js/common/UpgradablePackagesList/UpgradablePackagesList.test.jsx
@@ -7,7 +7,7 @@ import UpgradablePackagesList from './UpgradablePackagesList';
 describe('UpgradablePackagesList component', () => {
   it('should render the upgradable packages list', () => {
     const upgradablePackage = upgradablePackageFactory.build();
-    const { patches } = upgradablePackage;
+    const { original_patches } = upgradablePackage;
 
     const expectedInstalledPackage = `${upgradablePackage.name}-${upgradablePackage.from_version}-${upgradablePackage.from_release}.${upgradablePackage.arch}`;
     upgradablePackage.installed_package = expectedInstalledPackage;
@@ -18,7 +18,7 @@ describe('UpgradablePackagesList component', () => {
 
     expect(screen.getByText(expectedInstalledPackage)).toBeVisible();
     expect(screen.getByText(expectedLatestPackage)).toBeVisible();
-    patches.forEach(({ advisory }) => {
+    original_patches.forEach(({ advisory }) => {
       expect(screen.getByText(advisory)).toBeVisible();
     });
   });
--- a/assets/js/lib/test-utils/factories/upgradablePackage.js
+++ b/assets/js/lib/test-utils/factories/upgradablePackage.js
@@ -6,16 +6,14 @@ const releaseVersionFactory = () =>
   `${faker.number.int({ min: 100000, max: 160000 })}.${faker.system.semver()}`;
 
 export const upgradablePackageFactory = Factory.define(({ sequence }) => ({
-  from_epoch: faker.date.anytime(),
-  to_release: releaseVersionFactory(),
-  name: `${faker.word.noun()}${sequence}`,
-  from_release: releaseVersionFactory(),
-  to_epoch: faker.date.anytime(),
   arch: faker.airline.flightNumber(),
-  to_package_id: faker.number.int({ min: 2000, max: 5000 }),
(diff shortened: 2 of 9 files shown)


EMaksy (commented):
Just for the future, left a comment

assets/js/common/UpgradablePackagesList/UpgradablePackagesList.test.jsx
```diff
@@ -10,7 +10,9 @@ describe('UpgradablePackagesList component', () => {
     const { patches } = upgradablePackage;
 
     const expectedInstalledPackage = `${upgradablePackage.name}-${upgradablePackage.from_version}-${upgradablePackage.from_release}.${upgradablePackage.arch}`;
+    upgradablePackage.installed_package = expectedInstalledPackage;
```
arbulu89, line 13: nitpick: maybe something like this would look better rather than direct assignment:
```
const enrichedUpgradablePackage = {
    ...upgradablePackage,
    installed_package: expectedInstalledPackage,
    latest_package: expectedLatestPackage,
}
```

assets/js/pages/UpgradablePackagesPage/UpgradablePackages.jsx
```diff
@@ -15,14 +17,67 @@ export default function UpgradablePackages({
...
+    };
+  });
+  const displayedPackages = enrichedPackages.filter(
     ({ name, patches }) =>
       containsSubstring(name, search) ||
       patches
         .map(({ advisory }) => containsSubstring(advisory, search))
         .includes(true)
   );
+  const [csvURL, setCsvURL] = useState(null);
```
arbulu89, line 37: nitpick: I would move this `useState` above, to be together with the other `useState`

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.jsx
```diff
@@ -26,29 +26,40 @@ function HostRelevantPatches({ hostName, onNavigate, patches }) {
...
-              ),
-            ],
-            `${hostName}-patches.csv`,
-            { type: 'text/csv' }
+  const [csvURL, setCsvURL] = useState(null);
+
+  useEffect(() => {
+    setCsvURL(
+      patches.length > 0
+        ? URL.createObjectURL?.(
```
arbulu89, line 34: question: Why the `?.` thing? And not `URL.createObjectURL(...)`
General question for all the places where this is happening, which are a few

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.jsx
```diff
@@ -26,29 +26,40 @@ function HostRelevantPatches({ hostName, onNavigate, patches }) {
...
+                      'advisory_name',
+                      'advisory_synopsis',
+                      'update_date',
+                    ],
+                    data: patches,
+                  },
+                  { header: true }
+                ),
+              ],
+              `${hostName}-patches.csv`,
```
arbulu89, line 50: nitpick: maybe we could put `${hostName}-patches.csv` in a constant, as it is used twice, so we don't mess up if we change in this at some point

assets/js/pages/UpgradablePackagesPage/UpgradablePackages.jsx
```diff
@@ -15,14 +17,67 @@ export default function UpgradablePackages({
...
+    );
+
+    return () => {
+      if (csvURL) {
+        URL.revokeObjectURL(csvURL);
+      } else {
+        setCsvURL(null);
+      }
+    };
+  }, [patchesLoading]);
```
arbulu89, line 80: I would change this to `[enrichedPatches]` so the effect is executed everytime this value is changed.

assets/js/pages/UpgradablePackagesPage/UpgradablePackages.jsx
```diff
@@ -15,14 +17,67 @@ export default function UpgradablePackages({
...
+              { type: 'text/csv' }
+            )
+          )
+        : null
+    );
+
+    return () => {
+      if (csvURL) {
+        URL.revokeObjectURL(csvURL);
+      } else {
```
arbulu89, line 76: I think we can remove this `else`. If `csvURL` is empty the we don't need to set it again to empty

assets/js/pages/UpgradablePackagesPage/UpgradablePackages.jsx
```diff
@@ -15,14 +17,67 @@ export default function UpgradablePackages({
...
+                  {
+                    fields: ['installed_package', 'latest_package', 'patches'],
+                    data: hasPatches(enrichedPackages)
+                      ? createCsvPackages(enrichedPackages)
+                      : [],
+                  },
+                  { header: true }
+                ),
+              ],
+              `${hostName}-patches.csv`,
```
arbulu89, line 66: I would rename this to `${hostName}-upgradable-packages.csv`, and store it in a constant above, as we use it again later on

assets/js/pages/UpgradablePackagesPage/UpgradablePackages.jsx
```diff
@@ -15,14 +17,67 @@ export default function UpgradablePackages({
...
+  useEffect(() => {
+    setCsvURL(
+      hasPatches(enrichedPackages)
+        ? URL.createObjectURL?.(
+            new File(
+              [
+                Papa.unparse(
+                  {
+                    fields: ['installed_package', 'latest_package', 'patches'],
+                    data: hasPatches(enrichedPackages)
```
arbulu89, line 59: We don't need the `hasPatches` here. It is already checked above

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.jsx
```diff
@@ -90,11 +95,11 @@ function HostRelevantPatches({ hostName, onNavigate, patches }) {
             placeholder="Search by Synopsis"
             prefix={<EOS_SEARCH size="l" />}
           />
-          <a href={file} download={`${hostName}-patches.csv`}>
+          <a href={csvURL} download={`${hostName}-patches.csv`}>
             <Button
               className="w-max"
               type="primary-white"
-              disabled={displayedPatches?.length <= 0}
+              disabled={patches.length <= 0 || !csvURL}
```
arbulu89, line 102: I guess `disabled={!csvURL}` would be enough (and I hope we don't get never negative number of patches hehe)

assets/js/pages/UpgradablePackagesPage/UpgradablePackages.jsx
```diff
@@ -15,14 +17,67 @@ export default function UpgradablePackages({
...
       containsSubstring(name, search) ||
       patches
         .map(({ advisory }) => containsSubstring(advisory, search))
         .includes(true)
   );
+  const [csvURL, setCsvURL] = useState(null);
+
+  const hasPatches = (packages) =>
+    packages.some((packageDetails) => !isEmpty(packageDetails.patches));
+  const createCsvPackages = (packages) =>
```
arbulu89, line 41: I'm not sure about this function.
You are replacing the initial `patches` by the advisories included in the patches.
Are we sure we want that? Would be a bad thing to send the complete data of the `patches`?

Either way, do we need a function for this?
Wouldn't be better to simply transform the data into a variable and use it in the `useEffect` (and you could change the variable used to trigger the useEffect as well

assets/js/lib/test-utils/factories/upgradablePackage.js
```diff
@@ -6,16 +6,14 @@ const releaseVersionFactory = () =>
...
 
 export const upgradablePackageFactory = Factory.define(({ sequence }) => ({
+  arch: faker.airline.flightNumber(),
   from_epoch: faker.date.anytime(),
-  to_release: releaseVersionFactory(),
-  name: `${faker.word.noun()}${sequence}`,
   from_release: releaseVersionFactory(),
+  from_version: faker.system.semver(),
+  name: `${faker.word.noun()}${sequence}`,
+  patches: patchForPackageFactory.buildList(2),
```
arbulu89, line 14: This `patches` is suspicious here, as the endpoint doesn't include it. This combination is something done by the code.
Either one, it comes from before so...

arbulu89 (commented):
@EMaksy Some comments

assets/js/lib/test-utils/factories/upgradablePackage.js
dottorblaster: The combination between the content of the factory and the patches is made by other code, your choice if you wanna enrich the factory or keep it slim

dottorblaster (commented):
I wanted to ask @janvhs to provide one more test but you @EMaksy did it before :smile: 

Besides what @arbulu89 said (great comments BTW, really nothing to add), just a tiny suggestion on my side

assets/js/pages/UpgradablePackagesPage/UpgradablePackages.test.jsx
```diff
@@ -40,3 +42,78 @@ describe('UpgradablePackages', () => {
...
+    expect(csvButton).toBeDisabled();
+    expect(window.URL.createObjectURL).not.toHaveBeenCalled();
+    expect(window.URL.revokeObjectURL).not.toHaveBeenCalled();
+  });
+
+  it('Exports the packages in CSV format', () => {
+    const user = userEvent.setup();
+    const hostName = faker.string.uuid();
+    const upgradablePackage = [
+      upgradablePackageFactory.build({
```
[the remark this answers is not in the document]
EMaksy: i wanted to count the size very concrete in the test....  :)  
     size: 88,
```
    expect(window.URL.createObjectURL).toHaveReturnedWith({
      name: `${hostName}-upgradable-packages.csv`,
      size: 88,
    });
  });
  ```

PR #3112, opened 2024-11-04
Reduce ClusterDetailsUpdated events

nelsonkopliku:
# Description

This PR strips `lpa_<sid>_lpt` cluster node attributes during a cluster discovery, specifically when a `RegisterClusterHost` command is dispatched.

`lpa_<sid>_lpt` is a timestamp attribute attached to a cluster node that possibly changes at every discovery tick, resulting in many `ClusterDetailsUpdated` being emitted.

We want those possibly frequent changes to that attribute not be issuing the mentioned domain event.

Considerations:
- change is currently implemented for hana scale up clusters only. We will extend to other cluster types when/if detected it being necessary
- `lpa_<sid>_lpt` might be present for every cluster node, and it is being ignored for every cluster node
- we accept the tradeoff that for the time being the stripped node attribute is not shown anywhere in the UI until we decide differently

Supersedes https://github.com/trento-project/web/pull/3085

## How was this tested?

Automated tests.


gagandeepb (commented):
I get the sense that this attribute is sometimes meaningful, even if frequently changing/somewhat "noisy"-ish. Omitting this attribute, sounds to me like losing some possibly meaningful information, even if 80-90% of the time it will be non-informative. Would you say this understanding is accurate ? More concretely, why are frequent changes in this timestamp(-like) attribute not meaningful enough on the whole (i.e. taken together over some longer timescales), and make necessary the changes being proposed in this PR ?

nelsonkopliku:
> I get the sense that this attribute is sometimes meaningful
>
It is indeed meaningful for sap resource agents/pacemaker decisions.

> ...frequently changing/somewhat "noisy"-ish. Omitting this attribute, sounds to me like losing some possibly meaningful information, even if 80-90% of the time it will be non-informative. 
> Would you say this understanding is accurate? 
>
Accurate. 
Changes to that attribute should not be considered as a change in cluster configuration, so it is indeed pretty noisy when it leads to too many `ClusterDetailsUpdated` being emitted.

> More concretely, why are frequent changes in this timestamp(-like) attribute not meaningful enough on the whole (i.e. taken together over some longer timescales), and make necessary the changes being proposed in this PR ?
>
We as trento don't do anything with that attribute and its value, other than possibly showing it. 
It is also safe to assume that at this point in time its presence or absence from trento's state is equivalent.
Considering the quite invasive changes that are required to only show that attribute without its changes triggering too many unwanted events plus the limited knowledge of its counter part on clusters different than hana scale up, it's safer to keep the  changes at minimum (aka only make sure that not too many events are emitted) and defer further considerations, which at a certain point might come back.

lib/trento/infrastructure/commanded/middleware/enrich_register_cluster_host.ex
```diff
@@ -22,6 +34,36 @@ defimpl Trento.Infrastructure.Commanded.Middleware.Enrichable,
...
+  end
+
+  defp strip_irrelevant_details(_, _, details), do: details
+
+  defp strip_irrelevant_node_data(
+         %HanaClusterNode{attributes: attributes} = node,
+         sid
+       ) do
+    {_, new_attributes} =
+      Map.split_with(attributes, fn {key, _value} -> key == "lpa_#{String.downcase(sid)}_lpt" end)
```
arbulu89, line 62: Now that we don't want this get the modified version, maybe simply `new_attributes = Map.drop(attributes, ["lpa_#{String.downcase(sid)}_lpt")`

arbulu89 (approved):
Looks good!
Just a nitpick comment in the usage of `Enum.spli_with` and `Map.drop`

gagandeepb (approved):
LGTM. Thanks for the explanation!

PR #3119, opened 2024-11-06
Upgrade to Storybook 8.4

dottorblaster:
Just upgrading to Storybook 8.4 since we have been having it pending for some weeks and dependabot is just slow catching up with the whole set of upgrades


arbulu89 (approved):
Thank you!!

nelsonkopliku (approved):

EMaksy (approved):
Thanks for keeping storybook healthy!

PR #3144, opened 2024-11-13
Replaces Empty Favicon with Trento Logo

jagabomb:
# Description

This PR replaces the old favicon with the redesigned logo. 
Screenshot supplied below:
Screenshot 2024-11-13 at 12 01 57

There is also a minor update of the **trento.svg** file in the _priv/static/images_ folder since it still had the old logo.

## How was this tested?

- [x] **DONE** Visually


nelsonkopliku (approved):
image
👀

stefanotorresi (approved):

stefanotorresi (changes requested):
change needed here:
https://github.com/trento-project/web/blob/a5ec430c6dcb51d132466a4500d7e3715e049f37/mix.exs#L173

stefanotorresi:
@jagabomb also now that I re-think about it, are we sure we dont need anything done in the HTML header about this?
I wonder how would we bust users' browser cache in case we update the file in the future.

jagabomb:
> change needed here:
> 
> https://github.com/trento-project/web/blob/a5ec430c6dcb51d132466a4500d7e3715e049f37/mix.exs#L173

Hi @stefanotorresi, I ended up just replacing the SVG artwork with the redesigned logo. Here is a reference...
Screenshot 2024-11-15 at 08 52 59
Is this what you were expecting?

jagabomb:
> @jagabomb also now that I re-think about it, are we sure we don't need anything done in the HTML header about this? I wonder how would we bust users' browser cache in case we update the file in the future.

@stefanotorresi, so this is when I noticed that **Phoenix (elixir)** handles the favicon differently to conventional **HTML** standards. From what I can see, it seems like the favicon is served via the web server. So if I have to add it to the **HTML** `head`, then should I remove the reference of the favicon in the elixir code?

stefanotorresi:
@jagabomb nono, I thought you were removing the SVG in favor of the .ico file, sorry.

PR #3152, opened 2024-11-15
Add sap hana scale up cost optimized and performance optimized discovery

EMaksy:
# Description

This pr now adds a scenario discovery, with this, as a user I can see which HANA scale-up scenario I'm running on a cluster.

Main features.
- Multiple SID's are displayed in ClusterList, HostList, and ClusterDetails view. Links are redirecting to the correct sap database.
- Scenario discovery between Cost Opt. or Perf. Opt.
- Updated factory
- Additional sids in the overview are sorted by alphabet

Note: Scale out and ASCS/ERS untouched.

## Preview

ClusterList view:
image

HostList view: 
image

ClusterDetails View  Cost Opt: 
image

ClusterDetails View  Perf. Opt.
image

## How was this tested?

- Fixed existing tests
- Added frontend tests
- Added backend test for scale up scenario with payload
- Fixed brocken storybook stories and added new scenarios

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,3 @@ npm-debug.log
 
 # Local dev.local.exs file
 config/dev.local.exs
-
-# ignore scenario
-/test/fixtures/scenarios/cost_optimised_sap_scenario
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ npm-debug.log
 
 # Local dev.local.exs file
 config/dev.local.exs
+
+# ignore scenario
+/test/fixtures/scenarios/cost_optimised_sap_scenario
--- a/assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
+++ b/assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
@@ -52,9 +52,7 @@ function ClusterTypeLabel({
 }) {
   const clusterTypeLabel = getClusterTypeLabel(clusterType);
   const clusterTypeScenarioLabel = getClusterTypeScenarioLabel(clusterScenario);
-  const combinedClusterTypeLabel = `${clusterTypeLabel}${
-    clusterScenario ? ` ${clusterTypeScenarioLabel}` : ''
-  }`;
+  const combinedClusterTypeLabel = `${clusterTypeLabel} ${clusterTypeScenarioLabel}`;
 
   return (
     <span className="group flex items-center relative">
--- a/assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
+++ b/assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
@@ -45,11 +45,7 @@ const icons = {
   ),
 };
 
-function ClusterTypeLabel({
-  clusterType,
-  clusterScenario = '',
-  architectureType,
-}) {
+function ClusterTypeLabel({ clusterType, clusterScenario, architectureType }) {
   const clusterTypeLabel = getClusterTypeLabel(clusterType);
   const clusterTypeScenarioLabel = getClusterTypeScenarioLabel(clusterScenario);
   const combinedClusterTypeLabel = `${clusterTypeLabel} ${clusterTypeScenarioLabel}`;
--- a/assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
+++ b/assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
@@ -8,6 +8,7 @@ import {
   ANGI_ARCHITECTURE,
   CLASSIC_ARCHITECTURE,
   getClusterTypeLabel,
+  getClusterTypeScenarioLabel,
 } from '@lib/model/clusters';
 
 const MIGRATION_URL =
@@ -44,11 +45,21 @@ const icons = {
   ),
 };
 
-function ClusterTypeLabel({ clusterType, architectureType }) {
+function ClusterTypeLabel({
+  clusterType,
+  clusterScenario = '',
+  architectureType,
+}) {
+  const clusterTypeLabel = getClusterTypeLabel(clusterType);
+  const clusterTypeScenarioLabel = getClusterTypeScenarioLabel(clusterScenario);
+  const combinedClusterTypeLabel = `${clusterTypeLabel}${
+    clusterScenario ? ` ${clusterTypeScenarioLabel}` : ''
+  }`;
+
   return (
     <span className="group flex items-center relative">
       {icons[architectureType]}
-      {getClusterTypeLabel(clusterType)}
+      {combinedClusterTypeLabel}
     </span>
   );
 }
--- a/assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
+++ b/assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
@@ -8,7 +8,7 @@ import {
   ANGI_ARCHITECTURE,
   CLASSIC_ARCHITECTURE,
   getClusterTypeLabel,
-  getClusterTypeScenarioLabel,
+  getClusterScenarioLabel,
 } from '@lib/model/clusters';
 
 const MIGRATION_URL =
@@ -47,13 +47,13 @@ const icons = {
 
 function ClusterTypeLabel({ clusterType, clusterScenario, architectureType }) {
   const clusterTypeLabel = getClusterTypeLabel(clusterType);
-  const clusterTypeScenarioLabel = getClusterTypeScenarioLabel(clusterScenario);
-  const combinedClusterTypeLabel = `${clusterTypeLabel} ${clusterTypeScenarioLabel}`;
+  const clusterScenarioLabel = getClusterScenarioLabel(clusterScenario);
+  const clusterLabel = `${clusterTypeLabel} ${clusterScenarioLabel}`;
 
   return (
     <span className="group flex items-center relative">
       {icons[architectureType]}
-      {combinedClusterTypeLabel}
+      {clusterLabel}
     </span>
   );
 }
--- a/assets/js/lib/model/clusters.js
+++ b/assets/js/lib/model/clusters.js
@@ -1,8 +1,24 @@
+// Cluster types
 export const HANA_SCALE_UP = 'hana_scale_up';
 export const HANA_SCALE_OUT = 'hana_scale_out';
 export const ASCS_ERS = 'ascs_ers';
 
-export const clusterTypes = [HANA_SCALE_UP, HANA_SCALE_OUT, ASCS_ERS];
+// Hana scale up scenarios
+export const HANA_SCALE_UP_PERFORMANCE_SCENARIO = 'performance_optimized';
+export const HANA_SCALE_UP_COST_OPT_SCENARIO = 'cost_optimized';
+export const HANA_SCALE_UP_UNKNOWN_SCENARIO = 'unknown';
+
+export const clusterTypes = [
+  HANA_SCALE_UP,
+  HANA_SCALE_UP_PERFORMANCE_SCENARIO,
+  HANA_SCALE_OUT,
+  ASCS_ERS,
+];
+export const hanaClusterScenarioTypes = [
+  HANA_SCALE_UP_PERFORMANCE_SCENARIO,
+  HANA_SCALE_UP_UNKNOWN_SCENARIO,
+  HANA_SCALE_UP_COST_OPT_SCENARIO,
+];
 
 export const isValidClusterType = (clusterType) =>
   clusterTypes.includes(clusterType);
@@ -15,6 +31,14 @@ const clusterTypeLabels = {
 
 export const getClusterTypeLabel = (type) =>
   clusterTypeLabels[type] || 'Unknown';
+
+const clusterTypeScenarioLabels = {
+  [HANA_SCALE_UP_PERFORMANCE_SCENARIO]: 'performance optimized',
+  [HANA_SCALE_UP_COST_OPT_SCENARIO]: 'cost optimized',
+};
+
+export const getClusterTypeScenarioLabel = (type) =>
+  clusterTypeScenarioLabels[type] || '';
 
 export const ANGI_ARCHITECTURE = 'angi';
 export const CLASSIC_ARCHITECTURE = 'classic';
--- a/assets/js/lib/model/clusters.js
+++ b/assets/js/lib/model/clusters.js
@@ -4,20 +4,13 @@ export const HANA_SCALE_OUT = 'hana_scale_out';
 export const ASCS_ERS = 'ascs_ers';
 
 // Hana scale up scenarios
-export const HANA_SCALE_UP_PERFORMANCE_SCENARIO = 'performance_optimized';
 export const HANA_SCALE_UP_COST_OPT_SCENARIO = 'cost_optimized';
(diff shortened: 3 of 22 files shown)


.gitignore
```diff
@@ -42,3 +42,6 @@ npm-debug.log
 
 # Local dev.local.exs file
 config/dev.local.exs
+
+# ignore scenario
+/test/fixtures/scenarios/cost_optimised_sap_scenario
```
arbulu89, line 47: forgot to remove this?
we shouldn't have entries here for our internal tests

lib/trento/clusters/enums/hana_scenario.ex
```diff
@@ -0,0 +1,7 @@
+defmodule Trento.Clusters.Enums.HanaScenario do
+  @moduledoc """
+  Type that represents the supported HANA architecture types.
```
arbulu89, line 3: Update the docstring

lib/trento/discovery/payloads/cluster/cluster_discovery_payload.ex
```diff
@@ -176,6 +176,28 @@ defmodule Trento.Discovery.Payloads.Cluster.ClusterDiscoveryPayload do
     end)
   end
 
+  defp parse_cluster_additional_sids(%{
+         "cluster_type" => ClusterType.hana_scale_up(),
+         "cib" => %{"configuration" => %{"resources" => %{"primitives" => primitives}}}
+       }) do
+    primitives
```
arbulu89, line 183: Could we create a new function like `get_sapinstance_sids` or similar?
To avoid having the exact same code duplicated twice.

lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -172,7 +174,8 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
       stopped_resources: parse_cluster_stopped_resources(crmmon),
       nodes: parse_cluster_nodes(payload, sid),
       sbd_devices: parse_sbd_devices(sbd),
-      sites: parse_hana_scale_out_sites(HanaArchitectureType.classic(), cib, sid)
+      sites: parse_hana_scale_out_sites(HanaArchitectureType.classic(), cib, sid),
+      hana_scenario: parse_hana_scenario(crmmon, ClusterType.hana_scale_out())
```
arbulu89, line 178: Let's try to put the field in the same order. Above, you put it after `architecture_type`. Do the same everywhere

lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -1011,4 +1026,25 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
 
   defp parse_hana_cluster_health(%{sr_health_state: _, secondary_sync_state: _}),
     do: Health.critical()
+
+  defp parse_hana_scenario(crmmon, ClusterType.hana_scale_up()) do
```
arbulu89, line 1030: I don't get all of this. I thought we agreed that we were just checking if `additional_sids` have some element to decide if it is cost or performance optimized.
```
defp parse_hana_scenario([]), do: HanaScenario.performance_optimized()
defp parse_hana_scenario(_), do: HanaScenario.cost_optimized()
```
We don't need to check the resources again, we already did

lib/trento_web/openapi/v2/schema/cluster.ex
```diff
@@ -99,6 +100,11 @@ defmodule TrentoWeb.OpenApi.V2.Schema.Cluster do
             description: "HANA architecture type.",
             enum: HanaArchitectureType.values()
           },
+          hana_scenario: %Schema{
+            type: :string,
+            description: "Detected hana scenario",
```
arbulu89, line 105: I would simply put `HANA scenario type`.
PD: `HANA` always goes in capital letters

test/trento/discovery/policies/cluster_policy_test.exs
```diff
@@ -2008,6 +2015,7 @@ defmodule Trento.Discovery.Policies.ClusterPolicyTest do
                   cib_last_written: "Thu Feb 23 15:59:56 2023",
                   details: %HanaClusterDetails{
                     architecture_type: HanaArchitectureType.classic(),
+                    hana_scenario: HanaScenario.unknown(),
```
arbulu89, line 2018: Why `unknonwn`? Scale out should be performance optimized.
You can hardcode the type in the policy for this scenario

test/trento/discovery/policies/cluster_policy_test.exs
```diff
@@ -9,6 +9,7 @@ defmodule Trento.Discovery.Policies.ClusterPolicyTest do
   require Trento.Enums.Provider, as: Provider
 
   require Trento.Clusters.Enums.HanaArchitectureType, as: HanaArchitectureType
+  require Trento.Clusters.Enums.HanaScenario, as: HanaScenario
```
arbulu89, line 12: We need some new test for cost-optimized, right?

assets/js/lib/model/clusters.js
```diff
@@ -1,8 +1,24 @@
+// Cluster types
 export const HANA_SCALE_UP = 'hana_scale_up';
 export const HANA_SCALE_OUT = 'hana_scale_out';
 export const ASCS_ERS = 'ascs_ers';
 
-export const clusterTypes = [HANA_SCALE_UP, HANA_SCALE_OUT, ASCS_ERS];
+// Hana scale up scenarios
+export const HANA_SCALE_UP_PERFORMANCE_SCENARIO = 'performance_optimized';
```
arbulu89, line 7: Let's remove the `HANA_SCALE_UP` prefix from here.
Actually, Scale out are performance optimized as well

assets/js/lib/model/clusters.js
```diff
@@ -1,8 +1,24 @@
...
+export const HANA_SCALE_UP_COST_OPT_SCENARIO = 'cost_optimized';
+export const HANA_SCALE_UP_UNKNOWN_SCENARIO = 'unknown';
+
+export const clusterTypes = [
+  HANA_SCALE_UP,
+  HANA_SCALE_UP_PERFORMANCE_SCENARIO,
```
arbulu89, line 13: We careful, this shouldn't be a new cluster type

assets/js/common/ClusterTypeLabel/ClusterTypeLabel.jsx
```diff
@@ -44,11 +45,21 @@ const icons = {
...
 
-function ClusterTypeLabel({ clusterType, architectureType }) {
+function ClusterTypeLabel({
+  clusterType,
+  clusterScenario = '',
+  architectureType,
+}) {
+  const clusterTypeLabel = getClusterTypeLabel(clusterType);
+  const clusterTypeScenarioLabel = getClusterTypeScenarioLabel(clusterScenario);
+  const combinedClusterTypeLabel = `${clusterTypeLabel}${
```
arbulu89, line 55: Do we need this logic? we will always have a scenario, it cannot be empty

assets/js/pages/ClusterDetails/ClusterDetailsPage.test.jsx
```diff
@@ -10,13 +10,16 @@ import { ClusterDetailsPage } from './ClusterDetailsPage';
 
 describe('ClusterDetails ClusterDetailsPage component', () => {
   it.each([
-    { type: 'hana_scale_up', label: 'HANA Scale Up' },
+    { type: 'hana_scale_up', label: 'HANA Scale Up', scneario: 'unknown' },
```
arbulu89, line 13: Typo in `scneario`.
Anyway, this test looks really inconsistent. You shouldn't put the scenario only in one of the entries, and you should actually update the `label` as you are testing that, the label with different hana scenario, right?

assets/js/pages/ClusterDetails/ClustersList.jsx
```diff
@@ -79,7 +79,7 @@ function ClustersList() {
             const sapSystemData = getSapSystemBySID(allInstances, singleSid);
 
             return [
-              index > 0 && ', ',
+              index > 0 && <br />,
```
arbulu89, line 82: I wouldn't put a new line, maybe just a space. the newline breaks the table format in my opinion

assets/js/pages/ClusterDetails/HanaClusterDetails.jsx
```diff
@@ -150,14 +150,20 @@ function HanaClusterDetails({
...
+                content: enrichedSapSystems,
                 render: (content) => (
-                  <SapSystemLink
-                    sapSystemId={content?.id}
-                    systemType="databases"
-                  >
-                    {content?.sid}
-                  </SapSystemLink>
+                  <div>
+                    {content.map((system) => (
```
arbulu89, line 156: You should destruct the `id` and `sid` here: `{content.map(({ id, sid}) =>`

assets/js/pages/ClusterDetails/HanaClusterDetails.jsx
```diff
@@ -32,6 +32,9 @@ export const enrichNodes = (clusterNodes, hosts) =>
     ...hosts.find(({ hostname }) => hostname === node.name),
   }));
 
+const generateEnrichedSapSystem = (systems) =>
```
arbulu89, line 35: This new function, and the `enrichedSapSystems`, do they make any sense?
You are just putting id and sid in id and sid.
Why don't you simply pass `sapSystems` to the table?

arbulu89 (changes requested):
Looking good overall.
Some comments.
I see that tests and storybook updates are missing

assets/js/pages/ClusterDetails/ClustersList.jsx
EMaksy: The mockup showed a line break, we still can also have space, what do you think ? 
image

For me both works fine :)

assets/js/pages/ClusterDetails/HanaClusterDetails.jsx
```diff
@@ -150,15 +153,27 @@ function HanaClusterDetails({
...
-                  </SapSystemLink>
-                ),
+                content: enrichedSapSystems,
+                render: (content) =>
+                  content.every(
+                    (sapSystem) => sapSystem.id && sapSystem.sid
+                  ) ? (
+                    <div>
+                      {content.map(({ id, sid: sapSystemSid }) => (
+                        <span key={`${id}-${sapSystemSid}`}>
```
EMaksy, line 163: Had to add a key in the span and in SapSystemLink component to get rid off 
``` Each child in a list should have a unique "key" prop```
Any ideas how to handle it better ?

File: test/trento/discovery/policies/cluster_policy_test.exs (hunk 8 above)
Comment (EMaksy) [reply test/trento/discovery/policies/cluster_policy_test.exs:12]: Added test + payload

File: lib/trento/discovery/policies/cluster_policy.ex [hunk 17]
```diff
@@ -157,11 +160,13 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
            sbd: sbd,
            cluster_type: ClusterType.hana_scale_out(),
            hana_architecture_type: HanaArchitectureType.classic(),
+           additional_sids: additional_sids,
            sid: sid
          } = payload
        ) do
     %{
       architecture_type: HanaArchitectureType.classic(),
+      hana_scenario: parse_hana_scenario(additional_sids),
```
arbulu89, line 169: I think you could directly put `HanaScenario.performance_optimized()` here, as we don't have any scale out cost optimized.
I guess using the function will get us to the same result.
Up to you

lib/trento_web/openapi/v2/schema/cluster.ex
```diff
@@ -99,6 +100,11 @@ defmodule TrentoWeb.OpenApi.V2.Schema.Cluster do
             description: "HANA architecture type.",
             enum: HanaArchitectureType.values()
           },
+          hana_scenario: %Schema{
+            type: :string,
+            description: "Detected HANA scenario",
```
arbulu89, line 105: nitpick: maybe better `HANA scenario type` to be aligned of what we have aboe

assets/js/pages/ClusterDetails/HanaClusterDetails.jsx
```diff
@@ -51,10 +52,12 @@ function HanaClusterDetails({
   navigate = () => {},
 }) {
   const enrichedNodes = enrichNodes(details?.nodes, hosts);
-  const enrichedSapSystem = {
-    sid,
-    ...sapSystems.find(({ sid: currentSid }) => currentSid === sid),
-  };
+  const sortedSidList = [sid, ...additionalSids].sort();
```
arbulu89, line 55: We don't want to sort it. We want to have first the main sid and the QAS later on.
Just remove the `sort` function and rename the variable to `clusteredSids` maybe

assets/js/pages/ClusterDetails/HanaClusterDetails.jsx
```diff
@@ -150,15 +153,27 @@ function HanaClusterDetails({
...
-                  >
-                    {content?.sid}
-                  </SapSystemLink>
-                ),
+                content: enrichedSapSystems,
+                render: (content) =>
+                  content.every(
+                    (sapSystem) => sapSystem.id && sapSystem.sid
+                  ) ? (
+                    <div>
```
arbulu89, line 161: We don't need this sort of protection here.
Just using what you have withing the `div` should be enough. Use the `sid` as key, they cannot be repeated.
The `SapSystemLink` component will take care about the scenario where the `id` does not exist, it will go as undefined. Don't worry about that. You'll see in storybook.

arbulu89 (changes requested):
hey @EMaksy 
The backend is all good (maybe I wrote some small nitpick).
I have written comments about the frontend. Small details, but I would like to fix them

arbulu89 (approved):
Green light!

nelsonkopliku (approved):
Nothing to add.

Just for me to better understand: 
- perf opt and cost opt are only related to hana scale up (not scale out) 
- definition rule is: no additional sids -> perf opt, some additional sids -> cost opt

Am I getting it right?

arbulu89:
> Nothing to add.
> 
> Just for me to better understand:
> 
>     * perf opt and cost opt are only related to hana scale up (not scale out)
> 
>     * definition rule is: no additional sids -> perf opt, some additional sids -> cost opt
> 
> 
> Am I getting it right?

Yes, you are right!

assets/js/lib/model/clusters.js
```diff
@@ -16,6 +25,14 @@ const clusterTypeLabels = {
 export const getClusterTypeLabel = (type) =>
   clusterTypeLabels[type] || 'Unknown';
 
+const clusterScenarioLabels = {
+  [PERFORMANCE_SCENARIO]: 'Perf. Opt.',
+  [COST_OPT_SCENARIO]: 'Cost Opt.',
+};
+
+export const getClusterScenarioLabel = (type) =>
+  clusterScenarioLabels[type] || '';
```
dottorblaster, line 34: Do we need the `||`? Doesn't it work the same with just the access to the object?

dottorblaster (approved):
LGTM, just one minor thing that I happened to see :+1: 

Very good job!

assets/js/lib/model/clusters.js
dottorblaster: Fair enough :smile: 
For the future remember that we can also use `get` from lodash for better default semantics :+1: 

Thank you!

PR #3156, opened 2024-11-20
Allow the scheduled execution to check also ascs/ers clusters

nelsonkopliku:
# Description

While hackweeking I realized that the scheduled checks execution was not taking into account ascs/ers clusters for execution.

This PR fixes it.


arbulu89 (approved):
:tada: 

PD: The indentation of mox expects is so ugly XD

abravosuse (approved):
Good catch @nelsonkopliku . Thank you!

PR #3159, opened 2024-11-22
Omit an unnecessary query when requesting cluster checks execution

nelsonkopliku:
# Description

Just a tiny adjustment removing an unnecessary query when requesting a checks execution for ascs/ers clusters.

### DIFF
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -148,23 +148,6 @@ defmodule Trento.Clusters do
     |> select_merge([c, e], %{cib_last_written: e.cib_last_written})
   end
 
-  defp get_filesystem_type(%ClusterReadModel{} = cluster) do
-    filesystems_list =
-      cluster
-      |> Map.get(:details, %{})
-      |> Map.get("sap_systems", [])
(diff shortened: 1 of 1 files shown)


lib/trento/clusters.ex
```diff
@@ -148,16 +148,9 @@ defmodule Trento.Clusters do
...
   end
 
-  defp get_filesystem_type(cluster_id) do
-    query =
-      from(c in ClusterReadModel,
-        where:
-          c.id == ^cluster_id and c.type == ^ClusterType.ascs_ers() and is_nil(c.deregistered_at)
-      )
-
+  defp get_filesystem_type(%ClusterReadModel{} = cluster) do
```
arbulu89, line 151: nitpick: Could we move this `get_filesystem_type` function after `maybe_request_checks_execution`? To follow the order precedence

arbulu89 (approved):
Good!
Just a tiny nitpick :stuck_out_tongue_closed_eyes:

lib/trento/clusters.ex
nelsonkopliku: Here we go 696a3cc728fe87092d5981877a49c51b647bf21b

PR #3160, opened 2024-11-22
requiring target's selectable checks

nelsonkopliku:
# Description

During hackweek I attempted implementing a TUI (`trentoctl`) to interact with Trento.

I wanted to be able to:
- list hosts/clusters
- select an entry
- show selectable checks for such entry
- select desired checks
- issue a checks execution

Now, besides the fanciness and challenges of implementing a tui in rust and https://ratatui.rs/, it is mainly a matter of integrating with Trento APIs: serialize/deserialize json, making http requests, checking responses... pretty common things.

---

Let's focus on: **showing selectable checks for a target**

In order to answer the question _which are the selectable checks for a given target (cluster/host)?_ we need to query Wanda's catalog API with a specific set of parameters based on the target for which we are querying selectable checks.

For a host we need just a provider, while for a cluster we need its type, the architecture type if hana scale up or hana scale out, while we need ensa version and filesystem type when talking about an ascs/ers cluster...

This sounds like delegating too much to integrating clients (logic leakage), and leads to a lot of boilerplate code to achieve a well known outcome like _the selectable checks for a given target (cluster/host/maybemorecomingnext)_.

As someone integrating with trento I'd find it useful having something more straightforward to interact with in order to get that information.

Here is where I was considering the option of having two endpoints in trento doing this
- GET `/clusters/:cluster_id/available_checks`
- GET `/hosts/:id/available_checks`

What I like about this option:
- simplifies integrating with checks execution process within trento context
- reduces cognitive overhead of who integrates
- reduces logic duplication and possibility for errors
- allows for frontend code simplification

What I don't like much:
- proxy controller towards wanda (which anyway is what we do with SUMA integration)

Note that wanda as such remains untouched in it's agnosticism.

Happy to hear thoughts. If the team agrees on this approach the implementation will be finalized.

### DIFF
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -41,6 +41,21 @@ defmodule Trento.Clusters do
     end
   end
 
+  def list_selectable_checks(cluster_id) do
+    cluster_id
+    |> with_registered_cluster()
+    |> perform_operation(fn hosts_data, cluster ->
+      Logger.debug("Fetching selectable checks for cluster: #{cluster_id}")
+
+      cluster
+      |> build_env(hosts_data)
+      |> Checks.list_selectable_checks(:cluster)
+    end)
+    |> on_cluster_not_found(fn _ ->
+      Logger.error("Fetching selectable checks for a non-existing cluster: #{cluster_id}")
+    end)
+  end
+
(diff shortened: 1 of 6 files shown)


nelsonkopliku:
Closing as there seems to be not much interest.

PR #3165, opened 2024-11-28
Use prometheus targets

arbulu89:
# Description

Use new information about the prometheus targets to create the http service discovery data.
Pretty straightforward.

Related do: https://github.com/trento-project/agent/pull/356

PD:
I haven't updated the e2e and photofinish payloads to include the new field, because in reality we don't use them neither in the tests or in demo.
I think it is fine as is.
Otherwise, we could create other PR as well, to not bloat this one and make the review slightly simpler

## How was this tested?

UT added

### DIFF
--- a/lib/trento/hosts/host.ex
+++ b/lib/trento/hosts/host.ex
@@ -129,7 +129,7 @@ defmodule Trento.Hosts.Host do
     field :socket_count, :integer
     field :os_version, :string
     field :provider, Ecto.Enum, values: Provider.values()
-    field :prometheus_targets, :map, default: %{}
+    field :prometheus_targets, :map
     field :installation_source, Ecto.Enum, values: [:community, :suse, :unknown]
     field :heartbeat, Ecto.Enum, values: [:passing, :critical, :unknown]
     field :checks_health, Ecto.Enum, values: Health.values(), default: Health.unknown()
--- a/lib/trento/hosts/host.ex
+++ b/lib/trento/hosts/host.ex
@@ -129,6 +129,7 @@ defmodule Trento.Hosts.Host do
     field :socket_count, :integer
     field :os_version, :string
     field :provider, Ecto.Enum, values: Provider.values()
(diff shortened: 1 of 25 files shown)


lib/trento/hosts/host.ex
```diff
@@ -129,6 +129,7 @@ defmodule Trento.Hosts.Host do
     field :socket_count, :integer
     field :os_version, :string
     field :provider, Ecto.Enum, values: Provider.values()
+    field :prometheus_targets, :map, default: %{}
```
CDimonaco, line 132: I think we could omit the default here, like we don't put defaults in commands/events this could be easier when we want in the future spot what instances has the targets

lib/trento/hosts/projections/host_read_model.ex
```diff
@@ -32,6 +32,7 @@ defmodule Trento.Hosts.Projections.HostReadModel do
     field :provider, Ecto.Enum, values: Provider.values()
     field :provider_data, :map
     field :saptune_status, :map
+    field :prometheus_targets, :map
```
CDimonaco, line 35: considering we don't have a default in the read model I think we could omit in the aggregate

CDimonaco (approved):
Just some general comments, it looks beautiful LGTM

lib/trento/hosts/host.ex
arbulu89: yes, I did initially this because otherwise as the default value is `nil`, the json view conversion was failing. But I can add a condition there to return empty targets list if the `prometheus_targets` value is nil.
I will add a patch

arbulu89: @CDimonaco Change done.
I have removed the defaults. After doing this, I have done other tiny change, to still leave the "current" way of creating targets, just in case older agents are sending data, so we still have the current behaviour.
Removing will be super easy at any point, when we decide.

nelsonkopliku (approved):
Noice

PR #3169, opened 2024-12-02
Activity log autorefresh e2e

nelsonkopliku:
# Description

Even though tested via jest, some e2e might be valuable.

In fact I spotted a tiny issue that has been fixed 😄


balanza (approved):
Nice catch!

PR #3183, opened 2024-12-05
Limit user access in activity log filters

nelsonkopliku:
# Description

This PR makes sure that only relevant users are exposed as in activity log filter by user, based on the currently logged user's permissions.

A user with any of `all:all`, `all:users`, `activity_log:users` permissions can access all users in the activity log filters, otherwise only self and `system` are available.

Following up also the activity log API output and related view are going to be restricted based on caller permissions.

## How was this tested?

Automated tests.


balanza (approved):
Good implementation, thanks

nelsonkopliku:
@balanza thank you for the review. 
This PR is based on https://github.com/trento-project/web/pull/3180 where the new `activity_log:users` is added. 
Could you please review that as well? 😄

PR #3184, opened 2024-12-05
Add e2e test for hana scale up cost optimized scenario

EMaksy:
# Description

This PR is an ad-don to the newly introduced[ hana scale up scenario discovery ](https://github.com/trento-project/web/pull/3152). 
Besides the front and back end tests this pr introduces e2e tests.

Added a new cluster called hana_cost_opt which contains a Database called HDC and QAS
image

This scenario adds a hana cost opt scenario
image

## What has been changed ? 

- Refactor of the HostsList to remove the line break, this helps with the e2e tests and aligns like in other parts of the code
- Updated existing tests in order to check for multiple SID's which a cost opt scenario has
- Updated index of old tests as new hosts were added in the scenario
- Added new test for hana scale up cost opt
- Added two new nodes scenarios 2372b24f-3d7a-5d01-9b1a-a2c4c95c53d4 and fa7a5602-232b-5389-96a6-f5f5de6ff9a2


assets/js/pages/HostsList/HostsList.jsx
```diff
@@ -119,22 +119,18 @@ function HostsList() {
...
+            <span key={`${sids[index]}-${instance?.id}`}>
+              <SapSystemLink
+                key={`${getInstanceID(instance)}-${instance?.id}`}
+                systemType={instance?.type}
+                sapSystemId={getInstanceID(instance)}
+              >
+                {instance?.sid}
+              </SapSystemLink>{' '}
+            </span>
+          ));
```
dottorblaster, line 133: I'm not sure a PR introducing some tests should actually modify features' code too :smile: 


This refactor is just made to remove the line break right?

dottorblaster (commented):
The PR looks good, I just wanted to think more about the fact that we're turning the whole `27-nodes` landscape into `29-nodes`. I don't know if it's something we can avoid due to how the tests are structured.

@balanza @CDimonaco do you have any idea about any other possible approach?

assets/js/pages/HostsList/HostsList.jsx
EMaksy: Did the small refactor to remove the line break, it produced some issues with existing tests when additional sid was added. 
I guess i can revert it, but i may to modify existing tests more. 
What do you think ?

EMaksy:
The main reason for renaming the healthy-27-node-SAP-cluster  scenario to healthy-29-node-SAP-cluster is that we added two new nodes.. tbh i never thought about the name of healthy-27-node-SAP-cluster before i saw that it contains 27 systems. 
An alternative approach would be to keep the 27 scenario and add a new 29 scenario. Use only the healthy-29-node-SAP-cluster scenario where required, but it is at a lot of points tbh so not sure why to keep the duplicate.

EMaksy:
Actually it makes a lot of sense just to remove the number tbh of the scenario. 
@CDimonaco any additional ideas before i start making changes ?

assets/js/pages/HostsList/HostsList.jsx
CDimonaco: This change have some ui changes? Like in the appaerence of things

test/e2e/cypress/e2e/clusters_overview.cy.js
```diff
@@ -54,7 +54,9 @@ context('Clusters Overview', () => {
           .contains('th', 'SID')
           .invoke('index')
           .then((i) => {
-            cy.get('@clusterRow').eq(i).should('contain', cluster.sid);
+            cluster.sid.forEach((sid) => {
```
CDimonaco, line 57: Maybe `sids` instead of sid?

test/e2e/cypress/e2e/hana_cluster_details.cy.js
```diff
@@ -233,6 +234,102 @@ context('HANA cluster details', () => {
...
+        .next()
+        .contains(availableHanaClusterCostOpt.provider);
+    });
+
+    it('should have all cost optimized SID`s and correct database links', () => {
+      cy.get('.tn-cluster-details')
+        .contains('SID')
+        .parent()
+        .within(() => {
+          availableHanaClusterCostOpt.sid.forEach((sid, index) => {
```
CDimonaco, line 259: sids as the previous one

test/e2e/cypress/e2e/hosts_overview.cy.js
```diff
@@ -140,13 +140,19 @@ context('Hosts Overview', () => {
...
           .then((i) => {
             if (host.clusterId) {
-              cy.get('@hostRow').eq(i).should('contain', host.sapSystemSid);
-              cy.get('@hostRow').eq(i).click();
-              cy.location('pathname').should(
-                'eq',
-                `/databases/${host.sapSystemId}`
-              );
-              cy.go('back');
+              host.sapSystemSid.forEach((sid, counter) => {
```
CDimonaco, line 143: prural

test/e2e/cypress/fixtures/clusters-overview/available_clusters.js
```diff
@@ -2,55 +2,61 @@ export const availableClusters = [
   {
     id: '8a66f8fb-5fe9-51b3-a34c-24321271a4e3',
     name: 'drbd_cluster',
-    sid: '',
+    sid: [''],
```
CDimonaco, line 5: ofc prurals also here

CDimonaco (commented):
Just left some comments

CDimonaco:
I think we can stick with the numbers of nodes in the scenario name, we can think about it later.
Just to keep the convention alive until we don't need it anymore.

dottorblaster (approved):
LGTM, just apply comments by @CDimonaco and that's :fire:

assets/js/pages/HostsList/HostsList.jsx
EMaksy: No it looks the same and that is the fun thing about it :+1:

PR #3191, opened 2024-12-06
Show password toggle on login

balanza:
# Description

Add toggle to show password on login

|Screenshot from 2024-12-06 11-44-39|Screenshot from 2024-12-06 11-45-18|
|-|-|

### DIFF
--- a/assets/js/pages/Login/Login.test.jsx
+++ b/assets/js/pages/Login/Login.test.jsx
@@ -47,10 +47,12 @@ describe('Login component', () => {
 
     await waitFor(() => screen.getByText('Invalid credentials'));
 
-    ['username', 'password'].forEach((id) => {
-      const element = screen.getByTestId(`login-${id}`);
-      expect(element).toHaveClass('border-red-500');
-    });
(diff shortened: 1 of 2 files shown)


nelsonkopliku (commented):
Nice idea!

Have you considered using the `Password` component we already have?

stefanotorresi:
Great idea!

I know this will involve quite a lot of CSS wrestling but, for consistency, consider using the same UX we have for the SUMA settings password input, with a little eye icon embedded within the password input.

Screenshot 2024-12-06 115339

That said feel also free to first merge this and then follow up with the improved one at a later moment.

stefanotorresi:
oh okay, from Nelson's comment I seem to understand that we can even reuse the same exact component, so one more reason to do it. 😄

nelsonkopliku (approved):
LGTM

jagabomb:
@balanza Thanks for this! I think it really helps to create more consistency with the UX. I am willing to trade simple border colours for an overall better UX.

jagabomb (approved):
❤️ LGTM

PR #3195, opened 2024-12-09
Redact users in activity log

nelsonkopliku:
# Description

With this PR we make sure that a user without necessary permissions receives redacted usernames in the activity log, except for own username and system.

In addition a check has been added that allows such users to only be able to query by their own username or by system.


balanza (approved):
LGTM

PR #3196, opened 2024-12-10
Order users query

nelsonkopliku:
# Description

Sometimes test at `test/trento_web/controllers/v1/users_controller_test.exs:59` fails.

Seems to be related to ordering of the entries returned, see [this failing pipeline](https://github.com/trento-project/web/actions/runs/12252134618/job/34178144423#step:8:203).

By adding an order by clause to the query we get a deterministic result and the test should not fail.

### DIFF
--- a/lib/trento/users.ex
+++ b/lib/trento/users.ex
@@ -42,6 +42,7 @@ defmodule Trento.Users do
     |> where([u], is_nil(u.deleted_at))
     |> preload(:abilities)
     |> preload(:user_identities)
+    |> order_by(asc: :id)
     |> Repo.all()
   end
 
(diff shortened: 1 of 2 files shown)


balanza (approved):
I must say that adding something to production working code for the sake of a flecky test is at least controversial.

Anyway, the `list_users` function is used in contexts in which an order by id is of no harm.

LGTM

nelsonkopliku:
@balanza I see.

Options are:
- we keep the flaky test
- we change the test so that it checks whether the expected users are in the returned list, regardless of the order
- we add a deterministic ordering

If went for what made most sense to me. Would you rather change the test?

balanza:
As I said, the PR is fine and you can merge. Btw good catch, it never occurred to me that this test failed.

An alternative could be to fix the test to check results for membership, something like
```elixir
      assert Enum.find_value(resp, false, fn
               %{id: ^user_one_id} -> true
               _ -> false
             end)
```

The test is flaky because is based on the wrong assumption that the order is guaranteed by either Ecto or Postgres.

nelsonkopliku:
@balanza changed the test so that it relies on membership AND kept the order by which is not harmful for the result to be deterministic.

PR #3200, opened 2024-12-11
Improve e2e tests on Activity log

balanza:
# Description

Narrow down the selection of pagination elements on Activity Log e2e tests using the dedicated test ID.

By selecting only by text, being the text to query for just a number, sometimes the wrong element is clicked hence the script doesn't work ([example](https://github.com/trento-project/web/actions/runs/12122807801/job/33796890089?pr=3155#step:15:133)).


nelsonkopliku (approved):
Whoops, missed those. Thanks!

PR #3206, opened 2024-12-13
Hana Scale up cost opt checks in the trento console

EMaksy:
# Description

This pr will enrich the communication between web and wanda required for the  hana scale up  cost opt and performance opt scenario. By updating the env which is send to wanda, web receives the correct checks related to the hana scale up scenario. 
Additionally, updated the internal env building on the backend to inform about the correct hana scale up scenario and update the checks execution properly. 
Visually, the frontend displays the cluster type correctly according to the scenario in the ClusterDetailsPage and ChecksResultDetails view.

Additionally to this, the pr requires an addition to all existing checks:
https://github.com/trento-project/checks/pull/23
The existing checks require the new hana_scenario
and a small update to wanda docs
https://github.com/trento-project/wanda/pull/538

## How was this tested?

- updated and added frontend tests
- updated and added backend tests


dottorblaster (approved):
LGTM

PR #3208, opened 2024-12-17
Cypress E2E Tests parallelization

vicenteqa:
# Description
~~I am creating this PR to check in an isolated workflow changes that I may have to try to achieve Cypress parallelization.~~

~~As an starting point I copied the development made by @CDimonaco in his PR and took only the E2E part. I want to try this after some changes that I made in the E2E tests.~~

This PR implements Cypress Split plugin using 8 parallel threads.
To optimize test execution, I decoupled each test file, allowing it to run independently without relying on prior tests. This resolved issues where some tests required multiple executions of the "photofinish" main scenario, while others needed it to revert changes from previous tests. By executing "photofinish" once at the beginning of the test suite (either sequentially or in parallel in every CI thread) and removing unnecessary calls, overall test execution time has been reduced too.

Added logic to execute loadScenario twice when needed and removed that logic from pre-cypress code execution so we can benefit of having cy.loadScenario and let each test use it or not.

Added a step in the CI that needs previous parallel execution step to pass to meet the required check in the branch protection rule, this way we don't need to add a check for every thread and would be good if in the future the number of threads changes.

This PR also adds ability to run each test file independently from other files to be run before, so now test files are decoupled.

~~I also added a small delay in start agent hearbeat function as I've noticed that it reduces flakiness in hosts_overview tests.~~

Fixes # (issue)

- [X] **DONE**

## How was this tested?

- [ ] **DONE**

- [ ] **DONE**


.github/workflows/cypress_parallel_poc.yaml
```diff
@@ -0,0 +1,210 @@
...
+
+jobs:
+  elixir-deps:
+    name: Elixir ${{ matrix.mix_env }} dependencies
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        include:
+          - mix_env: dev
+          - mix_env: test
```
balanza, line 24: **question:** ‏Why the need to iterate on both environments>

vicenteqa: this was copied from Carmine's pr, this file is no longer available :)

test/e2e/cypress/plugins/index.js
```diff
@@ -15,18 +15,27 @@
...
 let heartbeatsIntervals = [];
 
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+  on('before:run', async () => {
+    const photofinishBinary =
+      await getPhotofinishBinaryAndGiveExecutablePermissions();
+    await runPhotofinishMainScenario(photofinishBinary);
+  });
```
balanza, line 32: **question:** ‏This means the scenario is loaded once per process, right?
So, given the target application is the same, the CI will be executed 8 times while on local execution just once.

test/e2e/cypress/plugins/index.js
```diff
@@ -37,13 +46,15 @@ module.exports = (on, config) => {
...
           .end();
 
-      agents.forEach((agentId) => {
-        heartbeat(agentId);
-        let interval = setInterval(
-          () => heartbeat(agentId),
-          heartbeat_interval
-        );
-        heartbeatsIntervals.push(interval);
+      sleep(500).then(() => {
```
balanza, line 49: **issue:** ‏Do you think we can better understand why it is needed, to remove it?
Sleeps on arbitrary time spans are often cause of flakiness, as they may depend on the running machine

balanza (commented):
Great job @vicenteqa! I'm very happy we came up with a solution that does not require an invasive refactor on code. To be honest, I wasn't so optimistic on that.

I left some comments on things I think we can improve. Besides that, I think we can merge. 

Maybe we want to _stage_ this PR for a while, to verify if tests are consistent after some changes to the codebase? We can rebase a couple of time and see.

test/e2e/cypress/plugins/index.js
```diff
@@ -66,3 +76,26 @@ module.exports = (on, config) => {
...
+  );
+  await runCommand(`chmod +x ${photofinishBinary}`);
+  return photofinishBinary;
+}
+
+async function runPhotofinishMainScenario(photofinishBinary) {
+  return runCommand(
+    `cd ../.. && ${photofinishBinary} run --url "http://localhost:4000/api/collect" healthy-29-node-SAP-cluster`
+  );
+}
```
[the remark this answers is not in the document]
vicenteqa: because when this piece of code is executed, cy commands are still not available

test/e2e/cypress/plugins/index.js
vicenteqa: I'm not a fan of arbitrary waits either and I know they should be avoided as a good practice, for the moment I don't have a conclusive answer tbh but I noticed that with a bit of delay the hosts_overview test flakiness is reduced a lot so I added it as a contingency measure, we can remove it if you want or even we can try the test flakiness tool designed by @gagandeepb then try with and without the arbitrary sleep and compare results.

test/e2e/cypress/plugins/index.js
vicenteqa: that's it, once per process, so if you execute the whole test suite locally (sequential) this will be executed just once at the very beginning.

test/e2e/cypress/plugins/index.js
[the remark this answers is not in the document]
balanza: > process.env 

Maybe it's better, we keep the same structure. 

Also, if we don't use the `loadScenario` command anymore, should we remove it?

vicenteqa:
> Maybe we want to _stage_ this PR for a while, to verify if tests are consistent after some changes to the codebase? We can rebase a couple of time and see.

I agree with this approach :)

test/e2e/cypress/plugins/index.js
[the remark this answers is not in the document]
vicenteqa: well, actually it is still used in some tests, specially to restore scenarios that previous test modified, and to load some other different scenarios.

test/e2e/cypress/support/commands.js
```diff
@@ -85,6 +85,25 @@ Cypress.Commands.add('updateApiKeyExpiration', (apiKeyExpiration) => {
   });
 });
 
+Cypress.Commands.add('preloadTestData', () => {
+  cy.isTestDataLoaded().then((isLoaded) => {
+    if (!isLoaded) cy.loadScenario('healthy-27-node-SAP-cluster');
+  });
+  cy.loadScenario('healthy-27-node-SAP-cluster');
```
balanza, line 92: **issue:** ‏Data is loaded regardless of the result of `isTestDataLoaded`. Also, in the case of `isLoaded==false`, it's loaded twice

balanza (changes requested):

test/e2e/cypress/support/commands.js
vicenteqa: in some tests it is needed to be loaded twice, so for some tests it will be loaded once and for some others it will be loaded twice. This is required for various reasons
1. Add ability to run single spec files
2. Restore changes that previous tests have done
3. To fully load the test scenario we need to load scenario twice as some data require some previous [pre conditions](https://suse.slack.com/archives/C02U4TWGKSQ/p1734447399649439?thread_ts=1734418842.256039&cid=C02U4TWGKSQ) in order to be loaded in trento.

balanza: Your point is good, I'm just afraid the actual code does not respect what you want to achieve.

As far as I understand, every time `preloadTestData` is executed it does the following:
* it always loads the scenario once (line 92)
* it checks data has been loaded (line 90)
* if not, it loads the scenario again (line 90)

Thus data is olways loaded once per spec file, and `isLoaded` will always be true. Isn't it?

Don't we just need to remove line 92?
```suggestion
  cy.isTestDataLoaded().then((isLoaded) => {
    if (!isLoaded) cy.loadScenario('healthy-27-node-SAP-cluster');
  });
```

vicenteqa: when running the full test suite (or a subset of tests), it will only be false the first time, that's for sure. With this logic what we avoid is to call loadScenario twice when it's not needed and to have a consistent test scenario, as there are things that in the first shot of photofinish, trento doesn't load but it does in the second time.

balanza: Given `preloadTestData` is called in 13 specs, the `loadScenario` command is executed 13 times when running the full test suite. 
Is this expected?

vicenteqa: yes, that was happening before this pr as well when calling `loadScenario` directly in every test, the problem with that was that some tests were failing when running isolated and to fix them the only solution was to add a second load scenario in those specific tests (which wasn't very elegant) or add this logic so `loadScenario` is called twice only if needed. Speaking with @arbulu89 we decided that this approach was better so we let every test decide if it needs to load an scenario or not, instead of executing this prior to launch cypress tests as I was initially doing. This way we can still use cypress helper and we are not "hiding" things to the tests by doing them before starting.

arbulu89: @vicenteqa
question: why do we need an additional `isTestDataLoaded` command? This command is not used in any tests. The commands are supposed to be "public" functions available for real tests.
Why don't you put the code directly in `preloadTestData`? I find it better.
Otherwise just a simple js function.
I don't think we should be exposing functions that we don't use anywhere else for "code structuring"

test/e2e/cypress/support/commands.js
```diff
@@ -85,6 +85,25 @@ Cypress.Commands.add('updateApiKeyExpiration', (apiKeyExpiration) => {
   });
 });
 
+Cypress.Commands.add('preloadTestData', () => {
```
arbulu89, line 88: nitpick: can we put a docstring comment to this?
Something like:
```
Preload required test data. It must run photofinish scenario twice as the order of sent payloads is relevant and the tests require a fully loaded scenario which only happens when the scenario is sent in the second time
```
At the end, this is a handicap we have and maybe a temporary solution.
```

arbulu89 (approved):
All good!
As commented, I would remove the `sleep` code, and if needed, work it in other PR.
The other code structuring is just a nitpick

test/e2e/cypress/support/commands.js
vicenteqa: totally agree about the cy command, I added that function for a better readability and understanding of  what `preloadTestData` command does so I would go and just make a it simple js function.

balanza: Sorry guys, I still think this code doesn't do what is expected. 

What is needed:
* test data to be loaded 2 times
* test data to be the same, no matter if we run the spec individually, in small batches, or all together (that is: it is always loaded twice)

What we have:
* test data is loaded once if the spec is executed independently
* test data is loaded N times, where N is the number of specs in the batch

Am I wrong?

arbulu89: @balanza I think the assumption misses some points.

> What is needed:
>
> -    test data to be loaded 2 times

Only if we don't have any previous data, the database is empty. If we already have data we don't need to load twice

> -   test data to be the same, no matter if we run the spec individually, in small batches, or all together (that is: it is always loaded twice)

We achieve this rerunning the 27 nodes at least once before each execution as this restores changes applied by previous test execution.

> What we have:
>
>  - test data is loaded once if the spec is executed independently

With the current change, the scenario is loaded twice only if we have an empty database, which satisfies our need
>   - test data is loaded N times, where N is the number of specs in the batch

We need this as well, as previous tests might have applied changes that we need to restore running the 27 nodes scenario again (this is what we do right now, before the parallelization)

balanza: Ok I didn't get it at first, thanks

balanza (approved):

PR #3212, opened 2024-12-23
Cost opt e2e

arbulu89:
# Description
Refactor HANA scale up cost-optimized e2e tests adding a new photofinish scenario instead of updating an already existing one.

Refactor of: https://github.com/trento-project/web/pull/3213


docker-compose.yaml
```diff
@@ -54,6 +54,7 @@ services:
       - host.docker.internal:host-gateway
   postgres:
     image: postgres:15
+    command: -c 'max_connections=200'
```
arbulu89, line 57: Adding this has fixed the e2e test that was failing always:
https://github.com/trento-project/web/actions/runs/12478685578/job/34826632861

It creates a `max_connections` hit error, the host takes a while to roll up, and the events sent to this host are excluded as it is in `rolling_up` state.

dottorblaster (approved):

PR #3215, opened 2024-12-27
Add job for flaky e2e tests detection

vicenteqa:
# Description

This pull request introduces a scheduled job to regularly identify and report on the flakiest tests in our suite.

**Implementation**

- A new GENERATE_REPORTS node environment variable controls the generation of Mocha JUnit reports, with a helper function ensuring unique timestamps for each report.
- The job utilizes a matrix strategy to run the test suite across multiple containers (defaulting to 100), allowing for parallel execution and efficient resource utilization.
- Each container uploads its test results as an artifact. These artifacts are then downloaded, organized, and analyzed using the script developed by @gagandeepb.
- The artifacts are automatically deleted after analysis.
- The analysis results are displayed in the job summary for easy review.

**Scheduling**

The job is currently scheduled to run weekly on Mondays at 3 AM.  **This schedule is open to discussion.**

**Performance**

The job typically completes within 1 hour and 30 minutes.  With 100 parallel threads, resource consumption peaks at around 20 machines working at the same time. 

No changes required in documentation.

Fixes # (issue)

- [x] **DONE**

## How was this tested?

- [x] **DONE**

- [X] **DONE**


.github/workflows/e2e-flaky-tests-detection.yaml
```diff
@@ -0,0 +1,263 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0    
+      - name: Download JUnit reports artifacts
+        uses: actions/download-artifact@v4
+        with:
+          path: junit-reports
+          pattern: e2e-junit-reports-*
+          merge-multiple: true
+      - name: Analize
```
gagandeepb, line 253: minor: typo

.github/workflows/e2e-flaky-tests-detection.yaml
```diff
@@ -0,0 +1,263 @@
+
+name: E2E Flaky Tests Analysis
+
+on:
+  schedule:
+    - cron: '0 3 * * 1' # Runs every monday at 3:00 AM UTC
```
gagandeepb, line 6: Question: I wonder if there should be a provision to run this workflow on-demand/by explicit opt-in ? For example, if a PR provides a fix for a known flaky test and would like to run `flaky test analysis` to verify that the fix possibly addressed the issue?

.github/workflows/e2e-flaky-tests-detection.yaml
```diff
@@ -0,0 +1,263 @@
...
+        with:
+          path: junit-reports
+          pattern: e2e-junit-reports-*
+          merge-multiple: true
+      - name: Analize
+        run: |
+          make venv-create
+          source .venv/bin/activate
+          make install-deps
+          make PATH-TO-JUNIT-FILES=../../junit-reports analyze-files 2>&1 | tail -n +2 >> $GITHUB_STEP_SUMMARY
```
gagandeepb, line 258: Question: I wonder if weekly reports/files saved in this location can be enumerated/aggregated/summarized at a later point in time, across the different weekly runs ?

gagandeepb (commented):
Thanks for looking into this! It is looking good. However, I have a few questions/comments.

.github/workflows/e2e-flaky-tests-detection.yaml
vicenteqa: Hey Gagan, Right now I am not sure for how long we can access the summaries (reports) of previous actions, otherwise maybe we can upload these reports to some storage server (S3?) so we can have a check in top of that to verify that the trend is fine.

.github/workflows/e2e-flaky-tests-detection.yaml
vicenteqa: that's a good point, the issue I see is that we consume a lot of resources (runners) for this check and we would leave other people without runners for keep doing their business as usual activities if we check that during regular working hours. A quick way I see for achieving that is to add a prefix in the branch that fixes a flaky test and run the workflow in pull request sync if the prefix matches.

balanza: > the issue I see is that we consume a lot of resources (runners)

+1, all the other PRs in the entire organization are queued for one hour when this workflow runs

> I wonder if there should be a provision to run this workflow on-demand

Isn't that what the `workflow_dispatch` clause does? Alternatively, we can activate the script based on a label or a comment

test/e2e/cypress.config.js
```diff
@@ -26,4 +29,12 @@ module.exports = defineConfig({
     testIsolation: false,
     baseUrl: 'http://localhost:4000',
   },
+  reporter:
+    process.env.GENERATE_REPORTS === 'true' ? 'mocha-junit-reporter' : 'spec',
+  reporterOptions:
+    process.env.GENERATE_REPORTS === 'true'
+      ? {
+          mochaFile: `/tmp/trento-e2e-junit-[hash]-${generateTimestamp()}.xml`,
```
balanza, line 37: **nitpick:** ‏We may want this path to come from an environment variable. This is because the path is usually meaningful for the context of the caller, not the script.
We may have a default, though.

.github/workflows/e2e-flaky-tests-detection.yaml
vicenteqa: yes, I left workflow_dispatch clause there in case we want to manually trigger it, which in the end is similar as having a label, branch prefix, or any other condition that triggers it. I don't have a strong opinion on how we should proceed, but we need to be careful when this job executes because it blocks everything else for some time.

.github/workflows/e2e-flaky-tests-detection.yaml
```diff
@@ -0,0 +1,263 @@
...
+          name: e2e-junit-reports-${{ matrix.container }}
+          path: /tmp/*.xml
+
+
+  flaky-tests-analysis:
+    name: Flaky Tests Analysis
+    needs: test-e2e
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Cancel Previous Runs
```
balanza, line 239: **question:** ‏Should we run this step at the beginning of the workflow, so we don't wait for the build steps?

.github/workflows/e2e-flaky-tests-detection.yaml
gagandeepb: I see. If it is an expensive run, it would also be fine to run it on schedule once a month.

gagandeepb: For running specific test suite on demand, there is the `--spec <specific-suite-one-wants-to-run>` Cypress option (also seen in the make target as `RUN_ONLY_*`). Maybe exposing that specific suite name as build parameter/input could be an option ?

stefanotorresi: If we need more to start paying for CI runners, let me know. Don't let this become a resourcing problem.

balanza (approved):

PR #3217, opened 2024-12-31
Update community container image to use fixed versions from leap15.6

arbulu89:
# Description
The build of the community image is failing since `tumbleweed` erlang version has been upgraded to `otp-27` and some our dependencies don't build with that version (something to be fixed at some point).

This PR simply sets opensuse Leap 15.6 as baseline and sets fixed versions of elixir and erlang.

Fixes https://github.com/trento-project/web/actions/runs/12560125083/job/35017432246

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,10 @@
-FROM opensuse/tumbleweed AS elixir-build
+FROM opensuse/leap:15.6 AS elixir-build
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
-RUN zypper -n in make gcc git-core elixir elixir-hex erlang-rebar3
+RUN zypper ar https://download.opensuse.org/repositories/devel:sap:trento:builddeps/15.6 buildeps
+RUN zypper -n --gpg-auto-import-keys ref
+RUN zypper -n in make gcc git-core elixir115 elixir-hex erlang26 erlang-rebar3
 COPY . /build
 WORKDIR /build
 ARG MIX_ENV=prod
@@ -24,10 +26,13 @@ ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 ARG MIX_ENV=prod
 ENV MIX_ENV=$MIX_ENV
+ENV MIX_HOME=/usr/bin
+ENV MIX_REBAR3=/usr/bin/rebar3
+ENV MIX_PATH=/usr/lib/elixir/lib/hex/ebin
 RUN mix phx.digest
 RUN mix release
 
-FROM opensuse/tumbleweed AS trento
+FROM opensuse/leap:15.6 AS trento
 LABEL org.opencontainers.image.source="https://github.com/trento-project/web"
 ARG MIX_ENV=prod
 ENV LANG en_US.UTF-8
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM opensuse/leap:15.6 AS elixir-build
-ENV LANG en_US.UTF-8
-ENV LANGUAGE en_US:en
-ENV LC_ALL en_US.UTF-8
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,8 +1,10 @@
-FROM opensuse/tumbleweed AS elixir-build
+FROM opensuse/leap:15.6 AS elixir-build
 ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
-RUN zypper -n in make gcc git-core elixir elixir-hex erlang-rebar3
+RUN zypper ar https://download.opensuse.org/repositories/devel:sap:trento:builddeps/15.6 buildeps
+RUN zypper -n --gpg-auto-import-keys ref
+RUN zypper -n in make gcc git-core elixir115 elixir-hex erlang26 erlang-rebar3
```
stefanotorresi, line 7: let's keep using version operators instead of explicit package names, it's arguably cleaner and it leaves room for an easier switch to more relaxed requirements.
```suggestion
RUN zypper -n in make gcc git-core elixir==1.15 elixir-hex erlang==26 erlang-rebar3
```

packaging/suse/container/Dockerfile
```diff
@@ -11,10 +11,10 @@ WORKDIR /build/web/assets
 RUN npm run tailwind:build
 RUN npm run build
 
-FROM bci/bci-base:15.4 AS release
+FROM bci/bci-base:15.6 AS release
 # Workaround for https://github.com/openSUSE/obs-build/issues/487
 RUN zypper --non-interactive in sles-release
```
stefanotorresi, line 16: This workaround is not needed any longer. Maybe take the chance to remove it?

packaging/suse/container/Dockerfile
```diff
@@ -11,10 +11,10 @@ WORKDIR /build/web/assets
...
-RUN zypper -n in make gcc git-core elixir>=1.15 elixir-hex erlang-rebar3
+RUN zypper -n in make gcc git-core elixir115 elixir-hex erlang26 erlang-rebar3
```
stefanotorresi, line 17: same as above
```suggestion
RUN zypper -n in make gcc git-core elixir==1.15 elixir-hex erlang==26 erlang-rebar3
```

packaging/suse/container/Dockerfile
```diff
@@ -11,10 +11,10 @@ WORKDIR /build/web/assets
 RUN npm run tailwind:build
 RUN npm run build
 
-FROM bci/bci-base:15.4 AS release
+FROM bci/bci-base:15.6 AS release
```
stefanotorresi, line 14: since we're at it, we should start using the full registry path
```suggestion
FROM registry.suse.com/bci/bci-base:15.6 AS release
```

packaging/suse/container/Dockerfile
```diff
@@ -28,7 +28,7 @@ ENV VERSION=%%VERSION%%
 RUN mix phx.digest
 RUN mix release
 
-FROM bci/bci-base:15.4 AS trento
+FROM bci/bci-base:15.6 AS trento
```
stefanotorresi, line 31: same as above (also no need for step alias since it's the last step)
```suggestion
FROM registry.suse.com/bci/bci-base:15.6
```

stefanotorresi (changes requested):
This is a good opportunity to sync up with some changes I made in the IBS stable snapshot just before the release, see [here](https://build.suse.de/projects/Devel:SAP:trento/packages/trento-web-image/files/Dockerfile?expand=1)

I never backported them because, as I mentioned previously, I intended to move all these packaging files in a separate repo, but since we're touching them, let's take the chance to do it.

arbulu89:
@stefanotorresi Thank you for the review and tips. I have applied them (i have created the commit myself, as some pieces were missing, but thank you for the straightforward change suggestions).

stefanotorresi (approved):

PR #3222, opened 2025-01-13
Activity log severity levels 

gagandeepb:
# Description

Adds support for activity log severity levels.


priv/repo/migrations/20250109130825_create_activity_log_severity_field_index.exs
```diff
@@ -0,0 +1,7 @@
+defmodule Trento.Repo.Migrations.CreateActivityLogSeverityFieldIndex do
```
nelsonkopliku, line 1: **suggestion**: we could squash the two migrations in just one. At the end they're strictly related to each other.

lib/trento/activity_logging/activity_log.ex
```diff
@@ -26,13 +26,262 @@ defmodule Trento.ActivityLog.ActivityLog do
...
+    |> validate_inclusion(:severity, @severity_min..@severity_max)
   end
+
+  def severity_level_to_integer(:debug), do: 5
+  def severity_level_to_integer(:info), do: 9
+  def severity_level_to_integer(:warning), do: 13
+  def severity_level_to_integer(:error), do: 17
+  def severity_level_to_integer(:critical), do: 21
+
+  defp map_severity_level(activity_type, metadata) do
```
nelsonkopliku, line 50: **question**: an alternative to having the severity detection in the schema could be introducing a module that does the required detection in the data transformation pipeline in `Trento.ActivityLog.Parser.ActivityParser.to_activity_log/2`, right after the metadata have been enriched.

How would that fit?

nelsonkopliku (commented):
Looking good, Gagan, leaving a couple of preliminary comments.

Additionally to what mentioned consider addressing the main aspects in different PRs:
- detecting and writing the severity in the log entry
- exposing the severity via the API
- enhancing the search to support severity
- the UI

Saying because if we attempt doing everything at once we might end up with a pretty big bulk of changes (considering of course the tests as well)

lib/trento/activity_logging/activity_log.ex
gagandeepb: Let's talk about this sync. My understanding is that there's some module dependency constraints imposed due to the bidirectional mapping from severity level as text and severity level as integer. (One direction during read ops and the other during write ops.)

gagandeepb:
@nelsonkopliku Thanks. I agree regarding the UI changes, I will remove those for now. I will try to reduce the size of this PR to the extent possible. However, in my understanding most of the BE changes would need to be in the same PR since once the context function starts returning an additional field (severity), adjustments would need to be made down the request lifecycle (OpenAPI schemas, etc.) to make tests/CI pass.

lib/trento/activity_logging/parser/activity_parser.ex
```diff
@@ -54,4 +61,51 @@ defmodule Trento.ActivityLog.Parser.ActivityParser do
 
   defp parse_domain_event_activity_info(:metadata, activity, activity_context),
     do: EventParser.get_activity_metadata(activity, activity_context)
+
+  defp map_severity_level(activity_type, metadata) do
```
nelsonkopliku, line 65: As agreed offline we can move these functions in `Trento.ActivityLog.SeverityLevel` so that the severity detection module can be tested in isolation.

lib/trento/activity_logging/parser/activity_parser.ex
```diff
@@ -54,4 +61,51 @@ defmodule Trento.ActivityLog.Parser.ActivityParser do
...
+    case SeverityLevel.severity_level_mapping()[activity_type] do
+      nil ->
+        # unmapped activity type found
+        SeverityLevel.severity_level_to_integer(:warning)
```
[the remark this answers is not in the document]
nelsonkopliku: if we decide so, I am fine with it.

Maybe with a fallback we are 100% safe. It could even be something like `unknown` if warning/info are not suitable.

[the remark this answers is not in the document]
abravosuse: If we decide to fall back, I'd rather do it to `info`. `unknown` would create confusion, in my opinion. But all in all, my preference is that we make the severity assignment part of the DoD for new entry types.

nelsonkopliku (approved):
Looks good!

We could add a few assertions in `test/trento/activity_logging/activity_logger_test.exs` about actual severities but up to you.

PR #3225, opened 2025-01-14
change base image and update dependencies

stefanotorresi:
Changes the base image so that we always use the same BCI everywhere, and fixes runtime issues with some libraries missing.
Also `tar` is in the base image now.

Sister PR in wanda coming up.

Fixes https://jira.suse.com/browse/TRNT-3457


stefanotorresi:
this was already tested in Devel:SAP:trento (IBS) when we encountered the runtime issue during the release window, and the images are now practically identical (this is why I changed the base image repo from opensuse to bci).

stefanotorresi:
superseded by #3232

PR #3227, opened 2025-01-14
Pin Chromatic action

balanza:
The latest version (11.23.0 at the time of writing) isn't working.

Pinning to the last working version.


gagandeepb (approved):
Thanks for looking into this!

PR #3210, opened 2024-12-20
Add hana scale up scenario as selection in checks catalog

EMaksy:
# Description
This pr enriches the checks catalog view by changing the Hana Scale Up filter to Perf Opt and adding a new filter with cost optimized scenario.  

## Demo *before* PR
image

## Demo *after* PR
image

New features:
- Rename "HANA Scale Up" as "HANA Scale Up Perf. Opt." in Cluster Types filter in Checks Catalog view
when selected, checks applicable to "HANA Scale Up Perf. Opt." clusters will be displayed in the view
- Add "HANA Scale Up Cost Opt." in Cluster Types filter in Checks Catalog view
when selected, checks applicable to "HANA Scale Up Cost Opt." clusters will be displayed in the view
Fixes # (issue)

## How was this tested?
- Fixed existing tests

PR #3229, opened 2025-01-14
Fix selected cluster type ui in Checks catalog view

EMaksy:
# Description
This pr will restore the previous ui behaviour in the checks catalog view. When selecting a cluster type in the checks catalog the green svg icon is not rendered.
With the latest changes in #3210 we now not only pass a string as value, which was handled by the built in headless ui `selected` boolean, but also an object `{ type, hanaScenario }` . Check out the official docs https://headlessui.com/v1/react/listbox#listbox-option
With using the lodash isEqual we now  use a "deep comparison"  to ensure accurate evaluation of selected options.

For this we use the "by" value provided by Listbox: 
```You can also pass your own comparison function to the by prop if you'd like complete control over how objects are compared```
Source: https://headlessui.com/v1/react/listbox#binding-objects-as-values

## Demo: 

https://github.com/user-attachments/assets/4d2fc8dc-e50a-47cb-9ebd-22ee862b4ae1

## Test

Enriched existing e2e test to test that the SVG is rendered

### DIFF
--- a/assets/js/common/Select/Select.jsx
+++ b/assets/js/common/Select/Select.jsx
@@ -11,7 +11,8 @@ export const OPTION_ALL = 'all';
 const defaultOnChange = () => {};
 const defaultRenderOption = (item) => item.value;
 
-const deepCompareSelection=(optionValue, value) => isEqual(optionValue, value);
+const deepCompareSelection = (optionValue, value) =>
+  isEqual(optionValue, value);
 
 function Select({
   optionsName,
--- a/assets/js/common/Select/Select.jsx
+++ b/assets/js/common/Select/Select.jsx
@@ -4,12 +4,14 @@ import { Listbox, Transition } from '@headlessui/react';
 import { CheckIcon, ChevronUpDownIcon } from '@heroicons/react/20/solid';
 
 import classNames from 'classnames';
-import { find, get } from 'lodash';
+import { find, get, isEqual } from 'lodash';
 
 export const OPTION_ALL = 'all';
 
 const defaultOnChange = () => {};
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: assets/js/common/Select/Select.jsx [hunk 1]
```diff
@@ -86,23 +86,26 @@ function Select({
...
-                        'font-normal': !isSelected,
-                      })}
-                    >
-                      {renderOption(option)}
-                    </span>
-                    {isSelected ? (
-                      <span className="absolute inset-y-0 right-2 end-1 flex items-center pl-3 text-green-600">
-                        <CheckIcon className="w-5 h-5" aria-hidden="true" />
+                {() => {
+                  const isSelected = isEqual(option.value, value);
```
EMaksy, line 90: @arbulu89 open for alternatives :)

arbulu89 (approved):
Sweet!

nelsonkopliku (approved):
Thanks @EMaksy!

test/e2e/cypress/e2e/checks_catalog.cy.js
```diff
@@ -93,12 +93,15 @@ context('Checks catalog', () => {
   });
 
   describe('Filtering', () => {
+    const selectedIconSelector =
```
arbulu89, line 96: question @EMaksy :
why did you create this constant and put in all the array entries?
I mean, you could simply have the constant and use it in the test, right?

EMaksy: Very fair point, when working on this first i thought the ui selector would be different, but nope. 
You are right we can just test directly, will create a follow up pr

PR #3232, opened 2025-01-15
change base image and update dependencies

stefanotorresi:
supersedes #3225


stefanotorresi:
FYI pr env wont trigger until after #3233 is merged and this is rebased.

stefanotorresi:
factory is now fine and dandy (except for Tumbleweed, which we shouldn't care about atm)

PR #3233, opened 2025-01-15
move some workflow env variables to GHA variables

stefanotorresi:
so that we can edit them from the GitHub settings


arbulu89 (approved):
Thank you!

PR #3240, opened 2025-01-20
Expose abilities as jwt claims

nelsonkopliku:
# Description

In order to allow service providers like wanda to do proper role based authorization, we expose abilities as part of the JWT claims.

### DIFF
--- a/lib/trento_web/plugs/app_jwt_auth_plug.ex
+++ b/lib/trento_web/plugs/app_jwt_auth_plug.ex
@@ -24,10 +24,7 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
   """
   def fetch(conn, _config) do
     with {:ok, jwt_token} <- read_token(conn),
-         {:ok, claims} <- validate_access_token(jwt_token) do
-      sub = claims["sub"]
-      abilities = claims["abilities"]
-
+         {:ok, %{"sub" => sub, "abilities" => abilities}} <- validate_access_token(jwt_token) do
       conn =
         conn
         |> Conn.put_private(:api_access_token, jwt_token)
--- a/lib/trento_web/plugs/app_jwt_auth_plug.ex
+++ b/lib/trento_web/plugs/app_jwt_auth_plug.ex
@@ -25,12 +25,21 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
   def fetch(conn, _config) do
     with {:ok, jwt_token} <- read_token(conn),
          {:ok, claims} <- validate_access_token(jwt_token) do
+      sub = claims["sub"]
+      abilities = claims["abilities"]
+
       conn =
         conn
         |> Conn.put_private(:api_access_token, jwt_token)
-        |> Conn.put_private(:user_id, claims["sub"])
+        |> Conn.put_private(:user_id, sub)
+        |> Conn.put_private(:abilities, abilities)
 
-      {conn, %{"access_token" => jwt_token, "user_id" => claims["sub"]}}
+      {conn,
+       %{
+         "access_token" => jwt_token,
+         "user_id" => sub,
+         "abilities" => abilities
+       }}
     else
       _ -> {conn, nil}
     end
@@ -42,7 +51,13 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
     The generated credentials will be stored in private section of the Plug.Conn struct
   """
   def create(conn, user, _config) do
-    claims = %{"sub" => user.id}
+    {:ok, user} = Users.get_user(user.id)
+
+    claims = %{
+      "sub" => user.id,
+      "abilities" => Enum.map(user.abilities, &%{name: &1.name, resource: &1.resource})
+    }
+
     access_token = AccessToken.generate_access_token!(claims)
     refresh_token = RefreshToken.generate_refresh_token!(claims)
 
--- a/lib/trento_web/plugs/app_jwt_auth_plug.ex
+++ b/lib/trento_web/plugs/app_jwt_auth_plug.ex
@@ -29,7 +29,6 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
         conn
         |> Conn.put_private(:api_access_token, jwt_token)
         |> Conn.put_private(:user_id, sub)
-        |> Conn.put_private(:abilities, abilities)
 
       {conn,
        %{
--- a/lib/trento_web/plugs/app_jwt_auth_plug.ex
+++ b/lib/trento_web/plugs/app_jwt_auth_plug.ex
@@ -49,13 +49,19 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
   def create(conn, user, _config) do
     {:ok, user} = Users.get_user(user.id)
 
-    claims = %{
-      "sub" => user.id,
-      "abilities" => Enum.map(user.abilities, &%{name: &1.name, resource: &1.resource})
+    default_claims = %{
+      "sub" => user.id
     }
 
-    access_token = AccessToken.generate_access_token!(claims)
-    refresh_token = RefreshToken.generate_refresh_token!(claims)
+    access_token_claims =
+      Map.put(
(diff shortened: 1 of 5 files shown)


lib/trento_web/plugs/app_jwt_auth_plug.ex
```diff
@@ -25,12 +25,21 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
   def fetch(conn, _config) do
     with {:ok, jwt_token} <- read_token(conn),
          {:ok, claims} <- validate_access_token(jwt_token) do
+      sub = claims["sub"]
```
CDimonaco, line 28: Could we destructure the claims on line 27, also the sub one, it's safer to do since we need to fail asap if they don't match with the with clause

test/support/factory.ex
```diff
@@ -1199,6 +1199,17 @@ defmodule Trento.Factory do
     }
   end
 
+  def user_with_abilities_factory do
+    user = build(:user)
+    ability = build(:ability)
+    build(:users_abilities, user_id: user.id, ability_id: ability.id)
```
EMaksy, line 1205: Why is this line required ? It is used to mock the database, right ?
Or any additional effects?

EMaksy (approved):
LGTM, just a left a question as a comment :)

test/support/factory.ex
nelsonkopliku: if you intend the specific line
```elixir
build(:users_abilities, user_id: user.id, ability_id: ability.id)
```
that is because of the relation between users and abilities.

lib/trento_web/plugs/app_jwt_auth_plug.ex
CDimonaco, line 1: On `attach_refresh_token_to_conn` on line 120, we generate a new access token during the renewal process, make sure you add the abilities also here

Before that pr we did not have any particular logic for creating the access token so a one liner with the user id in two places was ok, if you want create a private function to create the claims in the new format and use them in both create and renew functions to generate a proper access token

lib/trento_web/plugs/app_jwt_auth_plug.ex
```diff
@@ -42,7 +51,13 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
     The generated credentials will be stored in private section of the Plug.Conn struct
   """
   def create(conn, user, _config) do
-    claims = %{"sub" => user.id}
+    {:ok, user} = Users.get_user(user.id)
+
+    claims = %{
```
CDimonaco, line 56: Please use this claims only for access token, for refresh token on line 62 just use the sub as claims `%{"sub" => user.id}`, as you can see on line 83 are the only one we need when dealing with a refresh token, no further user infos are needed for this type of tokens

lib/trento_web/plugs/app_jwt_auth_plug.ex
```diff
@@ -25,12 +25,21 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
...
+      abilities = claims["abilities"]
+
       conn =
         conn
         |> Conn.put_private(:api_access_token, jwt_token)
-        |> Conn.put_private(:user_id, claims["sub"])
-
-      {conn, %{"access_token" => jwt_token, "user_id" => claims["sub"]}}
+        |> Conn.put_private(:user_id, sub)
+        |> Conn.put_private(:abilities, abilities)
```
CDimonaco, line 35: You can remove that from private, we add user_id for pow functioning, the abilities checking works without using the abilities from the private part of the conn, we can remove that field injecting

CDimonaco (commented):
Thanks man good job,

I left some comments to be addressed to have everything in the right spot for generating acess tokens also when renewing and with only the fields we really need

arbulu89 (approved):
Looks good to me!
I just did a question about loading the user in the `create`.
I guess there is not any other alternative, but I preferred to ask. Maybe @CDimonaco as well can answer

test/trento_web/plugs/app_jwt_auth_plug_test.exs
```diff
@@ -132,9 +133,9 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlugTest do
 
   describe "create/3" do
```
CDimonaco, line 134: When you add the abilities also to the renew token add some test cases to assert that

lib/trento_web/plugs/app_jwt_auth_plug.ex
```diff
@@ -42,7 +51,13 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlug do
     The generated credentials will be stored in private section of the Plug.Conn struct
   """
   def create(conn, user, _config) do
-    claims = %{"sub" => user.id}
+    {:ok, user} = Users.get_user(user.id)
```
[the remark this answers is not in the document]
CDimonaco: This is one of the possible approaches, the user in here comes from the Pow Context functions, we defined some custom context functions in `lib/trento/users.ex` you can take a look at `get_by` and `by_id` directly used by Pow because that module is defined as custom Pow context.

So we can think to modify that functions to also preload abilities but I think loading them here is a good approach, 99.9 of the times when Pow does his stuff, and believe me it calls that function a good amount of times, it does not need abilities so loading them here only when needed with another query seems pretty ok

Ofc we can change in the future

lib/trento_web/plugs/app_jwt_auth_plug.ex
nelsonkopliku: Done https://github.com/trento-project/web/pull/3240/commits/b6a9b12e52381933e62e6a672b9156cb78afe79b

lib/trento_web/plugs/app_jwt_auth_plug.ex
nelsonkopliku: Here we go https://github.com/trento-project/web/pull/3240/commits/a49e04ebca96631dce59f41c894bff017630fe85

lib/trento_web/plugs/app_jwt_auth_plug.ex
nelsonkopliku: Thanks for pointing out! Here we go https://github.com/trento-project/web/pull/3240/commits/b0c77b56b509ed4c5599c7ec6888ce5e3a89cbf2

lib/trento_web/plugs/app_jwt_auth_plug.ex
nelsonkopliku: Addressed here 48c29c6944a8eb0ae438cfa7aac5ffa0a5e4c849

test/trento_web/plugs/app_jwt_auth_plug_test.exs
nelsonkopliku: Right, at the end I just added an assertion to the same test 259cd3a16e91597602c7abe0961fff0b1ce47d61

CDimonaco (approved):
LGTM 🚀

test/trento_web/plugs/app_jwt_auth_plug_test.exs
```diff
@@ -132,9 +133,9 @@ defmodule TrentoWeb.Plugs.AppJWTAuthPlugTest do
...
     test "should add to the conn the access/refresh token pair and the expiration", %{conn: conn} do
-      user = %{id: 1}
+      %{id: user_id} = user = insert(:user_with_abilities)
 
-      assert {res_conn, ^user} = AppJWTAuthPlug.create(conn, user, @pow_config)
+      assert {res_conn, %{id: ^user_id}} = AppJWTAuthPlug.create(conn, user, @pow_config)
```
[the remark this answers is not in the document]
nelsonkopliku: hey @arbulu89 hope the last two commits help 259cd3a16e91597602c7abe0961fff0b1ce47d61 342fc53c74def493ce7833f2fa1852dfc7b8b828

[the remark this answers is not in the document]
arbulu89: Yes, thank you so much!

PR #3242, opened 2025-01-21
Adds Encapsulation for Dashboard Page

jagabomb:
# Description

Changes HTML structure of the **Dashboard** (_HomeHealthSummary_) to match the subpages encapsulation effect which has been established in previous commits. Encapsulation effect places most of the content within a single container with a white background and rounded corners. This is the usual effect for pages with a single section of content.

The **Table** component also required a change since it was displaying rounded corners. An additional parameter was added to the **Table** component to help control the displaying of the rounded corners for the **Table** header especially when no filters were present.

**Before**
Screenshot 2025-01-21 at 17 07 55

**Changes (Updated 22 Jan 11:47 CET)**
Screenshot 2025-01-22 at 12 37 33

- [x] **DONE**

## How was this tested?

- [x] **DONE** Visually

### DIFF
--- a/assets/js/common/HealthIcon/HealthIcon.jsx
+++ b/assets/js/common/HealthIcon/HealthIcon.jsx
@@ -23,22 +23,28 @@ function HealthIcon({
   size = 'l',
   isLink = false,
 }) {
+  const passingIcon = () =>
+    isLink ? EOS_CHECK_CIRCLE_FILLED : EOS_CHECK_CIRCLE_OUTLINED;
+  const PassingIcon = passingIcon();
+
+  const warningIcon = () =>
+    isLink ? EOS_WARNING_FILLED : EOS_WARNING_OUTLINED;
+  const WarningIcon = warningIcon();
+
+  const criticalIcon = () => (isLink ? EOS_ERROR_FILLED : EOS_ERROR_OUTLINED);
+  const CriticalIcon = criticalIcon();
+
+  const absentIcon = () => (isLink ? EOS_INFO_FILLED : EOS_INFO_OUTLINED);
+  const AbsentIcon = absentIcon();
+
   const hoverOpacityClass = {
     'hover:opacity-75': hoverOpacity,
     'hover:opacity-100': !hoverOpacity,
   };
   switch (health) {
     case 'passing':
-      return isLink ? (
-        <EOS_CHECK_CIRCLE_FILLED
-          size={size}
-          className={classNames(
-            hoverOpacityClass,
-            computedIconCssClass('fill-jungle-green-500', centered)
-          )}
-        />
-      ) : (
-        <EOS_CHECK_CIRCLE_OUTLINED
+      return (
+        <PassingIcon
           size={size}
           className={classNames(
             hoverOpacityClass,
@@ -47,16 +53,8 @@ function HealthIcon({
         />
       );
     case 'warning':
-      return isLink ? (
-        <EOS_WARNING_FILLED
-          size={size}
-          className={classNames(
-            hoverOpacityClass,
-            computedIconCssClass('fill-yellow-500', centered)
-          )}
-        />
-      ) : (
-        <EOS_WARNING_OUTLINED
+      return (
+        <WarningIcon
           size={size}
           className={classNames(
             hoverOpacityClass,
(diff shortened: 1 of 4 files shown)


assets/js/common/Table/Table.jsx
```diff
@@ -189,7 +190,7 @@ function Table({
             className={classNames(
               'min-w-fit shadow overflow-hidden rounded-b-lg',
               {
-                'rounded-t-lg': !hasFilters,
+                'rounded-t-lg': !hasFilters && roundedTop,
```
EMaksy, line 193: With this every Table will have now the roundedTop :) 
I guess you don't mind right?
I checked it looks good :+1:

nelsonkopliku (approved):

EMaksy (approved):
Hey @jagabomb thanks you make Trento look always better :+1: 
Just left a minor comment, but code wise fine for me. 

p.s adding env as label will create a pr env where you can directly check changes. Very usefull for code reviews, feel free to use it for your next pr

assets/js/common/Table/Table.jsx
jagabomb: Hey Eugen, the additional boolean here `roundedTop` helps yo override a scenario where a table has no filters and was previously displaying rounded corners. This was the case after rearranging the HTML elements for the **Dashboard**. Also there should be no impact for current tables but rather a new way for us to override the behaviour if we want the table to display with top rounded corners since this is the default behaviour. 😊 

Here is an example without the boolean:
Screenshot 2025-01-22 at 12 42 16

assets/js/common/HealthIcon/HealthIcon.jsx
```diff
@@ -17,14 +21,23 @@ function HealthIcon({
...
+  isLink = false,
 }) {
   const hoverOpacityClass = {
     'hover:opacity-75': hoverOpacity,
     'hover:opacity-100': !hoverOpacity,
   };
   switch (health) {
     case 'passing':
-      return (
+      return isLink ? (
```
nelsonkopliku, line 32: Hey @jagabomb I noticed not the latest changes.

In order to mitigate code repetition, an alternative approach here could be extrapolating the icon detection logic, which is what changes based on the `isLink` flag.

We could have the following functions allowing us to tell what's the actual icon to be used based on whether in a link context or not.

```js
const passingIcon = (isLink = false) =>
  isLink ? EOS_CHECK_CIRCLE_FILLED : EOS_CHECK_CIRCLE_OUTLINED;

const warningIcon = (isLink = false) =>
  isLink ? EOS_WARNING_FILLED : EOS_WARNING_OUTLINED;

const criticalIcon = (isLink = false) =>
  isLink ? EOS_ERROR_FILLED : EOS_ERROR_OUTLINED;

const absentIcon = (isLink = false) =>
  isLink ? EOS_INFO_FILLED : EOS_INFO_OUTLINED;
```

Then in the `switch` statement we can use those like this
```js
  switch (health) {
    case 'passing': {
      const PassingIcon = passingIcon(isLink);
      return (
        <PassingIcon
          size={size}
          className={classNames(
            hoverOpacityClass,
            computedIconCssClass('fill-jungle-green-500', centered)
          )}
        />
      );
    }
    case 'warning': {
      const WarningIcon = warningIcon(isLink);
      return (
        <WarningIcon
          size={size}
          className={classNames(
            hoverOpacityClass,
            computedIconCssClass('fill-yellow-500', centered)
          )}
        />
      );
    }
```
At the end only the icon reference changes, the rest seems to be the same.

jagabomb: Ah amazing! Thanks for this Nelson. I was wondering how to distinguish between the different icons in a smarter way without repeating so much code. I will make these changes. 😄

assets/js/common/Table/Table.jsx
EMaksy: Yeah, all good, i think i was not clear enough what i meant.  The default behavior won't change. 
You are right we now have the option to control it with `roundedTop = true` 
All fine here :)

nelsonkopliku (approved):
LGTM

abravosuse (approved):
Both changes look good to me, Jurgen. The encapsulation in the dashboard and the filled icons.

PR #3244, opened 2025-01-22
Fix for nil values in severity

gagandeepb:
Sometimes in the metadata to severity mapping, the health key in metadata is not found, since the metadata is an empty map. Here we handle this nil appropriately, by mapping it to the info level.
Added a unit test to cover the observed case.

### DIFF
--- a/lib/trento/activity_logging/severity_level.ex
+++ b/lib/trento/activity_logging/severity_level.ex
@@ -158,15 +158,21 @@ defmodule Trento.ActivityLog.SeverityLevel do
       :map_value_to_severity ->
         health_key_suffix = mapping.key_suffix
 
-        health_key =
+        maybe_health_key =
           keys
           |> Enum.map(&Atom.to_string/1)
           |> Enum.find(fn e -> String.ends_with?(e, health_key_suffix) end)
 
-        health_value = metadata[String.to_existing_atom(health_key)]
-        health_value_downcased = health_value |> Atom.to_string() |> String.downcase()
-        severity_default = Map.get(mapping.values, "*")
-        Map.get(mapping.values, health_value_downcased, severity_default)
+        case maybe_health_key do
+          nil ->
+            :info
+
+          health_key when is_binary(health_key) ->
+            health_value = metadata[String.to_existing_atom(health_key)]
+            health_value_downcased = health_value |> Atom.to_string() |> String.downcase()
+            severity_default = Map.get(mapping.values, "*")
+            Map.get(mapping.values, health_value_downcased, severity_default)
+        end
 
       :key_exists ->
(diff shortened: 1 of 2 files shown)


test/trento/activity_logging/severity_level_test.exs
```diff
@@ -96,5 +106,15 @@ defmodule Trento.ActivityLog.SeverityLevelTest do
 
       assert severity == :info
     end
+
+    test "should may conditional key suffix types with no metadata correctly" do
```
arbulu89, line 110: typo, `may` -> `map`

lib/trento/activity_logging/severity_level.ex
```diff
@@ -158,15 +158,21 @@ defmodule Trento.ActivityLog.SeverityLevel do
...
           |> Enum.map(&Atom.to_string/1)
           |> Enum.find(fn e -> String.ends_with?(e, health_key_suffix) end)
 
-        health_value = metadata[String.to_existing_atom(health_key)]
-        health_value_downcased = health_value |> Atom.to_string() |> String.downcase()
-        severity_default = Map.get(mapping.values, "*")
-        Map.get(mapping.values, health_value_downcased, severity_default)
+        case maybe_health_key do
+          nil ->
+            :info
```
arbulu89, line 168: Out of context, but I would put this severity levels in a `Enum`: https://github.com/trento-project/web/tree/main/lib/trento/enums
(The enum can be only for the Activitly logging concept as well)

nelsonkopliku (approved):
LGTM

lib/trento/activity_logging/severity_level.ex
```diff
@@ -158,15 +158,21 @@ defmodule Trento.ActivityLog.SeverityLevel do
...
+
+          health_key when is_binary(health_key) ->
+            health_value = metadata[String.to_existing_atom(health_key)]
```
arbulu89, line 171: question: the `health_key` atom is always there?
Otherwise this will fail as well

EMaksy (approved):
LGTM

lib/trento/activity_logging/severity_level.ex
gagandeepb: Here we need the ability to map to integer when storing in db and string when presenting as API response. But I will evaluate/look into this, if it is possible to use the Enums you mention as intermediate form/representation.

lib/trento/activity_logging/severity_level.ex
gagandeepb: I have observed that the keys in metadata field are usually atoms, application side. Would that be too strong of an assumption? I can look some more into this.

PR #3256, opened 2025-01-27
Add check customization call to action in checks selection

nelsonkopliku:
# Description

This PR adds an icon to start check customization in checks selection.

Notes:
- action is available only when a check is marked as customizable
- a `Coming Soon!` alert is shown currently

Unfortunately the PR env is not deploying checks, so can't really use it.

Here's a screenshot
image

### DIFF
--- a/assets/js/lib/test-utils/factories/index.js
+++ b/assets/js/lib/test-utils/factories/index.js
@@ -95,6 +95,18 @@ export const catalogExpectEnumExpectationFactory = Factory.define(
   })
 );
 
+export const catalogConditionFactory = Factory.define(() => ({
+  value: faker.lorem.word(),
+  expression: faker.lorem.sentence(),
+}));
+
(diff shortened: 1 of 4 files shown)


EMaksy (approved):
Hey codewise looks great !
Thank you.

EMaksy:
Hmm i just saw that we don't have a story for host checks selection  and cluster settings.

Did we have any reasons not to add ? 
Otherwise, we may consider adding it at a later point, when all ui changes were added/made.
What do you think ?

nelsonkopliku:
@EMaksy we have stories for the [ChecksSelection](https://github.com/trento-project/web/blob/main/assets/js/pages/ChecksSelection/ChecksSelection.stories.jsx) which should be enough
See also [here](https://643fc66e9747e47344899b8c-mwmnwgbdgr.chromatic.com/?path=/docs/patterns-checksselection--docs)

PR #3273, opened 2025-02-03
Fix cluster json view adapt_v1 function

arbulu89:
# Description

I have realized that the v1 adaption of the cluster view doesn't work as expected.
As the transformed data is loaded from the database, and the `details` field is a jsonb, it is loaded as plain maps.

The `to_atomized_map` function solves the issue and works when the `details` are loaded from the database and not. We need this as the transformation happens in 2 places:
- in the controller after loading the data from the db
- in the projection, where the details are coming from the ecto changeset as atom keys already

Other solution could be to change the `Trento.Support.Ecto.Payload` file to do this in the load, and set this type as the cluster `details` field in the read model. Like here: https://github.com/trento-project/wanda/pull/564
Just changing the `dump` function:

```
def dump(data) when is_list(data) or is_map(data), do: {:ok, to_atomized_map(data)}
```
I didn't do this, because we already use this `Payload` in other places and I'm not sure if we want to atomize all the keys.
We can postpone this decision as well and have some tech-debt task

## How was this tested?

UT

### DIFF
--- a/lib/trento/support/struct_helper.ex
+++ b/lib/trento/support/struct_helper.ex
@@ -33,4 +33,41 @@ defmodule Trento.Support.StructHelper do
   def to_map(value) when is_nil(value), do: value
   def to_map(value) when is_atom(value), do: Atom.to_string(value)
   def to_map(value), do: value
+
+  @doc """
+  Converts the string keys of a map to existing atoms.
+  If the key does not exist as atom it continues being a string
+  """
+  @spec to_atomized_map(map | [map] | struct | [struct]) :: map | [map]
+  def to_atomized_map(%NaiveDateTime{} = value), do: value
+  def to_atomized_map(%DateTime{} = value), do: value
+  def to_atomized_map(%Date{} = value), do: value
+
+  def to_atomized_map(struct) when is_struct(struct) do
+    struct
+    |> Map.from_struct()
+    |> to_atomized_map()
+  end
+
+  def to_atomized_map(map) when is_map(map) do
+    map
+    |> Enum.reject(fn
+      {:__meta__, _} -> true
+      {_, %Ecto.Association.NotLoaded{}} -> true
+      _ -> false
+    end)
+    |> Map.new(fn {k, v} -> {atomize_key(k), to_atomized_map(v)} end)
+  end
+
+  def to_atomized_map(list) when is_list(list) do
+    Enum.map(list, &to_atomized_map/1)
+  end
+
+  def to_atomized_map(value), do: value
+
+  defp atomize_key(key) do
+    String.to_existing_atom(key)
+  rescue
+    ArgumentError -> key
+  end
 end
--- a/lib/trento_web/controllers/v1/cluster_json.ex
+++ b/lib/trento_web/controllers/v1/cluster_json.ex
@@ -1,11 +1,13 @@
 defmodule TrentoWeb.V1.ClusterJSON do
+  alias Trento.Support.StructHelper
+
   def clusters(%{clusters: clusters}) do
     Enum.map(clusters, &cluster(%{cluster: &1}))
   end
 
   def cluster(%{cluster: cluster}) do
     cluster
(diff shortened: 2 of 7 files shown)


lib/trento/support/struct_helper.ex
```diff
@@ -33,4 +33,41 @@ defmodule Trento.Support.StructHelper do
...
+
+  def to_atomized_map(list) when is_list(list) do
+    Enum.map(list, &to_atomized_map/1)
+  end
+
+  def to_atomized_map(value), do: value
+
+  defp atomize_key(key) do
+    String.to_existing_atom(key)
+  rescue
```
arbulu89, line 70: This is the unique way I found of knowing if an atom exists or not.
Unfortunately I didn't find a way to remove the `try/rescue` option

lib/trento/support/struct_helper.ex
```diff
@@ -33,4 +33,41 @@ defmodule Trento.Support.StructHelper do
...
+    ArgumentError -> key
+  end
```
gagandeepb, line 72: It would be interesting to consider using a [Protocol](https://hexdocs.pm/elixir/1.15.8/Protocol.html) here. One benefit is that it would allow dispatching to the appropriate defimpl (protocol implementation) for different structs in a compile-time aware way. IIRC, one should get a compile error/warning upon calling a protocol function on a struct type that is not implemented/has no fallback.

gagandeepb (commented):
Thanks for this. Just one minor comment.

lib/trento/support/struct_helper.ex
arbulu89: Just to understand your point. How do you suggest to implement a Protocol here?
For what struct? `HanaClusterDetails`, `AscsErsClusterDetails`, etc ?

nelsonkopliku (approved):
Thanks Xabi, looks good!

I would love consistency about maps usage, and my personal preference would be with atomized keys.

However I am aware that it would be challenging streamlining in that direction because of dynamic maps where we have to consider `String.to_atom` vs `String.to_existing_atom` vs mixed atom/string keys.

I realize that even though I might not like it, string keys maps streamlining would be simpler to achieve.

Of course we can defer such a decision.

arbulu89:
> Of course we can defer such a decision.

About deferring the decision, I meant about putting this code in the `load` of an `Ecto.Type` or not.
Not about atomizing all the keys. For me `String.to_atom` is a no go unless it was strictly needed, which is not.
I think that having a best effort solution that will convert all the predefined keys to atoms is good (so all the ValueObjects we have share the same keys).
Unknown string keys were string keys since the first place, in the discovery payload, in the domain code, in the projection, etc

nelsonkopliku:
I see @arbulu89, thanks for clarifying.

In this case I would actually consider doing atomization in the type's load function and be fine with the fact that a key would fall back to a string when unable to find an existing atom.

lib/trento/support/struct_helper.ex
gagandeepb: This was a minor comment, that can be considered later too. For structs like ones you mention, we can add a default implementation that can be tapped into/customized, as needed/on-demand ([e.g. via `@derive`](https://hexdocs.pm/elixir/1.15.8/Protocol.html#derive/3)).

```elixir
defprotocol AtomizedMap do
  def to(arg)
  def from(arg)
end

defmodule User do
  @derive AtomizedMap
  defstruct name: nil, age: nil
end

defimpl AtomizedMap, for: Any do
  defmacro __deriving__(module, struct, options) do
    quote do
      defimpl AtomizedMap, for: unquote(module) do
        def to(arg) do
         ...
        end
        def from(arg) do
        ...
        end
      end
    end
  end
end
```

arbulu89: I see. Alright, I guess if we see more needs to keep improving this we can think of a more sophisticated solution like this one

PR #3274, opened 2025-02-04
Add modified pill for customized checks

nelsonkopliku:
# Description

This PR adds the ability to show a `MODIFIED` pill on a check that has been customized.
Wire up with API call will follow up.


EMaksy (approved):
Ah suuuper. 
Thanks for refactoring the factory!

PR #3275, opened 2025-02-05
Use embeds software updates

arbulu89:
# Description

Refactor SUMA software updates discovered struct to use known fields.
All of this started with the intention to remove the` Jason.decode!(keys: :atoms)` in https://github.com/trento-project/web/blob/main/lib/trento/software_updates/discovery.ex#L309
This piece of code is potentially dangerous as it creates not existing atoms with data coming from external sources (it uses `String.to_atom` behind the scenes).
It is true that this information is coming from SUMA and we know the sent values so it should be safe.
But knowing the values, having a embedded schema makes sense as it gives some more control and we don't need to relay in jsonb plain maps.

I have tested a transition from old stored code to the new and works fine. (`embeds_many` takes care of using already existing empty list and nil values, and always return a list).

As a bonus, we fix the openapi schema error, as the `to_package_id` field returns an integer and we define it as string. Now this is fixed.
We most probably had this issue as SUMA docs themselves define this as string, even though they send an integer and they confirmed that this is a docs bug:
https://documentation.suse.com/suma/4.3/api/suse-manager/api/system.html#apidoc-system-listLatestUpgradablePackages-loggedInUser-sid

## How was this tested?

UT and manual testing to test the regression


lib/trento/infrastructure/software_updates/adapter/mock_suma.ex
```diff
@@ -59,7 +59,7 @@ defmodule Trento.Infrastructure.SoftwareUpdates.MockSuma do
            to_version: "1.16.2",
            to_release: "1",
            to_epoch: "0",
-           to_package_id: 92_348_112_636
+           to_package_id: "92348112636"
```
arbulu89, line 62: This change will require to reset SUMA values in demo, as using an integer was not expected

nelsonkopliku (approved):
Much appreciated, thanks!

lib/trento/infrastructure/software_updates/adapter/mock_suma.ex
arbulu89: @nelsonkopliku @janvhs Do you know why we moved this value to integer in the past?
Here: https://github.com/trento-project/web/pull/2916
The `to_package_id` value is defined as string in the SUMA docs: https://documentation.suse.com/suma/4.3/api/suse-manager/api/system.html#apidoc-system-listLatestUpgradablePackages-loggedInUser-sid

With that change, if the value is really a string I don't really know how this worked in the first place...
And if it is an number, our backend code was wrong.

nelsonkopliku: not sure why it became an integer. In other places is considered a string, as it should be.
What happens if you just put back to strings in the mock definition?

arbulu89: Well, it works with my mocked data. I was wondering if the SUMA docs have some typo.
I will start a SUMA instance and try myself.
The code with different types is suspicious

nelsonkopliku: True, it looks like suma is actually returning an integer, at least on our testing instance running on 4.3

arbulu89: Yes. I'm going to change the field type to integer. Now everything works with a real SUMA instance.
I asked the SUMA team as well. 
I hope it is only a documentation error, otherwise we might need to do changes if they fix it

arbulu89:
The API bc check is failing due the change, but this is actually normal, as we fix an error with this.
It was a bug, so we don't need to do anything else.
I will force merge once everything else is green

nelsonkopliku (approved):
Alright, as long as things keep working lgtm 😄

PR #3276, opened 2025-02-05
Add parent cluster resource

arbulu89:
# Description

Add cluster resource parent field to cluster resources.
Some cluster resources are created under some grouping format. The groups can be master group, clone group or simply a normal group.
In order to keep the flat nature of the cluster resources, I have added a `parent` field putting their parent group data.
If they don't have any parent the field is set to nil.

We need this information in order to implement some new agent operations. Many operations will depend on knowing if the cluster is in maintenance or not. And in some cases, if specific resources of the cluster are in maintenance or not. Many times, the "parent" resource is set to maintenance, so we need to have this information.

For example to do SAP maintenance operations (like applying saptune): https://documentation.suse.com/sbp/sap-15/html/SLES4SAP-hana-sr-guide-PerfOpt-15/index.html#id-updating-sap-hana-seamless-sap-hana-maintenance

## How was this tested?

UT

### DIFF
--- a/lib/trento/clusters/value_objects/cluster_resource_parent.ex
+++ b/lib/trento/clusters/value_objects/cluster_resource_parent.ex
@@ -1,6 +1,12 @@
 defmodule Trento.Clusters.ValueObjects.ClusterResourceParent do
   @moduledoc """
-  Represents the parent of a cluste resource
+  Represents the parent of a cluster resource
+
+  `managed` represents the maintenance state of the resource.
+  `multi_state` represents the type of the group:
+    - `true` means a Master/Slave group
+    - `false` means a Clone group
+    - `nil` means a standard Group
   """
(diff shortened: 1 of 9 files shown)


lib/trento/clusters/value_objects/cluster_resource_parent.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Trento.Clusters.ValueObjects.ClusterResourceParent do
+  @moduledoc """
+  Represents the parent of a cluste resource
```
nelsonkopliku, line 3:
```suggestion
  Represents the parent of a cluster resource
```
**typo:** `cluste` -> `cluster`

nelsonkopliku (approved):
👀

PR #3277, opened 2025-02-05
Add Checks customization modal

EMaksy:
# Description

This PR will add a checks customization modal for host and cluster checks selection view.
In a future PR the placeholder "Value from Wanda" will be replaced with the custom check values from wanda

## What is msising?
- Modal must be hooked up with backend
- Missing e2e test, coming in a seperate pr

### Single check value unchecked

image

###  Single check value checked

image

### Multiple customization values

image

### Partial customizable values

image

### Many values with a very long name

image

## How was this tested ? 

Added tests for the Modal
Added Storybook stories


assets/js/common/CustomCheckModal/CustomCheckModal.jsx
```diff
@@ -0,0 +1,155 @@
...
+      }}
+    >
+      <p className="text-gray-500 text-sm font-normal tracking-wide pb-2">
+        {selectedCheckDescription}
+      </p>
+
+      <div className="flex items-center border border-yellow-400 bg-yellow-50 p-4 rounded-md text-yellow-600 mb-4">
+        <Input
+          type="checkbox"
+          checked={checked}
```
EMaksy, line 77: About the style of the checkbox. 
Tailwind provides a[ `accent-color` class: ](https://tailwindcss.com/docs/accent-color). 
[Sadly the check mark stays black.](https://github.com/tailwindlabs/tailwindcss-forms/issues/27)
In order to make it white we could use [tailwind forms libary](https://stackoverflow.com/questions/72226003/how-to-change-background-of-unchecked-checkbox-in-tailwind-3-if-ussing-accent-co)

Or build it from scratch we could use something like this 
```
const checkBoxCustomCss =
    "appearance-none w-4 h-5 border-2 border-yellow-500 rounded-md checked:bg-green-500 checked:border-green-500 flex items-center justify-center relative checked:after:content-[':heavy_check_mark:'] checked:after:absolute checked:after:left-1/2 checked:after:top-1/2 checked:after:-translate-x-1/2 checked:after:-translate-y-1/2 checked:after:text-white checked:after:text-lg checked:after:font-bold checked:after:leading-none";
```

 For me personally i would stick with the default style, or color it to just green

assets/js/common/CustomCheckModal/CustomCheckModal.jsx
```diff
@@ -0,0 +1,155 @@
...
+        />
+        <span className="font-semibold">{checkBoxWarningText}</span>
+      </div>
+      {selectedCheckValues?.map((value) => (
+        <div
+          key={`${value?.name}_${value?.default}`}
+          className="flex items-center space-x-2 mb-8"
+        >
+          <div className="flex-col w-1/3 min-w-[200px]">
+            {renderLabelWithTooltip(value?.name)}
```
EMaksy, line 92: This function allows us to add ellipsis(…) and a tooltip when the checks value becomes to long. 
With this the labels and input fields stay aligned

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
```diff
@@ -0,0 +1,159 @@
...
+              {renderLabelWithTooltip(value?.name)}
+              <Label>(Default: {value?.current_value})</Label>
+            </div>
+
+            <Input
+              className="w-full"
+              onChange={(inputEvent) =>
+                handleCustomValueInput(value?.name, inputEvent.target.value)
+              }
+              placeholder={value?.current_value}
```
nelsonkopliku, line 111: **thought:** we might not need a placeholder at all. There will always be either the current value or the custom value.

We can safely remove this imho.

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
```diff
@@ -0,0 +1,159 @@
...
+    labelContent
+  );
+};
+
+function CheckCustomizationModal({
+  open = false,
+  id,
+  values,
+  description,
+  provider = 'Unknown',
```
nelsonkopliku, line 44: **suggestion:** let's use `UNKNOWN_PROVIDER` exported form `@lib/model`
```suggestion
  provider = UNKNOWN_PROVIDER,
```

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.stories.jsx
```diff
@@ -0,0 +1,150 @@
+import React from 'react';
+import { useArgs } from '@storybook/preview-api';
+import CheckCustomizationModal from '.';
+
+const singleValue = [
+  {
+    name: 'ValueName1',
+    current_value: 'value1',
+    conditions: [],
```
nelsonkopliku, line 9: **suggestion:** we don't have `conditions` in `SelectableChecks`. Take a look at the api response and at the factories in `assets/js/lib/test-utils/factories/checks.js`.

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.stories.jsx
```diff
@@ -0,0 +1,150 @@
...
+export default {
+  title: 'Components/CheckCustomizationModal',
+  component: CheckCustomizationModal,
+  argTypes: {
+    open: {
+      description: 'Opens the modal',
+      control: 'boolean',
+    },
+    id: {
+      description: 'Selected Check ID',
```
nelsonkopliku, line 66: **suggestion:** let's remove the `selected` contextualization from stories as well.

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.test.jsx
```diff
@@ -0,0 +1,166 @@
...
+import CheckCustomizationModal from './CheckCustomizationModal';
+
+const mockOnClose = jest.fn();
+const mockOnSave = jest.fn();
+
+const expectedCheckModalValues = {
+  open: true,
+  id: '123',
+  values: [
+    catalogValueFactory.build({
```
nelsonkopliku, line 15: **suggestion:** we can use the appropriate factories here like `selectableCheckFactory`, `customizedValueFactory`, `nonCustomizedValueFactory`

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.stories.jsx
```diff
@@ -0,0 +1,150 @@
...
+
+export function SingleValue(args) {
+  const [{ open }, updateArgs] = useArgs();
+  const handleClose = () => updateArgs({ open: !open });
+
+  return (
+    <>
+      <button type="button" onClick={() => handleClose()}>
+        Toggle CheckCustomizationModal
+      </button>
```
nelsonkopliku, line 109:
```suggestion
```
When dealing with modals we usually rely on storybook's controls to set the `open` property
image

For consistency I would remove this trigger.

nelsonkopliku (commented):
Very good @EMaksy, just a couple of comments about factories, and that would be enough to merge this PR.

I believe that this is a good starting point and besides what you mentioned let's keep in mind we'll need to make sure to provide the correct ux based on th different types of values:
- string -> input text
- number -> input number
- boolean -> radio button

Example here.
image

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
EMaksy: I played a bit around with it. I prefer to keep the placeholder, because when the user clicks on the field and starts to enter values, he starts from scratch otherwise he starts on the previous value.
We can revisit this behavior on the follow up  for the different input types. 
What do you think ?

nelsonkopliku: I guess what you want to achieve is setting an initial value for the input 
```js
initialValue={value?.custom_value || value?.current_value}
```
See https://github.com/trento-project/web/commit/a709946fc843de61c8f452c5120c84a6c2ca1353#diff-7d61854c35d8df403e17910b9865976e0ca7a475035de0ad1fa98fa4fb28ffb3R107

EMaksy: Yeah it is basically which behavior we prefer: 
`initialValue={value?.custom_value || value?.current_value}`

https://github.com/user-attachments/assets/389d4aed-7096-4542-9212-b00c13b7868b

vs 

`placeholder={value?.current_value}`

https://github.com/user-attachments/assets/36359928-8f43-46fd-b631-73c69ba49d62

You prefer `initialValue={value?.custom_value || value?.current_value}`? 
No strong opinion here, but i would need to adjust slightly the test :)

nelsonkopliku: The fact is that a placeholder is semantically different from the current value of an input.

Additionally, something like `placeholder={value?.current_value}` would always show `current_value` and never the `custom_value`, when present.

nelsonkopliku (approved):
Besides the placeholder LGTM!

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
EMaksy: Yeah you are right, adjusted like suggested :)

EMaksy:
Fixed!

PR #3284, opened 2025-02-10
Improve read models associations

arbulu89:
# Description

Fix read models associations to reflect properly their relationship. I need this to implement some operation policies where I need to involve all the read models, so I can preload everything easily instead of querying them individually.

As the code says:
- A cluster has by multiple hosts
- A host has multiple application and database instances

The other direction is implemented with the `belongs_to` utility.

I have removed the json entries as we don't expose this associations externally.

PD: The json change in the clusters views is not strictly needed as the `to_atomized_map` already removes the not loaded associations, but it makes the removal more explicit

## How was this tested?

UT


nelsonkopliku (approved):
LGTM.
Just one question: how many queried does ecto trigger? Does it use joins for eager loading or does it fall into the n+1 query situation?

arbulu89:
> LGTM. Just one question: how many queried does ecto trigger? Does it use joins for eager loading or does it fall into the n+1 query situation?

The association doesn't decide that. This simply defines the the foreign keys and references between the tables.
Later on, everything depends on how you use the preload (this is the most normal way to associate tables). 
Check this for more info: https://hexdocs.pm/ecto/Ecto.Query.html#preload/3

EMaksy (approved):

PR #3285, opened 2025-02-10
Activity log severity levels filter

gagandeepb:
Adds new query parameter for filtering by severity level to the activity log endpoint. 

## How was this tested?

UT

### DIFF
--- a/lib/trento/activity_log.ex
+++ b/lib/trento/activity_log.ex
@@ -49,6 +49,8 @@ defmodule Trento.ActivityLog do
     end
   end
 
+  defdelegate map_severity_integer_to_text(int), to: Trento.ActivityLog.SeverityLevel
+
   defp maybe_exclude_user_logs(ActivityLog = q, true = _include_all_log_types?), do: q
 
   defp maybe_exclude_user_logs(ActivityLog = q, false = _include_all_log_types?) do
--- a/lib/trento/activity_log.ex
+++ b/lib/trento/activity_log.ex
@@ -131,9 +131,8 @@ defmodule Trento.ActivityLog do
            op: :>=,
            value:
              v
-             |> Enum.map(&String.to_existing_atom/1)
-             |> Enum.map(&SeverityLevel.severity_level_to_integer/1)
-             |> Enum.min()
+             |> (&String.to_existing_atom/1).()
+             |> (&SeverityLevel.severity_level_to_integer/1).()
          }}
 
       param ->
--- a/lib/trento/activity_log.ex
+++ b/lib/trento/activity_log.ex
@@ -10,6 +10,7 @@ defmodule Trento.ActivityLog do
   alias Trento.ActivityLog.ActivityLog
   alias Trento.ActivityLog.MetadataQueryParser
   alias Trento.ActivityLog.RetentionTime
+  alias Trento.ActivityLog.SeverityLevel
   alias Trento.Repo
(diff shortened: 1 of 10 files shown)


lib/trento_web/controllers/v1/activity_log_json.ex
```diff
@@ -41,6 +42,12 @@ defmodule TrentoWeb.V1.ActivityLogJSON do
     }
   end
 
+  defp map_severity_integer_to_text(n) when n >= 5 and n <= 8, do: "debug"
+  defp map_severity_integer_to_text(n) when n >= 9 and n <= 12, do: "info"
+  defp map_severity_integer_to_text(n) when n >= 13 and n <= 16, do: "warning"
+  defp map_severity_integer_to_text(n) when n >= 17 and n <= 20, do: "error"
+  defp map_severity_integer_to_text(n) when n >= 21 and n <= 24, do: "critical"
+
```
nelsonkopliku, line 50: These fit well in `Trento.ActivityLog.SeverityLevel` so we have bi directional translation in one place.

test/trento_web/controllers/v1/activity_log_controller_test.exs
```diff
@@ -279,6 +279,53 @@ defmodule TrentoWeb.V1.ActivityLogControllerTest do
       assert resp["data"] == []
       assert_schema(resp, "ActivityLog", api_spec)
     end
+
+    test "should return info and above severity level entries by default",
```
nelsonkopliku, line 283: **suggestion:** we might have controller tests deal with asserting the response matching the schemas.

I believe we could add a few tests in `lib/trento/activity_log.ex` where we assert about the filtered result is the actual one we expect.

nelsonkopliku (commented):
Looking good Gagan. Just a couple of comments.

test/trento_web/controllers/v1/activity_log_controller_test.exs
gagandeepb: The expectation should be at the controller level here since there's defaults being injected somewhere in the middle. I have added some context level tests too, for completeness, but the asserts on controller tests are stronger.

lib/trento/activity_log.ex
```diff
@@ -120,6 +124,17 @@ defmodule Trento.ActivityLog do
...
 
+      {:severity, v} ->
+        {:filters,
+         %{
+           field: :severity,
+           op: :>=,
+           value:
+             v
+             |> (&String.to_existing_atom/1).()
+             |> (&SeverityLevel.severity_level_to_integer/1).()
```
nelsonkopliku, line 135: question: 
Could these just be function calls?
```elixir
v
  |> String.to_existing_atom()
  |> SeverityLevel.severity_level_to_integer()
```

test/e2e/cypress/e2e/activity_log.cy.js
nelsonkopliku, line 1: I guess changes in this e2e test are not related to the changes in this PR.
I believe you could defer this when addressing ui concerns.

nelsonkopliku (approved):
Hey Gagan, looks good to me.
I believe you could defer changes to the e2e test when actually changing the ui and adding the new filter.

gagandeepb (commented):
E2E test/FE adjustments are necessary for CI to pass. (Discussed sync.)

PR #3294, opened 2025-02-12
Add Trento documentation to navigation bar

janvhs:
# Description

This commit adds the documentation link to the navigation bar

Fixes # (issue)

- [x] **DONE**

## How was this tested?

There are no tests for the navigation bar

### DIFF
--- a/assets/js/pages/Layout/Layout.jsx
+++ b/assets/js/pages/Layout/Layout.jsx
@@ -185,7 +185,7 @@ function Layout() {
                   href="https://www.trento-project.io/docs/"
                   target="_blank"
                   className={classNames(
-                    'flex gap-2 items-center text-jungle-green-800 font-bold bg-jungle-green-100 py-2 rounded-md hover:opacity-75',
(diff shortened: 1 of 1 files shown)


jagabomb (commented):
Thanks for this @janvhs. I realise now that we missed an opportunity with the collapsed sidebar state. I am considering a solution like this with an icon for the collapsed sidebar...
 
Screenshot 2025-02-13 at 10 04 13

EMaksy (commented):
Nice, but yeah like @jagabomb suggested, we need also something in the collapsed nav bar. 
Additionally it would be great to test it, maybe include it in the e2e test or a regular test. 
What do you think ?

jagabomb:
Sorry last request here... I would like the text and icon to be `jungle-green-800`. 😊  I think the initial mockup was too light and the current implementation has two different colours instead of a single colour.

New Proposal:
Screenshot 2025-03-04 at 11 45 51

jagabomb (approved):
LGTM

PR #3304, opened 2025-02-17
Add operations policy plug

arbulu89:
# Description
Add a pretty basic plug to standardize the way we authorize operations in the controllers.
This will help to avoid having the `with Policy.authorize(_) do` code in all the controller endpoints where we need to authorize things. It might look an overkill now, but once we more operations it will come handy.

To see a real implementation: https://github.com/trento-project/web/pull/3305

## How was this tested?
UT

### DIFF
--- a/lib/trento_web/controllers/fallback_controller.ex
+++ b/lib/trento_web/controllers/fallback_controller.ex
@@ -68,6 +68,13 @@ defmodule TrentoWeb.FallbackController do
     |> render(:"404")
   end
 
+  def call(conn, {:error, :operation_not_found}) do
+    conn
+    |> put_status(:not_found)
+    |> put_view(json: ErrorJSON)
+    |> render(:"404", reason: "Operation not found.")
+  end
+
   def call(conn, {:error, %Ecto.Changeset{} = changeset}) do
     conn
     |> put_status(:unprocessable_entity)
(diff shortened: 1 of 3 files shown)


gagandeepb (commented):
Thanks for this. 
Just one question: I wonder if this Plug can be implemented in terms of/wrapped around  [Bodyguard.Plug.Authorize](https://hexdocs.pm/bodyguard/Bodyguard.Plug.Authorize.html) ?
The two interfaces look similar-ish at first glance.

arbulu89:
> Thanks for this. Just one question: I wonder if this Plug can be implemented in terms of/wrapped around [Bodyguard.Plug.Authorize](https://hexdocs.pm/bodyguard/Bodyguard.Plug.Authorize.html) ? The two interfaces look similar-ish at first glance.

Yes, they are similar. This is based on that one.
The thing is that `Bodyguard` is strongly attached to the concept of the a user, which we don't have in operation policies.
I already commented that I discarded bodyguard policies usage for that same reason. 
At the end, the idea is not to put the loaded resource in the `user` field, which is out of place. And if you want to use the `params`, we would need to do uglier pattern matches in the `authorize` functions, as you wouldn't never be using the `user` field

gagandeepb (approved):

nelsonkopliku (approved):
LGTM. 

Only comment could be that since `resource` and `params` options are functions, we might consider names suggesting those are functions `resource_loader|resource_fun` or similar (same for params).
I am fine either way, though 👍

arbulu89:
> LGTM.
> 
> Only comment could be that since `resource` and `params` options are functions, we might consider names suggesting those are functions `resource_loader|resource_fun` or similar (same for params). I am fine either way, though 👍

I just followed bodyguard nomenclature. As this "replicates" that logic, I think having the same format looks ok

PR #3308, opened 2025-02-18
Add reset customization cta

nelsonkopliku:
# Description

This PR adds the reset check customization confirmation modal.

API call will be wired up via a custom hook as a follow up.


EMaksy (approved):
Great!

PR #3310, opened 2025-02-18
Add boolean input for checks customization modal

EMaksy:
# Description

This Pr enriches the existing CheckCustomization modal to support boolean input. As checks coming from wanda may contain boolean values in the future. 
For this i created two different sub components for boolean input and string/number input. As the logic differs a bit. Besides this the pr ensures that the payload datatype is correctly transformed from the input.

### Unchecked 

image

### After checks customization

Note: style was adjusted of the warning banner after rebase
image

## How was this tested?
Updated and enriched storybook
Refactored and enriched existing test


assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
```diff
@@ -51,6 +64,8 @@ function CheckCustomizationModal({
   const [checked, setChecked] = useState(customized);
   const [customValues, setCustomValues] = useState({});
   const canCustomize = customized || checked;
+  const userCustomizedValues =
```
nelsonkopliku, line 67: **question:** what cases is the extra `Object.keys(customValues).length === 0` check covering?
Asking because, logically, checking `!canCustomize` which is based on `customized` should be enough, unless I am missing something.

BTW `userCustomizedValues` does not seem to sound with the logic applied to it.
A better matching name could be `canSave` or `canCustomize` (as it initially was 😄 )

nelsonkopliku (commented):
Hey @EMaksy thanks for this. Going in the right direction.

I left a few comments about naming/organization, let me know what you think.

EMaksy (commented):
Okay naming is very very hard, thanks for the review. I removed the two different input components and merged them to one.

nelsonkopliku (commented):
Alright, almost there.

If you want we can defer the `inputTypeBool` thingy.
Let's address the other two minor things, tho.

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
```diff
@@ -51,6 +41,7 @@ function CheckCustomizationModal({
   const [checked, setChecked] = useState(customized);
   const [customValues, setCustomValues] = useState({});
   const canCustomize = customized || checked;
+  const canSave = Object.keys(customValues).length === 0 || !canCustomize;
```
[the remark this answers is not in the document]
EMaksy: Oh yeah much better!

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
```diff
@@ -21,20 +23,14 @@ const buildCustomCheckPayload = (checkID, values) => {
...
+const appliedValue = (value) => value?.custom_value ?? value?.current_value;
 
-  return name?.length > 25 ? (
-    <Tooltip zIndex="50" content={name}>
-      {labelContent}
-    </Tooltip>
-  ) : (
-    labelContent
-  );
+const whichInputType = (value) => {
```
EMaksy, line 28: @nelsonkopliku  in order to make the component more future prove we could use this helper function which allows us to decide which kind of input we want to render.

assets/js/common/CheckCustomizationModal/CheckCustomizationInput.jsx
```diff
@@ -0,0 +1,98 @@
...
+    handleInput(name, newValue);
+  };
+
+  const handleDefaultChange = (inputEvent) => {
+    handleInput(name, inputEvent.target.value);
+  };
+
+  const renderInput = (type) => {
+    switch (type) {
+      case 'boolean':
```
EMaksy, line 52: and here we can add more cases for what ever scenario we require. What do you think ?

nelsonkopliku: Sounds good

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
nelsonkopliku: Ok we can start with this. 
Consider also names like `detectType`, `getInputType` etc if they sound better to you.

nelsonkopliku (approved):
Thanks @EMaksy, I thing we can go with this.
Further improvement can be considered as we proceed.

PR #3312, opened 2025-02-19
Update `date-fns` to 4.1.0 and DateFilter refactoring

skrech:
# Description

Updated `date-fns` to 4.1.0. Its main addition is support for time-zones (provided by two separate packages, `@date-fns/utc` and `@date-fns/tz`, we use only the former one). By  updating to the newer version, old dependency for time-zone handling becomes obsolete, and it's removed (`date-fns-tz`). All the dependent code is adapted to work with the newer libraries.

During the implementation of this, I've encountered a bug, which manifests itself when using the DateFilter class on the ActivityLogPage -- when selecting one of the pre-configured options, wrong UTC strings were put into the searchParams caused by incorrect time-zone conversion. This bug is fixed by re-factoring `DateFilter` class to internally always work with UTC timestamps, while collaborating with external actors with plain Date objects. The interface of `DateFilter` class is not changed, but stronger guarantees are provided for when onChange events happen.

Fixes [TRNT-3327](https://jira.suse.com/browse/TRNT-3327)

## How was this tested?

I've run the automated front-end tests. Validated the changes manually, as well. All seems to be fine.
Some of the test files were changed to accommodate for the changed date-time handling in DateFilter.

No need for documentation update.


arbulu89 (approved):
It looks great @skrech 
We can merge from my side once we rebase with the fix and check the deployment

skrech:
Depends on https://github.com/trento-project/web/pull/3321

PR #3311, opened 2025-02-19
Saptune solution apply modal

arbulu89:
# Description

Add Saptune solution apply modal to the frontend.
The solutions per each host are based on having HANA or Application instances running. More details here: https://documentation.suse.com/sles-sap/15-SP5/html/SLES-SAP-guide/cha-tune.html#sec-saptune-configure-tuning

The `OperationModal` is a generic template to be used internally to create specific modals

image

PD: Once we have the warning box with the checkbox component we can replace it

## How was this tested?

UT and storybook

PR #3313, opened 2025-02-19
Create checkable warning messages

EMaksy:
# Description

This pr refactors the CheckCustomizationModal by exporting the warning banner to an own common component. With this it can be reused through the application if needed.

Checkout storybook to see how the component behaves

## How was this tested? 

Added unit test

### DIFF
--- a/assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
+++ b/assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
@@ -12,12 +12,7 @@ function CheckableWarningMessage({
   return (
     <div className="flex items-center border border-yellow-400 bg-yellow-50 p-4 rounded-md text-yellow-600 mb-4">
       {!hideCheckbox && (
-        <Input
-          type="checkbox"
-          checked={checked}
-          onChange={() => onChecked((prev) => !prev)}
-          className="w-5 h-5 border-2 border-yellow-400 rounded-md focus:ring-yellow-400"
-        />
+        <Input type="checkbox" checked={checked} onChange={onChecked} />
       )}
       <EOS_WARNING_OUTLINED
         size="l"
--- a/assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
+++ b/assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
@@ -7,7 +7,7 @@ function CheckableWarningMessage({
   hideCheckbox = false,
   warningText,
   checked,
-  setChecked = noop,
+  onChecked = noop,
 }) {
   return (
     <div className="flex items-center border border-yellow-400 bg-yellow-50 p-4 rounded-md text-yellow-600 mb-4">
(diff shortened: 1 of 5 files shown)


assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
```diff
@@ -0,0 +1,30 @@
...
+    <div className="flex items-center border border-yellow-400 bg-yellow-50 p-4 rounded-md text-yellow-600 mb-4">
+      {!hideCheckbox && (
+        <Input
+          type="checkbox"
+          checked={checked}
+          onChange={() => setChecked((prev) => !prev)}
+        />
+      )}
+      <EOS_WARNING_OUTLINED
+        size="xxl"
```
EMaksy, line 22: We could add more props to have more control over the component, but not sure if we need it right away to be honest. 
What do you think ?

EMaksy:
Maybe we can use the component in #3311

assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import Input from '@common/Input';
+import { noop } from 'lodash';
+import { EOS_WARNING_OUTLINED } from 'eos-icons-react';
+
+function CheckableWarningMessage({
+  hideCheckbox = false,
+  warningText,
+  checked,
+  setChecked = noop,
```
arbulu89, line 10: Maybe better `onChecked`?
And in `onChange` you simply do `onChange={onChecked}`.
It should be the user of this module responsabilite to decide what to do them

assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
arbulu89, line 22: Based on @jagabomb mock up the size of the icon is `l`.

assets/js/common/CheckableWarningMessage/CheckableWarningMessage.test.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { render, screen, fireEvent } from '@testing-library/react';
```
arbulu89, line 2: Don't use `fireEvent`, use `userEvent`

arbulu89 (commented):
Hey @EMaksy ,
Some comments to be closer to the proposed mock up.
Besides that, at least in my mock up the checkbox border and and selected state have the same yellow color of the box.
Could you check if we can apply those styles to the input?

assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
```diff
@@ -0,0 +1,26 @@
...
+import { EOS_WARNING_OUTLINED } from 'eos-icons-react';
+
+function CheckableWarningMessage({
+  hideCheckbox = false,
+  warningText,
+  checked,
+  onChecked = noop,
+}) {
+  return (
+    <div className="flex items-center border border-yellow-400 bg-yellow-50 p-4 rounded-md text-yellow-600 mb-4">
```
arbulu89, line 13: Maybe adding `text-sm` and `font-normal` makes it closer to the mock up we have.
@jagabomb what do you think?

arbulu89 (approved):
Perfect!
Just a comment about the font of the message.
Anyway, it looks good as it is as well.
@jagabomb can you review this?

nelsonkopliku (approved):
LGTM

assets/js/common/CheckableWarningMessage/CheckableWarningMessage.jsx
EMaksy: I would apply the suggested changes for now we can revisit this @jagabomb

PR #3319, opened 2025-02-20
Actions button

arbulu89:
# Description

Add `ActionsButton` component to be used in the operation framework. 
It receives a list of actions with `value`, `running`, `disabled` and `onClick` parameters that are added in a dropdown menu.

image

PD: About the usage of the `ref`: https://headlessui.com/react/menu#rendering-as-different-elements

## How was this tested?

UT and storybook

### DIFF
--- a/assets/js/common/ActionsButton/ActionsButton.jsx
+++ b/assets/js/common/ActionsButton/ActionsButton.jsx
@@ -1,0 +1,68 @@
+import React, { useRef, Fragment } from 'react';
+import classNames from 'classnames';
+import { some } from 'lodash';
+import {
+  Menu,
+  MenuButton,
+  MenuItems,
+  MenuItem,
+  MenuSeparator,
+} from '@headlessui/react';
+import { EOS_MORE_VERT, EOS_LOADING_ANIMATED } from 'eos-icons-react';
+
+import Button from '@common/Button';
+
+function ActionsButton({ actions }) {
+  const ref = useRef(null);
+  const someRunning = some(actions, { running: true });
+
+  return (
+    <Menu>
+      <MenuButton as={Fragment}>
+        <div className="flex" ref={ref}>
+          <Button
+            type="primary-white"
+            className="inline-block mx-0.5 border-green-500 border"
+            size="small"
+          >
+            <EOS_MORE_VERT className="inline-block fill-jungle-green-500" />{' '}
+            Actions
+          </Button>
+        </div>
+      </MenuButton>
+      <MenuItems
+        anchor={{ to: 'bottom start', gap: '5px' }}
+        as="div"
+        className="w-56 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 px-3 py-3 focus:outline-none"
+      >
+        {actions.map(({ value, running, disabled, onClick }) => (
+          <div key={value}>
+            <MenuItem>
+              <button
+                type="button"
+                onClick={onClick}
+                className={classNames(
+                  'w-full rounded-md text-left block px-4 py-2 text-sm',
+                  {
+                    'text-gray-700 hover:bg-gray-100':
+                      !disabled && !someRunning,
+                  },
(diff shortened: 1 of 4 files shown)


abravosuse (approved):
Code apart, it looks good to me @arbulu89

assets/js/common/ActionsButton/ActionsButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+  MenuButton,
+  MenuItems,
+  MenuItem,
+  MenuSeparator,
+} from '@headlessui/react';
+import { EOS_MORE_VERT, EOS_LOADING_ANIMATED } from 'eos-icons-react';
+
+import Button from '@common/Button';
+
+function ActionsButton({ actions }) {
```
EMaksy, line 15: Overall question, do we want common components to be more readable and destruct the required data fields or keep `actions` as it is more or less simple. But requires us to have a deeper look at the component.

assets/js/common/ActionsButton/ActionsButton.stories.jsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export function Default(args) {
+  return <ActionsButton {...args} />;
+}
+
+export function Disabled({ actions }) {
+  return (
+    <ActionsButton
+      actions={map(actions, (action, index) =>
```
EMaksy, line 40: Maybe exporting it to a const variable would improve the readability a bit

assets/js/common/ActionsButton/ActionsButton.stories.jsx
```diff
@@ -0,0 +1,55 @@
...
+        index === 0 ? { ...action, disabled: true } : action
+      )}
+    />
+  );
+}
+
+export function Running({ actions }) {
+  return (
+    <ActionsButton
+      actions={map(actions, (action, index) =>
```
EMaksy, line 50: or even have a simple helper function ? 
```

const modifyAction = (actions, key, value) =>
  actions.map((action, index) =>
    index === 0 ? { ...action, [key]: value } : action
  );

export function Disabled({ actions }) {
  return <ActionsButton actions={modifyAction(actions, 'disabled', true)} />;
}

export function Running({ actions }) {
  return <ActionsButton actions={modifyAction(actions, 'running', true)} />;
}
```

EMaksy (approved):
Hey man good job, i left some overall questions and minor suggestions :+1:

assets/js/common/ActionsButton/ActionsButton.jsx
```diff
@@ -0,0 +1,68 @@
...
+            Actions
+          </Button>
+        </div>
+      </MenuButton>
+      <MenuItems
+        anchor={{ to: 'bottom start', gap: '5px' }}
+        as="div"
+        className="w-56 rounded-md shadow-lg bg-white ring-1 ring-black ring-opacity-5 px-3 py-3 focus:outline-none"
+      >
+        {actions.map(({ value, running, disabled, onClick }) => (
```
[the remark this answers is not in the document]
arbulu89: I don't find any scenario where we would have the same entry twice

assets/js/common/ActionsButton/ActionsButton.jsx
arbulu89: How can you destruct array element values in the signature?

EMaksy: No you can't directly, or it would create more things to do. It makes here sense to have the array.

assets/js/common/ActionsButton/ActionsButton.stories.jsx
arbulu89: I will go with `Object.assign([], actions, {0: { ...actions[0], disabled: true }})`.
I don't want to create a function for a such a simple thing

jagabomb (approved):
LGTM

PR #3321, opened 2025-02-20
Fix operations policy get_params usage

arbulu89:
# Description
The `OperationsPolicyPlug` incorrect usage has caused an error that I was only reproduced when generating the release. I didn't see it running locally.

This is the error: https://github.com/trento-project/web/actions/runs/13430969298/job/37522721412

The `Keyword.get` function/macro cannot find a private function apparently.

I'm setting the `env` label to check if everything works fine

## How was this tested?

Creating the docker file locally and with the `env` label


CDimonaco (approved):
:O :lgtm:

PR #3323, opened 2025-02-21
Add reset customization cta inside modal

nelsonkopliku:
# Description

This PR:
- adds a `Reset Button` in the check customization modal actions
- properly updated the state so that non customized values are shown
- adds minor improvements

### DIFF
--- a/assets/js/pages/ChecksSelection/ChecksSelection.jsx
+++ b/assets/js/pages/ChecksSelection/ChecksSelection.jsx
@@ -92,11 +92,9 @@ function ChecksSelection({
           onReset={() => {
             onResetCheckCustomization(groupID, selectedCheck?.id);
             setResetConfirmationModalOpen(false);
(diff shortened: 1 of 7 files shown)


assets/js/pages/ChecksSelection/ChecksSelection.jsx
```diff
@@ -138,11 +136,9 @@ function ChecksSelection({
           customized={selectedCheck?.customized}
           selectedCheck={selectedCheck}
           provider={provider}
-          onClose={() => {
-            setIsCheckCustomizationModalOpen(false);
-            setSelectedCheck(null);
```
nelsonkopliku, line 143: Removing this `setSelectedCheck(null);` saves us from some flickering in the UI.
It is safe because every time we open the modal the current _selected check_ is overwritten

EMaksy (approved):
Hey man nice lgtm

PR #3331, opened 2025-02-25
Add OperationForbiddenModal

arbulu89:
# Description

Add operation forbidden modal. We can render any kind of markdown and have specific description messages for each operation. For example:

image

PD: Don't check too much the text after the banner. We will review that in other PRs, when it is really implemented. This was just an example

## How was this tested?

UT

### DIFF
--- a/assets/js/common/OperationModals/OperationForbiddenModal.test.jsx
+++ b/assets/js/common/OperationModals/OperationForbiddenModal.test.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { act, render, screen } from '@testing-library/react';
+import { act, render, screen, within } from '@testing-library/react';
 import '@testing-library/jest-dom';
 import userEvent from '@testing-library/user-event';
 import OperationForbiddenModal from './OperationForbiddenModal';
@@ -8,7 +8,11 @@ describe('OperationForbiddenModal', () => {
   it('should show forbidden operation modal', async () => {
     await act(async () => {
       render(
-        <OperationForbiddenModal operation="My operation" isOpen>
+        <OperationForbiddenModal
+          operation="My operation"
+          errors={['error1', 'error2']}
+          isOpen
+        >
           Some children
         </OperationForbiddenModal>
       );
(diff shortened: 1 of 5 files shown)


abravosuse (approved):
If we are going to review the banner text later on, then - code apart - this looks good to me.

arbulu89:
> If we are going to review the banner text later on, then - code apart - this looks good to me.

Yes, the text that comes after the banner will depend on the operation itself.
The unique thing that are hardcoded now are the title and banner content (except the operation name itself, which comes as variable)

abravosuse:
> Yes, the text that comes after the banner will depend on the operation itself. The unique thing that are hardcoded now are the title and banner content (except the operation name itself, which comes as variable)

No need to discuss it here but wondering if we are going to have a warning message tailored to the specific SAP workload(s) in the host or it's going to be a generic message trying to cover all possible cases. For example:

- SAP workload managed by cluster : 
- - ASCS/ERS workload
- - HANA workload
- SAP workload not managed by cluster
- - ASCS/ERS workload
- - Application Server Instance workload
- - HANA workload

arbulu89:
@abravosuse I'm putting this PR back in draft. We clarified offline that we are going to do some adjustments to get the exact reasons of why the operation is forbidden to generate the modal message

abravosuse:
It looks great @arbulu89 . As mentioned in Slack, my suggestion regarding SAP instance is:

- For HANA instances: Instance ## of HANA database XYZ is not stopped
- For application instances: Instance ## of SAP system XYZ is not stopped

Also, we might want to have a specific link per message: one link for cluster related messages, one link for SAP instances related messages.

jagabomb (approved):
LGTM

assets/js/common/OperationModals/OperationForbiddenModal.test.jsx
```diff
@@ -0,0 +1,56 @@
...
+        >
+          Some children
+        </OperationForbiddenModal>
+      );
+    });
+
+    expect(screen.getByText('Operation Forbidden')).toBeInTheDocument();
+    expect(
+      screen.getByText(
+        'Unable to run My operation operation. Some of the conditions are not met.'
```
EMaksy, line 24: operation operation two times ?
I mean sure code wise all perfect,  i was just confused on the first read by it :)

assets/js/common/OperationModals/SaptuneSolutionApplyModal.stories.jsx
```diff
@@ -15,7 +15,7 @@ export default {
       control: 'boolean',
     },
     isOpen: {
-      description: 'Model is open',
+      description: 'Modal is open',
```
EMaksy, line 18: oopsie, thanks

EMaksy (approved):
Great as always, just a little nitpick about the readability in a test. Everything else lgtm!

PR #3332, opened 2025-02-25
Refactor checkable warning message

EMaksy:
# Description

A small refactor of the CheckableWarningMessage component to make it more customizable by passing children instead of only text. With this we can reuse 

## How was this tested?

Updated existing test and storybook

### DIFF
--- a/assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
+++ b/assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
@@ -79,10 +79,11 @@ function CheckCustomizationModal({
       </p>
       <CheckableWarningMessage
         hideCheckbox={customized}
-        warningText={checkBoxWarningText}
         checked={checked}
         onChecked={() => setChecked((prev) => !prev)}
-      />
+      >
+        {checkBoxWarningText}
+      </CheckableWarningMessage>
       {values
         ?.filter(({ customizable }) => customizable)
         .map((value) => (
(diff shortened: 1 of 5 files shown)


assets/js/common/CheckCustomizationModal/CheckCustomizationModal.jsx
```diff
@@ -79,10 +79,11 @@ function CheckCustomizationModal({
       </p>
       <CheckableWarningMessage
         hideCheckbox={customized}
-        warningText={checkBoxWarningText}
         checked={checked}
         onChecked={() => setChecked((prev) => !prev)}
-      />
+      >
+        <span className="font-semibold">{checkBoxWarningText}</span>
```
arbulu89, line 85: What if we put the `span` in the component itself and only leave the text as children.
This way we don't need to replicate the span and classname, and after all, all the texts should have the same font in the component.
What do you think?

EMaksy: Yeah sure, i was thinking between this two options :) 
I'm going to adjust like you suggest it. I was not sure if we want more options to modify or not :)

assets/js/common/OperationModals/OperationModal.jsx
```diff
@@ -17,6 +17,8 @@ function OperationModal({
   onCancel = noop,
   children,
 }) {
+  const warningMessage = `Trento & SUSE cannot be held liable for damages if system is unable to
```
arbulu89, line 20: You could directly put the text as children. No real need for having a constant. Less to scroll hehe

arbulu89 (approved):
Thank you @EMaksy

PR #3333, opened 2025-02-25
Request and process operation events

arbulu89:
# Description

 - Request operation and send the event to wanda.
 - Process incoming operation events from wanda.

Wanda implementation: https://github.com/trento-project/wanda/pull/577

PD: forget about the API BC check. I just used a more appropriate response schema

## How was this tested?

UT

### DIFF
--- a/lib/trento/hosts.ex
+++ b/lib/trento/hosts.ex
@@ -139,14 +139,19 @@ defmodule Trento.Hosts do
     end
   end
 
-  @spec request_operation(atom(), String.t(), map()) :: :ok | {:error, any}
+  @spec request_operation(atom(), String.t(), map()) :: {:ok, String.t()} | {:error, any}
   def request_operation(:saptune_solution_apply, host_id, params) do
-    Operations.request_operation(
-      UUID.uuid4(),
-      host_id,
-      "saptuneapplysolution@v1",
-      [%{agent_id: host_id, arguments: params}]
-    )
+    operation_id = UUID.uuid4()
+
+    with :ok <-
+           Operations.request_operation(
+             operation_id,
+             host_id,
+             "saptuneapplysolution@v1",
+             [%{agent_id: host_id, arguments: params}]
+           ) do
+      {:ok, operation_id}
+    end
   end
 
   def request_operation(_, _, _), do: {:error, :operation_not_found}
--- a/lib/trento/hosts.ex
+++ b/lib/trento/hosts.ex
@@ -23,7 +23,10 @@ defmodule Trento.Hosts do
     SelectHostChecks
   }
 
-  alias Trento.Infrastructure.Checks
+  alias Trento.Infrastructure.{
+    Checks,
+    Operations
+  }
 
   alias Trento.Repo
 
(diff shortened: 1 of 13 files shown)


lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -191,15 +191,11 @@ defmodule TrentoWeb.V1.HostController do
...
 
-    Logger.info(
-      "Operation authorized. Send real saptune_solution_apply with #{solution} to #{host_id}"
-    )
-
-    # Request operation, get operation id when created and return in json
-
-    conn
-    |> put_status(:accepted)
-    |> json(%{operation_id: UUID.uuid4()})
```
arbulu89, line 202: At the end I'm removing the `operation_id` value from the response. We are not using it by now.

arbulu89: Forget about it. We are going to return the `operation_id`. It will be used at least by the activity logging

lib/trento/infrastructure/operations/operations.ex
```diff
@@ -0,0 +1,58 @@
...
+
+        error
+    end
+  end
+
+  defp map_value(map) when is_map(map) do
+    Enum.into(map, %{}, fn {key, value} -> {map_key(key), map_value(value)} end)
+  end
+
+  defp map_value(value) when is_number(value), do: %{kind: {:number_value, value}}
```
CDimonaco, line 50: Could we use for clarity the module `Trento.Operations.V1.OperationTarget.ArgumentsEntry` instead of the plain map?

CDimonaco: Of course to all the map value

lib/trento/hosts.ex
```diff
@@ -136,6 +139,23 @@ defmodule Trento.Hosts do
     end
   end
 
+  @spec request_operation(atom(), String.t(), map()) :: {:ok, String.t()} | {:error, any}
+  def request_operation(:saptune_solution_apply, host_id, params) do
+    operation_id = UUID.uuid4()
+
+    with :ok <-
```
CDimonaco, line 146: Could we use a `case` instead of a single step with? It improves readability and allow to read the code better when a single operation is involved and the error handling is more clear

lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -191,15 +192,12 @@ defmodule TrentoWeb.V1.HostController do
...
+    with {:ok, operation_id} <-
```
CDimonaco, line 195: I think we have this pattern with a single with in controllers, but if you want to make more explicit, the case statement also here is a good idea

CDimonaco (approved):
Good Job, I left some comments

lib/trento/infrastructure/operations/operations.ex
arbulu89: We cannot really use `ArgumentsEntry`. That's just an internal representation of a map given by the elixir protobuf implementation. In reality we still need to provide a plain map.
Either way, as commented offline, I found out that the new version of protobuf elixir library comes with a super util functions to abstract this logic: `to_map` and `from_map`

lib/trento_web/controllers/v1/host_controller.ex
arbulu89: I will keep this using the `with`. I think it is the standard way of using the fallback controller and we have it all around the controllers

nelsonkopliku (approved):
Noice

PR #3345, opened 2025-02-27
Add operations button host view

arbulu89:
# Description

Add operations buttion and Saptune solution apply modal to the host details view.
I added a selector function to get the data from redux, so we can put the solution in running state from there.
The disabled option is simple function checking if saptune is not set and the host has SAP instances.

I have passed the `operationsEnabled` from the backend, so we only display the operations button in dev.

## How was this tested?

UT and storybook


CDimonaco (approved):
LGTM 🚀

PR #3353, opened 2025-03-03
Rabbitmq explicit ssl support

CDimonaco:
# Description

This PR adds support for RabbitMQ TLS for development and testing purposes.

In production, the RabbitMQ connection is already configured using an environment variable that should be set to an AMQP URI. This URI can include TLS options to connect to secured RabbitMQ brokers. See [RabbitMQ URI Query Parameters](https://www.rabbitmq.com/docs/uri-query-parameters#example-b) for reference.

This PR introduces a TLS bundle—including a custom CA, server, and client certificates—to enable SSL connections in development and testing. By setting the environment variable USE_LOCAL_RABBIT_TLS, the connection switches to an amqps connection in these environments.
Example Usage

`USE_LOCAL_RABBIT_TLS=true iex -S mix`

This starts the IEx shell in development mode, using an SSL connection to the RabbitMQ instance exposed via Docker Compose. The broker supports both TLS and non-TLS connections simultaneously.

This capability will also be extended to other components of Trento. They will share the same CA, enabling interconnection between all components in development mode over SSL. Additionally, all RabbitMQ servers exposed through Docker Compose will support secure connections for all parties.

Fixes # (issue)

- [X] **DONE**

## How was this tested?

- [X] **DONE**

- [X] **DONE**


config/dev.exs
```diff
@@ -96,21 +96,28 @@ unless IEx.started?() do
     ]
 end
 
+amqp_connection =
+  if System.get_env("USE_LOCAL_RABBIT_TLS") do
```
arbulu89, line 100: No need for a `if`. We put the default value as 2nd parameter, so if the env value is not set it will be used

config/test.exs
```diff
@@ -51,14 +51,21 @@ config :trento,
   api_key_authentication_enabled: false,
   jwt_authentication_enabled: false
 
+amqp_connection =
+  if System.get_env("USE_LOCAL_RABBIT_TLS") do
```
arbulu89, line 55: same, no need for `if`

container_fixtures/rabbitmq/rabbitmq.conf
```diff
@@ -0,0 +1,6 @@
+ssl_options.cacertfile           = /etc/ssl/certs/ca_certificate.pem
+ssl_options.certfile             = /var/lib/rabbitmq/server_rabbit.trento.local_certificate.pem
+ssl_options.keyfile              = /var/lib/rabbitmq/server_rabbit.trento.local_key.pem
+ssl_options.verify               = verify_peer
+ssl_options.fail_if_no_peer_cert = true
+listeners.ssl.default = 5671
```
arbulu89, line 6: nitpick: we could indent the value to be the same as the others :see_no_evil:

arbulu89 (commented):
Nice!
I think we should update the connection in `wanda.exs` as well. At the end, this is when we will have the communication working. When we use this config, we tell `web` to connect to `wanda` rabbitmq installation. I think the easiest thing will be to copy the server and ca certificates to wanda docker installation, and simply create only wanda specific client certs.
What do you think?

CDimonaco:
> Nice! I think we should update the connection in `wanda.exs` as well. At the end, this is when we will have the communication working. When we use this config, we tell `web` to connect to `wanda` rabbitmq installation. I think the easiest thing will be to copy the server and ca certificates to wanda docker installation, and simply create only wanda specific client certs. What do you think?

That's exactly what I will do, copy to all the components the same ca and the same server certificates, each client will have different client certificates built with the same ca in order to work with all the rabbitmqs we have!

I will also update wanda conf

config/dev.exs
CDimonaco: Do you mean provide the whole connection string as default argument? My idea was to have a simple "switch" without passing explicitly the whole string but just a boolean thing to switch between the two

arbulu89: Yes, my fault. I read it to fast. I didn't realize that this is just a boolean.
All good!

arbulu89:
> > Nice! I think we should update the connection in `wanda.exs` as well. At the end, this is when we will have the communication working. When we use this config, we tell `web` to connect to `wanda` rabbitmq installation. I think the easiest thing will be to copy the server and ca certificates to wanda docker installation, and simply create only wanda specific client certs. What do you think?
> 
> That's exactly what I will do, copy to all the components the same ca and the same server certificates, each client will have different client certificates built with the same ca in order to work with all the rabbitmqs we have!
> 
> I will also update wanda conf

Ok! Considering that, could we change the `wanda.exs` entry to do the same `if` that we do in `dev`? Just to point to `wanda` rabbitmq as it does now. I guess we will need to "save" other port for wanda SSL connection. 
Just not to open a new PR only with that

container_fixtures/rabbitmq/rabbitmq.conf
CDimonaco: done!

CDimonaco:
> > > Nice! I think we should update the connection in `wanda.exs` as well. At the end, this is when we will have the communication working. When we use this config, we tell `web` to connect to `wanda` rabbitmq installation. I think the easiest thing will be to copy the server and ca certificates to wanda docker installation, and simply create only wanda specific client certs. What do you think?
> > 
> > 
> > That's exactly what I will do, copy to all the components the same ca and the same server certificates, each client will have different client certificates built with the same ca in order to work with all the rabbitmqs we have!
> > I will also update wanda conf
> 
> Ok! Considering that, could we change the `wanda.exs` entry to do the same `if` that we do in `dev`? Just to point to `wanda` rabbitmq as it does now. I guess we will need to "save" other port for wanda SSL connection. Just not to open a new PR only with that

Yep done, I chose the port 5676 I will make sure that will be the tls port on wanda

PR #3354, opened 2025-03-03
Log operation requested in activity log

arbulu89:
# Description

Log operation requested event in the activity log.
The `operation_id` comes in the response of the request, in raw format. That's why we need to decode the json. 

image

@abravosuse I'm putting the "internal" name of the operation here. Should we put something else? We could change the message to be `Operation Apply Saptune solution requested` for example. The value in the Data field is more annoying, as we would need to map the internal value to the "frontend label" which I don't want to do in the backend as well

## How was this tested?
UT

### DIFF
--- a/assets/js/lib/model/activityLog.js
+++ b/assets/js/lib/model/activityLog.js
@@ -9,6 +9,7 @@ import {
   uniq,
   values,
 } from 'lodash/fp';
+import { getOperationResourceType } from '@lib/operations';
 import { isPermitted } from './users';
 
 export const LOGIN_ATTEMPT = 'login_attempt';
@@ -112,6 +113,9 @@ export const DATABASE_ROLL_UP_REQUESTED = 'database_roll_up_requested';
 export const DATABASE_TENANTS_UPDATED = 'database_tenants_updated';
 export const DATABASE_TOMBSTONED = 'database_tombstoned';
 
+// Operations
+export const OPERATION_REQUESTED = 'operation_requested';
+
 export const resourceTypes = ['host', 'cluster', 'database', 'sap_system'];
 
 export const resourceTypesToNameKeyMap = {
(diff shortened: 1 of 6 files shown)


abravosuse:
'@abravosuse I'm putting the "internal" name of the operation here. Should we put something else? We could change the message to be Operation Apply Saptune solution requested for example. The value in the Data field is more annoying, as we would need to map the internal value to the "frontend label" which I don't want to do in the backend as well'

@arbulu89 yes, I like the idea of changing the message into a more "user friendly" format like the one you suggest. No need to touch the value in the Data field. That's meant to be raw data.

nelsonkopliku (approved):
LGTM.

Just a question for me to better understand: the params sent to `HostController.request_operation` controller are coming from wanda and then sent back to wanda via this endpoint that publishes a message on rabbit, correct?

gagandeepb (approved):
LGTM.

arbulu89:
> LGTM.
> 
> Just a question for me to better understand: the params sent to `HostController.request_operation` controller are coming from wanda and then sent back to wanda via this endpoint that publishes a message on rabbit, correct?

What do you mean, "are coming from wanda"? The params are coming from the frontend in the API request. Of course, each operation defines which parameters its needs, and Webs controller requires them declaring them in the body params of the request. So at the end is a shared knowledge web and wanda need to have to compose the messages.

nelsonkopliku:
Hey @arbulu89, what I meant is those values coming from the frontend to the api request are coming to the frontend from wanda, correct? Perhaps indeed enriched with information that web has.

arbulu89:
> Hey @arbulu89, what I meant is those values coming from the frontend to the api request are coming to the frontend from wanda, correct? Perhaps indeed enriched with information that web has.

No, they are not. The frontend is not consuming wanda to know which parameters to use. Each operation is implemented in a specific way, and the frontend and backend of Web know what parameters to send.

PR #3355, opened 2025-03-03
Fetch running operation on page load

arbulu89:
# Description

Update running operation state on page arrival. If we load trento in a moment when an operation was already running, this saga dispatch will query wanda to see if some operation is in running state for this group. If that's the case, it will set the operation in running, so we cannot request the operation again.

## How was this tested?
UT

### DIFF
--- a/assets/js/lib/api/operations.js
+++ b/assets/js/lib/api/operations.js
@@ -8,7 +8,7 @@ const defaultConfig = { baseURL };
 export const requestHostOperation = (hostID, operation, params) =>
   post(`/hosts/${hostID}/operations/${operation}`, params);
 
-export const getOperations = (params) =>
+export const getOperationExecutions = (params) =>
   networkClient.get('/api/v1/operations/executions', {
     ...defaultConfig,
     params,
--- a/assets/js/lib/api/operations.js
+++ b/assets/js/lib/api/operations.js
@@ -1,4 +1,15 @@
-import { post } from '@lib/network';
+import { networkClient, post } from '@lib/network';
+
+// eslint-disable-next-line no-undef
+const baseURL = config.checksServiceBaseUrl;
+
+const defaultConfig = { baseURL };
 
 export const requestHostOperation = (hostID, operation, params) =>
   post(`/hosts/${hostID}/operations/${operation}`, params);
+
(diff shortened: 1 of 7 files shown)


assets/js/lib/api/operations.js
```diff
@@ -1,4 +1,15 @@
-import { post } from '@lib/network';
+import { networkClient, post } from '@lib/network';
+
+// eslint-disable-next-line no-undef
+const baseURL = config.checksServiceBaseUrl;
```
arbulu89, line 4: I'm reusing this value as the at the end, the checks server and operations endpoints are in the same url.
I will open a tech-debt ticket to change this, but let's better do it before the release of the feature. It will require changing many things in deployment projects and documentation

assets/js/state/sagas/operations.test.js
```diff
@@ -27,96 +34,163 @@ describe('operations saga', () => {
...
-    );
-
-    expect(dispatched).toContainEqual(
-      setRunningOperation({ groupID, operation }),
-      notify({
-        text: `Operation ${operation} requested for ${hostname}`,
-        icon: '⚙️',
-      })
-    );
+  describe('request operation', () => {
```
arbulu89, line 37: It looks like there are many changes, but I simply grouped existing tests in their own describes and added the new `describe` for `update running operation` tests.

assets/js/state/sagas/operations.js
```diff
@@ -82,7 +86,27 @@ export function* completeOperation({ payload }) {
...
+
+  try {
+    const {
+      data: { items: operations, total_count: totalCount },
+    } = yield call(getOperationExecutions, {
+      group_id: groupID,
+      items_per_page: 1,
+    });
+    if (totalCount === 1 && operationRunning(operations[0])) {
+      const operation = getOperationInternalName(operations[0].operation);
```
CDimonaco, line 100: If we could destructure that in the signature the first item could be better, wdyt?

CDimonaco (approved):
LGTM

assets/js/state/sagas/operations.js
arbulu89: Done

PR #3358, opened 2025-03-04
Add modified check pill in checks execution results an detail

nelsonkopliku:
# Description

Adding the MODIFIED pill to checks in execution result and in result detail.

### DIFF
--- a/assets/js/common/ModifiedCheckPill/ModifiedCheckPill.jsx
+++ b/assets/js/common/ModifiedCheckPill/ModifiedCheckPill.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import Pill from '@common/Pill';
+
+function ModifiedCheckPill({
+  className,
+  content = 'MODIFIED',
+  customized = false,
+}) {
+  if (!customized) return null;
+  return (
+    <Pill
+      className={classNames(
+        'bg-white text-jungle-green-500 border border-jungle-green-500',
+        className
+      )}
+      size="xs"
+    >
+      {content}
+    </Pill>
+  );
+}
+
+export default ModifiedCheckPill;
--- a/assets/js/common/ModifiedCheckPill/ModifiedCheckPill.test.jsx
+++ b/assets/js/common/ModifiedCheckPill/ModifiedCheckPill.test.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import { screen, render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import ModifiedCheckPill from './ModifiedCheckPill';
+
+describe('ModifiedCheckPill component', () => {
+  it('should not render modified pill for a non customized check', () => {
+    render(<ModifiedCheckPill customized={false} />);
+
+    expect(screen.queryByText('MODIFIED')).toBeNull();
+  });
+
+  it('should render modified pill for a customized check', () => {
+    render(<ModifiedCheckPill customized />);
+
+    expect(screen.getByText('MODIFIED')).toBeVisible();
+  });
+
+  it('should render modified pill with custom content for a customized check', () => {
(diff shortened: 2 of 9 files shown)


assets/js/common/ModifiedCheckPill/ModifiedCheckPill.jsx
```diff
@@ -0,0 +1,21 @@
...
+  if (!customized) return null;
+  return (
+    <Pill
+      className={classNames(
+        'bg-white text-jungle-green-500 border border-jungle-green-500',
+        className
+      )}
+      size="xs"
+    >
+      MODIFIED
```
EMaksy, line 16: We could have a label prop with defaul value MODIFIED, so the component is a bit more reusable, what do you think ?

assets/js/common/ModifiedCheckPill/ModifiedCheckPill.test.jsx
```diff
@@ -0,0 +1,19 @@
...
+import { screen, render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+
+import ModifiedCheckPill from './ModifiedCheckPill';
+
+describe('ModifiedCheckPill component', () => {
+  it('should not render modified pill for a non customized check', () => {
+    render(<ModifiedCheckPill customized={false} />);
+
+    expect(screen.queryByText('MODIFIED')).toBeNull();
```
EMaksy, line 11: if you consider to add the label prop, we could test for default value and changed value

assets/js/common/ModifiedCheckPill/ModifiedCheckPill.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import Pill from '@common/Pill';
+
+function ModifiedCheckPill({ className, customized = false }) {
```
EMaksy, line 6: What about adding a story?

assets/js/pages/ExecutionResults/CheckResultDetail/ExpectedValues.stories.jsx
```diff
@@ -11,8 +10,8 @@ export const Default = {
   args: {
     isTargetHost: true,
     expectedValues: [
-      { name: faker.animal.bear(), value: faker.number.int() },
-      { name: faker.animal.bear(), value: faker.number.int() },
+      executionValueFactory.build({ customized: true }),
```
EMaksy, line 13: This is an overall question, do we want to use factories in our story? 
In case we introduce changes to the factory, we always need to check the stories where the dedicated factory was used. 
I'm not against it :+1:  , but on the other hand using static values in storybook is more trivial? 
What do you think ?

EMaksy (approved):
Hey man good job, left some suggestions and an overall question.

assets/js/common/ModifiedCheckPill/ModifiedCheckPill.jsx
nelsonkopliku: A specific value does not add much value, imho, as it is just a pill (already has stories) and is mainly showcased where used.

assets/js/pages/ExecutionResults/CheckResultDetail/ExpectedValues.stories.jsx
nelsonkopliku: Using factories in stories allows to have stories built with near-real data shape that components use at runtime.

It does not change much either way, imho, because if we change a component in a way that used data is affected we very likely need to change its related story static values or factories.

assets/js/common/ModifiedCheckPill/ModifiedCheckPill.jsx
nelsonkopliku: Done.

PR #3359, opened 2025-03-04
Fix: Verify mocks in settings_test.ex

skrech:
# Description

I've noticed that mocks are not verified in `settings_test.ex`. This is a one-liner fix for that.
I've double-checked that verify was *not* called from any of the `use` statements -- changing call count expectations in the mocks did not make a test to fail. After adding `verify_on_exit` and with modified expectations the tests start failing, so the fix is correct.

### DIFF
--- a/test/trento/settings_test.exs
+++ b/test/trento/settings_test.exs
@@ -30,6 +30,8 @@ defmodule Trento.SettingsTest do
     on_exit(fn -> Application.put_env(:trento, :flavor, "Community") end)
   end
 
+  setup :verify_on_exit!
+
   describe "installation_settings" do
     # TODO: remove InstallationSettings since premium does not exist anymore?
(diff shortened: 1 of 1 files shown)


test/trento/settings_test.exs
```diff
@@ -30,6 +30,8 @@ defmodule Trento.SettingsTest do
     on_exit(fn -> Application.put_env(:trento, :flavor, "Community") end)
   end
 
+  setup :verify_on_exit!
```
nelsonkopliku, line 33: Good catch! Let's just place this before the `setup` maybe

nelsonkopliku (approved):
Awesome, thanks!

test/trento/settings_test.exs
skrech: Moved it before the other setup.

PR #3360, opened 2025-03-04
Operation completed activity log

arbulu89:
# Description

Add `operation_completed` event activity log entry.
This is a new event type for the activity log. I have named it `queue_event`, as the event is received in the rabbitmq queue.
Basically it replicates what we do with connection and domain events.
It includes its own parser.

In this particular case, and to avoid sending the user/actor to wanda (it doesn't really need it), I'm simply querying the `operation_requested` activity log to get the actor. The both share the same operation id, and this is a UUID, so the unique possible return will be the request for this operation.
If for some reason, the request to wanda doesn't go through web, we return `system` as actor. At the end, we are logging `web` things.

image

## How was this tested?
UT

### DIFF
--- a/lib/trento/activity_logging/parser/queue_event_parser.ex
+++ b/lib/trento/activity_logging/parser/queue_event_parser.ex
@@ -1,0 +1,39 @@
+defmodule Trento.ActivityLog.Logger.Parser.QueueEventParser do
+  @moduledoc """
+  Queue event activity parser
+  """
+
+  alias Trento.Operations.V1.OperationCompleted
+
+  alias Trento.Infrastructure.Operations
+
+  alias Trento.ActivityLog
+
+  def get_activity_actor(:operation_completed, %OperationCompleted{operation_id: operation_id}) do
+    case ActivityLog.list_activity_log(%{type: "operation_requested", search: operation_id}) do
+      {:ok, [%{actor: actor}], _meta} -> actor
+      _ -> "system"
+    end
+  end
+
+  def get_activity_actor(_, _), do: nil
+
+  def get_activity_metadata(
+        :operation_completed,
+        %OperationCompleted{
+          operation_id: operation_id,
+          group_id: group_id,
+          operation_type: operation_type,
+          result: result
+        }
+      ) do
+    %{
+      resource_id: group_id,
+      operation: Operations.map_operation_type(operation_type),
+      operation_id: operation_id,
+      result: result
+    }
+  end
+
+  def get_activity_metadata(_, _), do: %{}
+end
--- a/assets/js/lib/model/activityLog.js
+++ b/assets/js/lib/model/activityLog.js
@@ -115,6 +115,7 @@ export const DATABASE_TOMBSTONED = 'database_tombstoned';
 
 // Operations
 export const OPERATION_REQUESTED = 'operation_requested';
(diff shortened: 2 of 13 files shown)


lib/trento/activity_logging/parser/queue_event_parser.ex
```diff
@@ -0,0 +1,39 @@
...
+  alias Trento.Infrastructure.Operations
+
+  alias Trento.ActivityLog
+
+  def get_activity_actor(:operation_completed, %OperationCompleted{operation_id: operation_id}) do
+    case ActivityLog.list_activity_log(%{type: "operation_requested", search: operation_id}) do
+      {:ok, [%{actor: actor}], _meta} -> actor
+      _ -> "system"
+    end
+  end
```
gagandeepb, line 17: Pattern matching on the number of elements in a list can be tricky/error-prone here, in my understanding. I am operating under the assumption that this function/ underlying query (`list_activity_log`) can return more than one result.

arbulu89: It should never return more than one result, as we are filtering by the `operation_id`, which is a UUID, so unique by definition.
I can add the tail of the list "for safety", but it shouldn't change anything.

nelsonkopliku (approved):
Hey @arbulu89 thanks for this, it's very nice!

Just a thought about the central part in `lib/trento/infrastructure/operations/amqp/processor.ex`
```elixir
event
  |> log_activity()
  |> handle()
```
Do you think it would make sense having a processor decorating all the others where it tries to log the message, if interesing, and then delegates to the actual processor.

Just thinking out loud as we will shortly add checks customization related messages.
Current implementation is just fine.

lib/trento/activity_logging/parser/queue_event_parser.ex
gagandeepb: I see. My understanding was that although the operation_id UUID is unique, we don't use it to uniquely identify write operations on the activity log. As I see it, there is some chance that we may still see some entries written more than once in some cases. Happy to sync, so I understand you better.

arbulu89: But the `operation_id` I use here is the `operation_id` that uniquely identifies a requested operation and it is part of its metadata, of a `operation_requested` entry. We should never have more than one operation with the same ID. We would have much more important issues than the activity log in that case

gagandeepb (approved):
Thanks for the clarification. LGTM.

arbulu89:
> Hey @arbulu89 thanks for this, it's very nice!
> 
> Just a thought about the central part in `lib/trento/infrastructure/operations/amqp/processor.ex`
> 
> ```elixir
> event
>   |> log_activity()
>   |> handle()
> ```
> 
> Do you think it would make sense having a processor decorating all the others where it tries to log the message, if interesing, and then delegates to the actual processor.
> 
> Just thinking out loud as we will shortly add checks customization related messages. Current implementation is just fine.

We could explore that option, but considering that we have only 2 processors, and that we need to run the `Contracts.from_event` function before even trying to log, or you put the current `process` function in the macro, or I don't see how this would work.
Besides that, even though the logging doesn't look like an intensive work, I don't think we should add it by default. If some processor messages don't need any logging, better not to add it. Less chances to introduce bugs or problems.

Anyway, we will see once we have the problem.

PR #3363, opened 2025-03-05
Center severity icon in table cell

nelsonkopliku:
# Description

OCD: centering the severity icon in the table cell 😅


gagandeepb (commented):
The following screenshot from this [PR's env](https://3363.prenv.trento.suse.com/activity_log?severity=info&first=20) is the reason I had switched to `w-auto`. Maybe thPR's envere is some other (third?) alternative that has not been considered yet?

image

nelsonkopliku:
Oh true @gagandeepb, thanks! Pushed a change. Let's see if that does the trick.

gagandeepb (approved):
Thanks for this. LGTM.

PR #3364, opened 2025-03-05
Operation policy messages

arbulu89:
# Description

Talking with @abravosuse , we agreed that having detailed information about why the operation is not authorized is a really good thing to have. So instead of returning a boolean in the authorization, this PR changes the code to return a list of forbidden errors if the operation is forbidden.
So the return now is a `:ok | {:error, [errors}` thing instead of a boolean.
In order to accumulate all the errors, I have create the `OperationsHelper.reduce_operation_authorizations` function. It is a pretty simple reduce usage that accumulates the errors.

The pattern matching gets a bit more difficult, as we cannot simply return false the first forbidden scenario, and we need to accumulate the errors, but it is not that much.

Now in the request api forbidden error, we return the list of the forbidden messages as well, after changing the plug.

Now, we can have fancy things like this in the frontend:
image

## How was this tested?

UT

### DIFF
--- a/lib/trento/operations/application_instance_policy.ex
+++ b/lib/trento/operations/application_instance_policy.ex
@@ -7,6 +7,8 @@ defmodule Trento.Operations.ApplicationInstancePolicy do
 
   require Trento.Enums.Health, as: Health
 
+  alias Trento.Support.OperationsHelper
+
   alias Trento.Clusters.Projections.ClusterReadModel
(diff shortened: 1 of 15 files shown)


CDimonaco (approved):
LGTM

I think it's a good idea to have this sorted like that, I was thinking of having a more rigid and defined structure directly in the error part of the views but it does not fit very well so having a dedicated module it's the best way to handle this.

PR #3368, opened 2025-03-06
Operation user authorization

arbulu89:
# Description
Authorize `request_operation` specific operations, `saptune_solution_apply` in this case.
It includes a new ability `saptune_solution_apply`.
We will have individual abilities always. Future examples `start:application_instance`, `stop:application_instance`, etc.
We will group them in the frontend by now as we do with other similar things.

I will remove the ability from the frontend temporarily in the next PR, based on frontend changes.

## How was this tested?

UT

### DIFF
--- a/lib/trento_web/controllers/v1/host_controller.ex
+++ b/lib/trento_web/controllers/v1/host_controller.ex
@@ -202,7 +202,7 @@ defmodule TrentoWeb.V1.HostController do
 
   def get_policy_resource(%{
         private: %{phoenix_action: :request_operation},
-        params: %{"operation" => operation}
+        path_params: %{"operation" => operation}
       }),
       do: %{operation: operation}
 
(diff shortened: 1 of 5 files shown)


EMaksy (approved):
Nicely done, lgtm, nothing to add!

lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -200,6 +200,12 @@ defmodule TrentoWeb.V1.HostController do
     end
   end
 
+  def get_policy_resource(%{
+        private: %{phoenix_action: :request_operation},
+        params: %{"operation" => operation}
```
CDimonaco, line 205: could we match on path_params instead of params, so we could match only the params we need and it's more understandable where we are getting the params

https://hexdocs.pm/plug/Plug.Conn.html

CDimonaco (approved):

lib/trento_web/controllers/v1/host_controller.ex
arbulu89: Alright, let me try that!

arbulu89: Yes, as easy as just replacing it. Changed!
Thank you!

PR #3369, opened 2025-03-06
Operation user authorization frontend

arbulu89:
# Description
Guard operations authorizations based on user abilities.
@abravosuse I have grouped the currently existing `saptune_solution_apply:host` ability to `operation:all`.
So we will display only the 2nd when we need to select the abilities. 
It is the same we do for others like cleanup. Otherwise, we would overflow the options when we have more abilities. In the future we can decide to remove the grouping, it is just a visualization thing. The abilities are still stored one by one.

Related to: https://github.com/trento-project/web/pull/3368

image

image

## How was this tested?

UT and storybook

### DIFF
--- a/assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.jsx
+++ b/assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { assign, find } from 'lodash';
+import { getFromConfig } from '@lib/config';
 
 import MultiSelect from '@common/MultiSelect';
 
@@ -48,9 +49,14 @@ const groupedAbilities = [
       'all:activity_logs_settings',
     ],
   },
+  {
+    ability: 'operation:all',
+    tooltip: 'Permits running operations in all resources',
+    groupAbilities: ['saptune_solution_apply:host'],
+  },
 ];
 
-const mapAbilities = (abilities) =>
+const mapAbilities = (abilities, operationsEnabled) =>
   abilities.reduce((acc, { id, name, resource, label }) => {
     const valueLabel = `${name}:${resource}`;
     const groupedAbility = find(groupedAbilities, ({ groupAbilities }) =>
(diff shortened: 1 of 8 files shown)


assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.jsx
```diff
@@ -82,14 +93,15 @@ function AbilitiesMultiSelect({
   userAbilities,
   placeholder,
   setAbilities,
+  operationsEnabled = getFromConfig('operationsEnabled'),
```
arbulu89, line 96: I'm just putting this here as a default value. 
Otherwise, I need to pass the variable at least in other 3 components, and it is really annoying.
It is just a temporary thing anyway, until we can remove this `operationsEnabled` flag

abravosuse (approved):
> I have grouped the currently existing saptune_solution_apply:host ability to operation:all.
> So we will display only the 2nd when we need to select the abilities.
> It is the same we do for others like cleanup. Otherwise, we would overflow the options > when we have more abilities. In the future we can decide to remove the grouping, it is just a visualization thing. The abilities are still stored one by one.

As discussed, this is good for a first approach. I will engage @jagabomb to see how we can enhance the UI to offer more granularity in the selection of permissions.

assets/js/common/OperationsButton/OperationsButton.jsx
```diff
@@ -11,8 +11,34 @@ import {
 import { EOS_MORE_VERT, EOS_LOADING_ANIMATED } from 'eos-icons-react';
 
 import Button from '@common/Button';
+import DisabledGuard from '@common/DisabledGuard';
 
-function OperationsButton({ operations }) {
+// The custom component is required to wrap in the DisabledGuard component
+// and pass the disabled value to the child
+function CustomMenuButton({ value, running, disabled, onClick }) {
```
EMaksy, line 18: What about moving to an own component  ?

assets/js/common/OperationsButton/OperationsButton.test.jsx
```diff
@@ -76,4 +94,27 @@ describe('OperationsButton', () => {
       .querySelector("[data-testid='eos-svg-component']");
     expect(svgEl).toBeInTheDocument();
   });
+
+  it('should forbid operation is the user does not have the correct abilities', async () => {
```
EMaksy, line 98: typo is --> if

EMaksy (approved):
Hey man besides a small typo and a suggestion nothing to add --> LGTM!

assets/js/common/OperationsButton/OperationsButton.jsx
arbulu89: I don't find it necessary. It is an internal technical decision and I don't want to export the component

PR #3371, opened 2025-03-06
Fix operation_completed activity log severity level

arbulu89:
# Description
Fix `operation_completed` activity log severity level usage.
I didn't know the values were lower cased.


gagandeepb (approved):
Let's monitor and see if is necessary to make values that are mapped to severity levels, case-insensitive on our end. I am not sure/fully aware if the sender of these values makes any guarantees about the case (upper/lower/mixed) of values sent.

PR #3372, opened 2025-03-06
Backend for User Analytics Settings

jagabomb:
# Description

**Update 28 March 2025**
This PR adds the capability to access and store an `analytics_opt_in` value in the db that retains the consent from each user for their preference regarding the collection of analytics data. This will then be configurable in the frontend in the next PR.

**Update 12 March 2025**
This PR will focus on the backend mostly with regards to feedback. A separate PR will be created for the frontend feedback.

~~This PR attempts to allow collection of Analytics when users have opted in on the Settings screen. This was the most logical starting point to enable Analytics since it gives the user the ability to opt-in or opt-out.~~

~~**Outstanding items for this PR**~~

~~1. Link to our Analytics Policy (Settings Modal)
2. Tests
3. Storybook for Analytics Settings Component
4. Potential Documentation?~~

~~**Future PRs**~~

~~1. Analytics Optin Modal on Home or After Initial Login
2. Adding variables like `installation_id`, `environment` and `version`  to the tracking events.
3. Using the correct Production vs Dev/Demo Posthog codes based on deployment.~~

## How was this tested?
- [ ] **TODO**

- [ ] **TODO?**

### DIFF
--- a/lib/trento/users/user.ex
+++ b/lib/trento/users/user.ex
@@ -40,6 +40,7 @@ defmodule Trento.Users.User do
     field :totp_enabled_at, :utc_datetime_usec
     field :totp_secret, EncryptedBinary, redact: true
     field :totp_last_used_at, :utc_datetime_usec
+    field :analytics_enabled_at, :utc_datetime_usec
     field :lock_version, :integer, default: 1
 
     many_to_many :abilities, Ability, join_through: UsersAbilities, unique: true
@@ -85,7 +86,7 @@ defmodule Trento.Users.User do
     |> pow_extension_changeset(attrs)
     |> validate_password()
     |> custom_fields_changeset(attrs)
-    |> cast(attrs, [:password_change_requested_at])
+    |> cast(attrs, [:password_change_requested_at, :analytics_enabled_at])
   end
 
   def totp_update_changeset(user, attrs) do
--- a/lib/trento_web/controllers/v1/users_json.ex
+++ b/lib/trento_web/controllers/v1/users_json.ex
@@ -16,6 +16,7 @@ defmodule TrentoWeb.V1.UsersJSON do
           password_change_requested_at: password_change_requested_at,
           user_identities: user_identities,
           totp_enabled_at: totp_enabled_at,
+          analytics_enabled_at: analytics_enabled_at,
           inserted_at: created_at,
           updated_at: updated_at
         }
@@ -30,6 +31,7 @@ defmodule TrentoWeb.V1.UsersJSON do
         idp_user: length(user_identities) > 0,
         password_change_requested_at: password_change_requested_at,
         totp_enabled_at: totp_enabled_at,
+        analytics_enabled_at: analytics_enabled_at,
         created_at: created_at,
         updated_at: updated_at
       }
--- a/lib/trento_web/controllers/v1/users_json.ex
+++ b/lib/trento_web/controllers/v1/users_json.ex
@@ -31,7 +31,7 @@ defmodule TrentoWeb.V1.UsersJSON do
         idp_user: length(user_identities) > 0,
         password_change_requested_at: password_change_requested_at,
         totp_enabled_at: totp_enabled_at,
-        analytics_enabled_at: analytics_enabled_at,
+        analytics_enabled: analytics_enabled_at != nil,
         created_at: created_at,
         updated_at: updated_at
       }
--- a/lib/trento/users.ex
+++ b/lib/trento/users.ex
@@ -93,7 +93,10 @@ defmodule Trento.Users do
     if username == admin_username() do
       {:error, :forbidden}
     else
-      updated_attrs = maybe_set_password_change_requested_at(attrs, true)
+      updated_attrs =
+        attrs
+        |> maybe_set_password_change_requested_at(true)
+        |> maybe_enable_analytics()
 
       user
       |> User.profile_update_changeset(updated_attrs)
@@ -249,6 +252,14 @@ defmodule Trento.Users do
   end
 
   defp maybe_set_password_change_requested_at(attrs, _), do: attrs
+
+  defp maybe_enable_analytics(%{analytics_enabled: true} = attrs) do
+    Map.put(attrs, :analytics_enabled_at, DateTime.utc_now())
+  end
+
+  defp maybe_enable_analytics(attrs) do
+    Map.put(attrs, :analytics_enabled_at, nil)
+  end
 
   defp insert_abilities_multi(multi, []), do: multi
 
--- a/lib/trento_web/controllers/v1/profile_json.ex
+++ b/lib/trento_web/controllers/v1/profile_json.ex
@@ -11,6 +11,7 @@ defmodule TrentoWeb.V1.ProfileJSON do
           password_change_requested_at: password_change_requested_at,
           totp_enabled_at: totp_enabled_at,
           user_identities: user_identities,
+          analytics_enabled_at: analytics_enabled_at,
           inserted_at: created_at,
           updated_at: updated_at
         }
@@ -24,6 +25,7 @@ defmodule TrentoWeb.V1.ProfileJSON do
         password_change_requested: password_change_requested_at != nil,
         totp_enabled: totp_enabled_at != nil,
         created_at: created_at,
+        analytics_enabled: analytics_enabled_at != nil,
         idp_user: length(user_identities) > 0,
         updated_at: updated_at
       }
--- a/lib/trento_web/openapi/v1/schema/user.ex
+++ b/lib/trento_web/openapi/v1/schema/user.ex
@@ -164,7 +164,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
           analytics_enabled: %Schema{
             type: :boolean,
             description: "Whether user analytics collection is enabled",
-            nullable: true
+            nullable: false
           }
         }
       },
@@ -199,6 +199,11 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
           password_confirmation: %Schema{
             type: :string,
             description: "User new password, should be the same as password field",
+            nullable: false
+          },
+          analytics_enabled: %Schema{
+            type: :boolean,
+            description: "Whether user analytics collection is enabled",
             nullable: false
           },
           abilities: AbilityCollection
@@ -238,6 +243,11 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             nullable: false
           },
           abilities: AbilityCollection,
+          analytics_enabled: %Schema{
+            type: :boolean,
+            description: "Whether user analytics collection is enabled",
+            nullable: false
+          },
           totp_disabled: %Schema{
             type: :boolean,
             description:
@@ -309,10 +319,9 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             description: "Date of user last update",
             nullable: true
           },
-          analytics_enabled_at: %Schema{
-            type: :string,
-            format: :"date-time",
-            description: "Date of analytics collection opt-in",
+          analytics_enabled: %Schema{
+            type: :boolean,
+            description: "Whether user analytics collection is enabled",
             nullable: false
           },
           totp_enabled_at: %OpenApiSpex.Schema{
--- a/lib/trento_web/openapi/v1/schema/user.ex
+++ b/lib/trento_web/openapi/v1/schema/user.ex
@@ -104,6 +104,11 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             description: "Password change is requested",
             nullable: false
           },
+          analytics_enabled: %Schema{
+            type: :boolean,
+            description: "Whether user analytics collection is enabled",
+            nullable: false
+          },
           totp_enabled: %Schema{
             type: :boolean,
             description: "TOTP is enabled",
@@ -155,6 +160,11 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             type: :string,
             description: "User new password, should be the same as password field",
             nullable: false
+          },
+          analytics_enabled: %Schema{
+            type: :boolean,
+            description: "Whether user analytics collection is enabled",
+            nullable: true
           }
         }
       },
@@ -298,6 +308,12 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             format: :"date-time",
             description: "Date of user last update",
             nullable: true
+          },
+          analytics_enabled_at: %Schema{
+            type: :string,
+            format: :"date-time",
+            description: "Date of analytics collection opt-in",
+            nullable: false
           },
           totp_enabled_at: %OpenApiSpex.Schema{
             type: :string,
--- a/lib/trento_web/openapi/v1/schema/user.ex
+++ b/lib/trento_web/openapi/v1/schema/user.ex
@@ -201,11 +201,6 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             description: "User new password, should be the same as password field",
             nullable: false
           },
-          analytics_enabled: %Schema{
-            type: :boolean,
-            description: "Whether user analytics collection is enabled",
-            nullable: false
-          },
           abilities: AbilityCollection
         },
         required: [:fullname, :email, :enabled, :password, :password_confirmation, :username]
@@ -243,11 +238,6 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             nullable: false
           },
           abilities: AbilityCollection,
-          analytics_enabled: %Schema{
-            type: :boolean,
-            description: "Whether user analytics collection is enabled",
-            nullable: false
-          },
           totp_disabled: %Schema{
             type: :boolean,
             description:
--- a/priv/repo/migrations/20250327095737_add_analytics_enabled_at_to_users.exs
+++ b/priv/repo/migrations/20250327095737_add_analytics_enabled_at_to_users.exs
@@ -1,0 +1,9 @@
+defmodule Trento.Repo.Migrations.AddAnalyticsEnabledAtToUsers do
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :analytics_enabled_at, :utc_datetime_usec
+    end
+  end
+end
--- a/test/support/factory.ex
+++ b/test/support/factory.ex
@@ -1203,7 +1203,8 @@ defmodule Trento.Factory do
       password_change_requested_at: nil,
       totp_enabled_at: nil,
       totp_secret: nil,
-      totp_last_used_at: nil
+      totp_last_used_at: nil,
+      analytics_enabled_at: nil
     }
   end
 
--- a/test/trento/users/user_test.exs
+++ b/test/trento/users/user_test.exs
@@ -47,5 +47,12 @@ defmodule Trento.Users.UsersTest do
       assert changeset.errors[:totp_enabled_at] ==
                {"is invalid", [validation: :inclusion, enum: [nil]]}
     end
+
+    test "changeset/2 validates analytics_enabled_at field is cast properly" do
+      changeset =
+        User.profile_update_changeset(%User{}, %{"analytics_enabled_at" => DateTime.utc_now()})
+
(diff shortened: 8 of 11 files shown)


abravosuse (approved):
Code apart, it looks good to me @jagabomb . Just one remark about documentation.

Not sure if you are referring to the upstream project documentation, the user-facing official documentation or both. I am referring to the official one. We will certainly have to include a reference to this new metrics collection in it, explaining its scope and purpose. But that addition is, in my opinion, beyond this PR. It should be handled via the standard documentation channels and involve Dmitri Popov, our official documentation gatekeeper.

jagabomb:
> Code apart, it looks good to me @jagabomb . Just one remark about documentation.
> 
> Not sure if you are referring to the upstream project documentation, the user-facing official documentation or both. I am referring to the official one. We will certainly have to include a reference to this new metrics collection in it, explaining its scope and purpose. But that addition is, in my opinion, beyond this PR. It should be handled via the standard documentation channels and involve Dmitri Popov, our official documentation gatekeeper.

Great, thanks for the clarification, @abravosuse . I was thinking about the points you raised when I created this PR. But as you say we can address the documentation related work in a future PR when we know the full extent of the UI metrics that we are collecting.

lib/trento/settings/analytics.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Trento.Settings.AnalyticsSettings do
+  @moduledoc """
+  Settings related to analytics.
+  """
+
+  use Ecto.Schema
+  use Trento.Support.Ecto.STI, sti_identifier: :installation_settings
```
arbulu89, line 7: Now we need to use the proper table. So use `analytics_settings`

lib/trento/settings/analytics.ex
```diff
@@ -0,0 +1,25 @@
+defmodule Trento.Settings.AnalyticsSettings do
```
arbulu89, line 1: Rename the file to `analytics_settings.ex`

lib/trento/settings/analytics.ex
```diff
@@ -0,0 +1,25 @@
...
+  use Ecto.Schema
+  use Trento.Support.Ecto.STI, sti_identifier: :installation_settings
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  schema "settings" do
+    field :analytics_optin, :boolean
```
arbulu89, line 15: Rename this to `optin`

lib/trento/settings/analytics.ex
```diff
@@ -0,0 +1,25 @@
...
+  """
+
+  use Ecto.Schema
+  use Trento.Support.Ecto.STI, sti_identifier: :installation_settings
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @primary_key {:id, :binary_id, autogenerate: true}
```
arbulu89, line 13: Most probably you want to add the jason encoding to avoid sending internals.
You can add `@derive {Jason.Encoder, except: [:__meta__, :__struct__]}` in top of the `primary_key` line

lib/trento/settings/analytics.ex
```diff
@@ -0,0 +1,25 @@
...
+  schema "settings" do
+    field :analytics_optin, :boolean
+    sti_fields()
+  end
+
+  @spec changeset(t() | Ecto.Changeset.t(), map) :: Ecto.Changeset.t()
+  def changeset(analytics_settings, attrs) do
+    analytics_settings
+    |> cast(attrs, [:analytics_optin])
```
arbulu89, line 22: Lets change this to:
```
analytics_settings
|> cast(attrs, [:optin])
|> validate_required([:optin]) # test this just in case, to see if it fails when false is giving, i hope it doesn't
|> sti_changes()
|> unique_constraint(:type)
```

This way, we validate all the required fields are coming and we only have one type of inherited table (you shouldn't be able to have 2, but just in case)

lib/trento/settings.ex
```diff
@@ -154,6 +155,38 @@ defmodule Trento.Settings do
     end
   end
 
+  # Analytics settings
+
+  @spec get_analytics_settings() ::
+          {:ok, AnalyticsSettings.t()} | {:error, :analytics_settings_not_configured}
+  def get_analytics_settings do
+    settings = Repo.one(AnalyticsSettings.base_query())
```
arbulu89, line 163: We can change this to:
```
case Repo.one(AnalyticsSettings.base_query()) do
    %AnalyticsSettings{} = settings -> {:ok, settings}
     nil -> {:error, :settings_not_configured}
end
```

lib/trento/settings.ex
```diff
@@ -154,6 +155,38 @@ defmodule Trento.Settings do
     end
   end
 
+  # Analytics settings
+
+  @spec get_analytics_settings() ::
+          {:ok, AnalyticsSettings.t()} | {:error, :analytics_settings_not_configured}
```
arbulu89, line 161: Change the return options to `{:error, :settings_not_configured}`.
If for some reason you need a specific value, you will need to handle this in the fallback controller as well.
I think the generic one is good enough

lib/trento/settings.ex
```diff
@@ -154,6 +155,38 @@ defmodule Trento.Settings do
...
+
+  @spec change_analytics_optin(boolean()) ::
+          {:ok, AnalyticsSettings.t()}
+          | {:error, :analytics_settings_not_configured}
+  def change_analytics_optin(value) do
+    case get_analytics_settings() do
+      {:ok, settings} ->
+        settings
+        |> AnalyticsSettings.changeset(%{
+          analytics_optin: value
```
arbulu89, line 180: Change this to `optin`

lib/trento_web/router.ex
```diff
@@ -192,6 +192,10 @@ defmodule TrentoWeb.Router do
         put "/activity_log", SettingsController, :update_activity_log_settings
         get "/activity_log", SettingsController, :get_activity_log_settings
 
+        # Analytics Settings
```
arbulu89, line 195: Let's protect this to go to prod by now:. Simply add the 2 new routes in a if:
```
if Application.compile_env!(:trento, :analytics)[:enabled] do
...
end
```

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -147,6 +147,44 @@ defmodule TrentoWeb.V1.SettingsController do
...
+      ok:
+        {"Analytics settings fetched successfully", "application/json",
+         Schema.Platform.AnalyticsSettings},
+      not_found: Schema.NotFound.response()
+    ]
+
+  def get_analytics_settings(conn, _) do
+    case Settings.get_analytics_settings() do
+      {:ok, settings} -> render(conn, :analytics_settings, %{settings: settings})
+      {:error, :analytics_settings_not_configured} -> {:error, :not_found}
```
arbulu89, line 163: Just put here the same error you put in the `settings.ex` `get_analytics_settings` function

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -147,6 +147,44 @@ defmodule TrentoWeb.V1.SettingsController do
...
+    end
+  end
+
+  operation :update_analytics_settings,
+    summary: "Updates the Analytics settings",
+    tags: ["Platform"],
+    request_body: {"AnalyticsSettings", "application/json", Schema.Platform.AnalyticsSettings},
```
arbulu89, line 170: I'm afraid this `request_body` cannot be the same as the response hehe.
You will need to create a new schema like `UpdateAnalyticsSettingsRequest` which will include the `optin` value.
By coincidence they match now, but we need 2 different schemas

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -147,6 +147,44 @@ defmodule TrentoWeb.V1.SettingsController do
...
+    responses: [
+      ok:
+        {"Analytics settings updated successfully", "application/json",
+         Schema.Platform.AnalyticsSettings},
+      unprocessable_entity: Schema.UnprocessableEntity.response()
+    ]
+
+  def update_analytics_settings(conn, _) do
+    %{analytics_optin: analytics_optin} = OpenApiSpex.body_params(conn)
```
arbulu89, line 179: Replace these to `optin`

lib/trento_web/openapi/v1/schema/platform.ex
```diff
@@ -84,6 +84,26 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Platform do
...
+
+    OpenApiSpex.schema(
+      %{
+        title: "AnalyticsSettings",
+        description: "Settings for Analytics",
+        type: :object,
+        additionalProperties: false,
+        properties: %{
+          analytics_optin: %Schema{
+            type: :boolean
```
arbulu89, line 98: Can we put a desciprion here?
```
description: "Whether platform analytics recollection is enabled"
```

arbulu89 (changes requested):
Alright @jagabomb 
A lot of comments, but good job in general.
Seeing now that we miss the testing as well, I think we should opt to leave the frontend part for other PR.

priv/repo/migrations/20250305100609_add_analytics_optin_field.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Trento.Repo.Migrations.AddAnalyticsOptinField do
+  use Ecto.Migration
+
+  def change do
+    alter(table(:settings)) do
+      add :analytics_optin, :boolean, default: false
```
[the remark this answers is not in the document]
jagabomb: I ended up renaming this to `:analytics_settings_opt_in` for consistency after seeing a Codespell issue

lib/trento/settings/analytics.ex
jagabomb: Renamed to `opt_in`

lib/trento_web/controllers/v1/settings_controller.ex
jagabomb: I ended up creating an `analytics_settings_not_configured ` in the Fallback Controller because the `settings_not_found` refers to SUMA settings only.

lib/trento_web/controllers/v1/settings_controller.ex
jagabomb: Cool, I have created an `AnalyticsSettingsUpdateRequest`.

lib/trento_web/openapi/v1/schema/platform.ex
jagabomb: This has been added.

lib/trento/settings.ex
jagabomb: This has been added to the fallback controller now because `:settings_not_configured` only refers to SUMA.

lib/trento_web/router.ex
jagabomb: Added

jagabomb:
FYI, I have found a bug which causes a loading loop when a user is not logged and it is trying to load the Analytics Settings from the API. I will need to find another solution for returning the Analytics Settings from the frontend for this scenario.

lib/trento/settings/analytics_settings.ex
```diff
@@ -0,0 +1,28 @@
...
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @derive {Jason.Encoder, except: [:__meta__, :__struct__]}
+  @primary_key {:id, :binary_id, autogenerate: true}
+  schema "settings" do
+    field :opt_in, :boolean, source: :analytics_settings_opt_in, default: false
+    sti_fields()
```
arbulu89, line 17: I think we can add the `timestamps(type: :utc_datetime_usec)` macro in top of `sti_fields` so we add the `created_at/updated_at` fields. It is a good to have thing.
Check other settings schemas for reference

lib/trento/settings.ex
```diff
@@ -154,6 +155,44 @@ defmodule Trento.Settings do
...
+    end
+  end
+
+  @spec change_analytics_opt_in(boolean()) ::
+          {:ok, AnalyticsSettings.t()}
+          | {:error, :analytics_settings_not_configured}
+  def change_analytics_opt_in(value) do
+    case Repo.one(AnalyticsSettings.base_query()) do
+      %AnalyticsSettings{} = settings ->
+        {:ok, settings}
```
arbulu89, line 175: You don't need this `{:ok, settings}`. You can simply remove it.

lib/trento_web/openapi/v1/schema/platform.ex
```diff
@@ -84,6 +84,48 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Platform do
...
+    )
+  end
+
+  defmodule AnalyticsSettingsUpdateRequest do
+    @moduledoc false
+
+    OpenApiSpex.schema(
+      %{
+        title: "AnalyticsSettingsUpdateRequest",
+        description: "Request body for analyrics settings update",
```
arbulu89, line 114: typo in `analyrics`

lib/trento/settings.ex
```diff
@@ -154,6 +155,44 @@ defmodule Trento.Settings do
...
+
+        settings
```
arbulu89, line 177: This update or insert usually is implemented in a different way. Basically, you don't use `update` and you use `insert` with the `on_conflicts` field.

```
 %AnalyticsSettings{}
|> AnalyticsSettings.changeset(%{
    opt_in: value
})
|> Repo.insert(conflict_target: :id, on_conflict, {:replace, [:opt_in]})
```

Anyway, I don't like much the `insert if not found` idea.

assets/js/lib/analytics/index.js
```diff
@@ -1,22 +1,39 @@
...
 
 export const capture = (event, payload) => {
   if (!analyticsEnabled) {
     return;
   }
   posthog.capture(event, { ...payload, installationID });
 };
+
+// Check if Analytics is enabled
+if (analyticsEnabled) {
```
EMaksy, line 27: This will be always executed when the file is imported, normally it is better to define functions to control the execution.
What do you think ?

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -330,6 +347,31 @@ function SettingsPage() {
...
+        <SettingsLoader
+          sectionName="Analytics"
+          status={calculateSettingsLoaderStatus(
+            analyticsSettingsLoading,
+            analyticsSettingsFetchError
+          )}
+          onRetry={() => fetchAnalyticsSettings()}
+        >
+          <AnalyticsConfig
+            analyticsOptin={analyticsSettings.opt_in}
```
EMaksy, line 361: analyticsOptin --> analyticsOptIn ?

EMaksy (changes requested):
Hey @jagabomb, you did great but you also took a hard task :). 

Maybe for the next time try to split the pr in multiple little pr's this would help you a lot to keep an overview.
One for the front-end and the other one for the back-end. So you can include stories and tests.
I left some comments with a bit more focus on the front end structure so maybe we can go step by step.
Overall still amazing even if we need to adjust some things :+1:

jagabomb:
> Hey @jagabomb, you did great but you also took a hard task :).
> 
> Maybe for the next time try to split the pr in multiple little pr's this would help you a lot to keep an overview. One for the front-end and the other one for the back-end. So you can include stories and tests. I left some comments with a bit more focus on the front end structure so maybe we can go step by step. Overall still amazing even if we need to adjust some things 👍

Thanks for all the feedback... I will follow your advice and create a separate PR to address the frontend feedback.

jagabomb:
> FYI, I have found a bug which causes a loading loop when a user is not logged and it is trying to load the Analytics Settings from the API. I will need to find another solution for returning the Analytics Settings from the frontend for this scenario.

This has been resolved by creating another API endpoint which just returns the required analytics value. This way if a user can still login without the page loading loop that was being triggered before.

lib/trento/settings.ex
jagabomb: This has been removed.

lib/trento_web/openapi/v1/schema/platform.ex
jagabomb: Fixed

lib/trento/settings.ex
jagabomb: I have changed this to use the `Repo.insert` approach.

arbulu89: Let' use `Repo.update` @jagabomb 
It is more semantically and technically correct.
I just wanted to show you how you implement a ` insert or update` thing, but here we are not going to insert never, so using `update` is better in all the cases

lib/trento_web/controllers/v1/settings_controller.ex
arbulu89: But you are not using it. You are returning `{:error, :not_found}` here, so the entry you added in the fallback controller is not going to be used

lib/trento_web/router.ex
```diff
@@ -88,6 +88,11 @@ defmodule TrentoWeb.Router do
   scope "/api", TrentoWeb.V1 do
     pipe_through [:api, :api_v1]
     get "/public_keys", SettingsController, :get_public_keys
+
+    # Analytics Settings
```
arbulu89, line 92: Duplicated. Let's remove this addition

arbulu89: Why do we even want that? Either way, we need to change it. We must not have the same endpoint repeated twice. We will need to think how to put it but duplicating is a no-go

lib/trento/users/user.ex
```diff
@@ -86,6 +87,7 @@ defmodule Trento.Users.User do
     |> validate_password()
     |> custom_fields_changeset(attrs)
     |> cast(attrs, [:password_change_requested_at])
+    |> cast(attrs, [:analytics_opt_in])
```
arbulu89, line 90: We can put the new field together with the previous one, no need to add a new `cast`

arbulu89 (commented):
Hey @jagabomb ,
We might want to change the value to `analytics_enabled_at`, so we have more information about the field. This way, we would know when it was enabled. And it would be `nil` if disabled.
If you check, we do exactly the same with the rest of the fields like `enabled_at`, `password_update_requested_at`, etc.

For the frontend, we can return the boolean as you do know though, as we don't need the datetime. You can simply do in the json view code something like ```
analytics_enabled: analytics_enabled_at != nil
```

What do you think?

**Comment by jagabomb** [comment]
> Hey @jagabomb , We might want to change the value to `analytics_enabled_at`, so we have more information about the field. This way, we would know when it was enabled. And it would be `nil` if disabled. If you check, we do exactly the same with the rest of the fields like `enabled_at`, `password_update_requested_at`, etc.
> 
> For the frontend, we can return the boolean as you do know though, as we don't need the datetime. You can simply do in the json view code something like
> 
> ```
> analytics_enabled: analytics_enabled_at != nil
> ```
> 
> What do you think?

Cool, thanks for this Xabi, I completely forgot about that detail when I changed the tables. I will make the reverts on the next commit. 😅

File: lib/trento/users/user.ex (hunk 22 above)
Comment (jagabomb) [reply lib/trento/users/user.ex]: This has been resolved. Thanks! 😄

File: lib/trento_web/controllers/v1/users_json.ex [hunk 23]
```diff
@@ -30,6 +31,7 @@ defmodule TrentoWeb.V1.UsersJSON do
         idp_user: length(user_identities) > 0,
         password_change_requested_at: password_change_requested_at,
         totp_enabled_at: totp_enabled_at,
+        analytics_enabled_at: analytics_enabled_at,
```
arbulu89, line 34: Let's do like the other. `analytics_enabled: analytics_enabled_at != nil`

arbulu89 (commented):
Great @jagabomb !
I would use `analytics_enabled` as boolean in the users endpoint as well.
Besides that, all good.
We could merge once we add tests

arbulu89 (approved):
Great @jagabomb !
All good from my side

EMaksy (approved):
Beautiful @jagabomb !

jagabomb:
Merging this PR as per approval by Xabi and Eugen. These changes should not affect demo and release yet.

PR #3377, opened 2025-03-10
Add hostname field to activity log metadata for operation entries

arbulu89:
# Description

Add additional `hostname` field to host related operations (operation requested and completed).

image

## How was this tested?

UT

### DIFF
--- a/lib/trento/activity_logging/parser/metadata_enricher.ex
+++ b/lib/trento/activity_logging/parser/metadata_enricher.ex
@@ -86,7 +86,7 @@ defmodule Trento.ActivityLog.Logger.Parser.MetadataEnricher do
        when activity in [:resource_tagging, :resource_untagging],
        do: {:ok, resource_id}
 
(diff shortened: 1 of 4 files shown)


lib/trento/activity_logging/parser/metadata_enricher.ex
```diff
@@ -82,6 +86,9 @@ defmodule Trento.ActivityLog.Logger.Parser.MetadataEnricher do
        when activity in [:resource_tagging, :resource_untagging],
        do: {:ok, resource_id}
 
+  defp detect_enrichment(:host, {_, %{resource_id: id, operation: :saptune_solution_apply}}),
```
arbulu89, line 89: Using `operation` instead of the activity name itself, as we need to differenciate between host/cluster/sap system resource at some  point.
Either way, we will see how to pattern match once we have more operations

gagandeepb (approved):
LGTM.

nelsonkopliku (approved):
LGTM

PR #3393, opened 2025-03-19
Alerting Settings WebUI

skrech:
# Description

Work-stream for the front-end part of the Alerting Settings. 

Fixes [TRNT-2711](https://jira.suse.com/browse/TRNT-2711) [TRNT-3723](https://jira.suse.com/browse/TRNT-3723)

### DIFF
--- a/assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
+++ b/assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
@@ -3,8 +3,12 @@ import { noop } from 'lodash';
 
 import Button from '@common/Button';
 import DisabledGuard from '@common/DisabledGuard';
+import Tooltip from '@common/Tooltip';
 
 const alertingSettingsPermittedFor = ['all:alerting_settings'];
+
+export const ENFORCED_FROM_ENV_MESSAGE =
+  'Alerting settings are enforced by environment variables';
 
 export default function AlertingSettingsConfig({
   settings: {
@@ -14,6 +18,7 @@ export default function AlertingSettingsConfig({
     smtpUsername = '.....',
     senderEmail = '...@...',
     recipientEmail = '...@...',
+    enforcedFromEnv = false,
   } = {},
   userAbilities = [],
   onEditClick = noop,
@@ -28,14 +33,21 @@ export default function AlertingSettingsConfig({
             userAbilities={userAbilities}
             permitted={alertingSettingsPermittedFor}
           >
-            <Button
-              type="primary-white-fit"
-              size="small"
-              aria-label="alerting-edit-button"
-              onClick={onEditClick}
+            <Tooltip
+              isEnabled={enforcedFromEnv}
+              content={ENFORCED_FROM_ENV_MESSAGE}
+              place="bottom"
             >
-              Edit Settings
-            </Button>
+              <Button
+                type="primary-white-fit"
+                size="small"
+                aria-label="alerting-edit-button"
+                onClick={onEditClick}
+                disabled={enforcedFromEnv}
+              >
+                Edit Settings
+              </Button>
+            </Tooltip>
           </DisabledGuard>
         </span>
       </div>
--- a/assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
+++ b/assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
@@ -23,6 +23,47 @@ export default function AlertingSettingsConfig({
   userAbilities = [],
   onEditClick = noop,
 }) {
+  const alertingFields = [
+    {
+      label: 'SMTP Server',
+      value: smtpServer,
+      ariaLabel: 'smtp-server',
+      addClasses: 'truncate pr-12',
+    },
+    {
+      label: 'SMTP Port',
+      value: smtpPort,
+      ariaLabel: 'smtp-port',
+    },
+    {
+      label: 'SMTP User',
+      value: smtpUsername,
+      ariaLabel: 'smtp-username',
+      addClasses: 'truncate pr-12',
+    },
+    {
+      label: 'Password',
+      value: '•••••',
+      ariaLabel: 'smtp-password',
+    },
+    {
+      label: 'Alerting Sender',
+      value: senderEmail,
+      ariaLabel: 'alerting-sender',
+      addClasses: 'truncate pr-12',
+    },
+    {
+      label: 'Alerting Recipient',
+      value: recipientEmail,
+      ariaLabel: 'alerting-recipient',
+    },
+    {
+      label: 'Send Email Alerts',
+      value: alertingEnabled ? 'Enabled' : 'Disabled',
+      ariaLabel: 'alerting-enabled',
+    },
+  ];
+
   return (
     <div className="container max-w-7xl mx-auto p-4 sm:p-6 lg:p-8 bg-white dark:bg-gray-800 rounded-lg">
       <div>
@@ -56,60 +97,31 @@ export default function AlertingSettingsConfig({
         Settings used to send Trento system e-mails.
       </p>
 
-      <div className="grid grid-cols-6 mt-5 items-center">
-        <div className="font-bold mb-3">SMTP Server</div>
-        <div
-          aria-label="smtp-server"
-          className="col-span-2 text-gray-500 mb-3 truncate pr-12"
-        >
-          {smtpServer}
-        </div>
-
-        <div className="font-bold mb-3">SMTP Port</div>
-        <div aria-label="smtp-port" className="col-span-2 text-gray-500 mb-3">
-          {smtpPort}
-        </div>
-
-        <div className="font-bold mb-3">SMTP User</div>
-        <div
-          aria-label="smtp-username"
-          className="col-span-2 text-gray-500 mb-3 pr-12"
-        >
-          {smtpUsername}
-        </div>
-
-        <div className="font-bold mb-3">Password</div>
-        <div
-          aria-label="smtp-password"
-          className="col-span-2 text-gray-500 mb-3"
-        >
-          •••••
-        </div>
-
-        <div className="font-bold mb-3">Alerting Sender</div>
-        <div
-          aria-label="alerting-sender"
-          className="col-span-2 text-gray-500 mb-3 truncate pr-12"
-        >
-          {senderEmail}
-        </div>
-
-        <div className="font-bold mb-3">Alerting Recipient</div>
-        <div
-          aria-label="alerting-recipient"
-          className="col-span-2 text-gray-500 mb-3"
-        >
-          {recipientEmail}
-        </div>
-
-        <div className="font-bold mb-3">Send Email Alerts</div>
-        <div
-          aria-label="alerting-enabled"
-          className="col-span-2 text-gray-500 mb-3 truncate pr-12"
-        >
-          {alertingEnabled ? 'Enabled' : 'Disabled'}
-        </div>
+      <div className="grid grid-cols-6 gap-y-3 mt-5 items-center">
+        {alertingFields.map(({ label, value, ariaLabel, addClasses }) => (
+          <AlertingSettingsRow
+            key={ariaLabel}
+            label={label}
+            value={value}
+            ariaLabel={ariaLabel}
+            addClasses={addClasses}
+          />
+        ))}
       </div>
     </div>
+  );
+}
+
+function AlertingSettingsRow({ label, value, ariaLabel, addClasses = '' }) {
+  return (
+    <>
+      <div className="font-bold">{label}</div>
+      <div
+        aria-label={ariaLabel}
+        className={`col-span-2 text-gray-500 ${addClasses}`}
+      >
+        {value}
+      </div>
+    </>
   );
 }
--- a/assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
+++ b/assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
@@ -29,7 +29,6 @@ export default function AlertingSettingsConfig({
             permitted={alertingSettingsPermittedFor}
           >
             <Button
-              className="mr-2"
               type="primary-white-fit"
               size="small"
               aria-label="alerting-edit-button"
--- a/assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
+++ b/assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
@@ -1,0 +1,104 @@
+import React from 'react';
+import { noop } from 'lodash';
+
+import Button from '@common/Button';
+import DisabledGuard from '@common/DisabledGuard';
+
+const alertingSettingsPermittedFor = ['all:alerting_settings'];
+
+export default function AlertingSettingsConfig({
+  settings: {
+    alertingEnabled = false,
+    smtpServer = 'https://.....',
+    smtpPort = 587,
+    smtpUsername = '.....',
+    senderEmail = '...@...',
+    recipientEmail = '...@...',
+  } = {},
+  userAbilities = [],
+  onEditClick = noop,
+}) {
+  return (
+    <div className="container max-w-7xl mx-auto p-4 sm:p-6 lg:p-8 bg-white dark:bg-gray-800 rounded-lg">
+      <div>
+        <h2 className="text-2xl font-bold inline-block">Email Alerts</h2>
+
+        <span className="float-right">
+          <DisabledGuard
+            userAbilities={userAbilities}
+            permitted={alertingSettingsPermittedFor}
+          >
+            <Button
+              className="mr-2"
+              type="primary-white-fit"
+              size="small"
+              aria-label="alerting-edit-button"
+              onClick={onEditClick}
+            >
+              Edit Settings
+            </Button>
+          </DisabledGuard>
+        </span>
+      </div>
+
+      <p className="mt-3 mb-3 text-gray-500">
+        Settings used to send Trento system e-mails.
+      </p>
+
+      <div className="grid grid-cols-6 mt-5 items-center">
+        <div className="font-bold mb-3">SMTP Server</div>
+        <div
+          aria-label="smtp-server"
+          className="col-span-2 text-gray-500 mb-3 truncate pr-12"
+        >
+          {smtpServer}
+        </div>
+
+        <div className="font-bold mb-3">SMTP Port</div>
+        <div aria-label="smtp-port" className="col-span-2 text-gray-500 mb-3">
+          {smtpPort}
+        </div>
+
+        <div className="font-bold mb-3">SMTP User</div>
+        <div
+          aria-label="smtp-username"
+          className="col-span-2 text-gray-500 mb-3 pr-12"
+        >
+          {smtpUsername}
+        </div>
+
+        <div className="font-bold mb-3">Password</div>
+        <div
+          aria-label="smtp-password"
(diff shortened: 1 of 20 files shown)


nelsonkopliku (commented):
Hey @skrech, good job!
Before going deeper into the code I noticed a few visual/behavioral improvement opportunities.

<details>

<summary>Edit Settings button misalignment</summary>

We might align the `Edit Settings` button with the ones in the other sections
misalignment

</details>

<details>

<summary>Missing feedback when saving/updating settings fails</summary>

When incorrect data is provided and submitted, there is not visual feedback for the user. This might make it unclear what is going on.

Options we have are:
- a generic `An error occurred`
- a per field validation error message (see what happens in SMLM settings)

Additionally there's the `NaN` apperaring when non numbers are provided as the port. We might leave the API validate that and show the related validation error message

https://github.com/user-attachments/assets/905cad66-4cb9-4088-a00e-3211f1214395

</details>

<details>

<summary>Stale modal state</summary>

The changes made to the edit form that are not successfully applied (in case the API fails or the modal is closed before submission) are retained between edit modal openings.
For a better UX I believe we need to consider this cases and possibly clean up the state.

https://github.com/user-attachments/assets/58f8dabb-d484-4b18-a330-b0bd7107db5e

</details>

Happy to answer any question.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+    );
+  });
+
+  it('renders all visual props supplied as arugments', () => {
+    const alertingEnabled = true;
+    const smtpServer = faker.internet.url();
+    const smtpPort = faker.number.int({ max: 60000 });
+    const smtpUsername = faker.animal.cow();
+    const senderEmail = faker.internet.email();
+    const recipientEmail = faker.internet.email();
```
nelsonkopliku, line 42: question: what about having a dedicated factory for alerting settings?

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.test.jsx
```diff
@@ -0,0 +1,109 @@
...
+    expect(screen.getByLabelText('alerting-recipient')).toHaveTextContent(
+      recipientEmail
+    );
+    expect(screen.getByLabelText('alerting-enabled')).toHaveTextContent(
+      'Enabled'
+    );
+  });
+
+  it('forbids editing when called with default arguments', async () => {
+    var spy = jest.spyOn(lodash, 'noop');
```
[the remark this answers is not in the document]
skrech: Looking at the test, I don't know why I chose a `spy`, will change it to a simple func mock. I suspect the first implementation was different and the mocked operation was needed to execute.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
```diff
@@ -0,0 +1,103 @@
...
+          </DisabledGuard>
+        </span>
+      </div>
+
+      <p className="mt-3 mb-3 text-gray-500">
+        Settings used to send Trento system e-mails.
+      </p>
+
+      <div className="grid grid-cols-6 mt-5 items-center">
+        <div className="font-bold mb-3">SMTP Server</div>
```
EMaksy, line 48: Hey what about reducing repetition with a small reusable component? 

This pattern is repeated a lot: 
```
<div className="font-bold mb-3">{Label}</div>
<div className="col-span-2 text-gray-500 mb-3">{VARIABLE}</div>
```

Maybe we can create a component something in this direction....

```
const AlertSetingRow = ({ label, value, ariaLabel }) => (
  <>
    <div className="font-bold mb-3">{label}</div>
    <div
      aria-label={ariaLabel}
      className="col-span-2 text-gray-500 mb-3 truncate pr-12"
    >
      {value}
    </div>
  </>
);
```

Then you could reuse
``<AlertSetingRow label="SMTP Server" value={smtpServer} ariaLabel="smtp-server" />``


Also consider using a const 

```
const fields = [
  { label: 'SMTP Server', value: smtpServer, ariaLabel: 'smtp-server' },
  <<All other values>>
  {
    label: 'Send Email Alerts',
    value: alertingEnabled ? 'Enabled' : 'Disabled',
    ariaLabel: 'alerting-enabled',
  },
];
```

You could use map to render everything inside this config. 

Soomething like this ? 
```
<div className="grid grid-cols-6 mt-5 items-center">
  {fields.map(({ label, value, ariaLabel }) => (
    <AlertSettingRow
      key={ariaLabel}
      label={label}
      value={value}
      ariaLabel={ariaLabel}
    />
  ))}
</div>
```
With this you would have less code, easy to maintain and extend :) 


What do you think ?

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.test.jsx
EMaksy: +1 to add a factory

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -299,36 +319,70 @@ function SettingsPage() {
...
+              userAbilities={abilities}
+              onEditClick={() => {
+                clearAlertingSubmitErrors()
+                setAlertingSettingsModalOpen(true);
+              }}
+            />
+          </SettingsLoader>
+
+          <AlertingSettingsModal
+            key={`alertingmodal-${Object.values(alertingSettings).join("-")}-${alertingSettingsModalOpen}`}
```
EMaksy, line 377: Maybe there is a better alternative then using Object.values as key ? It may cause unnecessary remounts, because of the unpredictable ordering and also requires some performance for the key calculation.
Something more like https://github.com/trento-project/web/blob/7dce2f71f5598fedfb61938786cffc1e86a7348a/assets/js/pages/SettingsPage/SettingsPage.jsx#L298. 

Also the readability is harder like this.

What do you think  ?

EMaksy (commented):
Hey  @skrech  nice work, left some ideas, tell me what you think :)

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
```diff
@@ -0,0 +1,104 @@
+import React from 'react';
+import { noop } from 'lodash';
+
+import Button from '@common/Button';
+import DisabledGuard from '@common/DisabledGuard';
+
+const alertingSettingsPermittedFor = ['all:alerting_settings'];
```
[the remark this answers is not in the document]
skrech: Addressed.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.test.jsx
skrech: Created factories and put the in use in the AlertingSettingsConfig and AlertingSettingsModal tests.

skrech:
> Hey @skrech, good job! Before going deeper into the code I noticed a few visual/behavioral improvement opportunities.
> Edit Settings button misalignment
> Missing feedback when saving/updating settings fails
> Stale modal state
> 
> Happy to answer any question.

Thank you, Nelson.
All of these UX issues are addressed.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
skrech: Thanks Eugen, looks like a good idea, I'm only concerned  about the CSS values because I know that they differ in some cases (last rows, if I recall correctly). Anyway, I think I know how to handle that. :)

assets/js/pages/SettingsPage/SettingsPage.jsx
skrech: Hmm... You mean to enumerate all the alerting settings fields (which are 7, and +1 for modality) here? 

I decided to use some dynamism here especially  to help on the readability front. 

As to the unnecessary re-mounts -- I don't think they are expected. I'm far from JS expert, but usually, in other languages, map keys are not guaranteed to be in order but should be consistent/`not change` once they are put into the map.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
EMaksy: You also could pass css as prop and give even more control :) 
The repeated css could be a default value, and you pass only if it differs or how ever you like. 
I'm sure you will find a good solution :)

nelsonkopliku (commented):
Hey @skrech very nice!

The main thing we need to address are the missing tests. For the rest it does not seem that much is missing.

There's something we might want to improve with openapi validation like making the port an integer only and adding a `format: :email` to email.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.test.jsx
```diff
@@ -0,0 +1,119 @@
...
+
+import AlertingSettingsConfig, {
+  ENFORCED_FROM_ENV_MESSAGE,
+} from './AlertingSettingsConfig';
+
+const adminUser = [{ name: 'all', resource: 'all' }];
+
+describe('AlertingSettingsConfig', () => {
+  beforeAll(() => {
+    jest.restoreAllMocks();
```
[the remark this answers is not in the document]
skrech: Good catch, no it's not.

assets/js/lib/test-utils/factories/alertingSettings.js
```diff
@@ -0,0 +1,19 @@
...
+  smtpUsername: faker.internet.username(),
+  senderEmail: faker.internet.email(),
+  recipientEmail: faker.internet.email(),
+}));
+
+export const alertingSettingsFactory = baseAlertingSettingsFactory.params({
+  enforcedFromEnv: false,
+});
+
+export const alertingSettingsSendFactory = baseAlertingSettingsFactory.params({
```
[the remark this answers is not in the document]
skrech: Yes, I had hard time coming up with a name for that. I didn't want to have 'request' in the name but since there is only one usage for that factory now, I'll go with your suggestion.

assets/js/common/AlertingSettingsModal/AlertingSettingsModal.jsx
```diff
@@ -0,0 +1,259 @@
...
+          </div>
+
+          <Label
+            className="col-span-2"
+            htmlFor="recipient-email-input"
+            required
+          >
+            Alert Recipient
+          </Label>
+          <div className="col-span-4">
```
EMaksy, line 225: Nitpick: 
This strucutre repeats more or less 7 times

```
<Label className="col-span-2" htmlFor="field-id" required>
  Field Label
</Label>
<div className="col-span-4">
  <Input
    id="field-id"
    value={...}
    onChange={...}
    error={...}
    ...
  />
  {error && <p className="text-red-500">...</p>}
</div>
```

We rely quite a lot on this structure also in other Input Modals like in SuseManagerSettingsModal. 

It's out of scope for this PR, but we could consider extracting it into a reusable FormField component. Might be worth tracking as tech debt what do you think?

assets/js/pages/SettingsPage/SettingsPage.jsx
```diff
@@ -299,36 +320,70 @@ function SettingsPage() {
...
+              userAbilities={abilities}
+              onEditClick={() => {
+                clearAlertingSubmitErrors();
+                setAlertingSettingsModalOpen(true);
+              }}
+            />
+          </SettingsLoader>
+
+          <AlertingSettingsModal
+            key={`alertingmodal-${Object.values(alertingSettings).join('-')}-${alertingSettingsModalOpen}`}
```
EMaksy, line 378: After having a look, do we really need the key here ? 
I think the use effect mounts properly and we can remove the key. What do you think ?

EMaksy (commented):
You did a great progress!
Just some more nitpicks and ideas.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
skrech: Done

assets/js/pages/SettingsPage/SettingsPage.test.jsx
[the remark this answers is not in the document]
skrech: Added these tests.

assets/js/pages/SettingsPage/hooks.js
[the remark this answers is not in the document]
skrech: Added these tests

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
```diff
@@ -0,0 +1,127 @@
...
+            ariaLabel={ariaLabel}
+            addClasses={addClasses}
+          />
+        ))}
+      </div>
+    </div>
+  );
+}
+
+function AlertingSettingsRow({ label, value, ariaLabel, addClasses = '' }) {
```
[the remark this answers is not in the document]
skrech: To me this AlertingSettingsRow feels like supplementary component, just to lower the duplication. I don't think it needs to be extracted any futher. Additionally, it won't be used in any other component.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
```diff
@@ -0,0 +1,127 @@
...
+import DisabledGuard from '@common/DisabledGuard';
+import Tooltip from '@common/Tooltip';
+
+const alertingSettingsPermittedFor = ['all:alerting_settings'];
+
+export const ENFORCED_FROM_ENV_MESSAGE =
+  'Alerting settings are enforced by environment variables';
+
+export default function AlertingSettingsConfig({
+  settings: {
```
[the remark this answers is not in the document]
skrech: For the destructuring part: I personally feel that leaving the destructuring in the function signature speaks more. In addition, I tried to make a quick search on the Internet on style guides and found this: https://github.com/airbnb/javascript?tab=readme-ov-file#destructuring which seems to classify destructuring in signatures as "best". Ultimately, this seems like a style/preferences type of question, maybe I could leave it like that?

For the second remark: It would also fail if we supply `1` or `"settings"`. For me, `undefined` (eg. not supplied at all) is the only case worth handling apart from providing the expected object. `undefined` is handled already in the function signature destructuring.

assets/js/common/AlertingSettingsModal/AlertingSettingsModal.jsx
skrech: I see your point. However, I lean more on the side of leaving this as it is now. Although repeated several times, these elements are fairly simple -- Label combined with an Input. Most of the code would have to be supplied as an argument, anyway. Almost half of the code is CSS fields that are concretely related to our current layout. The error part is a good candidate for "bundling" all these elements into their own component... but the current implementation depends on naming convention (for the error fields) that is different than the convention for JS variable names, meaning that we have to pass in these values separately.

All in all, we'll have fairly simple composite component that has a function signature accepting 10-20 arguments and at the same time be very rigid, working in only this specific case. I think, it's more beneficial to leave such elements as separate components which gives them more flexibility.

assets/js/pages/SettingsPage/SettingsPage.jsx
skrech: To be honest, I'm not entirely sure.
I think we need this, since  we are using the values in `alertingSettings` as default values in `useState` expressions in `AlertingSettingsModal`. However, changing these values is only possible if we modify settings via the modal, in which case it would close itself after successful change.
Maybe the only thing needed is the `alertingSettingsModalOpen`. I'll try it now.

assets/js/common/AlertingSettingsModal/AlertingSettingsModal.jsx
```diff
@@ -0,0 +1,259 @@
...
+              }}
+            />
+          </div>
+
+          <Label className="col-span-2" htmlFor="smtp-server-input" required>
+            SMTP Server
+          </Label>
+          <div className="col-span-4">
+            <Input
+              id="smtp-server-input"
```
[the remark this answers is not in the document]
skrech: To me, it seems too protective. For example, `smtp-sever-input` is used only 3 time. Two times for the bidirectinal HTML label attachment and as NAME for the input element. The NAME is the same only because I don't have a better name to put. :D It'd be fine for ID and NAME to be different.

skrech:
I think I've addressed every code-related issue and responded to every remark. So, requesting re-review. :)

All the changes are into their own commit for easier reviewing.

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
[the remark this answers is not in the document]
nelsonkopliku: nothing against destructuring in function signature

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.jsx
```diff
@@ -0,0 +1,127 @@
...
+    alertingEnabled = false,
+    smtpServer = 'https://.....',
+    smtpPort = 587,
+    smtpUsername = '.....',
+    senderEmail = '...@...',
+    recipientEmail = '...@...',
+    enforcedFromEnv = false,
+  } = {},
+  userAbilities = [],
```
nelsonkopliku, line 23: Not for this particular case but keep in mind this https://github.com/facebook/react/issues/18123#issuecomment-591978544

TL;DR: if an array prop is used inside a `useEffect` we'll have to define the empty array as a const

assets/js/pages/SettingsPage/hooks.test.jsx
```diff
@@ -319,3 +327,144 @@ describe('useApiKeySettings', () => {
...
+      pick([
+        'enabled',
+        'smtp_server',
+        'smtp_port',
+        'smtp_username',
+        'sender_email',
+        'recipient_email',
+      ]),
+      set('enforced_from_env', false)
+    )(alertingData);
```
nelsonkopliku, line 360: Besides naming not being immediately clear, to me at least, I am wondering: since these two functions are always called with the same constant arguments, couldn't they be just two more constants? 👀 

Or am I missing something?

```suggestion
  const fooNeedsBetterName = flow(
      pick([
        'alertingEnabled',
        'smtpServer',
        'smtpPort',
        'smtpUsername',
        'senderEmail',
        'recipientEmail',
      ]),
      set('enforcedFromEnv', false)
    )(saveAlertingSettings);

  const barNeedsBetterName = flow(
      pick([
        'enabled',
        'smtp_server',
        'smtp_port',
        'smtp_username',
        'sender_email',
        'recipient_email',
      ]),
      set('enforced_from_env', false)
    )(saveAlertingData);
```

nelsonkopliku (approved):
Hey @skrech, well done!

There are still a few rough edges that might be polished especially related to user input to request transformation and related validation, but I believe we can push this forward and possibly address them later in smaller, laser focused PRs.

EDIT: don't worry about the failing **API bc check (V1)** step in the pipeline, we'll force merge

✉️ 🔔

assets/js/common/Button/Button.jsx
```diff
@@ -35,7 +35,15 @@ const getButtonClasses = (type) => {
...
 };
 
-function Button({ children, className, type, size, disabled, ...props }) {
+function Button({
+  children,
+  className,
+  type,
+  size,
+  disabled,
+  isSubmitButton,
```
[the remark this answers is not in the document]
skrech: Changed the name, we'll leave the test for later :(

PR #3394, opened 2025-03-19
Discovery HANA+ASCS/ERS ers clusters

arbulu89:
# Description

Improve the cluster type detection to avoid having HANA cluster types when the cluster is managing HANA+ASCS/ERS clusters.
Even though this setup is not recommended some users have it. In the other hand, Trento doesn't support it, so this change simply identifies it as a different type to avoid doing things as it was a Hana cluster (like executing checks or showing details).

Funnily enough, for some error, some of our test payloads already were combining HANA and ASCS/ERS resources. Some copy/paste error I guess

demo

PD: It breaks the backward compatibility, but I don't really want to bump an API version just for adding a new enum, specially when it was an incorrect value. What do you think?

## How was this tested?

UT and manual testing with payloads given by customer


lib/trento/discovery/payloads/cluster/cluster_discovery_payload.ex
```diff
@@ -69,39 +78,12 @@ defmodule Trento.Discovery.Payloads.Cluster.ClusterDiscoveryPayload do
...
-        Enum.any?(resources, fn %{"agent" => agent} ->
-          agent == "ocf::heartbeat:SAPInstance"
-        end)
-      end)
-
-    do_detect_cluster_type(sap_instance_count)
-  end
-
-  defp parse_cluster_type(%{"crmmon" => %{"clones" => clones}}) do
+  defp parse_hana_cluster_type(%{"crmmon" => %{"clones" => clones}}) do
```
arbulu89, line 86: The most of the changes are basically due change in order of the functions in the file.
Now, instead of returning the first found type, we parse HANA and ASCS/ERS types as we did before, but based in the two values, we have the possibility to have a `hana_ascs_ers` cluster.

lib/trento/discovery/payloads/cluster/cluster_discovery_payload.ex
```diff
@@ -177,25 +213,16 @@ defmodule Trento.Discovery.Payloads.Cluster.ClusterDiscoveryPayload do
...
 
   defp parse_cluster_additional_sids(%{
-         "cluster_type" => ClusterType.hana_scale_up(),
-         "cib" => %{"configuration" => %{"resources" => %{"primitives" => primitives}}}
-       }),
-       do: get_sapinstance_sids(primitives)
-
-  defp parse_cluster_additional_sids(%{
-         "cib" => %{"configuration" => %{"resources" => %{"clones" => nil, "groups" => groups}}}
+         "cib" => %{
```
arbulu89, line 216: I have simplified this function to get additional sids from resources and grouped resources.

abravosuse:
@arbulu89 in the PR environment [3394.prenv.trento.suse.com](https://3394.prenv.trento.suse.com/) I don't see the HANA+ASCS/ERS use case:

image

abravosuse:
Now I can see it, @arbulu89 . Thank you! It looks good to me.

lib/trento/discovery/payloads/cluster/cluster_discovery_payload.ex
```diff
@@ -69,39 +78,9 @@ defmodule Trento.Discovery.Payloads.Cluster.ClusterDiscoveryPayload do
...
-        Enum.any?(resources, fn %{"agent" => agent} ->
-          agent == "ocf::heartbeat:SAPInstance"
-        end)
-      end)
-
-    do_detect_cluster_type(sap_instance_count)
-  end
-
-  defp parse_cluster_type(%{"crmmon" => %{"clones" => clones}}) do
+  # The cluster manages a HANA system when it manages a SAPHanaTopoligy resource
```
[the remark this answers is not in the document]
arbulu89: Yes. Thank you!

PR #3397, opened 2025-03-20
Identify clustered sap instances

arbulu89:
# Description

Identify clustered SAP instances as right now, all SAP application and database instances running in a cluster host are identified as clustered instances, which is not always the case.

With this, we can be precise in the `ApplicationInstanceMoved` event, and only do it when the instance is clustered.

All the new logic is almost implemented in the policies code. Everything else is just side-effect work.

Now that we have a more detailed SAP instances data, I have removed the `sid` and `additional_sids` entries from all the code, except the frontend.

In the next PR I want to deprecate the `sid` and `additional_sids` elements from the API (keep them for backward compatibility) and send the new SAP instances to the frontend. This way, we can display properly if the instance is clustered as well.

## How was this tested?

UT, regression test, manual testing...


lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -669,7 +715,7 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
        }) do
     Enum.concat(
       primitives,
-      Enum.flat_map(clones, &Map.get(&1, :primitives, [])) ++
+      Enum.flat_map(clones, &(&1 |> Map.get(:primitive, []) |> List.wrap())) ++
```
arbulu89, line 718: This was a small bug we had. We didn't notice as we were not using cloned resources coming from this function

lib/trento_web/controllers/v2/cluster_json.ex
```diff
@@ -75,4 +79,11 @@ defmodule TrentoWeb.V2.ClusterJSON do
   defp adapt_resources(resources) do
     Enum.map(resources, &Map.drop(&1, [:parent]))
   end
+
+  defp adapt_sids(%{sap_instances: sap_instances} = cluster) do
+    cluster
+    |> Map.put(:sid, SapInstance.get_hana_instance_sid(sap_instances))
+    |> Map.put(:additional_sids, SapInstance.get_sap_instance_sids(sap_instances))
+    |> Map.delete(:sap_instances)
```
arbulu89, line 87: Removing the `sap_instances` by now.
I will work on the frontend part in a next implementation and see how to give the information to the frontend

abravosuse (approved):
Tested using http://ghcr.io/trento-project/trento-web:3397-env.

lib/trento/clusters/value_objects/sap_instance.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  deftype do
+    field :name, :string
+    field :sid, :string
+    field :instance_number, :string
+    field :hostname, :string
+    field :resource_type, Ecto.Enum, values: SapInstanceResourceType.values()
+  end
+
+  @spec get_hana_instance_sid([__MODULE__.t()]) :: String.t() | nil
```
nelsonkopliku, line 19: **question**: does this ever return nil? 
Asking because we're giving `""` as a default return value to `Enum.find_value` 👀

nelsonkopliku (commented):
hey @arbulu89 the code looks good. These kind of changes require possibly a lot of changes indeed 😄 and I trust the testing that has been made.

Question about the modified events: why are we changing legacy events as well? And did not changes to events require upcasting? 🤔

lib/trento/clusters/value_objects/sap_instance.ex
arbulu89: You are right. I didn't update the spec properly

arbulu89:
> Question about the modified events: why are we changing legacy events as well? And did not changes to events require upcasting?

Indeed. Changing legacy events was needed. I thought that the params would be superseded, but they are not, we simply redirect to the new module. Good catch.

About the upcasting, as the `embeds_many` defaults to an empty list, everything works without it. It is true that we lose some partial data though. I will implement the upcasting just for completeness, but the result would be the same due eventual consistency.

arbulu89:
Hey @nelsonkopliku ,
About the upcasting again. I have done some coding and adding "advanced" upcasting code looks unnecessary again. At the end, we just simply update some partial incomplete data in the aggregate, that we don't even use there (it is there only for projection purposes). And this data will be overriden 100% cases when the next payload arrives. So it is not used for anything.

I think that I will avoid adding not needed extra complexity/code and simply bump the `version` of the event, at least to be informative (because if there is not any upcasting function, it doesn't really do anything).
What do you think?

nelsonkopliku:
Sounds reasonable @arbulu89, thanks for looking into it.

lib/trento/clusters.ex
```diff
@@ -159,6 +161,21 @@ defmodule Trento.Clusters do
...
+      from c in ClusterReadModel,
+        join: h in HostReadModel,
+        on: h.cluster_id == c.id,
+        where: h.id == ^host_id,
+        select: c.sap_instances
+
+    case Repo.one(query) do
+      nil -> []
+      sap_instances -> sap_instances
+    end
```
gagandeepb, line 176: This will raise an exception if there's more than one result, IIRC.

gagandeepb (approved):
Just one comment/side-note.

lib/trento/clusters.ex
arbulu89: This is the `host_id`, and UUID and primary key in the database. It is impossible by database definition

PR #3400, opened 2025-03-21
Allow Usage Of Podman

janvhs:
This change allows the usage of podman and podman compose without having to set up custom aliases in /etc/containers/registries.conf{,.d}. Furthermore, hard coding "docker compose" in mix.exs is probably something one should invoke manually, anyway.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -134,7 +134,6 @@ defmodule Trento.MixProject do
   defp aliases do
     [
       start: [
-        "cmd docker-compose up -d",
         "deps.get",
         "ecto.create",
         "ecto.migrate",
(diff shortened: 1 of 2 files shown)


vicenteqa (approved):
lgtm!

mix.exs
```diff
@@ -134,7 +134,6 @@ defmodule Trento.MixProject do
   defp aliases do
     [
       start: [
-        "cmd docker-compose up -d",
```
balanza, line 137: Let's keep this command out. We don't have to bind the start command to a specific local environment

balanza (changes requested):

mix.exs
balanza: You are right, forget my comment

balanza (approved):

PR #3405, opened 2025-03-24
Use custom function to check if data is loaded

arbulu89:
# Description

The regression suite was broken. Most probably since we change photofinish behaviour.
This change give the option to decide if we need to load photofinish data twice or not before the test for each suite.

For example, for the databases suite, we check if sap systems are present or not.


vicenteqa (approved):
:+1:

PR #3408, opened 2025-03-25
Apply PageObject pattern sap systems overview e2e tests

vicenteqa:
# Description

- Apply page object pattern to sap systems overview tests
- Modified test fixture to make it consistent with the expected scenario and be able to use it for assertions.
- Deleted functions from commands file that are not used anymore. That file should disappear by the end of the refactor

Fixes # (issue)

- [ ] **DONE**

## How was this tested?

- [ ] **DONE**

- [ ] **DONE**


test/e2e/cypress/fixtures/sap-systems-overview/available_sap_systems.js
```diff
@@ -95,7 +95,7 @@ export const availableSAPSystems = [
         features: 'HDB|HDB_WORKER',
         instanceNumber: '10',
         systemReplication: HANASystemReplicationModes.Primary,
-        systemReplicationStatus: 'ACTIVE',
+        systemReplicationStatus: '',
```
nelsonkopliku, line 98: This means that a non empty `systemReplicationStatus` is only for hana Secondary, correct?

nelsonkopliku (approved):
LGTM

test/e2e/cypress/fixtures/sap-systems-overview/available_sap_systems.js
vicenteqa: tbh I made an assumption here, because the response received shows 'active' only for one of the instances, which I understood is the right thing to happen, but this test data to compare against wasn't matching that, so I did this modification also to be able to perform a more exact assertion for that value, and use an exact one rather than contain. Answering your question, it means that only one of the instances can be active, and with the data we load for testing, it is the secondary one in this case. Maybe this screenshot clarifies this.

Screenshot from 2025-03-31 15-45-35

PR #3411, opened 2025-03-26
Run regression and integration e2e tests always in CI

arbulu89:
# Description

As we discussed recently, tests that are not executed always in CI, are not executed.
This change makes the regression and integration E2E tests run always.
It shouldn't increase the time of the execution as they are executed in parallel


gagandeepb (approved):
Thanks for this. LGTM.

vicenteqa (approved):

PR #3412, opened 2025-03-26
Support AWS without metadata scenarios

nelsonkopliku:
# Description

This PR makes sure scenarios where AWS metadata is missing is properly handled.
Related to https://github.com/trento-project/agent/pull/414

Some of the cloud discovery fixtures with `"Metadata": {}` have been adjusted to what the agent actually returns in those cases `"Metadata": null`

### DIFF
--- a/assets/js/pages/HostDetailsPage/ProviderDetails.test.jsx
+++ b/assets/js/pages/HostDetailsPage/ProviderDetails.test.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
 import '@testing-library/jest-dom';
+
+import { AWS_PROVIDER } from '@lib/model';
 import ProviderDetails from './ProviderDetails';
 
 describe('Provider Details', () => {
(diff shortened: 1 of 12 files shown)


assets/js/pages/HostDetailsPage/ProviderDetails.test.jsx
```diff
@@ -43,4 +45,80 @@ describe('Provider Details', () => {
...
+      });
+    });
+
+    it('should render without AWS', () => {
+      render(<ProviderDetails provider={AWS_PROVIDER} provider_data={null} />);
+
+      expect(screen.getByText('AWS')).toBeVisible();
+
+      metadataLabels.forEach(({ _, label }) => {
+        expect(screen.queryByText(label)).toBeNull();
```
arbulu89, line 120: nitpick: maybe `expect(screen.queryByText(label)).not.toBeInTheDocument()` would be more descriptive

assets/js/pages/HostDetailsPage/ProviderDetails.test.jsx
```diff
@@ -43,4 +45,80 @@ describe('Provider Details', () => {
...
+        label: 'Region',
+      },
+      {
+        key: 'vpc_id',
+        label: 'VPC ID',
+      },
+    ];
+
+    it('should render with AWS metadata', () => {
+      const awsMetadata = {
```
arbulu89, line 87: nitpick: ideally we would have this in a factory. But well, I'm not going to complain if you don't. You did more than others adding the test at least hehe

arbulu89 (approved):
Good!
I'm fine not showing any field. I don't know if Alberto or Jurgen have other opinions.
Anyway, again, I'm fine as a bug fix.
Up to you if you want to ask them :see_no_evil:

assets/js/pages/HostDetailsPage/ProviderDetails.test.jsx
nelsonkopliku: Here we go 535b7dd2f9c2f76b40d2e9b9f8270c0568ed5359

nelsonkopliku:
@abravosuse @jagabomb FYI

With the latest changes to the aws metadata extraction process wrt [this bug](https://bugzilla.suse.com/show_bug.cgi?id=1239534) and this PR https://github.com/trento-project/agent/pull/414, it could happen that it is not possible to have metadata at all.

We thought valuable to still be able to discover the provider, so this change makes sure to gracefully handle the scenario where aws metadata are not available.

Without this change

image

With this change

image

This solution has been driven by what we currently do in other cases where we do not have metadata: for instance on KVM

image

abravosuse:
Thanks for the update @nelsonkopliku . Good for me.

jagabomb:
@nelsonkopliku This change makes sense. Thanks for bringing it to my attention.

PR #3422, opened 2025-03-31
Handle unmounted clustered sap instance

arbulu89:
# Description

Some customers use a pacemaker setup where the ERS resource is not mounted using a Filesystem or SAPStartSrv cluster resource.

This means that they have the ERS installed in both nodes of the cluster, and the cluster starts and stops it instead of moving it.

In this case, we need to avoid emitting the `ApplicationInstanceMoved` event.
To check that, we need to know if the SAP instance is mounted checking if the `Filesystem` or `SAPStartSrv` resources are managed by the cluster. 

Some info about simple mount configuration: https://documentation.suse.com/sbp/sap-15/html/SAP-S4HA10-setupguide-simplemount-sle15/index.html

## How was this tested?

UT
- `simple_mount` fixture includes a cluster configuration using the Simple mount setup. There we have the `SAPStartSrv` resource instead of the traditional `Filesystem` one
- `unmounted` fixture includes a cluster configuration where the ERS is not mounted with `Filesystem` nor `SAPStartSrv` resources. This means that the ERS is not moved as it is installed in both cluster nodes.

### DIFF
--- a/test/fixtures/discovery/ha_cluster_discovery_ascs_ers_simple_mount.json
+++ b/test/fixtures/discovery/ha_cluster_discovery_ascs_ers_simple_mount.json
@@ -1,0 +1,672 @@
+{
+  "agent_id": "4b30a6af-4b52-5bda-bccb-f2248a12c992",
+  "discovery_type": "ha_cluster_discovery",
+  "payload": {
+    "DC": false,
+    "Provider": "azure",
+    "Id": "8bca366a6cb7816555538092a1ddd5aa",
+    "Cib": {
+      "Configuration": {
+        "Nodes": [
+          {
+            "Id": "1",
+            "Uname": "vmnwprd01",
+            "InstanceAttributes": null
+          },
+          {
+            "Id": "2",
+            "Uname": "vmnwprd02",
+            "InstanceAttributes": null
+          }
+        ],
+        "CrmConfig": {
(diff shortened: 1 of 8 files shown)


test/fixtures/discovery/ha_cluster_discovery_ascs_ers_simple_mount.json
```diff
@@ -0,0 +1,672 @@
...
+                    {
+                      "Id": "rsc_ip_NWP_ASCS00-instance_attributes-ip",
+                      "Name": "ip",
+                      "Value": "10.80.1.25"
+                    }
+                  ]
+                },
+                {
+                  "Id": "rsc_SAPStartSrv_NWP_ASCS00",
+                  "Type": "SAPStartSrv",
```
arbulu89, line 92: Here we use a `SAPStartSrv` resource instead of a `Filesystem` resource

nelsonkopliku (approved):

test/fixtures/discovery/ha_cluster_discovery_ascs_ers_simple_mount.json
EMaksy: SAP is fun!

EMaksy (approved):

PR #3440, opened 2025-04-07
Log user coming from messages

nelsonkopliku:
# Description

This PR allows extracting user from messages so that proper actor can be added to logged activity when necessary.

Notes:
- clean up of deleted users' usernames is moved to users
- metadata passing through activity log function pipeline has been streamlined so that a `metadata` entry is used when necessary

need to re-bump contracts when https://github.com/trento-project/contracts/pull/97 is merged


lib/trento/users.ex
```diff
@@ -18,7 +18,7 @@ defmodule Trento.Users do
   @spec by_id(id :: non_neg_integer()) :: {:ok, User.t()} | {:error, :not_found}
   def by_id(id) do
     case Repo.get(User, id) do
-      %User{} = user -> {:ok, user}
+      %User{} = user -> {:ok, User.polish_username(user)}
```
arbulu89, line 21: question: are we sure of this? This `by_id` is used by Pow in many places and it could have some side-effect.
I find it safer to use `polish_name` in the `get_activity_actor` function.

arbulu89 (commented):
Hey @nelsonkopliku ,
I'm just a bit worried with the change of `by_id`. This function is used by `Pow` in multiple of places. Do we know if the change could affect somehow?
Maybe it is better to simply change the username in the activity log part

lib/trento/users.ex
nelsonkopliku: whoops, let me fix that 🙈

arbulu89: You are right. I confused the function with `get_by`. 
We have `by_id`, `get_by` (used by `Pow`), `get_user`...
I don't know why we have `by_id` and `get_user`... but well.
Yes, i would prefer a more informative name maybe. I wouldn't have specific things in general naming functions

arbulu89 (approved):
As long as Pow is not affected I'm fine.
I personally prefer not having such a specific getting functions and have enriching functions. Anyway, as said, I'm fine

PR #3445, opened 2025-04-09
Implements newly defined severity level conditions

gagandeepb:
... for existing log types.
Certain existing log types need to be remapped in a more conditional/payload dependent manner than was the case before. This PR does so according to feedback received else-where as to what those precise conditional mappings are.


abravosuse (approved):
I cannot speak for the code. I leave that part to @nelsonkopliku . But I can see that the code addresses the four activity types where I requested a behavior change. If Nelson approves, then merge it and I will update my rolling environment to test this and other pending changes.

nelsonkopliku (approved):

PR #3449, opened 2025-04-10
Analytics Settings UI with Collection to Posthog

jagabomb:
# Description

This PR adds the Analytics Opt-in (`analytics_enabled`) field to the Profile screen where it can be managed by a normal user. There is also a view included for the Edit User Page where it **only shows** the `analytics_enabled` value for a specific user. These field are only visible if a user has set the `analytics_enabled` in the `dev config file` to true.

Posthog events have also been enabled with this PR. This is only working if the user has opted in using the Profile screen. If they logout and use another user without the optin then no events will be captured. Once the previous user logs in again then the analytics will continue. The `username` and `installation_id` parameters are sent with every tracking event.

Upcoming PRs:
There is still the need to add more tracking data like the environment variables (development, demo, production). Then the we might need to look at the super admin user and how to opt-in for that user since this user cannot update their profile by default. Also we still might want an opt-in banner outside of profile so that the user is prompted to opt-in in a more obvious way.

**Profile Screen**
Screenshot 2025-04-10 at 15 31 26

**Edit User Screen (Super Admin)**
Screenshot 2025-04-10 at 15 32 12

**Posthog Test/Dev Events**
Screenshot 2025-04-10 at 15 34 51

## How was this tested?

- [ ] **TODO**

- [ ] **DONE**

### DIFF
--- a/assets/js/lib/analytics/index.js
+++ b/assets/js/lib/analytics/index.js
@@ -17,6 +17,7 @@ if (analyticsEnabledConfig) {
   });
 }
 
+// TODO: Remove this when this feature is ready for production
 export const getAnalyticsEnabledConfig = () => analyticsEnabledConfig;
 
 export const identify = (analyticsEnabled, userID) => {


abravosuse (changes requested):
Thanks for the clarifications on today's design review, @jagabomb . As discuss let's investigate how we can anonymize username data, as it conflicts with the way Trento itself handles this kind of data.

jagabomb:
Thanks for the feedback @abravosuse and @arbulu89. I am going to change the status of this to ready to review since I have applied the feedback and I finished creating and amending some tests.

assets/js/lib/analytics/index.test.js
```diff
@@ -9,4 +11,14 @@ describe('analytics config', () => {
 
     expect(getFromConfig('analyticsEnabled')).toBeTruthy();
   });
+
+  it('it should allow analytics opt-in to be configurable', () => {
+    expect(posthog.has_opted_out_capturing()).toBeFalsy();
```
arbulu89, line 16: Why is false at the beginning? The opt out is the default thing, right?

assets/js/lib/analytics/index.js
```diff
@@ -1,22 +1,53 @@
...
+  }
+
+  const distinctUserID = uuidv5(userID.toString(), installationID);
+  posthog.identify(distinctUserID, {
+    installationID,
+  });
+};
+
+export const optinCapturing = (analyticsEnabled) => {
+  if (analyticsEnabled) {
```
arbulu89, line 34: nitpick:
i think you can use both conditions in same if:
```
if (analyticsEnabled && posthog.has_opted_out_capturing()) {
    posthog.opt_in_capturing();
} else {
    posthog.opt_out_capturing();
}
```

assets/js/lib/analytics/index.js
```diff
@@ -1,22 +1,53 @@
...
+      posthog.opt_in_capturing();
+    }
+  } else posthog.opt_out_capturing();
+};
+
+export const capture = (analyticsEnabled, event, payload) => {
   if (!analyticsEnabled) {
     return;
   }
   posthog.capture(event, { ...payload, installationID });
```
arbulu89, line 45: I think we can remove the `installationID` from this `capture`. It is already associated to the user

assets/js/pages/Layout/Layout.jsx
```diff
@@ -93,12 +96,22 @@ function Layout() {
...
+    email,
+    analytics_enabled: analyticsEnabled,
+  } = useSelector(getUserProfile);
 
   const sidebarIconColor = 'currentColor';
   const sidebarIconClassName = 'text-gray-400 hover:text-gray-300';
   const sidebarIconSize = '24';
 
+  useEffect(() => {
+    identify(analyticsEnabled, id, username);
```
arbulu89, line 111: Let's remove the `username` from `identify` here. It is not used.

assets/js/pages/Profile/ProfileForm.jsx
```diff
@@ -180,6 +184,38 @@ function ProfileForm({
               disabled
             />
           </div>
+          {analyticsEnabledConfig && (
```
arbulu89, line 187: I think you can wrap both entries in the sama condition:
```
{analyticsEnabledConfig && (
<>
<Label className="col-start-1 col-span-2 pt-2">
    Analytics Opt-in
</Label>
<div className="col-start-3 col-span-4">
              <div className="pt-2">
...
</>
)}
```

I mean, you don't need 2 "if" conditions

assets/js/pages/Profile/ProfileForm.jsx
```diff
@@ -180,6 +184,38 @@ function ProfileForm({
...
+            <Label className="col-start-1 col-span-2 pt-2">
+              Analytics Opt-in
+            </Label>
+          )}
+          {analyticsEnabledConfig && (
+            <div className="col-start-3 col-span-4">
```
arbulu89, line 193: @jagabomb you are the UI/UX designed, but I couldn't stop noticing how off the input looks in the page :see_no_evil: 
Why don't se use a toggle and put the text in a icon with a tooltip?
Like we do with the `Authenticator app`. It would look more coesive

assets/js/pages/Profile/ProfileForm.jsx
```diff
@@ -180,6 +184,38 @@ function ProfileForm({
...
+              <div className="pt-2">
+                <div className="flex items-center">
+                  <Input
+                    className="mr-2"
+                    type="checkbox"
+                    role="checkbox"
+                    checked={analyticsEnabledState}
+                    onChange={() => setAnalyticsState(!analyticsEnabledState)}
```
arbulu89, line 201: This `onChange` usually is done this way:
```
onChange={(value) => setAnalyticsState(value)}

File: assets/js/lib/analytics/index.js [hunk 8]
```diff
@@ -1,22 +1,53 @@
...
   posthog.init(analyticsKey, {
     api_host: analyticsUrl,
+    opt_out_capturing_by_default: true,
     capture_pageview: false,
+    disable_persistence: true,
   });
 }
 
-export const capture = (event, payload) => {
+export const getAnalyticsEnabledConfig = () => analyticsEnabledConfig;
```
arbulu89, line 20: Let's maybe put a reminder comment saying to remove this function. It only is needed because the feature is not enabled yet. Or it is fine if you remember hehe

arbulu89 (commented):
Hey @jagabomb 
2 major things, and strangely enough, related to UX/UI XD:
- The `input` usage of the new field looks out of place for me. I think something like the `Authenticator app` looks better.
- The same in the user form. I wouldn't use a select for a thing that is disabled always

assets/js/pages/Profile/ProfileForm.jsx
jagabomb: Hey @arbulu89, the use of the `Input` (checkbox) is to make it clear that the Analytics Opt-in is a consent based feature, unlike Authenticator App that is an security enhancement so I think it could use the `Toggle` and `Tooltip` works better. I think we just want to be as transparent as possible regarding the collection of Analytics otherwise we hide the explanation in a Tooltip and hope that users understand and enable this. 😅 

Reference:
Screenshot 2025-05-13 at 10 43 34

With regards, to the overall aesthetic, yes I think it could look better but I realize that I need use a `Checkbox` component instead of the `Input` check box since the input component is limited in styling.

arbulu89: I just wanted to create the discussion XD
I move more towards the 2nd one, and you are right, it is due aesthetic reasons.
If you have stronger opinions about the 1st option because it raises the options of the user accepting, I accept that reasoning.
@abravosuse what about you?

assets/js/pages/Users/EditUserPage.jsx
```diff
@@ -28,6 +29,8 @@ function EditUserPage() {
   const [updatedByOther, setUpdatedByOther] = useState(false);
   const [abilitiesState, setAbilities] = useState([]);
 
+  const analyticsEnabledConfig = getAnalyticsEnabledConfig;
```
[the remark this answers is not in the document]
jagabomb: Hey @arbulu89, so I had this added here so that I could test for the values in the `UserForm.test.jsx` file. This way I could turn on the analytics_enabled in the config file and test for it. Do you think I should also remove the changes from the `UserForm.test.jsx` file?

assets/js/lib/analytics/index.test.js
jagabomb: I just added it in the beginning because I was wanted to confirm the initial value. I have removed this line since it is tested at the end anyway.

assets/js/lib/analytics/index.js
jagabomb: I have made this change and it seems to be working. 😄

assets/js/pages/Profile/ProfileForm.jsx
jagabomb: Thanks for this. I was wondering how to do this with using a empty div.

assets/js/pages/Users/UserForm.jsx
```diff
@@ -287,6 +290,25 @@ function UserForm({
...
+              )}
+              {analyticsEnabledConfig && (
+                <span className="col-start-3 col-span-4">
+                  <Select
+                    disabled
+                    className="w-full"
+                    optionsName="analytics"
+                    options={['Enabled', 'Disabled']}
+                    value={analyticsState ? 'Enabled' : 'Disabled'}
+                    onChange={(value) => {
```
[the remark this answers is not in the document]
jagabomb: I have changed this to a `Switch` component and removed the `onChange` function.

assets/js/pages/Profile/ProfileForm.jsx
jagabomb: I tried this again now, and I can't get it to work with this Input component. 😅

abravosuse:
Hey @jagabomb, in regards to the double Analytics Opt-in field, one with the check box, another one with the toggle, I think that if we include the text "Allow the collection of anonymous metrics to help improve Trento" in the toolbox of the info icon, the first field (the one with the check-box) becomes redundant and unnecessary. But if you strongly think otherwise, I trust you opinion as the UI/UX gatekeeper.

assets/js/pages/Profile/ProfileForm.jsx
jagabomb: Cool, as we discussed, this has now been changed to a `Switch` with the `Tooltip` option.

arbulu89 (commented):
Looks really good @jagabomb 
Just some final nitpick comments and we can merge :see_no_evil:

assets/js/pages/Profile/ProfileForm.jsx
```diff
@@ -20,6 +22,8 @@ function ProfileForm({
   totpSecret = '',
   totpQrData = '',
   abilities = [],
+  analyticsEnabledConfig = getAnalyticsEnabledConfig(),
```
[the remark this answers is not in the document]
arbulu89: I didn't actually meant that XD
I meant to put this value in the `ProfilePage` file, and from there, pass to the `ProfileForm` in the properties.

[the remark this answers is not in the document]
jagabomb: Ah yes, I ended up making the change that you just mentioned here. Sorry for the confusion. 😅

arbulu89 (approved):
Fantastic!
Really well done. Proud of this PR at the end :muscle:

PR #3450, opened 2025-04-10
Log checks selected as connection activity

nelsonkopliku:
# Description

This PR makes sure Checks selection activities are logged considering they're connection aspect so that the proper user is attached to the log entry.

Additionally both cluster and checks selections are now of warning severity.

### DIFF
--- a/lib/trento/activity_logging/activity_catalog.ex
+++ b/lib/trento/activity_logging/activity_catalog.ex
@@ -12,6 +12,11 @@ defmodule Trento.ActivityLog.ActivityCatalog do
 
   alias Trento.Operations.V1.OperationCompleted
 
+  @excluded_events [
+    Trento.Hosts.Events.HostChecksSelected,
+    Trento.Clusters.Events.ChecksSelected
+  ]
+
   @type activity_type :: atom()
   @type connection_activity :: {controller :: module(), action :: atom()}
   @type domain_event_activity :: event_module :: module()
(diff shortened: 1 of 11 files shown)


abravosuse (changes requested):
I've tested this using the PR environment and noticed a couple of issues:

1. The message isn the entries for cluster check selection is now "cluster_check_selected". It was "Checks were selected for cluster" before.
2. If you set filter Type = Cluster Checks Selected, entries for host check selection are also displayed.
I'm attaching a couple of screenshots to illustrate the behavior.

image

image

nelsonkopliku:
Hey @abravosuse, not sure what was going on the pr env when you tested there 👀 , and i cannot check it now because of an expired certificate.

Would you mind testing the image `ghcr.io/trento-project/trento-web:3450-env`? It might be more accurate

lib/trento/activity_logging/activity_catalog.ex
```diff
@@ -130,7 +135,7 @@ defmodule Trento.ActivityLog.ActivityCatalog do
             false
         end)
         |> Enum.map(&Module.concat/1)
-        |> Enum.filter(&(not &1.legacy?()))
+        |> Enum.filter(&(not &1.legacy?() and &1 not in @excluded_events))
```
gagandeepb, line 138: Suggestion to make it more readable (In my understanding the following is equivalent and we avoid the double negation first from the filter and then from the `not`):
```elixir
|> Enum.reject(&(&1.legacy?() or &1 in @excluded_events))
```

Reasoning for the above:
```
filter(&(not &1.legacy?() and &1 not in @excluded_events))
= filter(not &(&1.legacy?() or &1 not in @excluded_events))  [ since (~p /\ ~q) = ~(p \/ q)]
= reject(&(&1.legacy?() or &1 not in @excluded_events))    [not eliminated due to reject usage]
```

gagandeepb (approved):
I have just one comment/suggestion.

lib/trento/activity_logging/activity_catalog.ex
nelsonkopliku: Yup, better. Thanks da6ae356b6ed477bc86d500f026e6366722c288d

PR #3455, opened 2025-04-15
Add wanda to ci for e2e

EMaksy:
# Description

This pr aims to add wanda with checks to our ci run. We need wanda to be able to test checks customization with e2e tests. Insteat of cloning and rebulding wanda from scratch it is better just to take the docker images of web and wanda. With this it stays up2date. As we use wanda in demo mode, we can even test the mocked agent checks execution results


.github/workflows/ci.yaml
```diff
@@ -643,7 +643,9 @@ jobs:
       - name: "Docker compose dependencies"
         uses: isbang/compose-action@v2.1.0
         with:
-          compose-file: "./docker-compose.yaml"
+          compose-file: |
```
nelsonkopliku, line 646: We might want to start by adding this only to `test-e2e-parallel` and possibly if needed add it to
- `regression-test-e2e`
- `integration-test-e2e`

Besides this, now that I am looking closer, we might also consider using docker compose profiles and have all the services definitions in one compose file instead of the dedicated `wanda-docker-compose.yaml`.

I don't mind eitherway.

wanda-docker-compose.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+services:
+
+  postgres:
+    volumes:
+      - ./container_fixtures/postgres/init_wanda_db.sh:/docker-entrypoint-initdb.d/init_wanda_db.sh:ro
+  checks: 
+      image: ghcr.io/trento-project/checks:rolling
+      volumes: 
+        - /usr/share/trento/checks
```
nelsonkopliku, line 11: **ocd alert:** newline

wanda-docker-compose.yaml
```diff
@@ -0,0 +1,29 @@
...
+      ACCESS_TOKEN_ENC_SECRET: s2ZdE+3+ke1USHEJ5O45KT364KiXPYaB9cJPdH3p60t8yT0nkLexLBNw8TFSzC7k
+      CORS_ORIGIN: "http://localhost:4000"
+    depends_on:
+      - postgres
+      - rabbitmq
+      - checks
+    ports:
+      - 4001:4000
+    entrypoint: /bin/sh -c "/app/bin/wanda eval 'Wanda.Release.init()' && /app/bin/wanda start"
+    volumes_from: [checks]
```
nelsonkopliku, line 29: let's keep this as
```yaml
volumes_from:
    - checks
```

.github/workflows/ci.yaml
EMaksy: Hey Nelson, thanks i put only included wanda for test-e2e-parallel now, thanks for the tip. 

About using profiles vs wanda-docker-compose.yaml, yes we could also use profile, my only problem with this is we need to run a setup script for wanda, we would need an extra profile for postgres. Which is fine but the docker-compose file becomes quite huge. As this is mostly done for the ci only i prefer to split into seperate documents. For me personally it is just easier to read. At least for my eyes :)

No hard opinion here, just prefere the simplicity, but open for change :)

nelsonkopliku: I understand, no hard opinion here as well. I am fine eitherway.

wanda-docker-compose.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+  checks: 
+      image: ghcr.io/trento-project/checks:rolling
+      volumes:
+        - trento-checks:/usr/share/trento/checks
+
+  wanda:
+    image: ghcr.io/trento-project/trento-wanda:demo
+    environment:
+      DATABASE_URL: ecto://postgres:postgres@postgres/postgres
```
nelsonkopliku, line 17: Let's use the username/password/db created for wanda

```suggestion
      DATABASE_URL: ecto://wanda_user:wanda_pass@postgres/wanda_db
```

nelsonkopliku (approved):
Looks good!

container_fixtures/postgres/init_wanda_db.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+set -e
+
+psql -v ON_ERROR_STOP=1 --username "$POSTGRES_USER" <<-EOSQL
+  CREATE DATABASE wanda_db;
+  CREATE ROLE wanda_user WITH LOGIN PASSWORD 'wanda_pass';
```
arbulu89, line 6: do we need other extra user?
this is dev, I think we could reuse the `postgres` user. 
It is easier to debug with tools like `dbeaver`

arbulu89 (commented):
@EMaksy I actually think that we should use profiles. Adding wanda database even if we don't have the proper profile is not harmful.
I think it makes everything more usable and we use the tools provided by docker in native way.

arbulu89 (approved):
Nice @EMaksy !
It looks great!

nelsonkopliku (approved):
🚀

PR #3458, opened 2025-04-16
Recreate and not remove checks container in PR env ci

arbulu89:
# Description

Recreate checks container in every update and keep it in cleanup. Other PR environments depends on it.
Otherwise we can get errors like this one: https://github.com/trento-project/web/actions/runs/14493729333/attempts/1


nelsonkopliku (approved):
Thanks!

PR #3459, opened 2025-04-16
Add a GenericError for displaying non-recoverable errors

janvhs:
# Description

I fucked up the other PR (#3448) during rebase, so here we go again. This PR adds proper handling of non-recoverable errors, so the user does not see an empty page when SUMA fails to respond.

## How was this tested?

Updated test cases for AdvisoryDetailsPage and added GenericError tests.

N/A

### DIFF
--- a/assets/js/common/GenericError/GenericError.jsx
+++ b/assets/js/common/GenericError/GenericError.jsx
@@ -11,7 +11,7 @@ function GenericError({ message = undefined }) {
               Sorry
             </h1>
             <p className="font-light font-sans text-center lg:text-left text-lg mt-8 text-gray-700">
-              An error occoured
+              An error occurred
             </p>
             <p className="font-light font-mono text-center lg:text-left text-lg text-gray-500">
               {message ?? 'No further information is known at this point'}


assets/js/common/GenericError/GenericError.jsx
```diff
@@ -0,0 +1,29 @@
...
+        <div className="container mx-auto px-6 flex flex-col-reverse lg:flex-row justify-between items-center relative">
+          <div className="w-full mb-16 md:mb-8 text-center lg:text-left pr-10">
+            <h1 className="font-light font-sans text-center lg:text-left text-5xl lg:text-8xl mt-12 md:mt-0 text-gray-700 ">
+              Sorry
+            </h1>
+            <p className="font-light font-sans text-center lg:text-left text-lg mt-8 text-gray-700">
+              An error occurred
+            </p>
+            <p className="font-light font-mono text-center lg:text-left text-lg text-gray-500">
+              {message ?? 'No further information is known at this point'}
```
EMaksy, line 17: What about moving the plain text to a variable ? We should try to avoid nested strings :D
 ` const defaultMessage = 'No further information is known at this point';`
 
  So you can reuse it here: 
```
<p className="font-mono font-light text-lg text-gray-500">
  {message || defaultMessage}
</p>
```
or 
`function GenericError({ message = 'No further information is known at this point' }) {
  ......
}`

or even move outside the function body
```
const DEFAULT_ERROR_MESSAGE = 'No further information is known at this point';
function GenericError({ message }) {
  return (
              {message || DEFAULT_ERROR_MESSAGE}
  );
}
```

PR #3461, opened 2025-04-17
Alerting settings schema/model

skrech:
# Description

- Added DB schema for the Alerting Settings. It includes the fields expected to changed from future work on a web UI. 
- Added operations to the Settings context that would allow getting, idempotent setting and partial update of the newly created schema for alerting settings.
- Added authorization policy for alerting settings.
 
Fixes [TRNT-2711](https://jira.suse.com/browse/TRNT-2711)

## How was this tested?

Unit tests.

- Added factory function for the alerting settings
- Added tests for getting/setting/updating alerting settings from the DB (schema tests)
- Authorization policy tests


nelsonkopliku (commented):
Hey @skrech thanks for splitting the PR, this is way simpler to review.

Overall good job!

The main doubt I have is the intent of the operations exposed: is it gonna be one upsert operation or two distinct ones for initial saving and later updates?
Currently seems like we have both of them 👀 and I believe we need to clarify this, because it drives the behavior we implement.

When it comes to implementation details, take a look at what happens for other settings, many usecases have been already encountered and might be a good starting point, source of ideas (about how to do or how **_not_** to do something 😄 )

---

Other couple of mainly personal preference based comments:
- I personally do value clarity over brevity in naming (example `chset` vs `changeset`)
- I tend to find more compelling names like `apply/save settings` other than `set settings`

However these are secondary and not blocking.

lib/trento/settings/alerting_settings.ex
```diff
@@ -0,0 +1,85 @@
...
+    alerting_settings
+    |> cast(changes, [
+      :enabled,
+      :sender_email,
+      :recipient_email,
+      :smtp_server,
+      :smtp_port,
+      :smtp_username,
+      :smtp_password
+    ])
```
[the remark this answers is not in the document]
skrech: I'm sure that somewhere in the documentation of Ecto I saw recommendation to explicitly enumerate all the cast fields and avoid using "introspection". So, I'll go with the sigil :D

lib/trento/settings/policy.ex
```diff
@@ -40,4 +46,7 @@ defmodule Trento.Settings.Policy do
 
   defp has_suma_settings_change_ability?(user),
     do: user_has_ability?(user, %{name: "all", resource: "suma_settings"})
+
+  defp has_alerting_settings_resource_ability?(user),
+    do: user_has_ability?(user, %{name: "all", resource: "alerting_settings"})
```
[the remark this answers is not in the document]
skrech: Ugh, I really missed that, thanks!

test/trento/settings_test.exs
```diff
@@ -791,4 +792,139 @@ defmodule Trento.SettingsTest do
       assert certificates == Settings.get_sso_certificates()
     end
   end
+
+  describe "Alerting settings from DB" do
```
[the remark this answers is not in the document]
skrech: I left intentionally the "from DB" part because in the future PRs I'll need to distinguish tests for the case where settings come from the DB or from the ENV. I'll remove them  for this PR, though.

test/trento/settings_test.exs
```diff
@@ -791,4 +792,139 @@ defmodule Trento.SettingsTest do
...
+    end
+
+    test "return error when trying to save with wrong port value" do
+      settings = build(:alerting_settings, smtp_port: 70_000)
+
+      assert {:error,
+              %Ecto.Changeset{
+                errors: [smtp_port: {"Invalid port number.", _}]
+              }} = Settings.set_alerting_settings(Map.from_struct(settings))
+    end
```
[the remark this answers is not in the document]
skrech: I'll follow your advice.

I tried to search for what is the idiomatic way to parameterize tests in ExUnit and found only references to using `unquote`. That's why I used that. 

I like the "scenarios" approach, mainly because it's more tidy. In the interest of truth, though, it necessitates the same level of deep Elixir inner-working knowledge as the `unquote` variant, IMO :) For example, I didn't know that module attributes could be re-bound per-test definition and proper code in-lining would happen in the test body on every iteration of the `for`.

lib/trento/settings.ex
```diff
@@ -161,17 +182,62 @@ defmodule Trento.Settings do
...
+
+  @spec update_alerting_settings(alerting_setting_update_t()) ::
+          {:ok, AlertingSettings.t()}
+          | {:error, :alerting_settings_not_configured}
+          | {:error, Ecto.Changeset.t()}
+  def update_alerting_settings(alerting_settings) do
+    changes =
+      AlertingSettings.update_changeset(%AlertingSettings{}, alerting_settings)
+      |> Map.fetch!(:changes)
+      |> Map.put(:updated_at, DateTime.utc_now())
```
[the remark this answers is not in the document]
skrech: It seems that Ecto.Repo has two types of API functions: higher-level changeset-based API and lower-level query-based API. `insert_all`, `update_all`, etc are from the lower-level API where, it's said in the docs, auto-generated fields are not handled and the developer have to take care of them explicitly. And as to why I don't use the changeset-based `update` function: `update` works by requiring the primary key to be known in advance for the update to be successful. However, we don't know the concrete PK of the alerting settings record. We know that it's a "singleton", though, so I have to construct the query manually and fallback to `update_all` for that reason.

lib/trento/settings/alerting_settings.ex
```diff
@@ -0,0 +1,85 @@
...
+    |> sti_changes()
+    |> validate_required([
+      :type,
```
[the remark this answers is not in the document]
skrech: I thought that it's a good idea to validate everything that we expect to hit the DB adapter, irrespective of whether it comes from the STI or not. However, I have no strong opinion for this, I'd remove it if you want.

lib/trento/settings.ex
```diff
@@ -35,6 +36,26 @@ defmodule Trento.Settings do
...
+
+  @type alerting_setting_update_t :: %{
+          enabled: boolean | nil,
+          sender_email: String.t() | nil,
+          recipient_email: String.t() | nil,
+          smtp_server: String.t() | nil,
+          smtp_port: String.t() | integer | nil,
+          smtp_username: String.t() | nil,
+          smtp_password: String.t() | nil
+        }
```
gagandeepb, line 57: Consider unifying these two type definitions.

lib/trento/settings.ex
```diff
@@ -161,14 +182,53 @@ defmodule Trento.Settings do
...
+        where: [type: :alerting_settings],
+        update: [set: ^changes],
+        select: s
+
+    try do
+      {_count, [settings]} = Repo.update_all(query, [])
+      {:ok, settings}
+    rescue
+      _exc ->
+        {:error, :alerting_settings_not_configured}
```
gagandeepb, line 231: What are your thoughts about simplifying this, by initially fetching the AlertingSettings (using the zero arity `get_alerting_settings`) and then using the id from that, updating using the higher level API? (i.e. ideally a get-and-update in the same transaction) This should be possible with the higher level Ecto API.

lib/trento/settings/alerting_settings.ex
```diff
@@ -0,0 +1,69 @@
...
+    field :smtp_password, Trento.Support.Ecto.EncryptedBinary,
+      source: :alerting_smtp_password,
+      redact: true
+
+    timestamps(type: :utc_datetime_usec)
+    sti_fields()
+  end
+
+  @spec save_changeset(t(), map()) :: Ecto.Changeset.t()
+  def save_changeset(alerting_settings, changes) do
```
gagandeepb, line 33: Renaming suggestion: `create_changeset`

lib/trento/settings.ex
```diff
@@ -161,14 +182,53 @@ defmodule Trento.Settings do
...
+    |> AlertingSettings.save_changeset(alerting_settings)
+    |> Repo.insert(
+      on_conflict: {:replace_all_except, [:id, :inserted_at]},
+      conflict_target: :type,
+      returning: true
+    )
+  end
+
+  @spec update_alerting_settings(alerting_setting_update_t()) ::
+          {:ok, AlertingSettings.t()}
```
gagandeepb, line 210: These should be refactored into separate create and update functions.

priv/repo/migrations/20250304172641_add_alerting_settings.exs
```diff
@@ -0,0 +1,37 @@
...
+      add :alerting_enabled, :boolean
+      add :alerting_sender_email, :string
+      add :alerting_recipient_email, :string
+      add :alerting_smtp_server, :string
+      add :alerting_smtp_port, :integer
+      add :alerting_smtp_username, :string
+      add :alerting_smtp_password, :binary
+    end
+  end
+
```
gagandeepb, line 20: Consider/evaluate separating the migrations for abilities and settings into different files.

lib/trento/settings/alerting_settings.ex
```diff
@@ -0,0 +1,69 @@
...
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @cast_fields ~w(enabled sender_email recipient_email smtp_server smtp_port smtp_username smtp_password)a
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  schema "settings" do
+    field :enabled, :boolean, source: :alerting_enabled
```
gagandeepb, line 17: General/broad comment that we should sync on internally: avoiding STI/overloading the settings table. This leads to a rather cluttered data model. IMO, embedded schemas with [polymorphic embeds](https://hexdocs.pm/polymorphic_embed/readme.html) is a better extensible choice.

gagandeepb (changes requested):

lib/trento/settings.ex
skrech: My guiding principle when working with DB is to try to minimize the round-trips. And the performance is not my main driver, rather it's  the "straight-forwardness" of the logic involved. In my head the thought process is something like the following: A user would usually first try to GET the settings. Then, he would change them locally (in webui, native app, etc) and send a request for update to the API/backend. Now, why would the API/backend try to get the settings again? In my opinion, it's better to issue the UPDATE directly with the supplied input.

Having to get the record from the DB just to update it seems sub-optimal to me. :/

lib/trento/settings/alerting_settings.ex
skrech: I agree with you on your remarks for STI.

lib/trento/settings.ex
skrech: I'm not sure what you mean. In the set/create/save case, all the fields are mandatory, while in the partial update case any fields on it's own is fine.

lib/trento/settings.ex
skrech: I don't understand why my form of "extended create" is so problematic. The Upsert opeartion, used here, does two things:
1. Create a new record if not already created. (This is the `create` function)
2. _Replace_ the record that have been created previously.

I emphasized replace because this is really different than partial update. In order for a replace to succeed the same input should be provided as if create operation is requested (all fields should be supplied). And since the settings are known to be a form of singleton object, the upsert operation seems perfect match for this case.

gagandeepb (approved):
Thanks for this! Looks much better.

test/trento/settings_test.exs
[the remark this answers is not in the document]
nelsonkopliku: I see. There might be variations of tests given the scenario.

lib/trento/settings/alerting_settings.ex
nelsonkopliku: Tend to agree as well 🙈 Let's just maybe consider bringing this up later

nelsonkopliku (approved):
Hey @skrech thanks for clarifying!

Summing up we can say we need 2 operations:
- initial creation with all fields
- subsequent updates with partial - or complete - changes

and this is mainly driven by the fact that, as you mentioned, at a certain point a user might want to change, let's say, the smtp server url but not the password and we're not going to expose a plain password via the read API for that to be provided back, I agree.

Whether then we map the two operations to a POST/PATCH or PUT/PATCH pair, I don't really have anything agains any of the two.

I hear you about a few less queries to be made in one case vs the other, and again I wouldn't be against it, just let's consider 
- whether we fall in some premature optimization and 
- general consistency with the rest of the system

I am fine with the current state, and in any case we can adjust where needed as we proceed. This is a good baseline IMO.

---

If you really want to go towards PUT/PATCH pair consider that we'd end up with 
- an operation that either creates an entry or completely updates it (PUT)
- another op that allows you to partially or completely update a previously created one (PATCH)

While valid, it sounds slightly redundant, to me at least, and that's probably where the preference for the POST/PATCH comes from.
Even though less optimized for certain aspects, POST/PATCH pair seems to unambiguously separate concerns about creation and updates, additionally to being more consistent with what we _currently_ (emphasizing because things might change of course) do elsewhere.

---

Anyway great job! A fresh eye on things helps looking at things from a fresh perspective and opens for the possibility to re-consider past choices, which is healthy. Thanks!

However we need to also be conscious about _when_ we could get advantage of the fresh insights.

Now for the PRs depending on this, and in light of the ongoing release, we have two choices:
- we keep things pending and wait to merge after the release
- we use this (or yet another) branch as a temporary place where we merge alerting related PRs and later we merge that into main. In this way we can still put all the things together

---

PS: For completeness I am now realizing that some of our currently existing PUT operations might not support initial creation, thus not really compliant with what the verb should support. But this might be a discussion for another venue 🙈

nelsonkopliku (approved):

PR #3463, opened 2025-04-17
Fix references to activity log settings

nelsonkopliku:
# Description

While reviewing Alerting settings PR I realized we needed to fix this.


EMaksy (approved):
Nice catch!

gagandeepb (approved):
I am a bit surprised that the CI didn't catch this.

nelsonkopliku:
@gagandeepb yes, that's unfortunate.

AFAIU there's two reasons:

1.  this catchall in the policy
```elixir
def authorize(_, _, _), do: true
```

2. it seems we can alias any `Foo.Bar` module even if that does not exist.  And if that also matches in the test, here we go with success

🙈

PR #3470, opened 2025-04-23
Show tooltip only when hoovered over the start execution button

EMaksy:
# Description
The tooltip for the Start Execution button in the Checks Selection view of a cluster or a host is permanent, rather than showing up when the user hovers the mouse over it. This pr fixes it. 
Check out the demo environemnt

## How was this tested ?
Updated tests


abravosuse (approved):
I have tested the behavior in https://3470.prenv.trento.suse.com/ and for me it's the right one:

- If no checks are selected, the Start execution button is disabled (greyed out) and of course nothing happens when the user hovers the mouse over it
- If there are checks selected, the Start execution button is enabled (green) and when the user hovers the mouse over it, the tooltip pops-up (otherwise the tooltip is not visible)

But I would like to have @jagabomb 's blessing as well.

jagabomb (approved):
Tested the UX and it works for me. 😄

nelsonkopliku (approved):
LGTM

PR #3472, opened 2025-04-24
Implement retry logic for SUMA requests

janvhs:
# Description

Fixes TRNT-3568 where big responses from SUMA would time out and result in a 422 response. This is avoidable and we should respond to SUMA taking a bit longer with retries instead of propagating the error to the user as it results in a worse UX. I'd rather wait a bit and get the information I want to know, than not getting it in the first place

## How was this tested?

I tested manually tested the code with the security advisory triggering the bug and it acts like expected. I'm unsure how to test this in our current unit test setup

### DIFF
--- a/lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
+++ b/lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
@@ -87,7 +87,7 @@ defmodule Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor do
 
   @behaviour Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor
 
-  defp request(method, url, body \\ "", headers \\ [], options \\ []) do
+  defp request(method, url, body, headers, options) do
     # See https://hexdocs.pm/httpoison/HTTPoison.Request.html
     options = Keyword.put_new(options, :timeout, 8_000)
     options = Keyword.put_new(options, :recv_timeout, 5_000)


gagandeepb (commented):
I understand we have a specific HTTP endpoint (or possibly multiple endpoints) (SUMA) that is possibly slow to respond/sends a large response blob that leads to (client) time outs. I am not fully convinced that simply retrying would reliably fix this observed issue since the underlying issue of large response bodies is unaddressed. I wonder what are your thoughts about instead making a streaming request from the specific endpoints in question?

lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
```diff
@@ -85,6 +85,30 @@ defmodule Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor do
...
+          options = Keyword.update!(options, :timeout, &(&1 * 2))
+          options = Keyword.update!(options, :recv_timeout, &(&1 * 2))
+          next_attempt = current_attempt + 1
+          request(method, url, body, headers, options, next_attempt)
+        else
+          timeout_error
+        end
+
+      default ->
+        default
```
gagandeepb, line 108: I believe there could be improved separation of concerns in this implementation. The HTTP request making could be separated from the retry logic. Additionally, the explicit recursion should be/could be avoided  altogether. Consider making use of [something like this library, for instance.](https://hexdocs.pm/retry/readme.html)

gagandeepb: I wonder why you think this library/an implementation based on it, will not work with HTTPoison ?

nelsonkopliku (commented):
Hey Jan, I appreciate the attempt, however there's a couple of considerations from my side.

Let's be careful in applying a retry logic to all the requests indiscriminately.

As an example consider that there is already in place a retry mechanism for auth token retrieval https://github.com/trento-project/web/blob/aef787798a6118251df12ace74452a8e4458b112/lib/trento/infrastructure/software_updates/suma_api.ex#L18

At best we do a retry inside a retry, which we don't want, and at worst we might be introducing some unexpected behaviour.

Besides this, my main concern about adding a global retry capability to all SUMA requests indiscriminately is that several of those requests happen inside the request-response context of a trento API, and many of them are concurrent with each other inside the same trento API execution (I am thinking of relevant patches/available packages overviews).

If we're looking for a quick solution, then a more relaxed timeout might be a reasonable starting point.

For a long term, more robust approach, I do think we might need to re-consider making that many requests to SUMA inside trento's API request-response context and move those in a background integration where of course retrial is a good candidate to handle many scenarios.

PR #3460, opened 2025-04-16
Checks customization e2e

EMaksy:
# Description

This pr adds first e2e tests for checks customization feature. With the addition of wanda in demo mode in https://github.com/trento-project/web/pull/3455 we are able to test api communication between web and wanda. With this we don't require a mocked catalog anymore. As wanda runs in demo mode we receive even mocked agent data, which can be used in a follow up test  for modified checks execution.

## How was this tested?

Added e2e test with po

PR #3476, opened 2025-04-25
Add checks execution with custom values test e2e

EMaksy:
# Description

This pr is a follow up to https://github.com/trento-project/web/pull/3460. With this pr we add an additional test for the checks execution

## How was this tested? 

Added new e2e test and enriched po

### DIFF
--- a/test/e2e/cypress/e2e/checks_customization.cy.js
+++ b/test/e2e/cypress/e2e/checks_customization.cy.js
@@ -1,5 +1,7 @@
 import * as checksSelectionPage from '../pageObject/checks_customization_po';
 
+const COROSYN_MAX_MESSAGES_CHECK_ID = '00081D';
+const CHECK_COROSYNC_TOKEN_TIMEOUT = '156F64';
 context('Checks customization', () => {
   before(() => {
     checksSelectionPage.preloadTestData();
@@ -21,7 +23,9 @@ context('Checks customization', () => {
     });
 
     it('should customize and reset a check through the modal successfully', () => {
-      checksSelectionPage.openCustomizationModalFirstCheck();
+      checksSelectionPage.openCheckCustomizationModal(
+        COROSYN_MAX_MESSAGES_CHECK_ID
+      );
       // Validate if initial check customization modal has the correct values
       checksSelectionPage.validateFirstCheckId();
       checksSelectionPage.validateFirstCheckDescription();
@@ -45,7 +49,9 @@ context('Checks customization', () => {
       checksSelectionPage.checkCustomizationSuccessToastIsShown();
       // Validate if check was customized
       checksSelectionPage.customizedCheckShouldHaveModifiedPill();
-      checksSelectionPage.openCustomizationModalFirstCheck();
+      checksSelectionPage.openCheckCustomizationModal(
+        COROSYN_MAX_MESSAGES_CHECK_ID
+      );
       checksSelectionPage.validateCustomizedValue();
       // Reset check in the modal
       checksSelectionPage.clickResetCheckModalButton();
@@ -59,7 +65,9 @@ context('Checks customization', () => {
     });
 
     it('should customize check values in the check customization modal and reset check in checks category overview', () => {
-      checksSelectionPage.openCustomizationModalFirstCheck();
+      checksSelectionPage.openCheckCustomizationModal(
+        COROSYN_MAX_MESSAGES_CHECK_ID
+      );
       // User interacts with modal
       checksSelectionPage.clickOnWarningCheckbox();
       checksSelectionPage.inputCustomCheckValue();
@@ -75,7 +83,9 @@ context('Checks customization', () => {
     });
 
     it('should customize check values after fixing wrong user input', () => {
-      checksSelectionPage.openCustomizationModalSecondCheck();
+      checksSelectionPage.openCheckCustomizationModal(
+        CHECK_COROSYNC_TOKEN_TIMEOUT
+      );
       // User interact with modal
       checksSelectionPage.validateSecondCheckId();
       checksSelectionPage.clickOnWarningCheckbox();
@@ -97,7 +107,9 @@ context('Checks customization', () => {
     });
 
     it('should not customize check values if the user input is invalid', () => {
-      checksSelectionPage.openCustomizationModalSecondCheck();
+      checksSelectionPage.openCheckCustomizationModal(
+        CHECK_COROSYNC_TOKEN_TIMEOUT
+      );
       // User interact with modal
       checksSelectionPage.clickOnWarningCheckbox();
       checksSelectionPage.modalWarningCheckBoxShouldBeChecked();
@@ -115,8 +127,13 @@ context('Checks customization', () => {
       checksSelectionPage.resetIconShouldNotExistInOverview();
     });
 
+  });
+
+  describe('Execution with customized check values', () => {
     it('should run a checks execution with customized check values', () => {
-      checksSelectionPage.openCustomizationModalFirstCheck();
+      checksSelectionPage.openCheckCustomizationModal(
+        COROSYN_MAX_MESSAGES_CHECK_ID
+      );
       checksSelectionPage.clickOnWarningCheckbox();
       checksSelectionPage.inputCustomCheckValue();
       checksSelectionPage.clickModalSaveButton();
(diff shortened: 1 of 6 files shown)


EMaksy (commented):
With this we have the checks execution which checks for modified values. 
I think for now the test is sufficient, but we could enrich it easily in the future if we have more checks which cover for example bool check values

test/e2e/cypress/e2e/checks_customization.cy.js
```diff
@@ -114,5 +114,23 @@ context('Checks customization', () => {
       checksSelectionPage.secondCustomizedCheckShouldNotHaveModifiedPill();
       checksSelectionPage.resetIconShouldNotExistInOverview();
     });
+
+    it('should execute a customized check', () => {
+      checksSelectionPage.openCustomizationModalFirstCheck();
```
nelsonkopliku, line 119: I am wondering whether `openCustomizationModalFirstCheck` could actually be something like `openCheckCustomizationModal(checkId)`.

It would allow to test more scenarios and it might prove itself a bit more future-proof because if the catalog changes it might happen that the _first_ check is not anymore what we expect.

nelsonkopliku (approved):
Hey @EMaksy, in general it looks good. It feels great having the possibility to fully test checks related features.

My only remark is that we might want to make the test structure more parametric so that we can reuse same utilities for more scenarios.

We could even have a dedicated line of work for such test improvements.

test/e2e/cypress/pageObject/checks_customization_po.js
```diff
@@ -102,11 +119,42 @@ const _setInputValue = (element, index, value) => {
...
   _setInputValue(modalInput, 1, customMixedValue);
 
+const _clickChecksSelectionToggle = (index) => {
+  cy.get(checkSelectionToggle).eq(index).click();
+};
+export const clickCorosyncSelectionToggle = (corosyncIndex = 0) => {
+  _clickChecksSelectionToggle(corosyncIndex);
+};
+// used for local development to rerun the test multiple times
+export const resetCorosyncCheckSelection = () => {
```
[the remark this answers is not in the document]
vicenteqa: I haven't tried, but don't you think it can be useful to leave it in beforeEach hook? I understand it shouldn't affect the test execution but it will help make sure your scenario is prepared as you want, am I wrong? as some other test can leave an undesired scenario that this method would reset.

test/e2e/cypress/e2e/checks_customization.cy.js
vicenteqa: at least I would refer to it as the check name that we want to test and not by how it's displayed, first, in this case. The issue I see is that in that view the checkId is not something that we have in the DOM so it's not useful to select an element. is that or check against the test data which check name has that checkId so then the selector can be built. Because I agree that from a business perspective it makes sense to see which checkID are we interacting with in the test.

nelsonkopliku: Nothing prevents us from adding an attribute to the html element with the check id, if that helps (and I do believe it helps 😄)

EMaksy: Yeah that suggestion makes sense, but it adds some maintenance. 
Right now, we use this selector for the customize icon:
```
const checksCustomizationSettingsIcon = 'a[class*="block"] button[aria-label="customize-check"] svg';
```
  
This element appears for the corosync category 16 times.
  
 We pick the wished icon based by the index:  `eq(index)`
```
export const openCustomizationModalFirstCheck = () =>
cy.get(checksCustomizationSettingsIcon).eq(0)

export const openCustomizationModalSecondCheck = () =>
cy.get(checksCustomizationSettingsIcon).eq(1)
```
  
 We could add an object
```
const checkIndexMap = {
'00081D': 0,
'156F64': 1,
};
```
   where a check id is mapped to an index which would work, but every time we add a new check we would need to revisit and update this key map. 
   

For now i would prefer to keep it simple, as we need only two checks, but open to change it. 

What do you think @nelsonkopliku  and @vicenteqa ?

nelsonkopliku: what about making the attribute like this?
```html
aria-label="customize-check-156F64"
```
Then if we do need a generic `customize-check` marker to be on each element, we can think of using another kind of attribute or a css class

EMaksy: Hmm adding a check id may help, having a look at it :)

And  with this we have an option to discuss the usage of id's in test :+1:

test/e2e/cypress/pageObject/checks_customization_po.js
```diff
@@ -76,15 +77,29 @@ const saveButtonModal = 'button:contains("Save")';
...
 export const clickCorosyncCategory = () => cy.get(corosyncCategory).click();
-export const openCustomizationModalFirstCheck = () =>
-  cy.get(checksCustomizationSettingsIcon).eq(0).click();
-export const openCustomizationModalSecondCheck = () =>
-  cy.get(checksCustomizationSettingsIcon).eq(1).click();
+export const openCheckCustomizationModal = (checkID) => {
+  const checksCustomizationSettingsIcon = `a[class*="block"] button[aria-label*="customize-check-${checkID}"]`;
+  cy.get(checksCustomizationSettingsIcon).click();
+};
 export const clickOnWarningCheckbox = () => cy.get(modalInput).eq(0).click();
```
vicenteqa, line 103: I see that this `modalInput` const gets the three inputs present on the customization modal. It is a bad practice to access elements by index and should be avoided if possible. If some other day we place this checkbox on some other place the test will fail, if instead of that selector and its index, you use `div input[type="checkbox"]` you will surely get that checkbox and the test will be robust if that element is eventually placed somewhere else. I saw the same issue with some other selectors, that you are getting a selector that returns many elements and then access them by index

test/e2e/cypress/e2e/checks_customization.cy.js
```diff
@@ -1,5 +1,7 @@
 import * as checksSelectionPage from '../pageObject/checks_customization_po';
 
+const COROSYN_MAX_MESSAGES_CHECK_ID = '00081D';
+const CHECK_COROSYNC_TOKEN_TIMEOUT = '156F64';
```
nelsonkopliku, line 4: Not sure making a constant for check ids helps a lot here.

Anyway, If we intend to keep the constants let's make the naming pattern consistent

nelsonkopliku (commented):
Hey @EMaksy, thanks for making the attribute parametrized, I think it will be more helpful.

I believe that once @vicenteqa's feedback have been addressed, we might move this forward.

Then we'd need to improve overall testing coverage for check related features.

test/e2e/cypress/e2e/checks_customization.cy.js
vicenteqa: in theory we discarded in the past the usage of data-testid attributes in our tests, that was in some slack thread that I can't find now :|

nelsonkopliku: alright, now that you mention I kind of remember 🙈 
Please feel free to share a reference for refresher.

Was it only for `data-testid` or for any `data-*` attribute?

test/e2e/cypress/e2e/checks_customization.cy.js
EMaksy: I will pass the values directly to the functions then :)
I guess it is more readable :)

test/e2e/cypress/e2e/checks_customization.cy.js
vicenteqa: `data-testid`

test/e2e/cypress/pageObject/checks_customization_po.js
```diff
@@ -56,56 +58,99 @@ const resetModalText =
...
+export const clickResetModalButton = () => cy.get(resetButton).click();
 export const clickResetCustomizedCheck = () =>
   cy.get(resetCustomizedCheckIcon).click();
-export const clickModalSaveButton = () => cy.get(saveButtonModal).eq(1).click();
-const _setInputValue = (element, index, value) => {
-  cy.get(element)
-    .eq(index)
+export const clickModalSaveButton = () => cy.get(saveButtonModal).click();
+const _setInputValue = (element, currentValue, newValue) => {
+  cy.get(element + `[value="${currentValue}"]`)
```
vicenteqa, line 122: isn't any other way to select this element rather than it's current value? it's a bit weird seeing the test step with the old value and the new one. I will try to help with this

test/e2e/cypress/pageObject/checks_customization_po.js
```diff
@@ -56,56 +58,99 @@ const resetModalText =
...
     .should('be.visible')
```
vicenteqa, line 123: also, this should be visible is redundant, as if it wasn't visible you would not be able to interact with it.

vicenteqa (approved):
very nice job @EMaksy ! :clap:

PR #3477, opened 2025-04-28
Relax timeouts

janvhs:
Successor of #3472 . Description generally applies, but the TLDR is the default HTTPoison timeouts are more than enough for this out case and more restrictive lead to HTTP time-outs, even when SUMA would have responded. This PR addresses it

### DIFF
--- a/lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
+++ b/lib/trento/infrastructure/software_updates/adapter/suma_http_executor.ex
@@ -97,7 +97,7 @@ defmodule Trento.Infrastructure.SoftwareUpdates.Suma.HttpExecutor do
       "#{base_url}/auth/login",
       payload,
       [{"Content-type", "application/json"}],
-      ssl_options(ca_cert) ++ timeout_options()
+      ssl_options(ca_cert)
     )
   end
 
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):
Thanks @janvhs!

Just take a look at https://github.com/trento-project/web/pull/3435, which seems basically the same thing, although slightly different.

I personally tend to prefer us to have explicitly set timeouts as in https://github.com/trento-project/web/pull/3435, but I could live with implicit library ones.

I am just wondering whether we should keep a lower `timeout` for TCP/SSL connection. 8000 seems way too much 😄 

doc excerpt https://hexdocs.pm/httpoison/HTTPoison.Request.html
> :timeout - timeout for establishing a TCP or SSL connection, in milliseconds. Default is 8000
:recv_timeout - timeout for receiving an HTTP response from the socket. Default is 5000

nelsonkopliku:
> I would argue that it doesn't hurt to have longer timeouts here, since we do not control SUMA's response time. 8 seconds is a lot, but as it's a timeout, the 8 seconds wouldn't be used when not required. In my opinion it's better to wait instead of failing, because failing wouldn't allow the user to see the data, but waiting does.

I am fine 👍

balanza (approved):
LGTM

It's a bit implicit but it's fine, it's not hidden information. 

Be careful that the `timeout` is increasing from 1s to 8s -  that is, the time to establish a SSL connection with the SUMA service. That would delay the failure in case of SUMA unavailability or misconfiguration. Not a big deal, though.

PR #3482, opened 2025-05-05
Expose cluster resource parent api

arbulu89:
# Description

Expose cluster resource `parent` field in the clusters api. It is only exposed in the `v2` of the `/clusters` endpoint.
This field is needed in order to put parent cluster resources in maintenance, as it is a pretty common thing.
I don't aim to display the value anywhere else, even though we could do it, in the same place where we show other cluster resources.

## How was this tested?

UT

### DIFF
--- a/lib/trento_web/openapi/v2/schema/cluster.ex
+++ b/lib/trento_web/openapi/v2/schema/cluster.ex
@@ -33,7 +33,11 @@ defmodule TrentoWeb.OpenApi.V2.Schema.Cluster do
             nullable: true,
             properties: %{
               id: %Schema{type: :string},
-              managed: %Schema{type: :boolean},
+              managed: %Schema{
+                type: :boolean,
+                nullable: true,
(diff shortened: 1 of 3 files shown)


lib/trento_web/openapi/v2/schema/cluster.ex
```diff
@@ -26,7 +26,26 @@ defmodule TrentoWeb.OpenApi.V2.Schema.Cluster do
...
+            additionalProperties: false,
+            nullable: true,
+            properties: %{
+              id: %Schema{type: :string},
+              managed: %Schema{type: :boolean},
+              multi_state: %Schema{
+                type: :boolean,
+                nullable: true,
+                description: """
+                Represents the type of the group.
```
arbulu89, line 41: This `true/false/null` values are what crm_mon use to define the value

EMaksy (approved):

PR #3485, opened 2025-05-06
Reorder start ordering

gagandeepb:
Start the `Task.Supervisor` before the endpoint since the ActivityLogging plug is dependent on this and the plug relies on the endpoint for logging.


arbulu89 (approved):
Thank you @gagandeepb 
We at least fix one of the final errors (even though not the root causes).

nelsonkopliku (approved):
Thanks!

PR #3486, opened 2025-05-06
Unify cluster details header code in new component

arbulu89:
# Description

Create a new `ClusterDetailsHeader` component to remove duplicated code.
It will come handy when we add the new operations button and code, as otherwise we would need to add it twice as well.

## How was this tested?

UT


EMaksy (approved):
Hey man thanks taking care of the refactor --> LGTM!

Just a comment: 
We could create a Storybook story for the ClusterDetailsHeader, but as it is included in AscsErs and HanaClusterDetails, it is not required and  it is more busy work, so i guess no need for it right now right ?

arbulu89:
> Hey man thanks taking care of the refactor --> LGTM!
> 
> Just a comment: We could create a Storybook story for the ClusterDetailsHeader, but as it is included in AscsErs and HanaClusterDetails, it is not required and it is more busy work, so i guess no need for it right now right ?

Yes, I didn't create it for that reason. After all, we have stories of this component indirectly.

nelsonkopliku (approved):

arbulu89:
Hey @nelsonkopliku @EMaksy 
After giving a though, I have changed my strategy here.
Instead of having a header component, I have created the `ClusterDetails` component which works as a container wrapper. It includes all the share components and it accepts as children the details of HANA and ASCS/ERS.
This way, we don't need to pass that many props from one to another. And it makes testing easier, as we don't have that many boilerplate arguments.

assets/js/pages/ClusterDetails/ClusterDetailsPage.jsx
```diff
@@ -81,47 +82,63 @@ export function ClusterDetailsPage() {
     case 'hana_scale_up':
     case 'hana_scale_out':
       return (
-        <HanaClusterDetails
+        <ClusterDetails
```
EMaksy, line 85: This works for me we only have some duplicated props that we need to pass [clusterID, hosts, lastExecution, navigate] which is fine, i guess testing is easier that way. 
Just a note about
```
(clusterDetails={cluster.details})
```
In the HanaClusterDetails component we use
```
(details={cluster.details})
```

they are the same right? 

Maybe we can keep same prop name ? 
So we can  skip` <ClusterDetails clusterDetails={props.details} {...props}>` in the story
```
function ContainerWrapper({ children, ...props }) {
  return (
    <div className="max-w-7xl mx-auto px-4 sm:px-6 md:px-8">
      <ClusterDetails clusterDetails={props.details} {...props}>
        {children}
      </ClusterDetails>
    </div>
```

EMaksy (commented):
Hey nice, it looks good to me, just left a comment about prop naming. Have a look :)

assets/js/pages/ClusterDetails/ClusterDetailsPage.jsx
arbulu89: okey

arbulu89: done

EMaksy (approved):
:+1:

nelsonkopliku (approved):

PR #3487, opened 2025-05-07
Add cluster maintenance change modal

arbulu89:
# Description

Add `ClusterMaintenanceChangeModal` component.
The components uses the cluster details to compose the list of nodes and resources to change the maintenance mode.
Depending on the type, the request we need to do is different, so I have added the `scope` concept here as well. Based on this value, we will send `node_id`, `resource_id` or neither of them in the request.

Find the entry in storybook.
cluster_maintenance_change
(we see 2 nodes entries as strings and resources as UUID)

## How was this tested?

UT and storybook

### DIFF
--- a/assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
+++ b/assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
@@ -6,7 +6,7 @@ import Switch from '@common/Switch';
 
 import OperationModal from './OperationModal';
 
-const NOT_SELECTED = 'Select a cluster resource';
+const NOT_SELECTED_LABEL = 'Select a cluster resource';
 const CLUSTER_LABEL = 'Cluster (full maintenance)';
 const NODE_MAINTENANCE_STATE = 'Maintenance';
 const CLUSTER_SCOPE = 'cluster';
@@ -15,7 +15,7 @@ const RESOURCE_SCOPE = 'resource';
 
 const NOT_SELECTED_OPTION = {
   value: {
-    id: NOT_SELECTED,
+    id: NOT_SELECTED_LABEL,
     maintenance: false,
   },
   key: 'not_selected',
--- a/assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
+++ b/assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
@@ -1,0 +1,128 @@
+import React, { useState } from 'react';
+import { concat, flatMap, flow, get, map, uniqBy } from 'lodash';
+import Label from '@common/Label';
+import Select from '@common/Select';
+import Switch from '@common/Switch';
+import OperationModal from './OperationModal';
+
+const NOT_SELECTED = 'Select a cluster resource';
+const CLUSTER_LABEL = 'Cluster (full maintenance)';
+const NODE_MAINTENANCE_STATE = 'Maintenance';
+const CLUSTER_SCOPE = 'cluster';
+const NODE_SCOPE = 'node';
+const RESOURCE_SCOPE = 'resource';
+
+const NOT_SELECTED_OPTION = {
+  value: {
+    id: NOT_SELECTED,
+    maintenance: false,
+  },
+  key: 'not_selected',
+};
+
+const renderOption = ({ value: { id } }) => id;
+
+function ClusterMaintenanceChangeModal({
+  clusterDetails,
+  isOpen = false,
+  onRequest,
+  onCancel,
+}) {
+  const [checked, setChecked] = useState(false);
+  const [resource, setResource] = useState(NOT_SELECTED_OPTION.value);
+  const [maintenanceState, setMaintenanceState] = useState(false);
+
+  const clusterMaintenance = get(clusterDetails, 'maintenance_mode', false);
+  const clusterNodes = get(clusterDetails, 'nodes', []);
+  const stoppedResource = get(clusterDetails, 'stopped_resources', []);
+
+  const clusterOption = {
+    value: {
+      id: CLUSTER_LABEL,
+      maintenance: clusterMaintenance,
+      scope: CLUSTER_SCOPE,
+    },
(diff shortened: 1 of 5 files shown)


abravosuse (approved):
LGTM, code apart and based on the clip provided.

assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
```diff
@@ -0,0 +1,128 @@
...
+  const allOptions = concat(
+    NOT_SELECTED_OPTION,
+    clusterOption,
+    nodeOptions,
+    resourceOptions
+  );
+
+  return (
+    <OperationModal
+      title="Cluster Maintenance"
```
EMaksy, line 81: i guess this could also be a const variable label put on top?

EMaksy: also goes for description and  operationText?

assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
```diff
@@ -0,0 +1,128 @@
...
+      description="Update cluster, node or resource maintenance state"
+      operationText="Maintenance"
+      applyDisabled={!checked || resource.maintenance === maintenanceState}
+      checked={checked}
+      isOpen={isOpen}
+      onChecked={() => setChecked((prev) => !prev)}
+      onRequest={() =>
+        // send node_id or resource_id only if the option is of that type
+        onRequest({
+          maintenance: maintenanceState,
```
EMaksy, line 91: Maybe moving this block to a seperate function could improve the maintability and readablity? 

```
const handleRequest = () => {
  const requestData = {
    maintenance: maintenanceState,
    ...(resource.scope === NODE_SCOPE && { node_id: resource.id }),
    ...(resource.scope === RESOURCE_SCOPE && { resource_id: resource.id }),
  };

  onRequest(requestData);
};


and just have 
<OperationModal
  ....
  onRequest={handleRequest} 
  ....
/>
```

It is more my personal preference so feel free :)

assets/js/common/OperationModals/ClusterMaintenanceChangeModal.stories.jsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+import { clusterFactory } from '@lib/test-utils/factories';
+
+import ClusterMaintenanceChangeModal from './ClusterMaintenanceChangeModal';
+
+const { details: clusterDetails } = clusterFactory.build();
+const { details: clusterMaintenanceDetails } = clusterFactory.build({
+  details: { maintenance_mode: true },
```
EMaksy, line 9: There is an issue with this, checking storybook the value is false and this is not applied
`details: { maintenance_mode: true },`
Maybe we need to update the factory ? 
Check the  factory, i guess we need to do something with the details
```
export const clusterFactory = Factory.define(({ sequence, params }) => {
  const { type = 'hana_scale_up' } = params;

  const details = (() => {
    switch (type) {
      case 'ascs_ers':
        return ascsErsClusterDetailsFactory.build(params.details);
      case 'hana_scale_up':
      case 'hana_scale_out':
      default:
        return hanaClusterDetailsFactory.build(params.details);
    }
  })();

  return {
    id: faker.string.uuid(),
    name: `${faker.person.firstName()}_${sequence}`,
    sap_instances: clusteredSapInstanceFactory.buildList(1),
    hosts_number: faker.number.int(),
    resources_number: faker.number.int(),
    type: clusterTypeEnum(),
    health: healthEnum(),
    selected_checks: [],
    provider: cloudProviderEnum(),
    cib_last_written: format(faker.date.recent(), 'EEE MMM d h:mm:ss yyyy'),
    details,
  };
});
```

EMaksy: image

EMaksy (changes requested):
Hey man great work!
I left mostly some nitpicks for the component(and a lot of personal preference), but there is an small issue with the factory usage in the storybook story. It may need some attention, please have a look.

assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
```diff
@@ -0,0 +1,128 @@
...
+      id: name,
+      maintenance: status === NODE_MAINTENANCE_STATE,
+      scope: NODE_SCOPE,
+    },
+    key: name,
+  }));
+
+  // flatMap all resources and parents from all nodes
+  // and concat stopped resources
+  const resourceOptions = flow([
```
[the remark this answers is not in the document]
arbulu89: Well, it is the same as in elixir. Basically, one entry output is the input for the next one.

assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
arbulu89: Let's keep as it is. They are not used anywhere else and are just text that don't have any other meaning

assets/js/common/OperationModals/ClusterMaintenanceChangeModal.jsx
arbulu89: I think it is fine as it is. It is a fairly simple thing

assets/js/common/OperationModals/ClusterMaintenanceChangeModal.stories.jsx
arbulu89: Actually, it is working, even though the `controls` data is different.
The thing is that I implemented the old `function` way instead of the `const` one, which doesn't work properly with the that.
Let me change it (and I will change the the existing modals storybook as well)

EMaksy (approved):
Hey man great job, lgtm!

EMaksy:
Only ci seems complains :)

EMaksy (approved):

PR #3488, opened 2025-05-08
Add the ability to request a saptune solution change

nelsonkopliku:
# Description

This PR introduces the ability to request a saptune solution change operation.

As of now the only thing that changes with regards to the _saptune solution apply_ operation is the operation name `:saptune_solution_change`, `"saptune_solution_change"` and the operator being used `saptunechangesolution@v1`.

Requesting a saptune solution change operation currently follows the same policies defined for a saptune solution apply.

What's gonna be next:
- UI to enable requesting the new operation https://github.com/trento-project/web/pull/3501
- adding the relevant operation to the catalog in wanda https://github.com/trento-project/wanda/pull/608
- wiring up the correct operator in workbench/agent https://github.com/trento-project/workbench/pull/16

Note: current proposal is the simplest possible, maybe even naive, that just adds the counterparts related to the new `saptune_solution_change` operation as in atoms/strings.
I am considering adding maybe an enum and some helpers to improve operation definition and mapping in web.

EDIT: PR here https://github.com/trento-project/web/pull/3499

## How was this tested?

Automated tests.


lib/trento/activity_logging/parser/metadata_enricher.ex
```diff
@@ -88,8 +88,9 @@ defmodule Trento.ActivityLog.Logger.Parser.MetadataEnricher do
        when activity in [:resource_tagging, :resource_untagging],
        do: {:ok, resource_id}
 
-  defp detect_enrichment(:host, {_, %{resource_id: id, operation: :saptune_solution_apply}}),
-    do: {:ok, id}
+  defp detect_enrichment(:host, {_, %{resource_id: id, operation: operation}})
+       when operation in [:saptune_solution_apply, :saptune_solution_change],
```
arbulu89, line 92: Maybe we want to create an enum in operations, like `HostOperations` (as example, and `ClusterOperations` if needed in the future, etc) where we can list them all. So we don't need to be changing this.
What do you think?

lib/trento/hosts.ex
```diff
@@ -140,13 +140,20 @@ defmodule Trento.Hosts do
   end
 
   @spec request_operation(atom(), String.t(), map()) :: {:ok, String.t()} | {:error, any}
-  def request_operation(:saptune_solution_apply, host_id, params) do
+  def request_operation(operation, host_id, params)
+      when operation in [:saptune_solution_apply, :saptune_solution_change] do
     operation_id = UUID.uuid4()
 
+    operator =
```
arbulu89, line 147: nitpick:
could we maybe have a `map_operation` kind of function?
we will continue adding more operation, and it will scale better

lib/trento/infrastructure/operations/operations.ex
```diff
@@ -46,5 +46,7 @@ defmodule Trento.Infrastructure.Operations do
   end
 
   def map_operation_type("saptuneapplysolution@v1"), do: :saptune_solution_apply
+  def map_operation_type("saptunechangesolution@v1"), do: :saptune_solution_change
+
```
arbulu89, line 50: nitpick: i would remove this new line (if it is not added by the formatter)

lib/trento_web/controllers/v1/host_controller.ex
```diff
@@ -186,14 +186,15 @@ defmodule TrentoWeb.V1.HostController do
...
-      }) do
+  def request_operation(%{assigns: %{host: host}} = conn, %{operation: operation})
+      when operation in ["saptune_solution_apply", "saptune_solution_change"] do
     %{solution: solution} = OpenApiSpex.body_params(conn)
     %{id: host_id} = host
 
     with {:ok, operation_id} <-
-           Hosts.request_operation(:saptune_solution_apply, host_id, %{solution: solution}) do
+           conn
+           |> get_operation()
```
arbulu89, line 196: In the `cluster_maintenance_change` code I'm working in, I have updated the `operations_policy_plug.ex` code (the `call` function) to assign the operation to `:operation`, so we can get it in the `assign` here, and we don't need to be converting more than once.
I think we can leverage that. If you want I can cherry-pick the change here as well.

test/trento/activity_logging/metadata_enricher_test.exs
```diff
@@ -219,28 +219,32 @@ defmodule Trento.ActivityLog.MetadataEnricherTest do
   end
 
   describe "enriching operation activities" do
-    test "should enrich operation completed events in hosts" do
-      %{id: host_id, hostname: hostname} = insert(:host)
+    for saptune_operation <- [:saptune_solution_apply, :saptune_solution_change] do
+      @saptune_operation saptune_operation
```
arbulu89, line 223: nitpick:
as this looks a generic test, maybe we can rename `saptune_operation` to `operation/current_operatrion`

arbulu89 (commented):
Looks great!
just some nitpicks.
let me know what you think about the `enum` and the change in the operations policy plug

lib/trento/activity_logging/parser/metadata_enricher.ex
nelsonkopliku: Glad you ask, @arbulu89 😄 take a look at this https://github.com/trento-project/web/pull/3499

lib/trento/hosts.ex
nelsonkopliku: Agreed. The liked PR should cover this

lib/trento_web/controllers/v1/host_controller.ex
nelsonkopliku: It's a good idea. While going through the code I was also thinking about that. Thanks!

lib/trento/infrastructure/operations/operations.ex
nelsonkopliku: 13c6ee8c722373a1d1f0a246246eb656f14fb2b2

test/trento/activity_logging/metadata_enricher_test.exs
nelsonkopliku: 13c6ee8c722373a1d1f0a246246eb656f14fb2b2

nelsonkopliku:
@arbulu89 now that we have also the enum here, I am starting to question whether it actually is useful 🤔 

I mean, if we look where we are doing the `in HostOperations.values()` guard it's only 2 places, but more than that the thing is that in those two places we might still need to be strict on the saptune operations.

What about postponing introduction of such enum to a later stage when more host operations come up?

arbulu89:
> @arbulu89 now that we have also the enum here, I am starting to question whether it actually is useful 🤔
> 
> I mean, if we look where we are doing the `in HostOperations.values()` guard it's only 2 places, but more than that the thing is that in those two places we might still need to be strict on the saptune operations.
> 
> What about postponing introduction of such enum to a later stage when more host operations come up?

Which are the scenarios where we need to restrict? As far as I see, those 2 functions are for all host operations.
I think it is fine. Anyway, if you prefer postponing, go ahead, it is OK. We can re-think in a later stage as well

lib/trento_web/controllers/v1/host_controller.ex
arbulu89: I will modify this in my `cluster_maintenance_change` PR. It is one line change, so it is fine as i

arbulu89 (approved):
@nelsonkopliku I really like how it is now.
I think it covers everything.
About the `Enum`, i find it totally correct, but if you are dubious about it, I'm fine removing it by now

nelsonkopliku:
Thanks @arbulu89, your confidence on the enum makes me feel more confident as well 😄.
Indeed those two usages are safe for now for any host operation, that we can think of at the moment.

Let's go with this, if needed we'll adapt.

EMaksy (approved):
LGTM!

PR #3489, opened 2025-05-08
SMTP allows talking to unauthenticated servers

skrech:
# Description

Enabled Swoosh elixir lib to talk to unauthenticated SMTP servers.

Fixes [TRNT-3709](https://jira.suse.com/browse/TRNT-3709)

## How was this tested?

Not tested

No need for doc update.

### DIFF
--- a/lib/trento/infrastructure/alerting/emails/templates/api_key_expiration.html.eex
+++ b/lib/trento/infrastructure/alerting/emails/templates/api_key_expiration.html.eex
@@ -403,12 +403,12 @@
 
                   <a href="https://www.trento-project.io/">
                     <img src="https://www.trento-project.io/images/trento-logo-stacked.svg" alt="Trento Project"
-                      width="100" height="100" border="0"
+                      width="100" height="75" border="0"
                       style="border:0; outline:none; text-decoration:none; display:block; margin:auto">
                   </a>
                   <br />
-                  <a href="https://www.suse.com"><img src="https://brand.suse.com/assets/img/brand/block-logo-big.png" alt="Trento Project"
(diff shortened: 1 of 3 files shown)


arbulu89 (approved):
@skrech it looks good.
Please test against SMTP server with and without authentication usage.
we want to be sure that sending/using `user/password` still works

nelsonkopliku (approved):
Looks good.

Not necessarily related to this PR, but we could consider adding a test smtp server like mailhog to the local docker setup. 
Similarly to what happens [here](https://github.com/trento-project/playground/blob/main/docker-compose.yaml)

skrech:
Yes, I've tested it with the descendant of Mailhog called Mailpit. 
IMO, we can leave this PR for hotfix, and add proper docker-compose profile, dev mail config in `config.ex` and maybe even a test somewhere.

nelsonkopliku:
@skrech sure, go ahead! If it wasn't clear, I also meant we can add those in a different PR 👍

lib/trento/infrastructure/alerting/emails/templates/api_key_expiration.html.eex
```diff
@@ -403,12 +403,12 @@
 
                   <a href="https://www.trento-project.io/">
                     <img src="https://www.trento-project.io/images/trento-logo-stacked.svg" alt="Trento Project"
-                      width="100" height="100" border="0"
+                      width="100" height="75" border="0"
```
arbulu89, line 406: nitpick:
I don't think we should do this kind of changes in unrelated PRs.

PR #3490, opened 2025-05-08
Bugfix for web as it fails to start when no IPV6 is available

EMaksy:
# Description

This pr provides a fix for web, currently when web is started and the machine does not have ipv6 the app fails to start.
This happens because the [Cowboy webserver Plug](https://hexdocs.pm/plug_cowboy/Plug.Cowboy.html) requires either an ipv4 or ipv6 address.  For this we can use the [erlang inet getaddr function ](https://www.erlang.org/docs/23/man/inet#getaddr-2)

Same change will be done in wanda

## How was this tested?
Tested locally by disabling ipv6

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -69,12 +69,14 @@ if config_env() in [:prod, :demo] do
       For example: yourdomain.example.com
       """
 
-  defp determine_ip() do
+  # Returns an IP address to bind the HTTP server, based on IPv6 support.
+  # This prevents startup failures on systems without IPv6 where the app would crash
+  # Falls back to IPv4 `{0, 0, 0, 0}` if IPv6 is not available.
+  ip =
     case :inet.getaddr(~c"localhost", :inet6) do
       {:ok, _addr} -> {0, 0, 0, 0, 0, 0, 0, 0}
       {:error, _} -> {0, 0, 0, 0}
     end
-  end
 
   config :trento, TrentoWeb.Endpoint,
     http: [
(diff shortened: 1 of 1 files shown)


EMaksy:
Alternatively we could also add a new env for the runtime exs and give the user the ability to decide, but i think it is to much effort and may even confuse the user :)
open for suggestions

nelsonkopliku (approved):
Thanks Eugen, looks good!

I see that allowing the user to decide requires extra changes and I am fine to defer that.

The user did not have a chance to choose so far, afterall, so it should be a good enough fix for now.

We can discuss whether allowing this choice and implement later.

config/runtime.exs
```diff
@@ -69,13 +69,22 @@ if config_env() in [:prod, :demo] do
       For example: yourdomain.example.com
       """
 
+  # Returns an IP address to bind the HTTP server, based on IPv6 support.
+  # This prevents startup failures on systems without IPv6 where the app would crash
+  # Falls back to IPv4 `{0, 0, 0, 0}` if IPv6 is not available.
+  ip =
+    case :inet.getaddr(~c"localhost", :inet6) do
+      {:ok, _addr} -> {0, 0, 0, 0, 0, 0, 0, 0}
+      {:error, _} -> {0, 0, 0, 0}
```
stefanotorresi, line 78: Hey Eugen, I see that Erlang inet provides a bunch of specific [POSIX error codes](https://www.erlang.org/doc/apps/kernel/inet.html#module-posix-error-codes), maybe we could be a little more specific about the error we're looking for here, which I guess would be `eafnosupport`. Could you test this?

stefanotorresi: After an eye-to-eye discussion we decided to disregard my suggestion and leave it as it is, because what happens with `:inet.getaddr` is also a domain resolution syscall, so there are a bunch of other errors that might happen and indicate the absence of ipv6 support. We'll continue investigating on more precise ipv6 detection approaches as a background research task.

PR #3493, opened 2025-05-09
remove unnecessary version field

stefanotorresi:
This is not actually needed and we don't want to keep updating it at every release.


stefanotorresi:
ahh damn, I guess I will need to do something about that (wrong) assertion failing.

vicenteqa (approved):

stefanotorresi:
superseded by #3509

PR #3494, opened 2025-05-09
Remove legacy catalog list coming from `runner` implementation

nelsonkopliku:
# Description

Sometimes it's hard to leave things go... however it might be time for this 🙈


EMaksy (approved):
Thanks for clean up :)

arbulu89 (approved):

PR #3495, opened 2025-05-09
Cleanup maintainers list

nelsonkopliku:
# Description

Tidy up team member refs.


stefanotorresi:
this variable is actually not used anyware and not needed any longer, in fact I just removed it in https://github.com/trento-project/web/pull/3496

PR #3496, opened 2025-05-09
Automate changelog generation

stefanotorresi:
# Description

This PR adds two new GHA workflows to ease the release process a little bit:
- **Pre-release**: runs every time `main` changes (i.e. when a PR is merged). It creates a draft release with automatically generated release notes and version, that only waits for the green "Publish" button to be clicked, to actually publish the release.
- **Post-release**: runs every time a release is actually published. It creates a Pull Request to keep the CHANGELOG.md file in sync with the content that gets automatically generated in the release notes in the above _Pre-release_ worfklow.

Coincidentally, this PR also includes a couple minor fixes to the existing CI workflow, namely the removal of an unused env var and some automatically applied stylistic changes.

- [x] **DONE**

## How was this tested?

I tested manually in my own fork https://github.com/stefanotorresi/web

- [x] **DONE**

Documentation changes will follow in the docs repo.

- [ ] **DONE**

### DIFF
--- a/.github/release_drafter.yaml
+++ b/.github/release_drafter.yaml
@@ -1,0 +1,36 @@
+name-template: $RESOLVED_VERSION
+tag-template: $RESOLVED_VERSION
+categories:
+  - title: Features
+    labels:
+      - feature
+      - enhancement
+  - title: Bug Fixes
+    labels:
+      - fix
+      - bugfix
+      - bug
+  - title: Maintenance
+    labels:
+      - chore
+      - ci
+      - tech-debt
+  - title: Dependencies
+    label: dependencies
+    collapse-after: 3
+version-resolver:
+  major:
+    labels:
+      - major
+      - breaking
+  minor:
+    labels:
+      - feature
+      - minor
+  patch:
+    labels:
+      - patch
+      - bug
+  default: patch
+exclude-labels:
+  - skip-changelog
--- a/.github/release_drafter.yaml
+++ b/.github/release_drafter.yaml
@@ -35,4 +35,8 @@ version-resolver:
 exclude-labels:
   - skip-release-notes
 template: |
+  ## What's Changed
+
   $CHANGES
+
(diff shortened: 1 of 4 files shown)


nelsonkopliku (commented):
Cool, curious to see it in action. 

If I understand it correctly the changelog file gets updated with the relevant content _after_ the actual release, am I getting it right?

If that is the case, how does that cope with OBS/IBS workflows?
If I remember correctly OBS/IBS need the changelog changes in the source. 
Is this still the case?

stefanotorresi:
> If I understand it correctly the changelog file gets updated with the relevant content _after_ the actual release, am I getting it right?

Correct. This is suboptimal, I know, but it's a chicken&egg problem: the release notes are finalized contextually with the release, so I couldn't find a way to update `CHANGELOG.md` _before_ a release.


> If that is the case, how does that cope with OBS/IBS workflows? If I remember correctly OBS/IBS need the changelog changes in the source. Is this still the case?

The OBS workflow gets the release notes from the GitHub release API, it doesn't interact with the `CHANGELOG.md` file, so it works just fine, it will have the latest release notes included as expected.

At the end, we keep `CHANGELOG.md` just as a very passive and historical record. The way it needed to be maintained before this was quite cumbersome and not fully automated, plus the toolchain was getting really dated.
I'd say it's better to have it not perfect but fully automated, rather than having to maintain it with some manual interaction (which, by the way, has historically produced suboptimal output because it's error prone, so it's not perfect either).

nelsonkopliku (approved):
Alright, thanks for clarifying.

> The OBS workflow gets the release notes from the GitHub release API, it doesn't interact with the CHANGELOG.md file
>
I guess I was misled by the assumption the interaction was with the `CHANGELOG.md`

And yes I tend to agree: if we have to choose between two imperfect options, then I'd also prefer the automated imperfect one 😄 

EDIT: as long as it works 😅

PR #3499, opened 2025-05-12
Improve operations mapping

nelsonkopliku:
# Description

Adding a `HostOperations` enum to improve operations mapping.

### DIFF
--- a/lib/trento/infrastructure/operations/amqp/processor.ex
+++ b/lib/trento/infrastructure/operations/amqp/processor.ex
@@ -18,6 +18,8 @@ defmodule Trento.Infrastructure.Operations.AMQP.Processor do
 
   alias Trento.Discovery
 
+  require Trento.Operations.HostOperations, as: HostOperations
+
   require Logger
 
   def process(%GenRMQ.Message{payload: payload} = message) do
@@ -75,7 +77,7 @@ defmodule Trento.Infrastructure.Operations.AMQP.Processor do
   end
 
   defp maybe_request_discovery(operation, :UPDATED, group_id)
-       when operation in [:saptune_solution_apply, :saptune_solution_change] do
+       when operation in HostOperations.values() do
     Discovery.request_saptune_discovery(group_id)
   end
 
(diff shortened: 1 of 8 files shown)


lib/trento/infrastructure/operations/amqp/processor.ex
```diff
@@ -75,7 +77,7 @@ defmodule Trento.Infrastructure.Operations.AMQP.Processor do
   end
 
   defp maybe_request_discovery(operation, :UPDATED, group_id)
-       when operation in [:saptune_solution_apply, :saptune_solution_change] do
+       when operation in HostOperations.values() do
```
arbulu89, line 80: I think we need to be more picky on this one, and leave it as it was.
Here we just want launch a saptune discovery, and most probably not all host operations need this

lib/trento/infrastructure/operations/operations.ex
```diff
@@ -50,4 +50,10 @@ defmodule Trento.Infrastructure.Operations do
   def map_operation_type("saptunechangesolution@v1"), do: :saptune_solution_change
 
   def map_operation_type(_), do: :unknown
+
+  def map_operation(:saptune_solution_apply), do: "saptuneapplysolution@v1"
```
arbulu89, line 54: nitpick:
i would remove the newlines in `map_operation` and `map_operation_type`.

lib/trento/operations/host_policy.ex
```diff
@@ -23,7 +25,7 @@ defmodule Trento.Operations.HostPolicy do
         } = host,
         _
       )
-      when operation in [:saptune_solution_apply, :saptune_solution_change] do
+      when operation in HostOperations.values() do
```
arbulu89, line 28: the same here, this is just for saptune operations, not most probably all host operations

arbulu89 (commented):
I think we need to do in certain places a distinction between host operation and saptune operation, as all saptune operations are host operations, but not the opposite

lib/trento/infrastructure/operations/amqp/processor.ex
nelsonkopliku: 0b6e4bd35f1ff67eb3a01e0f436f87e89dbf4bd8

lib/trento/operations/host_policy.ex
nelsonkopliku: 0b6e4bd35f1ff67eb3a01e0f436f87e89dbf4bd8

arbulu89 (approved):
great

PR #3501, opened 2025-05-12
Change saptune solution UI

nelsonkopliku:
# Description

This PR adds a new action to the hosts' operations to change a saptune solution.

Follow up: make sure the list of selectable solutions does not contain the already applied one.

## How was this tested?

Automated tests.


assets/js/pages/HostDetailsPage/HostDetails.test.jsx
```diff
@@ -388,138 +393,105 @@ describe('HostDetails component', () => {
   });
 
   describe('operations', () => {
-    it('should run Saptune apply operation when clicking apply button in modal', async () => {
```
nelsonkopliku, line 391: I had a hard time with this test particularly. I attempted many combinations of `await act(...)`, `await waitFor(...)` but I could not make it less flaky.

I thought that:
- at the end what was being tested here is already tested in `SaptuneSolutionOperationModal` tests
- the test that was attempted here sounds more suitable for a cypress e2e test

.github/workflows/pr_env.yaml
```diff
@@ -22,6 +22,13 @@ jobs:
         run: echo "create_env=${{ contains(fromJson(vars.PR_ENV_ALLOW_LIST), github.event.sender.login) && contains(github.event.pull_request.labels.*.name, vars.PR_ENV_LABEL) }}" >> "$GITHUB_OUTPUT"
 
   build-and-push-pr-image:
+    strategy:
```
arbulu89, line 25: Why this change? I don't think we need we need an additional temporary image per build.
I guess this is a side-effect of not having the operations enabled in prod.
Let's wait until having them once we have the release

assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.jsx
```diff
@@ -52,7 +52,10 @@ const groupedAbilities = [
   {
     ability: 'operation:all',
     tooltip: 'Permits running operations in all resources',
-    groupAbilities: ['saptune_solution_apply:host'],
+    groupAbilities: [
+      'saptune_solution_apply:host',
```
arbulu89, line 56: not your fault at all, but I will need to fix many conflict resolutions in my `cluster_maintenance_change` code that is based in other PRs that I cannot merge due the merge freeze.
just to release my frustration back again :sweat_smile:

arbulu89 (commented):
I don't think we should touch anything in the `pr_env` workflow. This is again needed due the slow release process we are suffering that doesn't allow us to remove the operations flag.

Besides that, the rest of the code looks great

.github/workflows/pr_env.yaml
nelsonkopliku: Yes, I don't intend to merge changes to the pipeline. Was trying things out.
Will revert it

arbulu89 (approved):
:rocket:

EMaksy (approved):
Great job man, sorry for the late review :)

PR #3502, opened 2025-05-14
Alerting settings from env coerced correcly

skrech:
# Description

Fixes a bug where we don't coerce values coming from OS env-vars to proper types for the AlertingSettings struct.

## How was this tested?

Added new UT covering this case.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -153,9 +153,18 @@ if config_env() in [:prod, :demo] do
   # See https://hexdocs.pm/swoosh/Swoosh.html#module-installation for details.
 
   config :trento, :alerting,
-    enabled: System.get_env("ENABLE_ALERTING"),
+    enabled:
+      (case System.get_env("ENABLE_ALERTING") do
+         nil -> nil
+         "true" -> true
+         _ -> false
+       end),
     smtp_server: System.get_env("SMTP_SERVER"),
-    smtp_port: System.get_env("SMTP_PORT"),
+    smtp_port:
+      (case Integer.parse(System.get_env("SMTP_PORT", "")) do
+         :error -> nil
+         {number, _} -> number
+       end),
     smtp_username: System.get_env("SMTP_USER"),
     smtp_password: System.get_env("SMTP_PASSWORD"),
     sender_email: System.get_env("ALERT_SENDER"),
--- a/lib/trento/settings.ex
+++ b/lib/trento/settings.ex
@@ -242,9 +242,7 @@ defmodule Trento.Settings do
 
   defp get_alerting_settings_from_app_env do
     explicitly_set =
-      Application.get_env(:trento, :alerting)
-      |> Enum.filter(fn {_key, value} -> value != nil end)
-      |> Enum.map(&coerce_alerting_env_value/1)
+      Enum.filter(Application.get_env(:trento, :alerting), fn {_key, value} -> value != nil end)
 
     settings =
       struct!(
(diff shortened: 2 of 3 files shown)


arbulu89 (commented):
@skrech I don't think this is a good idea. We shouldn't deal with command line and env variables logic in the domain logic in my opinion. The runtime.exs code is responsible of giving valid values.  The "anomaly" of getting boolean values as string is a environment variables limitation, and it should be fixed in the place where we get them.
What do you think?

skrech:
> @skrech I don't think this is a good idea. We shouldn't deal with command line and env variables logic in the domain logic in my opinion. The runtime.exs code is responsible of giving valid values. The "anomaly" of getting boolean values as string is a environment variables limitation, and it should be fixed in the place where we get them. What do you think?

Yes, I agree actually. I don't know why I'm reluctant to put logic in `runtime.exs`. Maybe because it can't be tested. Anyway, when thinking about it, it seems a good place to coerce values coming from OS env-vars. 

I'll re-work it.

config/runtime.exs
```diff
@@ -153,9 +153,18 @@ if config_env() in [:prod, :demo] do
...
-    enabled: System.get_env("ENABLE_ALERTING"),
+    enabled:
+      (case System.get_env("ENABLE_ALERTING") do
+         nil -> nil
+         "true" -> true
+         _ -> false
+       end),
     smtp_server: System.get_env("SMTP_SERVER"),
-    smtp_port: System.get_env("SMTP_PORT"),
+    smtp_port:
```
arbulu89, line 163: This could be string according the defaults as well

lib/trento/settings.ex
```diff
@@ -61,7 +61,8 @@ defmodule Trento.Settings do
     smtp_username: "",
     smtp_password: "",
     sender_email: "alerts@trento-project.io",
-    recipient_email: "admin@trento-project.io"
+    recipient_email: "admin@trento-project.io",
+    enforced_from_env: true
```
arbulu89, line 65: Is this `enforced_from_env` still needed? (the same for the test file)

arbulu89 (approved):
Nice! I like the solution

lib/trento/settings.ex
skrech: Yes, it's needed. Actually, it was bug to not have it.

config/runtime.exs
skrech: I'm not sure if I get that right. But if you're asking why I've set `""` as default value for the `get_env` call for smtp port -- it's a little trick to skip using an `if` expression. `Integer.parse` doesn't work with `nil` value, raises and exception. So, I've made it such that if SMTP_PORT is not defined, the argument to Integer.parse would be `""` and it would return `:error`, then final value would be `nil`. This null will be converted to default value in the domain  logic.

If you think it's confusing, I could change it to be more explicit.

arbulu89: No, I'm saying the `smtp_port` is of `String.t() | number` in the internal domain code. 
I guessed that this means that using strings for that value is ok

skrech: Ah... this is the type when creating/updating settings from the API. This would be converted to integer in Ecto `cast`. When getting the settings, we expect that to be integer-only. 
This is mainly for consistency. If I'm not mistaken, Swoosh library works with both int and string for smtp_port. The OpenAPI spec for a GET request also expect integer-only smtp_port.

arbulu89: Thank you for having a look

PR #3509, opened 2025-05-16
Version bump automation

stefanotorresi:
# Description

This PR reduces the process of performing a release on GitHub to just editing one single `VERSION` file and then commit/merge the change to main.

To this effect, several changes were performed. This PR:
- Introduces the new `VERSION` file, a static in-tree source of truth for production releases. Note that this value is still overridable by a corresponding `VERSION `environment variable, and this is in-fact leveraged for non-stable builds to compose a detailed version string like `0.1.2+git.25.1747221238.6a63a0f7`, which is generated via the existing dedicated shell script. 
- Improves said `get_version_from_git.sh` script with a more precise regex and better error handling, which was actually done months ago contextually to an attempt to extract the script into the [CD container image](https://github.com/trento-project/continuous-delivery/blob/main/scripts,). I wasn't able to work around the presence of a copy of the script in this repo at this time, due to GitHub Actions quirks, but eventually we'll be able to get rid of it by extracting it in a dedicated custom Action. Saved for later.
- Addresses https://github.com/trento-project/web/pull/735#discussion_r934579845 by introducing the VERSION docker build argument in the Dockerfile, which is then propagated as a build environment variable. This makes the version value overridable without running that shell script at build time, thus completely decoupling it from the mix configuration.
- Updates the consumption of `gh_release_to_obs_changeset.py` to use the shared copy in https://github.com/trento-project/continuous-delivery/blob/main/scripts, hence removing it from this repo.
- Changes the e2e test of the About page, because it was asserting the version number against `tests/e2e/package.json`, which was never a source of truth for the version number set at build time. To make the test pass, we previously had to also update the that file at every release. The test now correctly reads this value from the new root `VERSION` file.
- Extracts container builds and OBS jobs in respective dedicated reusable GitHub Actions workflows. This was needed to be able to build new artifacts on release without repeating all the tests (which was a gigantic chicken-and-egg problem as long the build jobs were coupled with the tests in the same workflow. The tests now happen always _before_ the release, not after.).
  This coincidentally also makes the main `ci.yaml` workflow file marginally more digestible. There is still room for improvement, though.
- Embeds the previously introduced `pre-release.yaml` workflow into the new `release.yaml` one, while also fully addressing the shortcomings of that workflow by introducing fully automatic `CHANGELOG.md` updates, as originally intended.
- Adds new `CHROMATIC_ENABLED` and `OBS_ENABLED` repository variables so that third-party services can be explicitly enabled/disabled. This is convenient when doing development in a fork, and to work around third-party outages.
- Changes the OBS flow so that commits triggered by main branch pushes and releases are now decoupled, targeting entirely different projects set by the new repository variables `OBS_PROJECT_STABLE` and `OBS_PROJECT_ROLLING`. This way, the `main` branch merge freeze doesn't need to outlast a release to avoid undesired OBS commits, thus greatly reducing the freeze window.

- [x] **DONE**

## How was this tested?

Manually in my own fork https://github.com/stefanotorresi/web

- [x] **DONE**

Documentation will come separately.

- [ ] **DONE**


mix.exs
```diff
@@ -237,12 +236,16 @@ defmodule Trento.MixProject do
...
+  defp get_version do
+    version = System.get_env("VERSION", get_version_from_file())
 
-  defp get_version_from_git do
-    case File.cwd!() |> Path.join("hack/get_version_from_git.sh") |> System.cmd([]) do
-      {version, 0} -> version |> String.trim("\n")
-      _ -> @version
+    case version do
+      "" -> get_version_from_file()
+      _ -> version
```
nelsonkopliku, line 244: **question:** what is the need for this `case`?

I am assuming it covers the scenario where there is a `VERSION` env var, however it's content is an empty string.

In that case what we could do is the following:
```elixir
case System.get_env("VERSION", "") do
  "" -> get_version_from_file()
  version -> version
end
```

.github/workflows/ci.yaml
```diff
@@ -269,7 +273,7 @@ jobs:
     name: Chromatic deployment
     needs: [elixir-deps, npm-deps]
     runs-on: ubuntu-24.04
-    if: github.event_name != 'repository_dispatch' && github.secret_source != 'None'
+    if: github.event_name != 'repository_dispatch' && github.secret_source != 'None' && vars.ENABLE_CHROMATIC
```
nelsonkopliku, line 276: **question:** I could not find `ENABLE_CHROMATIC` in the project's settings. It iss still to be added, correct?

nelsonkopliku (commented):
Very nice, thanks for this!

There's many moving parts, but that's what it is 😄 

We're going to apply the `VERSION` file strategy across the board, correct?

.github/workflows/ci.yaml
stefanotorresi: yes, I will add the repository variable once I pull the draft; I'm still unsure about the name, but the intent is to add a variable for each third party service we need to interact with, so that in forks all of them are disabled by default, because they won't work by default. Unless I'm missing some, at the moment we have OBS, Chromatic and Coveralls.

mix.exs
stefanotorresi: I wasn't sure about the entire thing, I'll follow the suggestion!

stefanotorresi:
> We're going to apply the `VERSION` file strategy across the board, correct?

yes

mix.exs
```diff
@@ -237,12 +236,14 @@ defmodule Trento.MixProject do
...
   end
 
-  defp get_version, do: System.get_env("VERSION", get_version_from_git())
-
-  defp get_version_from_git do
-    case File.cwd!() |> Path.join("hack/get_version_from_git.sh") |> System.cmd([]) do
-      {version, 0} -> version |> String.trim("\n")
-      _ -> @version
+  defp get_version do
+    case System.get_env("VERSION") do
```
nelsonkopliku, line 240: Be careful here: `System.get_env("VERSION")` returns `nil` if the env var `VERSION` is not found.

I guess we actually need `System.get_env("VERSION", "")` which returns an empty string if `VERSION` is not defined as en env var, effectively covering all cases.

stefanotorresi: ohh nice catch.

.github/workflows/ci.yaml
stefanotorresi: renamed to `CHROMATIC_ENABLED` to follow the convention of having a common prefix for variables related to the same service.

stefanotorresi:
I don't think the pr env worfklow is intended to work in PR made from forks, so I'm afraid we can't test that until the changes land in main. Besides, not much has changed there, the container build _should_ work as expected, because it works in the other workflows and the only difference is the tag used in the input.

nelsonkopliku (approved):
:shipit:

PR #3514, opened 2025-05-21
Update host saptune status on SAP instances de/registration

arbulu89:
# Description

Update host saptune status when a database/application instance is registered or deregistered.
The saptune status depends on having a SAP workload running in the host. That's why, some SAP changes in a host are not properly updated in our domain.
This happens because saptune discovery data is not often sent, only every 30 minutes.

With this event handler, we listen to events that can have some impact on this.
It reuses the already stored saptune status, the difference is the `sap_running` flag, which is set to true if some instance is running in the host.

## How was this tested?

UT and some additional manual testing


lib/trento/infrastructure/commanded/event_handlers/saptune_update_event_handler.ex
```diff
@@ -0,0 +1,210 @@
...
+        },
+        _
+      ) do
+    ApplicationInstanceReadModel
+    |> where([s], s.sap_system_id == ^sap_system_id)
+    |> Repo.all()
+    |> handle_instances_deregistered(true)
+  end
+
+  defp maybe_dispatch_update_saptune_statue(nil, _), do: :ok
```
nelsonkopliku, line 134: I was imagining statues being thrown all over the places ehehe
```suggestion
  defp maybe_dispatch_update_saptune_status(nil, _), do: :ok
```

nelsonkopliku (approved):
Hey @arbulu89 it looks good!

Besides a typo, I believe we have the opportunity to move some functions in respective contexts

gagandeepb (approved):

nelsonkopliku (approved):
Thanks!

PR #3516, opened 2025-05-21
More ci fixes

stefanotorresi:
# Description

Fixes the pr_env worfkflow and makes the reusable build-containers one usable in other repos.


arbulu89 (approved):
Nice!
It looks it is fine now.
We can merge once everything in the pipeline is green

PR #3517, opened 2025-05-22
Table top margin

arbulu89:
# Description

The Table top margin usage was creating some issues. Depending on having filters or not it was different.

This PR changes the behaviour on that, only putting certain styles if the filters are available. Besides that, now the Table component accepts external class names. This means, that I have to pass the top margin to some tables.

These are all the tables we have. I have added the top margin in these ones:
**When I put top margin added, it means that I change the code to stay as it was until**
- ActivityLogOverview - Top margin added
- PatchList - Top margin added
- UpgradablePackagesList - Top margin added
- AscsErsClusterDetails  - Top margin added
- AttributeDetails (in cluster details) - Top margin added
- ClustersList - We don't need a margin, filters are used
- HanaClusterSite - Now, the error we have in the top is fixed - Check now if it is correct
- DatabasesOverview - We don't need a margin, filters are used
- ExecutionResults - Now, the error we have in the top is fixed - Check now if it is correct
- HomeHealthSummary - I don't think we need additional margin
- HostDetails - Top margin added in the 2 tables
- HostsList - We don't need a margin, filters are used
- SapSystemDetails/DatabaseDetails - Top margin added
- SapSystemOverview - We don't need a margin, filters are used
- Users - Top margin added

## How was this tested?

Visual inspection

### DIFF
--- a/assets/js/common/ActivityLogOverview/ActivityLogOverview.jsx
+++ b/assets/js/common/ActivityLogOverview/ActivityLogOverview.jsx
@@ -112,6 +112,7 @@ function ActivityLogOverview({ activityLog, loading = false }) {
         config={activityLogTableConfig}
         data={activityLog.map(toRenderedEntry)}
         emptyStateText={loading ? 'Loading...' : 'No data available'}
+        roundedTop={false}
       />
     </>
   );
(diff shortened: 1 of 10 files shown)


jagabomb (approved):
LGTM... For some of the changes, I can't see the difference when comparing this **PR Env** to the **Demo** environment. Regardless, I think the improvement fix is looking good. Thanks!

arbulu89:
> LGTM... For some of the changes, I can't see the difference when comparing this **PR Env** to the **Demo** environment. Regardless, I think the improvement fix is looking good. Thanks!

I guess the ones missing are PatchList and UpgradablePackagesList.
Don't worry, i have checked them and they stays as they were before

PR #3529, opened 2025-05-26
Cluster maintenance operation frontend

arbulu89:
# Description
Add cluster maintenance change operation frontend code.

cluster_maintenance_pr

## How was this tested?

UT

### DIFF
--- a/assets/js/pages/ClusterDetails/ClusterDetails.jsx
+++ b/assets/js/pages/ClusterDetails/ClusterDetails.jsx
@@ -1,6 +1,6 @@
 import React, { useState } from 'react';
 
-import { get } from 'lodash';
+import { get, noop } from 'lodash';
 
 import classNames from 'classnames';
 import {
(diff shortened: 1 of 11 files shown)


assets/js/pages/ClusterDetails/ClusterDetails.jsx
```diff
@@ -27,12 +38,19 @@ function ClusterDetails({
   hasSelectedChecks,
   hosts,
   lastExecution = {},
+  operationsEnabled = false,
+  runningOperation = {},
   selectedChecks,
   userAbilities,
   onStartExecution = () => {},
+  onRequestOperation = () => {},
+  onCleanForbiddenOperation = () => {},
```
nelsonkopliku, line 47: **nit:** not a biggie, however I am wondering whether we should be consistent and either use a `() => {}` or lodash's `noop`. I force myself to use `noop`, but no strong opinions.

nelsonkopliku (approved):
Noice! Hope we don't get much flakiness in tests 🙈

EMaksy (approved):

PR #3533, opened 2025-05-27
Show relevant saptune solution options

nelsonkopliku:
# Description

Also, include relevant motivation and context for this PR. List any dependencies that are required for this change.

Fixes # (issue)

- [ ] **DONE**

## How was this tested?

- [ ] **DONE**

- [ ] **DONE**


assets/js/pages/HostDetailsPage/HostDetails.jsx
```diff
@@ -117,6 +117,7 @@ function HostDetails({
   const saptuneVersion = get(saptuneStatus, 'package_version');
   const saptuneConfiguredVersion = get(saptuneStatus, 'configured_version');
   const saptuneTuning = get(saptuneStatus, 'tuning_state');
+  const currentlyAppliedSolution = get(saptuneStatus, 'enabled_solution.id');
```
nelsonkopliku, line 120: I am wondering whether we should use `applied_solution` here

assets/js/common/OperationModals/SaptuneSolutionOperationModal.jsx
```diff
@@ -54,25 +56,39 @@ function SaptuneSolutionOperationModal({
...
       checked={checked}
       isOpen={isOpen}
       onChecked={() => setChecked((prev) => !prev)}
-      onRequest={() => onRequest(solution)}
-      onCancel={() => {
-        onCancel();
-        setSolution(NOT_SELECTED);
-        setChecked(false);
-      }}
+      onRequest={() => pipe(onRequest, () => setChecked(false))(solution)}
```
arbulu89, line 90: curiosity:
why do we need a pipe here if the result of one is not used in the next.
cannot we just not use it and avoid its execution?
The same in the `onCancel`

assets/js/common/OperationModals/SaptuneSolutionOperationModal.jsx
```diff
@@ -54,25 +56,39 @@ function SaptuneSolutionOperationModal({
...
+  };
 
   return (
     <OperationModal
       title={title}
       description="Select Saptune tuning solution"
       operationText="Saptune solution"
-      applyDisabled={!checked || solution === NOT_SELECTED}
+      applyDisabled={
+        !checked || solution === NOT_SELECTED || solution === currentlyApplied
```
arbulu89, line 85: Do we really need the `solution === currentlyApplied`?
You cannot never select the current solution, it is disabled

assets/js/common/OperationModals/SaptuneSolutionOperationModal.jsx
```diff
@@ -54,25 +56,39 @@ function SaptuneSolutionOperationModal({
...
+    filter(({ available }) => available(isHanaRunning, isAppRunning)),
+    map((solutionOption) => ({
+      ...solutionOption,
+      disabled: solutionOption.value === currentlyApplied,
+    }))
+  )(solutions);
+
+  const resetSolution = () => setSolution(currentlyApplied || NOT_SELECTED);
+
+  useEffect(() => {
```
arbulu89, line 69: I'm not sure if we should have this `useEffect`. Here, if the solution changes for example from `null` to something else, the modal updates, but the whole concept changes as well, as we the user might not even notice that the scenario is different, as we maybe are not in a position to apply, and we are in a position to change.

arbulu89 (commented):
Hey @nelsonkopliku ,
Looking the code, I'm wondering in having the `useEffect` and updating the modal in the fly depending on the currently applied solution is a good idea.
If the user is in the modal, and the state changes, maybe we should even close the modal, or do nothing and let the backend fail.

I imagine:
- I'm in a Solution apply modal
- Suddenly, the solution changes in the backend (somebody doing command line things) and the modal options changes at this point
- First of all, the user might get surprised and confused, and second, the user still is in a `Apply` modal (with its title, etc), but we changed the solutions in the select with `solution change` logic.

I don't know...
Maybe we should differentiate the solution entries with something more static and don't have sudden updates.

What do you think?
Asking to @jagabomb @abravosuse as well

assets/js/common/OperationModals/SaptuneSolutionOperationModal.jsx
nelsonkopliku: That is a fair point I was fighting against as well.
For simplicity I agree that we can avoid such reactiveness.
Thanks

assets/js/common/OperationModals/SaptuneSolutionOperationModal.jsx
```diff
@@ -9,7 +10,7 @@ const solutions = [
   {
     value: NOT_SELECTED,
     key: 'not_selected',
-    available: () => true,
+    available: (_hana, _app, currentlyApplied) => !currentlyApplied,
```
arbulu89, line 13: I think we should add the usage of the 3rd variable in all `available` entries, even if it is a `_`.

assets/js/common/OperationModals/SaptuneSolutionOperationModal.test.jsx
```diff
@@ -148,4 +148,49 @@ describe('SaptuneSolutionOperationModal', () => {
...
+          isHanaRunning
+          currentlyApplied="HANA"
+        />
+      );
+    });
+
+    await user.click(screen.getByRole('checkbox'));
+    await user.click(screen.getByRole('button', { name: 'HANA' }));
+
+    expect(screen.getByRole('option', { name: 'HANA' })).toHaveAttribute(
```
arbulu89, line 191: Doesn't `.toBeDisabled()` work?

assets/js/pages/HostDetailsPage/HostDetails.jsx
arbulu89: Yes, maybe it would be better. But I think it will be the same always

arbulu89 (approved):

assets/js/common/OperationModals/SaptuneSolutionOperationModal.test.jsx
nelsonkopliku: not really as it is a div, not an actual button

PR #3535, opened 2025-05-28
Harden saptune operation policy

nelsonkopliku:
# Description

This PR hardens the host policy with regards to saptune operations.

Makes sure to not issue operations when the current state does not allow it
- it is not possible to apply a solution when there is an already applied one
- it is not possible to change a solution when there isn't another one applied

Checking whether the requested solution matches or not the one in the host is left to the operator.

EDIT:
the main change is in the first commit ddd7ed5ce6002c5435b6b8291b1169c452edf677
The secon commit fd1ab5d046cdf492fca8d6d4352f6d3c78f622c6 is mainly side effect required adjustments

Of course there is an API bc break https://github.com/trento-project/web/actions/runs/15303383262/job/43049718634?pr=3535 but since the previous state was just wrong, not sure it is worth a bump

### DIFF
--- a/lib/trento/operations/host_policy.ex
+++ b/lib/trento/operations/host_policy.ex
@@ -71,38 +71,28 @@ defmodule Trento.Operations.HostPolicy do
 
   defp get_saptune_operation_resource_id(_), do: nil
 
-  defp authorize_saptune_solution_operation(
-         :saptune_solution_apply,
-         saptune_status
-       ) do
-    case saptune_status do
-      nil ->
-        :ok
+  defp authorize_saptune_solution_operation(:saptune_solution_apply, nil), do: :ok
 
-      %{applied_solution: nil} ->
-        :ok
+  defp authorize_saptune_solution_operation(:saptune_solution_apply, %{applied_solution: nil}),
+    do: :ok
 
-      _ ->
-        {:error,
-         ["Cannot apply the requested solution because there is an already applied on this host"]}
-    end
-  end
+  defp authorize_saptune_solution_operation(:saptune_solution_apply, _),
+    do:
+      {:error,
+       ["Cannot apply the requested solution because there is an already applied on this host"]}
 
-  defp authorize_saptune_solution_operation(
-         :saptune_solution_change,
-         saptune_status
-       ) do
-    case saptune_status do
-      %{applied_solution: applied_solution} when not is_nil(applied_solution) ->
-        :ok
+  defp authorize_saptune_solution_operation(:saptune_solution_change, %{
+         applied_solution: applied_solution
+       })
+       when not is_nil(applied_solution),
+       do: :ok
 
-      _ ->
-        {:error,
-         [
-           "Cannot change the requested solution because there is no currently applied one on this host"
-         ]}
-    end
-  end
+  defp authorize_saptune_solution_operation(:saptune_solution_change, _),
+    do:
+      {:error,
+       [
+         "Cannot change the requested solution because there is no currently applied one on this host"
+       ]}
 
   # Classic SAP HANA setup
   defp find_resource_id(%{type: "ocf::suse:SAPHana", parent: %{id: id}}), do: id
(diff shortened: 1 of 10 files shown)


lib/trento/operations/host_policy.ex
```diff
@@ -64,6 +71,39 @@ defmodule Trento.Operations.HostPolicy do
 
   defp get_saptune_operation_resource_id(_), do: nil
 
+  defp authorize_saptune_solution_operation(
```
arbulu89, line 74: nitpick:
```
authorize_saptune_solution_operation(:saptune_solution_apply, nil), do: :ok
authorize_saptune_solution_operation(:saptune_solution_apply, %{applied_solution: nil}), do: :ok
authorize_saptune_solution_operation(:saptune_solution_apply, _), do: {:error,
         ["Cannot apply the requested solution because there is an already applied on this host"]}
authorize_saptune_solution_operation(:saptune_solution_change, %{applied_solution: applied_solution}) when not is_nil(applied_solution) -> :ok
authorize_saptune_solution_operation(:saptune_solution_change, _), do: {:error,
         [
           "Cannot change the requested solution because there is no currently applied one on this host"
         ]}
```

test/trento/operations/host_policy_test.exs
```diff
@@ -14,11 +14,33 @@ defmodule Trento.Operations.HostPolicyTest do
     assert {:error, ["Unknown operation"]} == HostPolicy.authorize_operation(:unknown, host, %{})
   end
 
+  describe "Forbidding saptune operations" do
```
arbulu89, line 17: Letfover describe?

test/trento_web/controllers/v1/host_controller_test.exs
```diff
@@ -378,12 +378,36 @@ defmodule TrentoWeb.V1.HostControllerTest do
...
         |> assert_schema("NotFound", api_spec)
       end
+    end
 
-      test "should forbid operation '#{saptune_operation}' if conditions are unmet", %{
-        conn: conn,
-        api_spec: api_spec
-      } do
-        %{id: host_id} = insert(:host)
+    scenarios = [
```
arbulu89, line 383: nitpick: is this not an overkill? I mean, this is already tested in the policy code.
I would say that testing with `saptune_status: build(:saptune_status)` would be enough.
We don't neither test with app instances, not app instances, db instances, etc

lib/trento_web/controllers/v1/host_json.ex
```diff
@@ -60,7 +62,7 @@ defmodule TrentoWeb.V1.HostJSON do
   def saptune_status_updated(%{
         host: %{id: id, saptune_status: status, hostname: hostname}
       }),
-      do: %{id: id, status: status, hostname: hostname}
+      do: %{id: id, status: StructHelper.to_atomized_map(status), hostname: hostname}
```
arbulu89, line 65: This concerns me. Why we are converting something to atom-keys when actually, phoenix will reconvert this back to json/string?

arbulu89 (commented):
Hey @nelsonkopliku 
The hardening part looks good.

I'm wondering why you changed the `json` view code thing.
The phoenix view doesn't care if he gets atoms/strings as keys. It will always convert them to json. So it doesn't make much sense to convert to atoms.

I guess that all of this is caused for adding new things to the factory. All saptune status data coming from the database should already have atomized keys.

lib/trento/operations/host_policy.ex
nelsonkopliku: thanks aab2e6405a0867f73ed5ca7323c1ba1d83e70eaa

lib/trento_web/controllers/v1/host_json.ex
nelsonkopliku: Because of what you mentioned
> I guess that all of this is caused for adding new things to the factory. All saptune status data coming from the database should already have atomized keys.

nelsonkopliku: here we go https://github.com/trento-project/web/pull/3535/commits/eeda9c4f095e372cfc02bd2cd5d6278b2b2a1e34

PR #3537, opened 2025-05-29
Fix start/stop AgentHeartbeat logic in before/after Each hooks in e2e tests

vicenteqa:
# Description

Use beforeEach instead of before so startAgentHeartbeat is called per every test, apply the same to afterEach and stopAgentHeartbeat. This fixes the issue with the cypress config crashing.

Bonus: Add always() to analyze step so it is executed even if some threads failed. Decrease timeout for every thread to 25 min (the whole test suite should take around 20 normally) 

Fixes # (issue)

- [ ] **DONE**

## How was this tested?

- [ ] **DONE**

- [ ] **DONE**


gagandeepb (approved):
Thanks for this! LGTM.

PR #3565, opened 2025-06-11
SAP instance start/stop policy

arbulu89:
# Description

**Spoiler alert: This is a pretty complex implementation**

Add SAP instance start/stop operations policies.
Some instance depend on others and the database to start/stop.
In our trento internal usage: `passing` means `running` and `uknown` means `stopped/gray`.
These are the conditions:

`sap_instance_start`:
- Message server instance `MESSAGESERVER|ENQUE` (or ASCS): it can start without any dependency
- Enque replicator `ENQREP` (or ERS):  it needs the Message server running
- Other instances (`ABAP|...`, `J2EE|...`): they need the Message server and the database associated to the system running

`sap_instance_stop`:
- Message server instance: all the remaining instances must be stopped
- Other instances: they can stop without any dependency
- **the stop operation doesn't have any dependency with the database)**

`cluster_maintenance`:
For both start/stop, the `cluster_maintenance` policy must be changed (which impacts and improves other operations, like `saptune_solution_apply`, which have the same need).
Now, instead of just looking for the whole cluster maintenance value, we check for the `SAPInstance` resource managed state as well.
As `ASCS/ERS` clusters have at least 2 of this resources, we need to match with the current application instances `sid/instance_number` to find the resource.

## How was this tested?

UT

### DIFF
--- a/lib/trento/operations/application_instance_policy.ex
+++ b/lib/trento/operations/application_instance_policy.ex
@@ -19,17 +19,42 @@ defmodule Trento.Operations.ApplicationInstancePolicy do
   def authorize_operation(
         :maintenance,
         %ApplicationInstanceReadModel{} = application_instance,
-        params
+        _params
       ) do
     OperationsHelper.reduce_operation_authorizations([
       instance_running(application_instance),
-      cluster_maintenance(application_instance, params)
+      cluster_maintenance(application_instance)
     ])
   end
 
-  def authorize_operation(operation, %ApplicationInstanceReadModel{}, _)
-      when operation in [:sap_instance_start, :sap_instance_stop] do
-    :ok
+  # instance start operation authorized when:
+  # - other instances in the system are started
+  # - database is started
+  # - cluster is in maintenance
+  def authorize_operation(
+        :sap_instance_start,
+        %ApplicationInstanceReadModel{} = application_instance,
+        _params
+      ) do
+    OperationsHelper.reduce_operation_authorizations([
+      other_instances_started(application_instance),
+      database_started(application_instance),
+      cluster_maintenance(application_instance)
+    ])
+  end
+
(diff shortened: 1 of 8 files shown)


abravosuse (commented):
Hey @arbulu89 . The description of the PR looks good to me. Other than that I am not able to provide a more valuable review at the moment. Maybe when/if a PR environment becomes available in the future, I will be able to test it out and give you a more insightful feedback..

arbulu89:
> Hey @arbulu89 . The description of the PR looks good to me. Other than that I am not able to provide a more valuable review at the moment. Maybe when/if a PR environment becomes available in the future, I will be able to test it out and give you a more insightful feedback..

Yes, I wanted confirmation of the rules I wrote in the PR description

arbulu89:
> Hey @arbulu89 . The description of the PR looks good to me. Other than that I am not able to provide a more valuable review at the moment. Maybe when/if a PR environment becomes available in the future, I will be able to test it out and give you a more insightful feedback..

@abravosuse anyway, I will produce a PR with this once I have the frontend thing sorted out, so we can test in real infra.

arbulu89:
> Since we do have the sap system id, what do you think about actually loading the sap system to start with and work with its related entities like the instances and what is needed.
> 
> At the end the _entry point_ is the sap system, and we might save some possibly unnecessary gymnastics.

The thing is that I need to return an `ApplicationInstanceReadModel`, so if I load a SAP system, a still need to get the instance itself to be returned. I need to do this as the policy is about this particular instance.
And at the end, the joins must exist one way or another (the cluster information is needed).

What I can do is to just instead of preloading the `application_instances`, just put the data that was loaded initially in the first query. 
Besides that, the number of queries and joins is exactly the same whether we load the instances or system first.

PR #3566, opened 2025-06-11
Add transparent prop to operations button

arbulu89:
# Description

Add `transparent` prop to `OperationsButton` to make it usable in contextual operation places.
I added just one prop and not all the class options as I don't expect the need to configure the component further, so fine grained classes don't look needed.

Example (not implemented in this PR, this is just a usage example):
op-button-transparent

## How was this tested?

UT


EMaksy (approved):
Nice!

PR #3575, opened 2025-06-16
Add operations button to SAP instance in table

arbulu89:
# Description

Add operations button to the SAP instance in table.

I continue adding things step by step, instead of a big change in one PR. The request and the running operation display are not that trivial.

As the `GenericSystemDetails` is a shared component between SAP systems and HANA databases, I have added a new prop` (`getInstanceOperations`), so we can pass the operations for each of the independently.
This way, we don't need to touch the table thing and do different things for each of both types.

I think it gets pretty clean and easy to use.

image

image

## How was this tested?

UT.
As in previous tests of operations button, I have tried to test the `onClick` action, but it is extremely flaky and it doesn't work often...
I think we will test this modals better once we add e2e tests

### DIFF
--- a/assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
+++ b/assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
@@ -1,15 +1,20 @@
 import React, { useState } from 'react';
 
+import { noop } from 'lodash';
+
 import {
   EOS_APPLICATION_OUTLINED,
   EOS_DATABASE_OUTLINED,
 } from 'eos-icons-react';
 
+import { SAP_INSTANCE_START, SAP_INSTANCE_STOP } from '@lib/operations';
 import { APPLICATION_TYPE, getEnsaVersionLabel } from '@lib/model/sapSystems';
 
 import ListView from '@common/ListView';
 import Table from '@common/Table';
 import PageHeader from '@common/PageHeader';
+import { SapInstanceStartStopModal } from '@common/OperationModals';
+
 import DeregistrationModal from '@pages/DeregistrationModal';
 
 import {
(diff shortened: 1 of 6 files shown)


assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
```diff
@@ -30,28 +35,46 @@ const getUniqueHosts = (hosts) =>
...
+    useState(modalInitialState);
+  const [currentOperationInstance, setCurrentOperationInstance] =
+    useState(undefined);
 
   const onCleanUpClick = (instance) => {
     setCleanUpModalOpen(true);
     setInstanceToDeregister(instance);
   };
 
+  const curriedGetInstanceOperations = getInstanceOperations(
```
EMaksy, line 70: Interesting, i think it is a nice usage of "currying"

assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
```diff
@@ -30,28 +35,46 @@ const getUniqueHosts = (hosts) =>
...
+    null, // placeholder for runningOperations to be used at some point
```
EMaksy, line 71: Just wondering in which direction will this placeholder go in the future ? 
What will it do ?

assets/js/pages/SapSystemDetails/sapOperations.js
```diff
@@ -0,0 +1,33 @@
+import { curry } from 'lodash';
```
EMaksy, line 1: I like the idea of using it, but to be honest i had first read about it :)
It is cleaner then writing some wrappers!

EMaksy (approved):
Hey man good job, thanks for bringing `curry` in, it is an interesting and very clean way.  
Code wise looks good to me!

assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
arbulu89: I will pass the `runningOperations` parameter to set operations in running state.
I wanted to allocate the place, as `curry` function uses the paremeter in the order that are given, and I wanted to have the `runningOperations` first.
You will see in the next PR hehe

abravosuse (approved):
Based on [3575.prenv.trento.suse.com](https://3575.prenv.trento.suse.com/). the flow to stop an instance looks OK to me.

PR #3577, opened 2025-06-16
Request SAP instance operation

arbulu89:
# Description

Add operation request usage to SAP instances. It includes the `runningOperation` usage, so we see the spinning icon in the selected operation.

In order to use the new API call, which has different params, I have moved the logic to the saga itself, and now, we need to pass the `requestParams` from the react components, as different resources need different params to compose the api call.

**TODO**:
- The running operation is not loaded in page refresh.

## How was this tested?
UT


abravosuse (approved):

EMaksy (approved):
LGTM! 
Nothing really to add!

nelsonkopliku:
Sorry @arbulu89 for being late to the party, was in the middle of the review 🙈 

No big deal. Just a clarification question: `sap instance` refers to only application instances?

PR #3578, opened 2025-06-16
Request pacemaker enable/disable

nelsonkopliku:
# Description

This PR adds the ability to request enable/disable pacemaker on a specific cluster node.

With this comes also a proposal for extension of the `OperationsPolicyPlug`, in order to make the operations policy pipeline a bit more flexible:
- resource loading function can be operation specific
- resource loading function also supports ok/error tuple being returned
- params loading function can be operation specific

With regards to general rules the idea is that:
- the operation can be issued only if the cluster exists and it is not deregistered
- the specific host is actually part of the cluster and it is not deregistered

I have doubts about the fact that we might return a 404 when failing to load the cluster, while 403 when the rules about the host are not respected, and this happens because part of this check is in the policy.

I am open to revisit this part.

What's next:
- abilities https://github.com/trento-project/web/pull/3581
- UI
- improved discovery carrying also systemd relevant information to allow disallow a specific operation
- logging relevant entries in the activity log

### DIFF
--- a/test/trento_web/controllers/v1/cluster_controller_test.exs
+++ b/test/trento_web/controllers/v1/cluster_controller_test.exs
@@ -267,7 +267,7 @@ defmodule TrentoWeb.V1.ClusterControllerTest do
     for operation <- ["pacemaker_enable", "pacemaker_disable"] do
       @operation operation
 
-      test "requesting #{operation} should return not found not found when the cluster does not exist",
+      test "requesting #{operation} should return not found when the cluster does not exist",
            %{
              conn: conn,
              api_spec: api_spec
@@ -278,7 +278,7 @@ defmodule TrentoWeb.V1.ClusterControllerTest do
         |> assert_schema("NotFound", api_spec)
       end
 
-      test "requesting #{operation} should return not found not found when the cluster is deregistered",
+      test "requesting #{operation} should return not found when the cluster is deregistered",
            %{
              conn: conn,
              api_spec: api_spec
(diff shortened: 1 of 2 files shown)


test/trento_web/controllers/v1/cluster_controller_test.exs
```diff
@@ -253,6 +253,134 @@ defmodule TrentoWeb.V1.ClusterControllerTest do
...
+      conn
+      |> post("/api/v1/clusters/#{UUID.uuid4()}/hosts/#{UUID.uuid4()}/operations/unknown")
+      |> json_response(:not_found)
+      |> assert_schema("NotFound", api_spec)
+    end
+
+    for operation <- ["pacemaker_enable", "pacemaker_disable"] do
+      @operation operation
+
+      test "requesting #{operation} should return not found not found when the cluster does not exist",
```
arbulu89, line 270: typo: double `not found`
the same in the next test as well

nelsonkopliku:
Hey @arbulu89 thanks a lot for the feedback, I did not think about simply anticipating the assignment of the operation in the pipeline. That's what reviews are for, right? 😄 

That said I also had my doubts about the host belonging check in the policy, so I am fine in streamlining to returning 404 in the mentioned cases.

Honestly I did like the possibility to access to the resolved operation straight away instead of having to pattern match the nested shape of the conn, at least sometimes 🙈 
However I believe this might be sufficient for now and if needed we can revisit and enhance in the future.

The last 4 commits address your feedback 😉

PR #3580, opened 2025-06-17
Fix SAP instance cluster maintenance authorization

arbulu89:
# Description

Fix a mistake I did in the application instance authorization policies. The `cluster_maintenance` must be authorized in SAP instances if the application instance is not clustered (even though other instance in the same host is clustered).

## How was this tested?

UT


nelsonkopliku (approved):
Hey Xabi, not much to say. As long as the behaviour is the expected one looks good.

PR #3581, opened 2025-06-17
Pacemaker enable disable policy

nelsonkopliku:
# Description

This PR adds the relevant abilities for the enable/disable pacemaker operations


arbulu89 (approved):
Nice!
I find `cluster` better than `cluster_host` as the resource, as we don't have the `cluster_host` resource anywhere in our code.

And at the end, the operation affects the cluster itself. I don't see needed the split if at some point in time we provide all the operation abilities individually for the user to choose. 

For example, we don't have this particular ability to put a cluster node/host in maintenance.

nelsonkopliku:
Alright @arbulu89, makes sense. In the doubt of cluster vs host I did a mix 😄 

I am ok with  just `cluster`

PR #3585, opened 2025-06-19
App instance operation activity log

arbulu89:
# Description

Add application instances operation request in activity log.

I have done some improvements in the general usage of the operations in the activity log. 

Now, instead of using `resource_id` to reference the resource where the operation is done, the field name is based in the operation type itself (`host_operation_requested` adds `host_id` for example).
With this, and adding additional metadata fields depending on the type (for application instances, we add `host_id` and `instance_number`), the already existing metadata enricher does its magic and adds everything related to SAP systems, clusters, hosts, etc. It is super nice!
Besides, new operations don't need any change on this (except adding additional params that are not the `id`)

About the operation completed, as we receive a very limited number of information on the rabbitmq message, I have added a logic to extract the params from the operation requested entry itself, based in the `operation_id`. This way, the operation completed entry has the same number of params than the request. Which complements it perfectly.

Have a look into the current PR env, you will see some examples.

Operation requested:

image

Operation completed:

image

## How was this tested?

UT


arbulu89:
> Additionally while it might look smart querying the content of the activity log to populate the activity log itself, it looks a bit hacky and has the drawback of relying on data that might have been deleted by the retention time based clean up.

Activity log of an operation requested 1/5 minutes ago deleted by the retention time clean up? Can this really happen?

I just wanted to implement the activity log in the context of the activity log. No need to send more data in the rabbitmq (data that is only used for the activity log and some extremely minor frontend things, which we don't maybe even need).

I could argue that sending data that is only meant to be used by the activity log to wanda can be hacky as well.
Up to you folks. For me this was a nice idea and solution. No hard feelings, we can ditch it.

gagandeepb (approved):

nelsonkopliku:
Hey @arbulu89, for the sake of completeness and transparency let me bring an extra point we discussed offline:
the minimum settable retention time currently is 1 day meaning that the the system always retains between 24 to 48 hours worth of logs.
This makes it effectively quite impossible to not have an operation requested entry when an operation completed needs to be logged.

I am still not a big fan of such assumption, but yes it is a quite safe one, so my main argument does not stand anymore.

Anyway let's keep the metadata propagation option under the radar to see if that could be useful otherwise.

PR #3586, opened 2025-06-19
Request pacemaker enable UI

nelsonkopliku:
# Description

Disclaimer: 50% of this PR verbosity are tests for `HanaClusterSite` and `AscsErsClusterDetails`.

---

This PR introduces the ability to run pacemaker enable/disable operations on pacemaker cluster hosts.
Following scenarios are covered:
- hana
  - with sited nodes
  - with unsited nodes
- ascs/ers

Since the lock resource for these operations is the cluster (aka group id == cluster id) and at the same time what we are operating on is a subresource - the cluster host -, relevant changes to the state/sagas/selectors have been made to support the following:
- show an operation specific message when unable to reconcile all information on operation completion (related to metadata propagation and not showing `unknown`)
- show the operation running for the correct cluster host (made sure it is consistent also on page reload)

Additional notes:
- `ClusterDetails` has been slightly refactored to reduce some duplication
- checking whether the operations are enabled or not is intentionally not done, hoping we can remove that check soon for everything 👀

### DIFF
--- a/assets/js/state/sagas/operations.js
+++ b/assets/js/state/sagas/operations.js
@@ -1,5 +1,5 @@
 import { all, call, put, select, takeEvery } from 'redux-saga/effects';
-import { map, noop, get } from 'lodash';
+import { map, noop } from 'lodash';
 
 import {
   HOST_OPERATION,
@@ -30,19 +30,14 @@ import {
 import { getHost } from '@state/selectors/host';
 import { getCluster } from '@state/selectors/cluster';
 
-function* fetchHostName(hostID) {
-  return (yield select(getHost(hostID)))?.hostname || 'unknown';
-}
-
-function* getResourceName(groupID, resourceType, requestParams = {}) {
+function* getResourceName(groupID, resourceType) {
   switch (resourceType) {
     case HOST_OPERATION:
     case APPLICATION_INSTANCE_OPERATION:
-      return yield fetchHostName(groupID);
+      return (yield select(getHost(groupID)))?.hostname || 'unknown';
     case CLUSTER_OPERATION:
-      return (yield select(getCluster(groupID)))?.name || 'unknown';
     case CLUSTER_HOST_OPERATION: {
-      return yield fetchHostName(requestParams.hostID);
+      return (yield select(getCluster(groupID)))?.name || 'unknown';
     }
     default:
       return 'unknown';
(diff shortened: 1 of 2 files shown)


assets/js/state/selectors/runningOperations.js
```diff
@@ -1,20 +1,30 @@
...
-export const isOperationRunning = (runningOperationsList, groupID, operation) =>
-  !!find(runningOperationsList, { groupID, operation });
+export const isOperationRunning = (
+  runningOperationsList,
+  groupID,
+  operation,
+  matcher = defaultMatcher
+) =>
+  pipe(
+    castArray,
```
arbulu89, line 27: I don't think we should have this `castArray`. The function expects a list, otherwise the param name above is totally missleading.

nelsonkopliku: Would it be enough to change the param from `runningOperationsList` to something like `runningOperationItemOrList`?

nelsonkopliku: It wouldn't be harmful, but here it is 772726d2a90e2506bc1d6a655370c7132e5e8b1a

assets/js/state/sagas/operations.js
```diff
@@ -28,13 +30,20 @@ import {
 import { getHost } from '@state/selectors/host';
 import { getCluster } from '@state/selectors/cluster';
 
-function* getResourceName(groupID, resourceType) {
+function* fetchHostName(hostID) {
+  return (yield select(getHost(hostID)))?.hostname || 'unknown';
+}
+
+function* getResourceName(groupID, resourceType, requestParams = {}) {
```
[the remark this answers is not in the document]
nelsonkopliku: I know the limitations and I understand your point, however having a possibly unused `requestParams` in the completion flow looked a reasonable and possibly temporary tradeoff - anyway we are protected against the `unknown` cases because the `resourceName` is not user where not needed.

Having the involved operations attached to che cluster - ie `CLUSTER_OPERATION` instead of `CLUSTER_HOST_OPERATION` would require checking inside `callRequest`'s `CLUSTER_OPERATION` case if we want to call a `requestClusterOperation` or a `requestClusterHostOperation` and we'd do that based on the content of `requestParams` and yes, `requestParams` is already part of the `callRequest` signature.

That said, it would mean missing the opportunity to:
- have relevant operations attached to the relevant resource type (because yes it is neither cluster nor host, but something in between 😄 🤷 )
- have some degree of flexibility we anyway need given the various differences in operations nature

If the unused `requestParams` in the completion flow is a big concern, I could further refactor to make sure that does not happen, it would mean some more code though ehehe

arbulu89 (approved):
Nice! Thank you

PR #3590, opened 2025-06-23
Architecture based checks execution

janvhs:
This is the first of a few PRs. This PR implements the required backend changes to the execution environment and updates the frontend to filter out non-applicable checks.


arbulu89 (commented):
Looking good!
We need to change the `ExecutionResultsPage`.
I have proposed a code snippet.
Please, give a local running test to see that it sends the correct filter only in host executions

PR #3592, opened 2025-06-24
Add Architecture to HostInfoBox

janvhs:
# Description

I think it might be a good addition to see the architecture of a host in both the checks-selection and checks-result page. We have space there anyway and by coincidentally it gets rid of the ugly 'just put an element here for the layout' hack  

## How was this tested?

Unit tests

### DIFF
--- a/assets/js/common/HostInfoBox/HostInfoBox.jsx
+++ b/assets/js/common/HostInfoBox/HostInfoBox.jsx
@@ -3,7 +3,7 @@ import React from 'react';
 import ListView from '@common/ListView';
 import ProviderLabel from '@common/ProviderLabel';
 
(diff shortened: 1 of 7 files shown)


abravosuse:
@arbulu89 yes, it would be nice to show in demo a combination of hosts running on x86_64 and ppc64le. But with certain rationale:

- for any given database, all hosts should have the same architecture
- for any SAP systems, all hosts in the application layer should have the same architecture

arbulu89:
@abravosuse the question was more about having the `arch` field in the check selection view

jagabomb:
@arbulu89 @janvhs @abravosuse In terms of the placement of the UI it looks right to me.

PR #3594, opened 2025-06-24
Add a cumulative photofinish demo scenario

nelsonkopliku:
# Description

I found myself in need for a near to demo local environment and the set of scenarios that are used in the pipeline were useful.

I added a cumulative `demo` photofinish scenario both for local usage and pipeline usage.

### DIFF
--- a/.photofinish.toml
+++ b/.photofinish.toml
@@ -1,3 +1,15 @@
+[demo]
+
+directories = [
+    "./test/fixtures/scenarios/healthy-27-node-SAP-cluster",
+    "./test/fixtures/scenarios/hana-scale-out",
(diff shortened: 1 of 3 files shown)


.photofinish.toml
```diff
@@ -444,4 +444,16 @@ files = [
 
 files = [
     "./test/fixtures/scenarios/healthy-27-node-SAP-cluster/9a3ec76a-dd4f-5013-9cf0-5eb4cf89898f_sap_system_discovery.json"
+]
+
+[demo]
```
arbulu89, line 449: nitpick:
i would put this on top, as it is not "just yet another" fixture. It is important, so it doesn't get lost in between other new entries

arbulu89 (approved):

PR #3595, opened 2025-06-24
Reduce wss updates based on cib last written

nelsonkopliku:
# Description

This change reduces the amount of non strictly necessary updates caused by unchanged `cib_last_written` information.

Why is this useful? 
Besides reducing the amount of unnecessary network traffic it reduces the amount of the consequent rerenders of a cluster detail mitigating undesired side effect like the inability to properly use cluster host operations (ie enable/disable pacemaker).


arbulu89 (commented):
-

PR #3596, opened 2025-06-25
Alerting Settings events in the ActivityLog

skrech:
# Description
Including events (of type connection events) for Alerting Settings in the ActivityLog.

## How was this tested?

BE UT, FE UT and E2E.

No docs updated


assets/js/lib/model/activityLog.js
```diff
@@ -19,6 +19,8 @@ export const API_KEY_GENERATION = 'api_key_generation';
 export const SAVING_SUMA_SETTINGS = 'saving_suma_settings';
 export const CHANGING_SUMA_SETTINGS = 'changing_suma_settings';
 export const CLEARING_SUMA_SETTINGS = 'clearing_suma_settings';
+export const CREATING_ALERTING_SETTINGS = 'creating_alerting_settings';
+export const UPDATING_ALERTING_SETTINGS = 'updating_alerting_settings';
```
nelsonkopliku, line 23: What do you think about keeping consistency with suma settings related worfing? 
I mean it can be also that previous is adapted to the new one, but at least it's consistent 😅

nelsonkopliku (commented):
Hey @skrech nice! Just one remark about wording consistency.

assets/js/lib/model/activityLog.js
skrech: Thanks, Nelson. Do you mean naming them `saving_alerting_settings` and `changing_alerting_settings`? If so, yes, I can do that. However, I'll change the names only of activity-related variables and won't change trento_web controller routing names because that might cause alot of unexpected errors.

nelsonkopliku: Good for me

PR #3597, opened 2025-06-25
Update Scenario Fixtures to Include Architecture

janvhs:
# Description

I used an automated way to include the new architecture payload attribute in the host discovery scenarios.

While this works flawlessly, it normalised the formatting of our checks as well. Because of this, the indenting of 2 checks are changed to align with the others and a netmask array formatting got normalized as well.

This is the script I used: https://paste.sr.ht/~janvhs/f5f3b8c43be734be9860e139d3443e8ac7767c25


arbulu89 (commented):
@janvhs I'm afraid we cannot have this automatic value setting :see_no_evil: 
Because it doesn't show a "real" setup.
All machines in the same SAP system must have the same `arch`.
What I would do:
- Put `x86_64` in all hosts
- Select one single HANA setup (which one would be good @abravosuse , `hana-scal-up-cost-opt` for example), and put `ppc64le` in them.

question: the formatting of the last line I guess is something done by `jq`, right?

PR #3598, opened 2025-06-25
Add chevron collapsible table

arbulu89:
# Description

Update Table collapsible rows usage to add a chevron to collapse/expand instead of clicking anywhere in the row.
When a table has collapsible rows, the first cell of the row is reserved to add chevron. 

collapsible_table

## How was this tested?

UT and e2e

### DIFF
--- a/assets/js/pages/ExecutionResults/ExecutionResults.jsx
+++ b/assets/js/pages/ExecutionResults/ExecutionResults.jsx
@@ -42,7 +42,7 @@ const resultsTableConfig = {
       render: (checkID, { customized, onClick }) => (
         <div className="flex whitespace-nowrap text-jungle-green-500 justify-between">
           <span
-            className="inline-flex leading-5"
+            className="inline-flex leading-5 cursor-pointer"
             aria-hidden="true"
             onClick={(e) => {
               e.stopPropagation();
--- a/test/e2e/cypress/e2e/checks_customization.cy.js
+++ b/test/e2e/cypress/e2e/checks_customization.cy.js
@@ -127,12 +127,12 @@ context('Checks customization', () => {
       checksSelectionPage.clickSaveChecksSelectionButton();
       checksSelectionPage.clickStartExecutionButton();
       checksSelectionPage.waitForCustomizedCheckElements();
-      checksSelectionPage.clickOnCheckResultDescription();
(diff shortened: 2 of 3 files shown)


test/e2e/cypress/pageObject/databases_overview_po.js
```diff
@@ -117,9 +117,11 @@ export const cleanUpButtonIsNotDisplayed = () => {
 
 // UI Interactions
 
-export const clickHdqDatabaseRow = () => cy.get(hdqDatabaseCell).click();
+export const clickHdqDatabaseRow = () =>
+  cy.get(`${hdqDatabaseCell} > td:eq(0)`).click();
```
EMaksy, line 121: Hey could we extract this as own selector ? 
Someting like this ? 
```
export const hdqDatabaseRowFirstCell = `${hdqDatabaseCell} > td:eq(0)`;
export const hddDatabaseRowFirstCell = `${hddDatabaseCell} > td:eq(0)`;
```

And put the selectors on top of the file so we have code like this 
```
export const clickHdqDatabaseRow = () => cy.get(hdqDatabaseRowFirstCell).click();

export const clickHddDatabaseRow = () => cy.get(hddDatabaseRowFirstCell).click();

```
We use this approach through the different page objects. https://github.com/trento-project/web/blob/9ddb29f3eb832c2f38f77cce29d7838a4703c6f8/test/e2e/cypress/pageObject/databases_overview_po.js#L40 and https://github.com/trento-project/web/blob/9ddb29f3eb832c2f38f77cce29d7838a4703c6f8/test/e2e/cypress/pageObject/hana_cluster_details_po.js#L41
Checkout the adr from @vicenteqa  https://github.com/trento-project/docs/blob/main/adr/0019-e2e-testing-practices.md#decisions

EMaksy (commented):
Hey @arbulu89 great job, it looks good to me only check out my comment about the selectors. 

Quick question: why did you go back to using just the chevron? Would it make sense to keep both — the chevron to indicate the table is collapsible, and also keep the row clickable to expand it?
I’m just thinking in terms of usability — making it easier for users to interact with either element but sure @jagabomb  had his reasons :D

jagabomb:
Minor feedback, can we change the padding-left of the arrow/chevron icon to 1em? I think its pl-4 😅. I think it will create a better balance in the row.

PR #3599, opened 2025-06-25
Removal of unused dependency

gagandeepb:
# Description

It appears we are not using this dependency `fun_with_flags` and it emits noisy debug logs, in local dev environment.

## How was this tested?
N/A


nelsonkopliku:
mh, maybe we need to adjust also 
- `test/fixtures/regression/sap_system_split/trento_dev.sql`
- `test/fixtures/regression/process_manager_rename/trento_dev.sql`

PR #3600, opened 2025-06-26
Correlation of operations activity log entries

gagandeepb:
# Description

Implements correlation of operations related activity log entries. Makes use of/re-uses the operation id as a correlation id. This leaves client/FE code agnostic to correlated activity log entry type.

## How was this tested?

UT

### DIFF
--- a/lib/trento/activity_logging/parser/phoenix_conn_parser.ex
+++ b/lib/trento/activity_logging/parser/phoenix_conn_parser.ex
@@ -191,9 +191,12 @@ defmodule Trento.ActivityLog.Logger.Parser.PhoenixConnParser do
              :cluster_operation_requested,
              :host_operation_requested
            ] do
+    operation_id = resp_body |> Jason.decode!() |> Map.get("operation_id")
+
     %{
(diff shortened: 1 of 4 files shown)


nelsonkopliku (approved):
Hey @gagandeepb there is nothing much concerning me about this very specific and restricted solution.

As long as we can gain at least some value from this, and I believe we do, I am fine.

I cannot but highlight that for the long term we might need to extend current correlation capabilities, for instance we might need to think about propagating the correlation id cross boundaries so that it can be reconciled at a certain point and thus build a solid history of related entries.

Of course there would be cases where we would face challenges, and we will 😄 

Anyway, so far so good!

abravosuse (approved):
Hey @gagandeepb I have deployed the web image from https://3600.prenv.trento.suse.com/ in my rolling environment and run the following operations:

- Change Saptune Solution
- Stop Instance
- Start Instance

In all cases I can correlate the corresponding request and completion entries in the activity log by clicking on the Show events link in the Activity Details modals. Good job!

PR #3605, opened 2025-06-30
Expose cluster resources

arbulu89:
# Description

**Spoiler**: Don't worry about the code changes added. 99% of them are just tests. This change actually simplifies things, and if we were to remove the deprecated fields it would have less code than before.

In order to properly display the new resources table, rellocate the resources entries in the read models so they are flat-mapped directly in `details`. We don't need anymore `stopped_resources` nor `resources` inside `nodes/sap_systems`. These fields are now deprecated (i hope we can remove them before release 3.x.x, as deprecated fields number is increasing).

At the end, the plan is to have something like this in the Frontend (not included in this PR):

image

## How was this tested?

UT

### DIFF
--- a/lib/trento/clusters.ex
+++ b/lib/trento/clusters.ex
@@ -159,16 +159,10 @@ defmodule Trento.Clusters do
   end
 
   @spec resource_managed?(ClusterReadModel.t(), String.t()) :: boolean()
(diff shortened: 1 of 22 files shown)


nelsonkopliku (approved):
Looking good.

I also hope we can drop deprecated field by version 3 😄

PR #3606, opened 2025-07-01
Correlation of heartbeat entries after API key regeneration

gagandeepb:
# Description

Correlates heartbeat entries with API key regeneration entry, until the (default) TTL of 15s is reached. The 15s time starts ticking after successful key regeneration request from the UI. After the TTL is reached, no further heartbeat entries are correlated with the API key regeneration entry previously mentioned.

## How was this tested?
 
UT adjustment and extension


abravosuse (approved):
I have updated my rolling environment with the web image from https://3606.prenv.trento.suse.com/. Then I have generated a new api key. As result, the heartbeat of all registered agents have failed. When I then go to the activity log, open the entry corresponding to the new key generation and click on Show events, I see all associated heartbeat failure entries.

arbulu89 (changes requested):
Hey @gagandeepb ,
I'm not sure about this implementation (and being honest, about the feature itself).
This correlates **all** events happening in the discovery as far as I see. And certainly, as event in host A being correlated to an event in cluster B being correlated doesn't looks appropriate.
The correlation of events must be much more strict that this in my opinion, and not to this "almost irrelevant" api-key generation.
We need to be aware that we only have 1 correlation id, and using this one for all discovery events looks dubius.

I'm flagging red so we discuss this team-wise before moving forward.

abravosuse (commented):
Hey @gagandeepb I have deployed the updated image from https://3606.prenv.trento.suse.com/ in my testing environment. Then I have:

1. generated a new API key
2. gone to the activity log and opened the details view of the "API Key was generated" entry
3. clicked on Show events

That has filtered out everything from the activity log except for:

1. the "API Key was generated" entry
2. the "Host heartbeat failed" entries
3. one "Host health changed" entry

One and two above were expected, But why three? See screenshot:

image

gagandeepb:
> 1. the "API Key was generated" entry
> 2. the "Host heartbeat failed" entries
> 3. one "Host health changed" entry
> 
> One and two above were expected, But why three?

@abravosuse In my understanding, there was a pre-existing correlation between 2 and 3 that pre-dates/pre-exists-from-before the changes introduced in this PR. 
See for example in this [linked search in the demo environment.](https://demo.trento-project.io/activity_log?severity=info&severity=warning&severity=critical&search=b9cb7374-0af1-4fb1-9e7e-084aec4067f1+&first=20)

abravosuse:
> @abravosuse In my understanding, there was a pre-existing correlation between 2 and 3 that pre-dates/pre-exists-from-before the changes introduced in this PR. See for example in this [linked search in the demo environment.](https://demo.trento-project.io/activity_log?severity=info&severity=warning&severity=critical&search=b9cb7374-0af1-4fb1-9e7e-084aec4067f1+&first=20)

I don't know. If that is the case, I wonder why - out of six hosts - the "Host health changed" was only capture for one of them. Unless the other five were already in critical status. Maybe that's why. I will check it out.

abravosuse (approved):
@gagandeepb that was indeed the situation: all hosts but one were in critical status before generating the API key. All good then. Thanks and great job!

arbulu89 (commented):
Hey @gagandeepb 
Are we sure we want to add cache system only for this?
I think we could have something more simpler doing.
1. Get new correlation id in the settings controller and add to a process. The same we do now in the plug:
```
correlation_id = UUID.uuidv4()
Process.put(:api_key_correlation_id, correlation_id)
```
2. With this and knowing when the api was changed (we have this in the settings database table, in the `created_at` field have something like:
```
def dispatch_heartbeat_failed_commands(date_service \\ DateService) do
    expired_heartbeats = get_all_expired_heartbeats(date_service)
    if length(expired_heartbeats) > 0 do
        is_api_key_updated = # check here if the api key created_at + heartbeat_interval is bigger than utc_now
    else 
        is_api_key_udpated = false
    end

    Enum.each(get_all_expired_heartbeats(date_service), fn %{agent_id: agent_id} ->
      dispatch_command(agent_id, :critical, is_api_key_updated)
    end)
  end
```
Of course, the code can be totally different, but the idea is to use the `created_at` (or `updated_at`) database field to know if the api key was changed.

3. In dispatch_command:

```
dispatch_command(.., ..., true), add the new correlation id to the command
```

I mean, the whole idea is to have the core logic of this particular correlation contained in the same logic of the code. I think we have all the information needed to avoid adding a generic cache, that doesn't fullfill anything we cannot do without it.

What do you think?

lib/trento/heartbeats.ex
```diff
@@ -82,7 +84,7 @@ defmodule Trento.Heartbeats do
   defp dispatch_command(agent_id, heartbeat) do
     case %{host_id: agent_id, heartbeat: heartbeat}
          |> UpdateHeartbeat.new!()
-         |> commanded().dispatch() do
+         |> maybe_correlated_dispatch() do
```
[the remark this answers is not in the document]
gagandeepb: This will only correlate if there is a correlation id set to non-nil value in the cache under the appropriate key, which is only the case in the 15 seconds following an API key regeneration, when the heartbeats are very likely (certain, maybe?) to be critical/fail. This `maybe_correlated_dispatch` will be like the default dispatch in every other case. The behaviour is still what you suggest, albeit a bit differently implemented due to the usage of the cache key and its ttl.

arbulu89:
> Hence I need to set it to a globally available store

Okey, fair point. The usage of `Process` is out of question. Either way, I think we could still have the same logic without cache (querying previous activity log entries for Api key generation entries to get the correlation ID, for example). Just my previous experiences fear of using caches, which are a source of problems hehe

Either way, I don't want to derail the implementation.
Let me re-review again with current cache usage

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -97,6 +98,16 @@ defmodule TrentoWeb.V1.SettingsController do
           AuthenticateAPIKeyPlug.generate_api_key!(updated_settings)
         )
 
+      correlation_id = Process.get(:correlation_id)
+
+      key = ActivityLog.correlation_key(:api_key)
```
arbulu89, line 103: I don't really understand this usage. Why do we need a key to get other key?
More when we have an initial hardcoded key.

If it is for testing, doesn't make more sense to make the tests sync instead of adding this piece of code?

lib/trento/activity_logging/correlations.ex
```diff
@@ -0,0 +1,42 @@
...
+  @spec get_correlation_id(binary()) :: binary() | nil
+  def get_correlation_id(key) do
+    with {:ok, maybe_value} <- Cachex.get(@cache_name, key) do
+      maybe_value
+    end
+  end
+
+  @spec expire_correlation_id(binary(), non_neg_integer()) :: :ok
+  def expire_correlation_id(key, ttl \\ @default_ttl)
+      when is_binary(key) do
```
arbulu89, line 37: Do we need this `is_binary`? `dialyzer` won't allow to put a non binary thing. Anyway, if we do, it will fail with a pattern match func error

lib/trento/heartbeats.ex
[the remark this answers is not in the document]
arbulu89: But certainly, if the user is fast enough to change the api key in a host within this 15 seconds, the positive heartbeat would be associated, woudn't be?
Anyway, corner case (but it could be avoided)

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -97,6 +98,16 @@ defmodule TrentoWeb.V1.SettingsController do
...
+
+      _ =
+        ActivityLog.put_correlation_id(key, correlation_id)
+
+      _ =
```
arbulu89, line 108: I personally would put here a explicit expiration time value. Using the 15 seconds value doesn't give enough context about what we are doing in my opinion

test/trento/heartbeats_test.exs
```diff
@@ -17,131 +18,141 @@ defmodule Trento.HeartbeatsTest do
...
       :dispatch,
-      fn command ->
-        assert %UpdateHeartbeat{
-                 host_id: ^agent_id,
-                 heartbeat: :critical
-               } = command
-
-        :ok
+      fn
+        command, _ ->
```
arbulu89, line 124: Shouldn't we test that we are adding the correlation?
Otherwise, do we even test that the correlation_id is added somewhere?

arbulu89 (approved):
Hey @gagandeepb ,
Just some "nitpick" comments.
As commented, I'm dubious about using a cache, but if you are confident about it, green light from me

lib/trento_web/controllers/v1/settings_controller.ex
gagandeepb: > doesn't make more sense to make the tests sync instead of adding this piece of code?

The function call `ActivityLog.correlation_key(:api_key)` returns a static key (in prod/application code) but a dynamic key set from the calling side in test code (if the calling side opts-in/uses `Process.put(:correlation_key, "some-uuid")`). This is necessary as an anti-test flaking measure, since with a static key the global mutable state is shared across test modules/ leading to flaky test failures.  There are multiple test modules which exercise the code path involving the cache, so letting any one (or even all) of them run sync does not resolve the global/shared-ness of the state. The implementation of `ActivityLog.correlation_key` is one way for individual test cases to own their own keys that map to a unique correlation id. As a verification, of anti-flakiness, I have run the BE Flaky test pipeline against this PR's feature branch, and observed no additional flakiness than what existed before.

lib/trento/activity_logging/correlations.ex
gagandeepb: I don't feel too strongly about this one. In my understanding/experience, dialyzer is a bit hit or miss with enforcing desired types. The `Cachex` function call inside this function can accept any term as a key, I just wanted to specialize that interface to only accept binaries at the outer function level or fail if a test case/similar surfaces that it was called with a non-binary, e.g. an atom.

lib/trento_web/controllers/v1/settings_controller.ex
gagandeepb: I addressed this by adding an explicit ttl value from the calling side. [25e642](https://github.com/trento-project/web/pull/3606/commits/25e642ee3abb37f491dacfa695574e9da0f1bc58)

nelsonkopliku (commented):
Hey @gagandeepb, we had a face to face review and @arbulu89 is raising same feedback I also raised privately 😅 

Now, the options are:
1. we make all tests using correlation and cache sync
2. we try using a different cache in tests
3. we add an abstraction for the correlations/correlator so that we can 
    - mock/stub it up as needed and where needed
    - keep tests async
    - test the actual implementation in isolation

_I_ would personally find option 3 to be the cleanest, however I understand it might feel more complex, so I would be fine with an attempt of option 2.
If we can avoid forcing sync tests as in option 1 because of this feature, I'd be happier.

The 4th option would be ignoring the feedback and just go on. Up to you.

PR #3608, opened 2025-07-02
Resources table frontend

arbulu89:
# Description
Add `Resources` table in frontend for cluster details view.

image

Have a look in the PR env to see different clusters.

Related to: https://github.com/trento-project/web/pull/3605

PD: I have adapted the `ClusterMaintenanceChange` modal to have the tests passing, but don't look that much there. The modal goes away in the next change

## How was this tested?

UT and e2e


assets/js/pages/ClusterDetails/ClusterDetails.jsx
```diff
@@ -119,6 +119,21 @@ function ClusterDetails({
             }}
             onCancel={closeOperationModal}
           />
+          <SimpleAcceptanceOperationModal
```
arbulu89, line 122: Just moving this inside `{operationsEnabled && (` condition, so we can easier disable if needed. (some buttons are still not guarded by this variable)

abravosuse (approved):
I've had a look at the different cluster types in the PR. It's all aligned with what we discussed yesterday in our follow up. The only thing that caught my attention - by comparison with the output of crm status - is that the stopped resources in the hana scale-out cluster don't have an assigned location. Which for me is totally fine, since you can figure it out by discarding the nodes where the resources are actually running. If it can be added, good. If it cannot, good as well.

assets/js/common/Table/CollapsibleTableRow.jsx
```diff
@@ -1,51 +1,63 @@
...
           >
-            <EOS_KEYBOARD_ARROW_DOWN
-              className={classNames(
-                'cursor-pointer self-center fill-gray-500',
-                {
-                  'transform rotate-180': rowExpanded,
-                }
-              )}
-            />
+            {isCollapsible && (
```
nelsonkopliku, line 35: In which cases the outer `collapsibleDetailRenderer` would be true, but `isCollapsible` false?
Meaning that we could have collapsible table rows without the icon 🤔

assets/js/common/Table/CollapsibleTableRow.jsx
```diff
@@ -1,51 +1,63 @@
...
+                  }
+                )}
+              />
+            )}
           </td>
         )}
         {renderCells(columns, item)}
       </tr>
-      {collapsibleDetailRenderer && (
+      {isCollapsible && wrapCollapsedRowInCell ? (
```
nelsonkopliku, line 49: I was not understanding the `wrapCollapsedRowInCell`, then I tried locally by setting it to true in `Resources` component and so I understood 😂

nelsonkopliku (approved):
Hey Xabi, it looks really nice! Well done!

assets/js/common/Table/CollapsibleTableRow.jsx
nelsonkopliku: I see, thanks!

EMaksy (approved):
Alright lgtm!
Nicely done. 

Just 2 cents, it is nice that we can reuse the Table component, but it is becoming a hard component to enrich/change as touches so much :+1:

PR #3614, opened 2025-07-03
Enable/disable pacemaker systemd ops based on current state

nelsonkopliku:
# Description

This PR makes use of the state exposed in https://github.com/trento-project/web/pull/3610 to enable/disable relevant call to actions based on current state.

### DIFF
--- a/assets/js/lib/model/hosts.test.js
+++ b/assets/js/lib/model/hosts.test.js
@@ -16,11 +16,18 @@ describe('hosts', () => {
           { name: 'foo.service', unit_file_state: 'enabled' },
           { name: 'bar.service', unit_file_state: 'unknown' },
           { name: 'baz.service', unit_file_state: 'unrecognized' },
+          { name: 'qux.service', unit_file_state: '' },
+          { name: 'quux.service', unit_file_state: null },
+          { name: 'corge.service' },
         ],
       });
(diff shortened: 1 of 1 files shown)


assets/js/lib/model/hosts.test.js
```diff
@@ -0,0 +1,48 @@
...
+      });
+      expect(canEnableUnit(host, 'foo.service')).toBe(true);
+    });
+
+    it('should not be able to enable a unit', () => {
+      const host = hostFactory.build({
+        systemd_units: [
+          { name: 'foo.service', unit_file_state: 'enabled' },
+          { name: 'bar.service', unit_file_state: 'unknown' },
+          { name: 'baz.service', unit_file_state: 'unrecognized' },
```
arbulu89, line 18: nitpick:
maybe more than `unrecognized`, `{ name: 'baz.service' }` and a test using `unknown.service` would cover more untested scenarios

arbulu89 (approved):
Nice!!

assets/js/lib/model/hosts.test.js
nelsonkopliku: Covered more cases here f7f2228a9375b5456b068f704e4df43498073018

PR #3618, opened 2025-07-07
Add cluster status for host

balanza:
Add `cluster_status` field to the host read model to map whether a host is online or offline in the cluster.

The field is set as `online` for every host that belongs to a cluster, to mimic the actual behaviour.

### DIFF
--- a/lib/trento/hosts/projections/host_read_model.ex
+++ b/lib/trento/hosts/projections/host_read_model.ex
@@ -95,5 +95,4 @@ defmodule Trento.Hosts.Projections.HostReadModel do
       true -> changeset
     end
   end
-
 end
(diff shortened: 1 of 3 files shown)


lib/trento/hosts/projections/host_read_model.ex
```diff
@@ -55,6 +55,8 @@ defmodule Trento.Hosts.Projections.HostReadModel do
       type: Ecto.UUID,
       where: [deregistered_at: nil]
 
+    field :cluster_status, Ecto.Enum, values: [:online, :offline]
```
nelsonkopliku, line 58: What would happen if a host is not part of a cluster?

nelsonkopliku (commented):
I second Xabi's feedback.

The change per se does not look neither good or bad.

I would also find it more helpful a domain first approach.

PR #3619, opened 2025-07-07
Add system replication field to database details view

arbulu89:
# Description
Add system replication field to Database details view. It is a totally relevant field. More, when we start adding operations.
This is one of the first efforts to add HANA native system replication data to the details view. More to come.

image

Without system replication:
image

## How was this tested?

UT and e2e


arbulu89:
> As discussed, this initial proposal is no scalable to mutli-tier/target scenarios. A site-based section that reflects the existing db pairs linked by replication might work better.

Yes. Moving to draft by now until deciding the next steps

PR #3620, opened 2025-07-07
Enforce pacemaker operation policy to rely on discovered state

nelsonkopliku:
# Description

Make sure the api returns relevant policy errors when attempting to enable/disable an already enabled/disabled pacemaker systemd unit.

It does not really fix much, as the operator is idempotent, it is just for consistency of the API.


lib/trento/operations/cluster_policy.ex
```diff
@@ -44,9 +46,49 @@ defmodule Trento.Operations.ClusterPolicy do
...
+    } = Enum.find(hosts, &(&1.id === host_id))
+
+    unit_file_state =
+      Enum.find_value(systemd_units, fn
+        %{name: "pacemaker.service", unit_file_state: unit_file_state} -> unit_file_state
+        _ -> nil
+      end)
+
+    desired_unit_state = desired_unit_state(operation)
+    already_applied_state = already_applied_unit_state(operation)
```
arbulu89, line 67: I don't really get this `already_applied_state`. Is this not `unit_file_state`?
I mean, `unit_file_state` can be `enabled/disabled/nil`. so we can 3 things in the case, without the need of `already_applied_unit_state` function:
```
    case unit_file_state do
      ^desired_unit_state ->
        :ok

      nil ->
        {:error,
         [
           "Pacemaker service unit state is unrecognized on host #{hostname}"
         ]}

      current_state ->
        {:error,
         [
           "Pacemaker service on host #{hostname} is already #{current_state}"
         ]}

      
    end
```

arbulu89: Yes, yes, i get it. I'm not sure it forbin any action if we have a not expected value is a good thing or not, but it is fine. I don't expect we will go from that code path ever (hopefully XD)

PR #3623, opened 2025-07-09
Handle offline host in cluster

balanza:
With this PR, a `RegisterOfflineClusterHost` command is added to the cluster aggregate.

The command maps the cases in which a host is discovered and recognised to belong to a cluster, but it's currently offline. Common user scenarios might be a host taken down for maintenance or a host not started yet.

The cluster aggregate handles the new command very similarly to a non-DC node regarding the registration/deregistration policies.

Finally, the emitted `HostAddedToCluster` event brings the information of whether the host is online or not in the cluster.


lib/trento/clusters/events/host_added_to_cluster.ex
```diff
@@ -5,8 +5,13 @@ defmodule Trento.Clusters.Events.HostAddedToCluster do
 
   use Trento.Support.Event
 
-  defevent do
+  defevent version: 2 do
     field :cluster_id, Ecto.UUID
     field :host_id, Ecto.UUID
+    field :cluster_status, Ecto.Enum, values: [:online, :offline]
```
balanza, line 11: **Important** We may want to think twice about wording before committing.
Is "online/offline" a good definition for a node in a cluster? Is it a generally recognised lingo?

lib/trento/clusters/cluster.ex
```diff
@@ -173,7 +174,8 @@ defmodule Trento.Clusters.Cluster do
       },
       %HostAddedToCluster{
         cluster_id: cluster_id,
-        host_id: host_id
+        host_id: host_id,
+        cluster_status: :online
```
arbulu89, line 178: Can we add a new enum in `trento/clusters/enums/online_status.ex` as `Trento.Clusters.Enums.OnlineStatus`.
So we can use the enum instead of putthing the atom in every plase.

lib/trento/clusters/commands/register_offline_cluster_host.ex
```diff
@@ -0,0 +1,19 @@
+defmodule Trento.Clusters.Commands.RegisterOfflineClusterHost do
+  @moduledoc """
+  Register a cluster node to the monitoring system.
```
arbulu89, line 3: nitpick: maybe change the docstring to specify that we are talking about an offline node

lib/trento/clusters/events/host_added_to_cluster.ex
arbulu89: I think it is

lib/trento/clusters/cluster.ex
```diff
@@ -200,7 +202,35 @@ defmodule Trento.Clusters.Cluster do
...
         cluster_id: cluster_id,
+        host_id: host_id,
+        cluster_status: :online
+      }
+    ]
+  end
+
+  # When an Offline node is discovered, a cluster is registered and the host is added to the cluster.
+  # The cluster details are left as unknown, and filled once a message from the DC node is received.
+  def execute(%Cluster{cluster_id: nil}, %RegisterOfflineClusterHost{
```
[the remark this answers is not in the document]
balanza: We discussed the solution already, we are going to parse the cibadmin configuration files in the discovery

lib/trento/clusters/cluster.ex
```diff
@@ -418,6 +481,46 @@ defmodule Trento.Clusters.Cluster do
...
+  def apply(
+        %Cluster{hosts: hosts, offline_hosts: offline_host} = cluster,
+        %HostAddedToCluster{
+          host_id: host_id,
+          cluster_host_status: ClusterHostStatus.offline()
+        }
+      ) do
+    %Cluster{
+      cluster
+      | hosts: [host_id | hosts],
```
nelsonkopliku, line 493: **questions:** 
- why do we add an offline host to the `hosts` list?
- when do we remove an offline host from the `hosts` list?

The broader question would actually be: what is now `hosts`? The list of online+offline hosts? Only the online hosts?

nelsonkopliku (commented):
@balanza mainly questions from me at this stage.

test/trento/clusters/events/host_added_to_cluster.exs
```diff
@@ -0,0 +1,26 @@
+defmodule Trento.Cluster.Events.HostAddedToClusterTeest do
```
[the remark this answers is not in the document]
arbulu89: And typo in `Teest` by the way XD

lib/trento/clusters/cluster.ex
nelsonkopliku: I see, however I am still failing to understand the reason behind the need to have also `offline_hosts` inside the main `hosts`.

With regards to _when do we remove an offline host from the hosts list_, it seems to be covered on `HostRemovedFromCluster` - assuming a `DeregisterClusterHost` is dispatched both for online and offline hosts.

> The ideal solution would have been to turn hosts into a collection of (id, status)
>
I actually agree on this, however I refrained from suggesting it to not add more work for you 😅 
But if you also think it would be better I am ok if we address that.

> that would imply migrating the aggregate state.
>
Not sure I understand this. The aggregate state is all in memory built from events in the event store.
What would need to be migrated?

arbulu89: Exactly. Changing the aggregate is really delicate thing. If we have a snapshot, and the first event populates the aggregate. That's why you need to implement a upcasting for it.
In my opinion, not worth it. It adds the chance to add some error while we don't solve almost any complexity.

nelsonkopliku: Roll up is a fair concern indeed, agreed.
Yes, not really worth adding extra complexity to properly handle snapshot state and aggregate state change at this point.

Maybe something to revisit in the future.

PR #3625, opened 2025-07-10
Analytics Add Version Code

jagabomb:
# Description

This PR adds the `version` code of **Trento Web** ~~and an environment value~~ for each Autocapture and PageView event. The purpose of these addition is to help detect which version of **Trento Web** is in use. ~~Then the environment value helps us to cross check that development, demo and production events are clearly distinguishable. See below for visual reference.~~

**Assumptions:**
1. Every official build of Trento Web will have the **production** `environment` value.
2. Every demo build will have either **demo** `environment` value.
3. Project source will use the **dev** `environment` value.
4. Official **production** Posthog instance will be for **production events only**.
5. Dev/Demo events will use a separate or private instance of Posthog for development purposes.

## How was this tested?

- [x] **Manually** (Need to find an automated solution) 😅 

Screenshot 2025-07-23 at 10 58 35

### DIFF
--- a/assets/js/lib/analytics/pageview.js
+++ b/assets/js/lib/analytics/pageview.js
@@ -1,17 +1,20 @@
 import { useEffect } from 'react';
 import { useLocation } from 'react-router-dom';
 import { capture } from '@lib/analytics';
+import { getFromConfig } from '@lib/config';
 import { useSelector } from 'react-redux';
 import { getUserProfile } from '@state/selectors/user';
 
 export default function PostHogPageView() {
   const location = useLocation();
   const { analytics_enabled } = useSelector(getUserProfile);
+  const webversion = getFromConfig('webversion');
 
   // Track pageviews
   useEffect(() => {
     capture(analytics_enabled, '$pageview', {
       $current_url: window.location.href,
+      webversion,
     });
   }, [location]);
 
--- a/lib/trento_web/controllers/page_html/index.html.heex
+++ b/lib/trento_web/controllers/page_html/index.html.heex
@@ -14,7 +14,8 @@
       ssoLoginUrl: '<%= raw @sso_login_url %>',
       ssoCallbackUrl: '<%= raw @sso_callback_url %>',
       ssoEnrollmentUrl: '<%= raw @sso_enrollment_url %>',
-      operationsEnabled: <%= @operations_enabled %>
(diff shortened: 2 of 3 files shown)


lib/trento_web/controllers/page_html/index.html.heex
```diff
@@ -1,4 +1,8 @@
-<div id="trento" />
+<div
```
arbulu89, line 1: Why don't we put the `version` in the `config`?
It is not a config, that's true, but it would enable to use the `getFromConfig` function.
And after all, the value is exposed in both cases.

Devils advocate question: does exposing the server version to not logged users expose some "you need to be logged" info?

assets/js/lib/analytics/pageview.js
```diff
@@ -7,11 +7,19 @@ import { getUserProfile } from '@state/selectors/user';
 export default function PostHogPageView() {
   const location = useLocation();
   const { analytics_enabled } = useSelector(getUserProfile);
+  const webversion = document
```
arbulu89, line 10: I think we can use the `getFromConfig` function outside of the component itself, as we do in many other cases

arbulu89 (commented):
Hey @jagabomb ,

assets/js/lib/analytics/pageview.js
```diff
@@ -1,17 +1,20 @@
...
 export default function PostHogPageView() {
   const location = useLocation();
   const { analytics_enabled } = useSelector(getUserProfile);
+  const webversion = getFromConfig('webversion');
 
   // Track pageviews
   useEffect(() => {
     capture(analytics_enabled, '$pageview', {
       $current_url: window.location.href,
+      webversion,
```
arbulu89, line 17: I personally prefer `version` or `appVersion` more than `webversion`.
What do you think?
pd: if you change, don't forget changing the `data-ph-...` field as well

arbulu89 (approved):
Looks good!
I pointed out only a naming preference XD

assets/js/lib/analytics/pageview.js
arbulu89: Lowercase is important: this means `webversion: webversion` in what we send.
Javascript recommends shorthanding (removing right side) if the value is the same.

PR #3626, opened 2025-07-10
Advanced system replication fields

arbulu89:
# Description

Add new System replication fields into Database instances data.
The parsing is super straightforward. The fields come in the `hdbnsutil` command output, that we already passed in the `sap_system_discovery` payload.

Once the parsing is done, everything else is basic `commanded` chores.

Right now there isn't any change in the frontend, but we will use this fields to group the instances per site and put the values there.

## How was this tested?

UT

### DIFF
--- a/lib/trento_web/openapi/v1/schema/database.ex
+++ b/lib/trento_web/openapi/v1/schema/database.ex
@@ -49,6 +49,31 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Database do
             type: :string,
             description: "System Replication Status"
           },
+          system_replication_site: %Schema{
+            type: :string,
+            nullable: true,
+            description: "System Replication Site"
+          },
+          system_replication_mode: %Schema{
+            type: :string,
+            nullable: true,
+            description: "System Replication Mode"
+          },
+          system_replication_operation_mode: %Schema{
+            type: :string,
+            nullable: true,
+            description: "System Replication Operation mode"
+          },
+          system_replication_source_site: %Schema{
+            type: :string,
+            nullable: true,
+            description: "System Replication Source site where replication is coming from"
+          },
+          system_replication_tier: %Schema{
+            type: :integer,
+            nullable: true,
+            description: "System Replication Site tier number"
+          },
           health: ResourceHealth,
           absent_at: %Schema{
             type: :string,


lib/trento/databases/commands/register_database_instance.ex
```diff
@@ -33,6 +33,11 @@ defmodule Trento.Databases.Commands.RegisterDatabaseInstance do
     field :start_priority, :string
     field :system_replication, :string
     field :system_replication_status, :string
+    field :system_replication_site, :string
+    field :system_replication_mode, :string
+    field :system_replication_operation_mode, :string
+    field :system_replication_source_site, :string
+    field :system_replication_tier, :integer
```
nelsonkopliku, line 40: **thought:** have you considered grouping system replication info in an embedded object?

Not that I have a strong opinion on that, just wondering whether it would add a sense of cohesion of information plus possibly reducing the changes in the PR.

Same goes in the other places where we have these information.

lib/trento/databases/events/database_instance_registered.ex
```diff
@@ -20,6 +20,14 @@ defmodule Trento.Databases.Events.DatabaseInstanceRegistered do
...
+    field :system_replication_site, :string
+    field :system_replication_mode, :string
+    field :system_replication_operation_mode, :string
+    field :system_replication_source_site, :string
+    field :system_replication_tier, :integer
     field :health, Ecto.Enum, values: Health.values()
   end
+
+  def upcast(params, _, 3),
+    do: Map.put(params, "system_replication_tier", 0)
```
nelsonkopliku, line 32: Very curious about where upcasting to version 2 is 😅

arbulu89: In `Trento.Databases.Event.Upcaster.UpcastDatabaseId`

lib/trento/databases/commands/register_database_instance.ex
arbulu89: The same than the current cluster aggregate change.
These fields are stored in the aggregate in a flat format. If I change that to am embedded schema, I need to implement the upcasting.
Besides, we already have 2 of these fields in our API schema, and moving to a "embedded object" implies I need to start moving things around, deprecate other fields, etc

PR #3635, opened 2025-07-14
Fix Host OpenApi spec

balanza:
`cluster_id` is, in fact, nullable for hosts, as they do not always belong to a cluster.

### DIFF
--- a/lib/trento_web/openapi/v1/schema/host.ex
+++ b/lib/trento_web/openapi/v1/schema/host.ex
@@ -103,7 +103,8 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Host do
           cluster_id: %Schema{
             type: :string,
             description: "Identifier of the cluster this host is part of",
-            format: :uuid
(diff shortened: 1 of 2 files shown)


nelsonkopliku (approved):
Good catch!

gagandeepb (commented):
An additional caveat is that that we should not rely too much on nullability via OpenAPI. [The reason is that there is no support for explicitly nullable fields in OpenAPI version 3.1.](https://github.com/OAI/OpenAPI-Specification/pull/2246)

balanza:
> An additional caveat is that that we should not rely too much on nullability via OpenAPI. [The reason is that there is no support for explicitly nullable fields in OpenAPI version 3.1.](https://github.com/OAI/OpenAPI-Specification/pull/2246) (It should be fine for now since we are on a version of the spec that is < 3.1 that has support for nullability and the library we use has limited/probably no upgrade path to spec version 3.1.)

Should we go for `[string, null]` instead?

PR #3637, opened 2025-07-14
Sort system replication tier

arbulu89:
# Description
Sort database instances by system replication tier.
This way, the primary instances will come before secondary instances.
This only affects the overview pages. 
The details page needs to be worked yet.

## How was this tested?

e2e


EMaksy (approved):
Nice!

PR #3640, opened 2025-07-16
SMLM settings save/change correlation

gagandeepb:
# Description

Implements correlation of activity log entries linked with a save/change of SMLM settings. This has been done by extending the previously used mechanism in use for api_key_regen line of correlation. Additionally some refactoring has been done to isolate test versions of a cache related module from the application/prod version of the same. Finally, a new test dependency has been added along with a detailed code comment that motivates the need for such a dependency (`ProcessTree`).

## How was this tested?

UT adjustment and extension.


gagandeepb (commented):
image
Correlation upon initial save of SMLM settings.

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -350,4 +345,32 @@ defmodule TrentoWeb.V1.SettingsController do
...
   end
+
+  defp propagate_correlation_id(ctx, correlation_ttl)
+       when ctx in [:api_key, :suse_manager_settings] do
+    correlation_id = Process.get(:correlation_id)
+
+    key = ActivityLog.correlation_key(ctx)
+
+    case key do
+      nil ->
```
nelsonkopliku, line 356: **question:** can `key` ever be `nil`? 
It could be nil when the correlation implementation is the testing one, but it should not happen in non test mode.

test/trento/settings_test.exs
```diff
@@ -128,402 +129,384 @@ defmodule Trento.SettingsTest do
...
-              }} = Settings.change_suse_manager_settings(change_submission)
-    end
+            "&Trento.Settings.change_suse_manager_settings/1" ->
+              expect(Trento.SoftwareUpdates.Discovery.Mock, :clear, 2, fn -> :ok end)
 
-    test "should properly update ca_cert and its upload date when a new cert is provided" do
-      now = DateTime.utc_now()
+              # there is a path dependence on needing to save first
+              # before trying to change settings
+              Settings.save_suse_manager_settings(settings)
```
nelsonkopliku, line 372: **suggestion:** we might leverage `insert_software_updates_settings` to insert initial settings instead of `Settings.save_suse_manager_settings/1`. 
It might be beneficial to also keep the previous expectation to 1 function call, which would be more realistic

nelsonkopliku (commented):
Hey Gagan, nice progress!

As mentioned privately I understand the struggle of testing and the intricacies required.

I still have some reserves on `Correlations.correlation_key/1` however if we don't have any alternative, so be it.

I left a few comments but the main point would be moving what's being called `ScopedCorrelations` closer to tests, as it effectively is a test implementation for the correlations.

---

For completeness, I attempted a slightly different approach in [this branch](https://github.com/trento-project/web/compare/smlm-settings-update-correlation...alternative-correlation?expand=1).

I'd say it works to a certain extent and of course I stumbled upon challenges in testing... 🙈 

I believe that with the deserved attention it should work, however it could require some restructuring of tests wrt mocking/stubbing combinations. 

Not something I can address right now. 
If it makes any sense for you, feel free to pick that up, otherwise we can resume at due time.

lib/trento/activity_logging/correlations/unscoped_correlations.ex
```diff
@@ -0,0 +1,39 @@
+defmodule Trento.ActivityLog.Correlations.UnscopedCorrelations do
```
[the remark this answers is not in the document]
nelsonkopliku: Additionally, we should also test this module in isolation.

lib/trento_web/controllers/v1/settings_controller.ex
gagandeepb: This can be nil only in test mode. Never in non-test/prod mode.

lib/trento/activity_logging/correlations/unscoped_correlations.ex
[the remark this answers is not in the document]
gagandeepb: `ScopedCorrelations` is now renamed and moved closer to tests. IMO, this name (`UnscopedCorrelations`) is pretty descriptive of module behavior. That said, I am open to naming suggestions.

test/trento/settings_test.exs
```diff
@@ -128,402 +129,384 @@ defmodule Trento.SettingsTest do
...
-                ca_uploaded_at: ^initial_ca_uploaded_at
-              }} = Settings.change_suse_manager_settings(change_submission)
-    end
+            "&Trento.Settings.change_suse_manager_settings/1" ->
+              expect(Trento.SoftwareUpdates.Discovery.Mock, :clear, 1, fn -> :ok end)
 
-    test "should properly update ca_cert and its upload date when a new cert is provided" do
-      now = DateTime.utc_now()
+              # there is a path dependence on needing to save first
+              # before trying to change settings
```
nelsonkopliku, line 371: this comment can be safely removed since now we're relying on `insert_software_updates_settings`

lib/trento_web/controllers/v1/settings_controller.ex
nelsonkopliku: Alright, as expected. 

Since it's test related, can we possibly have the test correlation implementation handle this and return `:ok` in relevant case?

lib/trento/activity_logging/correlations.ex
```diff
@@ -1,41 +1,20 @@
...
-    # ttl unit is milliseconds
-    _ = Cachex.expire(@cache_name, key, ttl)
-    :ok
-  end
+  defp impl,
+    do:
+      Application.get_env(
+        :trento,
+        :correlations,
+        Trento.ActivityLog.Correlations.UnscopedCorrelations
```
nelsonkopliku, line 18: I know it does not change much, however for consistency I'd move this to the config.

Something like
```elixir
config :trento, correlations: Trento.ActivityLog.Correlations.UnscopedCorrelations
```

nelsonkopliku (approved):
Forgot to approve 🙈

PR #3644, opened 2025-07-17
Sap system operation route

arbulu89:
# Description

Add new `/sap_systems/:id/operations/:op` route to handle SAP system operations.
Basically it replicates what we already do for other operations.

It includes the `sap_system_start` and `sap_system_stop` operations.

The unique important detail is that in order to get the target to run the operation, I check for hosts that have the heartbeat passing. We don't want to sent the operation exactly to the host which is not running...
I don't know if we should copy this behaviour in some other operation.

PD:
The operation, user policies and activity logging are not implemented yet. I will add them afterwards.

## How was this tested?

UT


lib/trento/sap_systems.ex
```diff
@@ -104,6 +112,40 @@ defmodule Trento.SapSystems do
...
 
+  @spec request_operation(atom(), Ecto.UUID.t(), map()) ::
+          {:ok, String.t()} | {:error, any}
+  def request_operation(operation, sap_system_id, params)
+      when operation in SapSystemOperations.values() do
+    operation_id = UUID.uuid4()
+
+    # Look for 1st running host to send the operation including the instance_number
+    # Not checking if the SAP system is deregistered. That must be done by the function user
+    agents =
```
arbulu89, line 123: I'm wondering if I put the "policy" to have at least one running host.
Because I want to forbid the operation if this happens.
The other option is to return 404, but it doesn't fully represent the reality.
Thought?

arbulu89: Perhaps I just send the first available host otherwise (if no running host if found). Just to not change how we send requests.
It would timeout in the worst case.
At the end, we do the same with any other operation.

test/trento/sap_systems_test.exs
```diff
@@ -230,4 +248,143 @@ defmodule Trento.SapSystemsTest do
...
+      %{
+        operation: :sap_system_stop,
+        expected_operator: "sapsystemstop@v1"
+      }
+    ]
+
+    for %{operation: operation} = scenario <- scenarios do
+      @scenario scenario
+
+      test "should request #{operation} operation" do
```
nelsonkopliku, line 272: **suggestion:** we might consider extending by using different payloads that turns to different messages being published.

lib/trento/sap_systems.ex
nelsonkopliku: Alright. In that case what do you think about just removing the `without targets` related test and add a comment in the function implementation?

nelsonkopliku (approved):
Thanks!

PR #3656, opened 2025-07-22
SAP start/stop operation activity log

arbulu89:
# Description
Record SAP start/stop operations in the activity log.

image

image

## How was this tested?
UT


nelsonkopliku (approved):
I am happy that it's become _boring_ enhancing the activity log 😄

PR #3676, opened 2025-07-28
Polish openapi spec

nelsonkopliku:
# Description

Since we're heading to version 3.0 it might make sense polishing the system from deprecations.
This PR aims to address polishing OpenAPI spec.

What's in it:
- deleted deprecated endpoints/specs
- improve current tagging (to be agreed upon and finalized)

TODO:
- clean up unused database tables/fields, if any


gagandeepb (approved):

lib/trento_web/controllers/session_controller.ex
```diff
@@ -90,41 +89,9 @@ defmodule TrentoWeb.SessionController do
     end
   end
 
-  operation :show,
```
arbulu89, line 93: question:
this was superseded by the `/api/profile` call, right?

arbulu89 (approved):
Nice! Good job
Maybe at some point we should do the same with deprecated fields in some payloads as well.

lib/trento_web/controllers/session_controller.ex
nelsonkopliku: yep, indeed

PR #3700, opened 2025-08-05
Cluster host start/stop operations user policy

balanza:
Add new cluster host start/stop operations user policies.

Depends on #3691

### DIFF
--- a/test/trento_web/controllers/v1/cluster_controller_test.exs
+++ b/test/trento_web/controllers/v1/cluster_controller_test.exs
@@ -264,7 +264,7 @@ defmodule TrentoWeb.V1.ClusterControllerTest do
       |> assert_schema("NotFound", api_spec)
     end
 
-    pacemaker_operations_scenarios = [
+    operations_scenarios = [
       %{
         operation: "pacemaker_enable",
         host_units: [
@@ -276,10 +276,18 @@ defmodule TrentoWeb.V1.ClusterControllerTest do
         host_units: [
           build(:host_systemd_unit, name: "pacemaker.service", unit_file_state: "enabled")
         ]
+      },
+      %{
+        operation: "cluster_host_start",
+        host_units: []
+      },
+      %{
+        operation: "cluster_host_stop",
+        host_units: []
       }
     ]
 
-    for %{operation: operation, host_units: host_units} <- pacemaker_operations_scenarios do
+    for %{operation: operation, host_units: host_units} <- operations_scenarios do
       @operation operation
       @host_units host_units
 
@@ -397,7 +405,7 @@ defmodule TrentoWeb.V1.ClusterControllerTest do
       end
     end
 
-    forbidden_pacemaker_operations_scenarios = [
+    forbidden_operations_scenarios = [
       %{
         name: "enable already enabled pacemaker",
         operation: "pacemaker_enable",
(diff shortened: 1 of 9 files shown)


test/trento_web/controllers/v1/cluster_controller_test.exs
```diff
@@ -276,10 +276,18 @@ defmodule TrentoWeb.V1.ClusterControllerTest do
...
+        host_units: []
+      },
+      %{
+        operation: "cluster_host_stop",
+        host_units: []
       }
     ]
 
-    for %{operation: operation, host_units: host_units} <- pacemaker_operations_scenarios do
+    for %{operation: operation, host_units: host_units} <- operations_scenarios do
```
gagandeepb, line 290: This is a good pattern for parameterizing tests. I think we should have an internal conversation in adopting it whenever we need to run some variation on the same test body. One library that I have liked that makes this pattern easier to use [is this one.](https://hexdocs.pm/parameterized_test/readme.html)
Example usage from library docs: 

```elixir
setup context do
  # context.permissions gets set by the param_test below
  permissions = Map.get(context, :permissions, nil)
  user = AccountsFixtures.user_fixture(permissions: permissions)
  %{user: user}
end

param_test "users with editor permissions or better can edit posts",
           """
           | permissions | can_edit? | description                     |
           |-------------|-----------|---------------------------------|
           | :admin      | true      | Admins have max permissions     |
           | :editor     | true      | Editors can edit (of course!)   |
           | :viewer     | false     | Viewers are read-only           |
           | nil         | false     | Anonymous viewers are read-only |
           """,
           %{user: user, permissions: permissions, can_edit?: can_edit?} do
  assert Posts.can_edit?(user) == can_edit?, "#{permissions} permissions should grant edit rights"
end
```

gagandeepb (approved):
Looks good. Just one comment (that is more for motivating internal conversation).

test/trento_web/controllers/v1/cluster_controller_test.exs
balanza: I used to be a huge fan of that, until I had to actually manain that 🤗

gagandeepb: Can you please explain your response? I am not sure I understood. Are you referring to the maintenance of the library I linked to? Or usage of the pattern exposed by the library? Or something else?

gagandeepb: Anyway, in cases where we have static values, it should work/be useful, and there's quite a few places like that. There's also some places elsewhere where this parameterization is happening inside the test body. There would certainly be exceptional cases where this pattern would not be useful/feasible to use like @arbulu89 mentioned. In any case, this table pattern is equivalent to the pattern @balanza has actually used and follows from some macro based helpers exposed by the library I mentioned. The example I specified above is actually equivalent to the following which is what we would end up implementing in any case, although in a slightly more verbose way:

```elixir
setup context do
  permissions = Map.get(context, :permissions, nil)
  user = AccountsFixtures.user_fixture{permissions: permissions}
  %{user: user}
end

for {permissions, can_edit?, description} <- [
        {:admin,  true,  "Admins have max permissions"},
        {:editor, true,  "Editors can edit (of course!)"},
        {:viewer, false, "Viewers are read-only"},
        {nil,     false, "Anonymous viewers are read-only"}
      ] do
    @permissions permissions
    @can_edit? can_edit?
    @description description

    @tag permissions: @permissions
    @tag can_edit?: @can_edit?
    @tag description: @description
    test "users with at least editor permissions can edit posts — #{@description}", %{user: user} do
      assert Posts.can_edit?(user) == @can_edit?
    end
  end
end
```

PR #3701, opened 2025-08-05
Analytics Eula Modal

jagabomb:
# Description

This PR adds the **Analytics Eula Modal** which prompts the user to accept or decline to the collection of analytics. This is initiated once the user has logged and it is displayed on the Dashboard (Home) screen. The modal only displays for users that are **not** the default admin user. 

**Scenarios**
1. If the user, clicks the **Accept Analytics Collection** button then it stops showing the modal and analytics is enabled. 
2. If the user, clicks **Continue without Analytics** then it continues to show every time the user logs in or returns to the Dashboard (home) screen.
3. If the user, enables the checkbox (Never show this again) and clicks **Continue without Analytics** then the modal with not show again and the analytics will remain disabled. 
Screenshot 2025-08-18 at 12 48 25

## How was this tested?

- [x] **DONE** Tests the presence of the `AnalyticEulaModal` component
- [x] **DONE**  Tests the value of the `analytics_eula_accepted` value in the backend

- [ ] **DONE** This will be done in a future PR

### DIFF
--- a/assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
+++ b/assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
@@ -15,13 +15,13 @@ export default function AnalyticsEula() {
     return null;
   }
 
-  const { analytics_eula_accepted, username } = useSelector(getUserProfile);
+  const user = useSelector(getUserProfile);
+  const { analytics_eula_accepted } = user;
   const [analyticsEulaModalOpen, setAnalyticsEulaModalOpen] = useState(
     !analytics_eula_accepted
   );
   const [checked, setChecked] = useState(false);
 
-  const user = { username };
   const isDefaultAdmin = isAdmin(user);
   const dispatch = useDispatch();
 
--- a/assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
+++ b/assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
@@ -20,7 +20,6 @@ export default function AnalyticsEula() {
   const [analyticsEulaModalOpen, setAnalyticsEulaModalOpen] = useState(
     !analytics_eula_accepted
   );
-  const [checked, setChecked] = useState(false);
 
   const isDefaultAdmin = isAdmin(user);
   const dispatch = useDispatch();
@@ -36,8 +35,6 @@ export default function AnalyticsEula() {
   return (
     <AnalyticsEulaModal
       isOpen={!isDefaultAdmin && analyticsEulaModalOpen}
-      checked={checked}
-      onChecked={() => setChecked((prev) => !prev)}
       onEnable={() => {
         setAnalyticsEulaModalOpen(false);
         updateAnalyticsEula({
@@ -45,7 +42,7 @@ export default function AnalyticsEula() {
           analytics_eula_accepted: true,
         });
       }}
-      onCancel={() => {
+      onCancel={(checked) => {
         setAnalyticsEulaModalOpen(false);
         if (checked) {
           updateAnalyticsEula({
--- a/assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
+++ b/assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
@@ -1,0 +1,58 @@
+import React, { useState } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+
+import { isAdmin } from '@lib/model/users';
+import { editUserProfile } from '@lib/api/users';
+import { getFromConfig } from '@lib/config/config';
+import { setUser } from '@state/user';
+import { getUserProfile } from '@state/selectors/user';
+import AnalyticsEulaModal from './AnalyticsEulaModal';
+
+const analyticsEnabledConfig = getFromConfig('analyticsEnabled');
+
+export default function AnalyticsEula() {
+  if (!analyticsEnabledConfig) {
+    return null;
+  }
+
+  const { analytics_eula_accepted, username } = useSelector(getUserProfile);
+  const [analyticsEulaModalOpen, setAnalyticsEulaModalOpen] = useState(
+    !analytics_eula_accepted
+  );
+  const [checked, setChecked] = useState(false);
+
+  const user = { username };
+  const isDefaultAdmin = isAdmin(user);
+  const dispatch = useDispatch();
+
+  const updateAnalyticsEula = (params) => {
+    editUserProfile(params)
+      .then(({ data: userData }) => {
+        dispatch(setUser(userData));
+      })
+      .catch(() => {});
+  };
+
+  return (
+    <AnalyticsEulaModal
+      isOpen={!isDefaultAdmin && analyticsEulaModalOpen}
+      checked={checked}
+      onChecked={() => setChecked((prev) => !prev)}
+      onEnable={() => {
+        setAnalyticsEulaModalOpen(false);
+        updateAnalyticsEula({
+          analytics_enabled: true,
+          analytics_eula_accepted: true,
+        });
+      }}
+      onCancel={() => {
+        setAnalyticsEulaModalOpen(false);
+        if (checked) {
+          updateAnalyticsEula({
+            analytics_eula_accepted: checked,
+          });
+        }
+      }}
+    />
+  );
+}
--- a/assets/js/pages/AnalyticsEula/AnalyticsEulaModal.jsx
+++ b/assets/js/pages/AnalyticsEula/AnalyticsEulaModal.jsx
@@ -1,0 +1,52 @@
+import React from 'react';
+
+import Modal from '@common/Modal';
+import Button from '@common/Button';
+import Input from '@common/Input';
+
+function AnalyticsEulaModal({
+  isOpen = false,
+  checked = false,
+  onChecked,
+  onEnable,
+  onCancel,
+}) {
+  return (
+    <Modal
+      className="!w-1/2"
+      title="Collection of Anonymous Metrics"
+      open={isOpen}
+      onClose={() => null}
+    >
+      Allow the collection of{' '}
+      <a
+        className="text-jungle-green-500 hover:opacity-75"
+        href="https://trento-project.io/docs"
+        target="_blank"
+        rel="noreferrer"
+      >
+        anonymous metrics
+      </a>{' '}
+      to help improve Trento. This can also be managed in your Profile.
+      <div>
+        <Input
+          className="inline-block py-4 pr-2"
+          type="checkbox"
+          checked={checked}
+          onChange={onChecked}
+        />
+        Never show this message again.
+      </div>
+      <div className="flex flex-row space-x-2">
+        <Button type="default-fit" onClick={onEnable}>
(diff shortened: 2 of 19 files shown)


lib/trento/users.ex
```diff
@@ -261,6 +262,14 @@ defmodule Trento.Users do
     Map.put(attrs, :analytics_enabled_at, nil)
   end
 
+  defp maybe_accept_analytics_eula(%{analytics_eula_accepted: true} = attrs) do
+    Map.put(attrs, :analytics_eula_accepted_at, DateTime.utc_now())
+  end
+
+  defp maybe_accept_analytics_eula(attrs) do
```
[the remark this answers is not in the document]
jagabomb: Thanks for these changes @arbulu89. I have applied them and removed the ability to set the  `analytics_eula_accepted_at ` to false. I have also added a **test** to check for the `analytics_eula_accepted_at` value.

lib/trento/users.ex
```diff
@@ -257,10 +258,18 @@ defmodule Trento.Users do
...
   end
 
-  defp maybe_enable_analytics(attrs) do
+  defp maybe_enable_analytics(%{analytics_enabled: false} = attrs) do
     Map.put(attrs, :analytics_enabled_at, nil)
   end
 
+  defp maybe_enable_analytics(attrs), do: attrs
+
+  defp maybe_accept_analytics_eula(%{analytics_eula_accepted: true} = attrs) do
```
arbulu89, line 267: nitpick:
maybe for our future selves, we can put a comment on top of `maybe_accept_analytics_eula` saying that we only can put this value to `true` and not false, as this is one time change (at least by now).
What do you think?

lib/trento_web/openapi/v1/schema/user.ex
```diff
@@ -165,6 +170,11 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             type: :boolean,
             description: "Whether user analytics collection is enabled",
             nullable: false
+          },
+          analytics_eula_accepted: %Schema{
+            type: :boolean,
+            description: "Whether user analytics eula is accepted",
```
arbulu89, line 176: nitpick:
maybe here as well, in the `description`, we could say that setting this value to `false` has no effect

test/trento/users_test.exs
```diff
@@ -114,6 +114,26 @@ defmodule Trento.UsersTest do
...
+
+      refute analytics_eula_accepted_at == nil
+
+      assert {:ok,
+              %User{
+                analytics_eula_accepted_at: analytics_eula_accepted_at
+              }} =
+               Users.update_user_profile(user, %{analytics_eula_accepted: false})
+
+      assert analytics_eula_accepted_at == nil
```
arbulu89, line 135: we could maybe add a 3rd scenario where we don't set the `analytics_eula_accepted_at` and check that the value doesn't change:
```
assert {:ok,
    %User{
        analytics_eula_accepted_at: analytics_eula_accepted_at
      }} = Users.update_user_profile(user, %{})

assert analytics_eula_accepted_at == nil
```

arbulu89 (commented):
Hey @jagabomb ,
All the backend code looks good.
I have written 3 small nitpick comments. 
Anyway, this part is all good.

PD: If you want to merge them, you can create its own PR only with backend code and we could merge, without waiting to the frontend part. Up to you.
I'm fine either way

lib/trento/users.ex
```diff
@@ -257,10 +258,20 @@ defmodule Trento.Users do
     Map.put(attrs, :analytics_enabled_at, DateTime.utc_now())
   end
 
-  defp maybe_enable_analytics(attrs) do
+  defp maybe_enable_analytics(%{analytics_enabled: false} = attrs) do
     Map.put(attrs, :analytics_enabled_at, nil)
   end
 
+  # The value for 'analytics_eula_accepted' can only be set
```
arbulu89, line 265: @jagabomb the comment needs to go above the next `maybe_accept_analytics_eula`. Not here

assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
```diff
@@ -0,0 +1,58 @@
+import React, { useState } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import { setUser } from '@state/user';
```
arbulu89, line 3: nitpick:
can we put this `import { setUser }` together with the next bunch.

assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
```diff
@@ -0,0 +1,58 @@
...
+import AnalyticsEulaModal from './AnalyticsEulaModal';
+
+const analyticsEnabledConfig = getFromConfig('analyticsEnabled');
+
+export default function AnalyticsEula() {
+  if (!analyticsEnabledConfig) {
+    return null;
+  }
+
+  const { analytics_eula_accepted, username } = useSelector(getUserProfile);
```
arbulu89, line 18: nitpick: 
maybe this looks more understandable, instead of creating a user const below:
```
const user = useSelector(getUserProfile);
const { analytics_eula_accepted, username } = user;
...
  const isDefaultAdmin = isAdmin(user);
```
What do you think?

assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
```diff
@@ -0,0 +1,58 @@
...
+  const [analyticsEulaModalOpen, setAnalyticsEulaModalOpen] = useState(
+    !analytics_eula_accepted
+  );
+  const [checked, setChecked] = useState(false);
```
arbulu89, line 22: A proposal on this `setChecked` usage. Let's see what do you think.
As this is a merely modal thing, we could do the next.

- Remove the `const [checked, setChecked] = useState(false);` and put it in the `AnalyticsEulaModal` component, before the `return`. With this, we don't need to pass the `checked` and `setChecked` params to the modal
- In the modal, where you do  `<Button type="primary-white-fit" onClick={onCancel}> (the same for enable if you want, but doesn't look needed), we could do this `onClick={() => onEnable(checked)}`. This means that the checked value is passed to the `onCanel` function
- Like this, in the this file, we could do:
```
onCancel={(checked) => {
    setAnalyticsEulaModalOpen(false);
    if (checked) {
        updateAnalyticsEula({
            analytics_eula_accepted: checked,
         });
      }
}}
```

All of this would help use to have the `useState` and its params as closest as possible to the usage, without asking other components to manage it.

assets/js/pages/AnalyticsEula/AnalyticsEulaModal.jsx
```diff
@@ -0,0 +1,51 @@
...
+      onClose={() => null}
+    >
+      Allow the collection of{' '}
+      <a
+        className="text-jungle-green-500 hover:opacity-75"
+        href="https://trento-project/docs"
+      >
+        anonymous metrics
+      </a>{' '}
+      to help improve Trento. This can also be managed also be managed in your
```
arbulu89, line 28: typo:
`be managed` is duplicated

assets/js/pages/AnalyticsEula/AnalyticsEulaModal.jsx
```diff
@@ -0,0 +1,51 @@
...
+  return (
+    <Modal
+      className="!w-1/2"
+      title="Collection of Anonymous Metrics"
+      open={isOpen}
+      onClose={() => null}
+    >
+      Allow the collection of{' '}
+      <a
+        className="text-jungle-green-500 hover:opacity-75"
```
arbulu89, line 23: question:
I guess we will need to have a proper page on analytics, right?

assets/js/pages/AnalyticsEula/AnalyticsEulaModal.stories.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+export default {
+  title: 'Patterns/AnalyticsEulaModal',
+  component: AnalyticsEulaModal,
+  argTypes: {
+    checked: {
+      type: 'boolean',
+      description: 'Checks checkbox',
+    },
+    isOpen: {
```
arbulu89, line 14: let's add the `onEnabled` and `onCancel` entries here.

assets/js/pages/AnalyticsEula/AnalyticsEulaModal.stories.jsx
```diff
@@ -0,0 +1,48 @@
...
+      type: 'boolean',
+      description: 'Sets the visibility of the modal',
+    },
+  },
+};
+
+function ButtonToOpenModal({ ...rest }) {
```
arbulu89, line 21: question:
do you think we need all this to show the modal itself?
We could simply have a static modal view, we don't need to test that it shows up when a button is clicked.
Take the `ActivityLogDetailModal` as reference for example.

We could simply have:
```
export const Default = {
  args: {
    open: false,
  },
};
```

arbulu89 (commented):
Some comments @jagabomb 

Besides this, we need to update/add tests for the `sagas/state` changes.
And the tests of the modal are not really thorough neither. We would like to test that we trigger the correct action when the buttons are clicked, with the correct checked value.

assets/js/pages/AnalyticsEula/AnalyticsEulaModal.jsx
jagabomb: I can't believe, I forgot about this one... 🤣

assets/js/pages/AnalyticsEula/AnalyticsEulaModal.jsx
jagabomb: Yes, we don't have a proper Analytics page in the docs yet. This will be added as a separate task. I also noticed the url was broken so I fixed it.

assets/js/pages/AnalyticsEula/AnalyticsEulaModal.stories.jsx
jagabomb: There is a combination of solutions in our current Storybook for modal behaviour (`ApiKeySettingsModal` VS `ActivityLogDetailModal`). I am happy to change this one to function like you are suggesting.

assets/js/pages/AnalyticsEula/AnalyticsEula.jsx
jagabomb: Cool, I have made this change. I think it reduces some confusion in the code.

assets/js/pages/AnalyticsEula/AnalyticsEulaModal.test.jsx
```diff
@@ -0,0 +1,70 @@
...
+    render(<AnalyticsEulaModal isOpen onCancel={mockOnCancel} />);
+
+    await user.click(
+      screen.getByRole('button', { name: 'Continue without Analytics' })
+    );
+
+    expect(mockOnCancel).toHaveBeenCalledWith(false);
+  });
+
+  it('it should return true when checkbox is checked and Close button is clicked ', async () => {
```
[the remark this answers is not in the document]
EMaksy: Alright

PR #3702, opened 2025-08-05
Add database and site operation buttons

arbulu89:
# Description

wip

## How was this tested?

UT


assets/js/pages/SapSystemDetails/tableConfigs.jsx
```diff
@@ -79,7 +79,7 @@ export const getSystemInstancesTableConfiguration = ({
     {
       title: '',
       key: 'actions',
-      className: 'w-40',
```
arbulu89, line 82: This file changes align instance operations button and site operation button in the same horizontal position, closer to the right margin

EMaksy (approved):
Hey this is great code looks great!

PR #3705, opened 2025-08-06
Add cluster host start/stop to the trento UI

balanza:
Add `Start host` and `Stop host` operation buttons to the cluster page

image

The two operations are enabled when the host is down or up, respectively.


assets/js/lib/operations/index.js
```diff
@@ -6,12 +6,15 @@ export const HOST_OPERATION = 'host';
 export const CLUSTER_OPERATION = 'cluster';
 export const SAP_SYSTEM_OPERATION = 'sap_system';
 export const APPLICATION_INSTANCE_OPERATION = 'application_instance';
+export const CLUSTER_HOST_START_OPERATION = 'cluster_host_start';
```
arbulu89, line 9: We don't need these 2 new scopes.
These initial entries are the operation scopes, and we already have the `CLUSTER_HOST_OPERATION`, which I think includes the new operations

assets/js/lib/operations/index.js
```diff
@@ -38,6 +43,8 @@ const OPERATION_LABELS = {
 const OPERATION_INTERNAL_NAMES = {
   'saptuneapplysolution@v1': SAPTUNE_SOLUTION_APPLY,
   'saptunechangesolution@v1': SAPTUNE_SOLUTION_CHANGE,
+  'clusterhoststart@v1': CLUSTER_HOST_START_OPERATION,
```
arbulu89, line 46: This should be what we return from wanda, and I think the value is `crmclusterstart@v1`.
And the value in the map should be `CLUSTER_HOST_START`, without the `OPERATION`

assets/js/lib/operations/index.js
```diff
@@ -52,6 +59,8 @@ const OPERATION_INTERNAL_NAMES = {
 const OPERATION_RESOURCE_TYPES = {
   [SAPTUNE_SOLUTION_APPLY]: HOST_OPERATION,
   [SAPTUNE_SOLUTION_CHANGE]: HOST_OPERATION,
+  [CLUSTER_HOST_START_OPERATION]: CLUSTER_HOST_OPERATION,
```
arbulu89, line 62: This should be `[CLUSTER_HOST_START]: CLUSTER_HOST_OPERATION`

assets/js/pages/ClusterDetails/clusterOperations.js
```diff
@@ -134,6 +140,42 @@ export const getClusterHostOperations = curry(
         setOperationModelOpen({ open: true, operation: PACEMAKER_DISABLE });
       },
     },
+    {
+      value: 'Start host',
```
arbulu89, line 144: nitpick:
Maybe `Start cluster host` ?

arbulu89 (changes requested):
Really good!
I think we just need to remove the `CLUSTER_HOST_START_OPERATION` constant and use `CLUSTER_HOST_START` in many places, and everything will fall in place.

assets/js/pages/ClusterDetails/clusterOperations.js
balanza: Well, we are acting on a host inside a cluster page, I guess it's redundant

arbulu89: @abravosuse @jagabomb what do you think?
Just to double check, i wouldn't be surprised of creating false expectations :sweat_smile: 
I would at least put some additional information in the modal. Something making super clear that we are talking about start/stop the cluster in that host.
Something like `Start cluster in host` or whatever

assets/js/lib/operations/index.js
```diff
@@ -8,10 +8,11 @@ export const SAP_SYSTEM_OPERATION = 'sap_system';
 export const APPLICATION_INSTANCE_OPERATION = 'application_instance';
 export const CLUSTER_HOST_OPERATION = 'cluster_host';
 export const DATABASE_OPERATION = 'database';
-
```
arbulu89, line 11: nitpick:
could we restore this new line?
To have a logical division between operation scopes and operations themselves

arbulu89 (approved):
Great!
Thank you!
I have commented a nitpick thing about restoring a removed new line

gagandeepb (approved):

PR #3709, opened 2025-08-07
Handle null node attributes nodes list

arbulu89:
# Description

In some cluster scenarios, for example, when ASCS/ERS cluster nodes are stopped/started, some crm node attributes are sent as `null`.

```
"NodeAttributes": {
    "Nodes": null
},
````

In this case the payload decoding raises an error as it expects a list:
```
"reason": "{:validation, %{crmmon: %{node_attributes: %{nodes: [\"is invalid\"]}}}}",
```

Or code expects a list there in `Nodes` as usually we always receive at least one attribute, the normal things being:
```
"NodeAttributes": {
  "Nodes": [
    {
      "Name": "vmnwdev02",
        "Attributes": [
          {
            "Name": "runs_ers_NWD",
            "Value": "1"
          }
        ]
      }
    ]
  }
```

This change simply converts a possible `nil` value to a list, and does nothing if a list is already coming.

## How was this tested?

UT


balanza (approved):
LGTM

Just a clarification: this is happening because a normal behaviour of Ascs/Ers cluster, right?

abravosuse (approved):
Tested using PR image in rolling environment. It works.

PR #3715, opened 2025-08-11
Refactors dispatch functions and their usage

gagandeepb:
# Description

Previously, many top level context modules had their own definitions of commanded command dispatch functions. This PR refactors this usage by centralizing the definitions in a single module and adds tests for the case of correlated/uncorrelated command dispatches. In addition to less repetition, we also get some tested scenarios of when to expect a correlated vs uncorrelated dispatch of commands from certain contexts.

## How was this tested?

UT

### DIFF
--- a/lib/trento/support/commanded_utils.ex
+++ b/lib/trento/support/commanded_utils.ex
@@ -2,7 +2,7 @@ defmodule Trento.Support.CommandedUtils do
   @moduledoc false
   alias Trento.ActivityLog
 
-  def commanded, do: Application.fetch_env!(:trento, Trento.Commanded)[:adapter]
+  def dispatch(command), do: commanded().dispatch(command)
 
   def maybe_correlated_dispatch(command, ctx \\ :uncorrelated_dispatch) do
     case ctx == :uncorrelated_dispatch do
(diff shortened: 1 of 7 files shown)


lib/trento/support/commanded_utils.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Trento.Support.CommandedUtils do
+  @moduledoc false
+  alias Trento.ActivityLog
+
+  def commanded, do: Application.fetch_env!(:trento, Trento.Commanded)[:adapter]
```
arbulu89, line 5: nitpick:
maybe we could have a `def dispatch` function to simply dispatch, instead of putting this `commanded` thing as a public thing. And `commaned` could be a private function.
At the end, it looks like a "detail".
This way, we would only export "useful" things like `dispatch` and `dispatch_correlated`, etc

lib/trento/support/commanded_utils.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  def maybe_correlated_dispatch(command, ctx \\ :uncorrelated_dispatch) do
```
arbulu89, line 7: What about renaming this to `correlated_dispatch/2` as well?
The `maybe_` thing looks a bit odd in public functions.

In fact, do we have any use case where we don't set the 2nd parameter?
I don't see any, in that case, they user could use `dispatch` itself, right?

arbulu89 (commented):
Nice @gagandeepb !

Some comments about maybe making `dispatch` public instead of `commanded` and the usage of the `maybe_..` function

arbulu89 (approved):
Fantastic!
I love it!

PR #3717, opened 2025-08-11
Make system/database operations self-exclusive and disable some options

arbulu89:
# Description

Make system/database operations exclusive.

Specifically these scenarios:
- Only 1 operation per database site is allowed. If one is running, the other site operations are disabled
- Distinguish properly SAP instance operations using the instance number. We can have multiple instances in the same host, so we need to match the instance number as well
- When a SAP instance operation is running, SAP system operations are disabled
- When a SAP system operation is running, SAP instance operations are disabled

## How was this tested?

UT

### DIFF
--- a/assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
+++ b/assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
@@ -141,13 +141,9 @@ export function GenericSystemDetails({
     ])
   );
   const operationsDisabled = !isEmpty(systemRunningOperations);
-  const instanceOperationsDisabled = find(systemRunningOperations, {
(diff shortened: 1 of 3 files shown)


abravosuse (approved):
`Distinguish properly SAP instance operations using the instance number. We can have multiple instances in the same host, so we need to match the instance number as well: in a set up with multiple instances of the same system in the same host, any operation request while there is an operation already running will fail; therefore the frontend should not allow the use to request it`
Tested with new PR image: now it doesn't allow you to request a operation on an instance if there is an operation already running for another instance of the same system in the same host

EMaksy (approved):
Nice, lgtm, i could not find anything :)

PR #3719, opened 2025-08-12
Database start/stop policy

arbulu89:
# Description

Add operation policies for `database_start` and `database_stop` operations.

These are the rules:
`database_start`:
- If the database instances are clustered with pacemaker, check cluster is in maintenance (or `ocf::suse:SAPHana/ocf::suse:SAPHanaController` are in maintenance
- If there is not system replication enabled, it is authorized
- If the request is for the primary site, it is authorized
- If the request is for a secondary site, check the primary site is running

`database_stop`:
- If the database instances are clustered with pacemaker, check cluster is in maintenance (same as above)
- If there is not system replication enabled, it is authorized
- If the request is for a secondary site, it is authorized
- If the request is for the primary site, check secondary sites are stopped
- If the database is not used by a SAP system, it is authorized
- If the database is used by a SAP system, check app instances (abap/j2ee) are stopped

@abravosuse to simply confirm what I wrote above is correct and I didn't miss anything

## How was this tested?

UT

### DIFF
--- a/lib/trento/operations/database_instance_policy.ex
+++ b/lib/trento/operations/database_instance_policy.ex
@@ -29,13 +29,13 @@ defmodule Trento.Operations.DatabaseInstancePolicy do
         },
         _params
       ) do
-    is_clustered =
+    is_clustered? =
       Enum.any?(sap_instances, fn
         %{sid: ^sid, instance_number: ^instance_number} -> true
         _ -> false
       end)
 
-    if is_clustered do
+    if is_clustered? do
       resource_id = get_cluster_resource_id(cluster)
 
       ClusterReadModel.authorize_operation(:maintenance, cluster, %{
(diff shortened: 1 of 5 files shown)


lib/trento/operations/database_instance_policy.ex
```diff
@@ -13,17 +13,50 @@ defmodule Trento.Operations.DatabaseInstancePolicy do
...
+        _params
+      ) do
+    is_clustered =
+      Enum.any?(sap_instances, fn
+        %{sid: ^sid, instance_number: ^instance_number} -> true
+        _ -> false
+      end)
+
+    if is_clustered do
+      resource_id = get_cluster_resource_id(cluster)
```
gagandeepb, line 39: minor: s/`is_clustered`/`is_clustered?`  since it is a boolean valued variable.

gagandeepb (approved):

lib/trento/operations/database_instance_policy.ex
arbulu89: Done

PR #3722, opened 2025-08-14
Adds support/compatibility with Elixir 1.18, Erlang/OTP 27 and Nodejs 22

gagandeepb:
# Description

Bumps upwards certain application/application stack dependencies, while maintaining compatibility with older versions of the BEAM stack. Depends on https://github.com/trento-project/contracts/pull/113 .


lib/trento/support/type.ex
```diff
@@ -152,10 +152,7 @@ defmodule Trento.Support.Type do
       defp polymorphic_fields,
         do:
           Enum.filter(__MODULE__.__schema__(:fields), fn field ->
-            case __MODULE__.__schema__(:type, field) do
-              {:parameterized, PolymorphicEmbed, _} -> true
-              _ -> false
-            end
+            field in [:details, :provider_data, :metadata, :polymorphic]
```
gagandeepb, line 155: Workaround (in place of using reflection) to make code type check. Open to other suggestions that also type check/don't emit warnings on compilation.

priv/repo/migrations/20240423110214_attach_abilities_to_users.exs
```diff
@@ -3,8 +3,8 @@ defmodule Trento.Repo.Migrations.AttachAbilitiesToUsers do
 
   def change do
     create table(:users_abilities) do
-      add :user_id, references(:users, primary_key: true)
-      add :ability_id, references(:abilities, primary_key: true)
+      add :user_id, references(:users)
+      add :ability_id, references(:abilities)
```
gagandeepb, line 7: From what I see in the docs the `primary_key: true` option is not present (neither now nor in the previous version of the EctoSQL dep, but it is now with the newer version of the dep erroring out while running this specific migration.
I am still evaluating the impact (if any) of editing this old migration file on existing Trento Installation's upgrade path to a newer version of Trento.

lib/trento/infrastructure/prometheus/adapter/prometheus_api.ex
```diff
@@ -162,7 +162,7 @@ defmodule Trento.Infrastructure.Prometheus.PrometheusApi do
          {:ok, samples} <- ChartIntegration.query_values_to_samples(query_values) do
       {:ok, samples}
     else
-      %HTTPoison.Response{status_code: status_code, body: body} ->
+      {:ok, %HTTPoison.Response{status_code: status_code, body: body}} ->
```
gagandeepb, line 165: Dialyzer caught this one.

lib/trento/infrastructure/commanded/middleware/enrich_register_application_instance.ex
```diff
@@ -27,7 +27,7 @@ defimpl Trento.Infrastructure.Commanded.Middleware.Enrichable,
         on: di.host_id == h.id,
         where:
           ^db_host in h.ip_addresses and
-            fragment("? @\\? '$.name \\? (@ == ?)'", d.tenants, literal(^tenant)) and
+            fragment("? @\\? '$.name \\? (@ == ?)'", d.tenants, identifier(^tenant)) and
```
gagandeepb, line 30: Deprecation of `literal` in favour of `identifier`.

lib/trento/support/type.ex
nelsonkopliku: Sounds reasonable to me, until we have a more robust solution.

nelsonkopliku (commented):
Hey Gagan, it sounds reasonable deferring further changes to the `type` module.
It is quite a central part of the codebase and let's be sure about it.

The other points mentioned in the previous review https://github.com/trento-project/web/pull/3722#pullrequestreview-3154552064 remain

arbulu89 (approved):
The migration looks great

PR #3742, opened 2025-08-22
[TRNT-3845] Run linters on the API spec and fix the errors (general info)

antgamdia:
# Description

This PR continues the work in #3676 and starts passing some linters (`redocly`, `vacuum` and `spectral`) over the API spec file. For now, the main goal is just to reduce the number of errors being found. Reaching zero errors/warnings is not yet intended.

Particularly, this PR improves the `info` section in the OAS file and starts improving some messages from the HTTP status codes.

Related # TRNT-3845

## How was this tested?

`openapi-diff`


nelsonkopliku (approved):
Awesome, Antonio! Thanks a lot!

PR #3743, opened 2025-08-22
[TRNT-3845] Run linters on the API spec and fix the errors (schemas)

antgamdia:
# Description

This PR continues the work in #3742 and starts passing some linters (`redocly`, `vacuum` and `spectral`) over the API spec file. For now, the main goal is just to reduce the number of errors being found. Reaching zero errors/warnings is not yet intended.

Particularly, this PR improves the schemas by using longer descriptions and adding examples.

Related # TRNT-3845

## How was this tested?

`openapi-diff`

PR #3744, opened 2025-08-22
[TRNT-3845] Run linters on the API spec and fix the errors (controllers)

antgamdia:
# Description

Particularly, this PR improves the controllers by using longer descriptions and adding examples.

There are some incompatible changes (and a bunch of them in another PR); we'll need to check them more closely.

Related # TRNT-3845

## How was this tested?

`openapi-diff`


nelsonkopliku:
Antonio, thanks for highlighting the improvement areas of the open api specification.

I realize we should have been more meticulous with those. I share your pain! 

Now, we have to find a way forward towards a better place. 
Let's talk about it and see whether we just have to address everything straightaway or if we can narrow the scope of changes at the beginning.

PR #3749, opened 2025-08-25
Revoke Personal Access Tokens

nelsonkopliku:
# Description

This PR adds the capability to revoke a user's api key.

Notes:
- in this specific change it means removing references to a key
- this specific endpoint is for a user revoking own PATs

Follow ups:
- changes to the auth mechanism so that a revoked key actually does not allow making requests to trento anymore
- user admin specific revoking endpoint (`DELETE /users/:id/tokens/:jti`)

## How was this tested?

Automated tests.

### DIFF
--- a/lib/trento_web/controllers/v1/personal_access_tokens_controller.ex
+++ b/lib/trento_web/controllers/v1/personal_access_tokens_controller.ex
@@ -16,6 +16,8 @@ defmodule TrentoWeb.V1.PersonalAccessTokensController do
 
   operation :create_personal_access_token,
     summary: "Creates a new Personal Access Token",
+    description:
+      "Creates a new Personal Access Token for the currently authenticated user, allowing programmatic access to the APIs.",
     tags: ["Profile"],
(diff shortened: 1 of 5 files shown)


lib/trento_web/controllers/v1/personal_access_tokens_controller.ex
```diff
@@ -53,4 +53,33 @@ defmodule TrentoWeb.V1.PersonalAccessTokensController do
...
+        in: :path,
+        required: true,
+        type: %OpenApiSpex.Schema{
+          type: :string,
+          format: :uuid,
+          example: "550e8400-e29b-41d4-a716-446655440000"
+        }
+      ]
+    ],
+    responses: [
```
arbulu89, line 71: Do we have a `not found` return?
I mean, it can happen that the coming jti doesn't exist or it is already removed.

arbulu89 (approved):
Nice!
only the `not_found` response missing in the controller spec

PR #3758, opened 2025-09-05
[TRNT-3845] Run linters on the API spec and fix the errors (controllers: type to schema)

antgamdia:
# Description

Related # TRNT-3845

## How was this tested?

`openapi-diff`


nelsonkopliku (approved):
I guess `type` was just wrong as it is not part of the Parameter Object as [defined in the spec](https://swagger.io/specification/#parameter-object).

And that causes validation issues.

If this is the case I am fine with the breaking change. Well done

antgamdia:
> I guess `type` was just wrong as it is not part of the Parameter Object as [defined in the spec](https://swagger.io/specification/#parameter-object).
> 
> And that causes validation issues.

Yes, that's exactly the issue. We incorrectly used this `type` property and the CI is complaining about it... but it is not allowed according to the spec, so it's ok to go ahead and merge it.

PR #3759, opened 2025-09-05
[TRNT-3845] Run linters on the API spec and fix the errors (controllers: content-type)

antgamdia:
# Description

Related # TRNT-3845

## How was this tested?

`openapi-diff`


nelsonkopliku (approved):
Alright, this is basically related to empty responses on 204/202.

If adding the content-type and schema for an empty response is mandatory for validation, then I am fine with the breaking change.

Only one thing: what do you think about defining an `EmptyResponse` schema to be reused?

nelsonkopliku:
I see. 
So if by closing this we don't need other changes and validation is ok then we can close it to reduce the changes surface.

PR #3761, opened 2025-09-05
[TRNT-3845] Run linters on the API spec and fix the errors (controllers: improve schema)

antgamdia:
# Description

Related # TRNT-3845

## How was this tested?

`openapi-diff`


nelsonkopliku (approved):
Nice. The API spec does now reflect better the reality.

I believe the breaking changes are acceptable here as well.

arbulu89 (approved):

PR #3762, opened 2025-09-05
[TRNT-3845] Run linters on the API spec and fix the errors (controllers: unprocessable_entity)

antgamdia:
# Description

Related # TRNT-3845

## How was this tested?

`openapi-diff`

### DIFF
--- a/lib/trento_web/openapi/v1/schema/sap_system.ex
+++ b/lib/trento_web/openapi/v1/schema/sap_system.ex
@@ -293,7 +293,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.SAPSystem do
           clusters_health: %Schema{
             deprecated: true,
             allOf: [
-              ResourceHealth,
+              ResourceHealth
             ]
           },
           application_cluster_health: ResourceHealth,


arbulu89 (dismissed):

antgamdia:
Thanks @nelsonkopliku , @arbulu89 for your feedback here. 

In fact, they are pretty similar schemas (even I initially thought they were 1:1 interchangeable)... but the [OpenApiSpex.JsonErrorResponse](https://github.com/open-api-spex/open_api_spex/blob/master/lib/open_api_spex/json_error_response.ex) is missing some stuff, and this makes the linter complain: 


```console
Running spectral linter...
Found 109 rules (81 enabled)
Linting /mnt/c/repos/trento-web/openapi.json
"\\.$"."
/mnt/c/repos/trento-web/openapi.json:3732:28 warning docs-description ""Unprocessable Entity" must match the pattern "\\.$"."
/mnt/c/repos/trento-web/openapi.json:5802:30 warning docs-description ""v1_JsonErrorResponse.description" property must be truthy."
/mnt/c/repos/trento-web/openapi.json:5802:30 error must-define-example-schema "Every DTO must define at least one example; DTO is lacking an example #/components/schemas/v1_JsonErrorResponse"

...

openapi.json:501:35  | warning  | media type is missing `examples` or `example`                          | oas3-missing-example  | Examples     | $.paths['/api/v1/clusters/{cluster_id}/checks'].post.respons...
openapi.json:5802:7  | warning  | `schemas` component `v1_JsonErrorResponse` is missing a description    | component-description | Descriptions | $.components.schemas['v1_JsonErrorResponse']
openapi.json:5804:11 | warning  | media type schema property `errors` is missing `examples` or `example` | oas3-missing-example  | Examples     | $.components.schemas['v1_JsonErrorResponse'].properties['err...
```

In fact, since the linter issues appear so frequently (23 occurrences with the first linter and 43 with the second), it makes it difficult to identify other legitimate issues. Ignoring the rules is not an option, IMO, as validating that we are providing examples as part of our schemas is a critical aspect. API clients (and AI agents in particular) need the examples to better understand how to use the API.

That said, I think we should, at least, maintain our own `json_error_response`, adding all the properties we require (eventually, contribute upstream). Or just make the `unprocessable_entity` model have the same fields as `json_error_response`, but make them optional, so that the API controller can decide which fields it should populate or not.

Also, I don't quite understand why we are using `Schema.UnprocessableEntity.response()` in some cases and another schema in other endpoints (for the same `unprocessable_entity`). With an API client perspective, I would expect the same error model for every request. Idem with `not_found`.

nelsonkopliku:
@antgamdia @arbulu89 we could have our `UnprocessableEntity` also define a `source` property as it happens in `OpenApiSpex.JsonErrorResponse` BUT not make that prop required.

This way we could always use `UnprocessableEntity` and cover both cases.

```elixir
OpenApiSpex.schema(
  %{
    type: :object,
    additionalProperties: false,
    properties: %{
      errors: %Schema{
        type: :array,
        items: %Schema{
          type: :object,
          properties: %{
            title: %Schema{type: :string, example: "Invalid value"},
            source: %Schema{ # source added as per JsonError
              type: :object,
              properties: %{
                pointer: %Schema{type: :string, example: "/data/attributes/petName"}
              },
              required: [:pointer]
            },
            detail: %Schema{type: :string, example: "null value where string expected"}
          },
          required: [:title, :detail] # source NOT required
        }
      }
    },
    required: [:errors]
  },
  struct?: false
)
```

Yes, a bit of a hack/workaround but sounds reasonable, to me at least.

What do you think?

nelsonkopliku (approved):
Noice, thanks!

Let's rebase and fix conflicts and we're good.

gagandeepb (commented):
Hey! I am a bit surprised the linter doesn't complain about UnprocessableEntity being renamed to UnprocessableContent [according to RFC-9110](https://httpwg.org/specs/rfc9110.html#status.422). There is an open PR on the open_api_spex repo to address this evolution. WDYT?

antgamdia:
> Hey! I am a bit surprised the linter doesn't complain about UnprocessableEntity being renamed to UnprocessableContent [according to RFC-9110](https://httpwg.org/specs/rfc9110.html#status.422). 

Well, all that the linter sees is a "422", so we're good on that side.

```json
          "422": {
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/v1_UnprocessableEntity"
                }
              }
            },
            "description": "A detailed error response indicating that the server cannot process the request due to semantic errors, such as invalid or missing values in the payload."
          }
```

> There is an open PR on the open_api_spex repo to address this evolution. WDYT?

No, I just performed a quick search... and nothing popped up. Since 422 has also been named as "Unprocessable Entity" in WebDAV (see [RFC-4918](https://www.rfc-editor.org/rfc/rfc4918.html#section-11.2))... I guess some implementations just use the wrong nomenclature.

PR #3763, opened 2025-09-05
[TRNT-3845] Run linters on the API spec and fix the errors (controllers: properties)

antgamdia:
# Description

Related # TRNT-3845

## How was this tested?

`openapi-diff`


lib/trento_web/controllers/v1/database_controller.ex
```diff
@@ -138,6 +138,7 @@ defmodule TrentoWeb.V1.DatabaseController do
         required: true,
         type: %OpenApiSpex.Schema{
           type: :string,
+          enum: ["database_start", "database_stop"],
```
arbulu89, line 141: Let's not do this by now.
Putting enums in schemas is a pain afterwards, as we will add new operations and it will complain that we are not backward compatible.
I would only put enums in places where we are pretty sure we are not going to add new items

arbulu89 (changes requested):
Hey @antgamdia ,
It looks good, but let's avoid putting that enum, it will be troublesome in the near future, as we might soonish add more operations, and it will break compatibility

lib/trento_web/controllers/v1/database_controller.ex
arbulu89: We can put the available options in the description entry if needed

antgamdia: Ok! Understood. Let's do it that way then, adding the options in the description, but avoiding using the `enum` field. I'll push the change soon.

arbulu89 (approved):

PR #3764, opened 2025-09-05
[TRNT-3845] Run linters on the API spec and fix the errors (controllers: routes)

antgamdia:
# Description

Particularly, this PR performs some breaking changes that need a closer review.

Related # TRNT-3845

## How was this tested?

`openapi-diff`


nelsonkopliku:
@antgamdia what about leveraging an explicit `operation_id` in the operation definition?

gagandeepb (commented):
This looks better overall to me. I particularly like that the put/patch overloading is now addressed. I think/guess there's some minor changes needed in the `ActivityCatalog` module, possibly, to fix the failing BE tests. The failing FE tests are probably a flake.

arbulu89:
> This looks better overall to me. I particularly like that the put/patch overloading is now addressed. I think/guess there's some minor changes needed in the `ActivityCatalog` module, possibly, to fix the failing BE tests. The failing FE tests are probably a flake.

Yes, I'm fixing that now.
I agree, we can go with this. It is not much boilerplate, and we can always put the options in a constant if we want to reuse them.

arbulu89:
@gagandeepb @antgamdia Activity log tests fixed. Luckily it was pretty easy to include the new variants.
I think we can move forward with this.
PD: I'm not sure why it breaks the BC compatibility, as in essence, we have not included any relevant change. Maybe it checks that the `operationId` is different now?

gagandeepb (approved):
LGTM.

PR #3767, opened 2025-09-11
Add host reboot operation policy on sap system

balanza:
Addendum to https://github.com/trento-project/web/pull/3723

Add a policy on reboot on 

```
If there is an SAP workload in the host, it is stopped (HANA instance, SAP instance)
```

### DIFF
--- a/test/trento/operations/host_policy_test.exs
+++ b/test/trento/operations/host_policy_test.exs
@@ -349,7 +349,9 @@ defmodule Trento.Operations.HostPolicyTest do
       host =
         build(:host,
           cluster: nil,
-          cluster_id: nil
+          cluster_id: nil,
+          application_instances: [],
+          database_instances: []
         )
 
       assert :ok == HostPolicy.authorize_operation(:reboot, host, %{})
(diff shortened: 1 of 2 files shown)


test/trento/operations/host_policy_test.exs
```diff
@@ -551,10 +565,60 @@ defmodule Trento.Operations.HostPolicyTest do
...
+          cluster: nil,
+          cluster_id: nil,
+          application_instances: [
+            build(:application_instance, health: Health.unknown()),
+            build(:application_instance, health: Health.passing())
+          ],
+          database_instances: []
+        )
+
+      {:error, _} = HostPolicy.authorize_operation(:reboot, host, %{})
```
arbulu89, line 606: I would find nice to check the specific error message as well.

arbulu89 (approved):
Cannot say is the piece of code that I most like with all that consecutive `if`s :rofl: 
but hey, it does the work.

test/trento/operations/host_policy_test.exs
balanza: Done it!

PR #3768, opened 2025-09-12
Add host reboot button

balanza:
Allow the users to reboot the host from the host detail page.

The operation is enabled if the host requirements are met (no sap system running in the host)

Depends on https://github.com/trento-project/web/pull/3771

### DIFF
--- a/assets/js/pages/HostDetailsPage/HostDetails.jsx
+++ b/assets/js/pages/HostDetailsPage/HostDetails.jsx
@@ -16,6 +16,7 @@ import {
   getOperationForbiddenMessage,
 } from '@lib/operations';
 import { APPLICATION_TYPE, DATABASE_TYPE } from '@lib/model/sapSystems';
+import { isValidClusterType } from '@lib/model/clusters';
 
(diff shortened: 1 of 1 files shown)


assets/js/pages/HostDetailsPage/HostDetails.jsx
```diff
@@ -67,6 +69,20 @@ export const buildCidrNotation = (ipAddresses, netmasks) =>
     (address, netmask) => `${address}${netmask ? `/${netmask}` : ''}`
   );
 
+const clusterCanReboot = (cluster) => {
+  if (!cluster) {
+    return true;
+  }
+  switch (get(cluster, 'type')) {
```
arbulu89, line 76: We have the `isValidClusterType` function in `lib/model/cluster.js`. It could help:
```
... // instead of the swtich
return isValidClusterType(get(cluster, 'type'))
```

arbulu89 (approved):
Alright!
It worries me that the user might not understand why the operation is enabled/disabled because it is not obvious, and we don't have anything explaining it.

PR #3772, opened 2025-09-15
Update trento docs link in layout

arbulu89:
# Description

Update the trento docs link in the layout (bottom left corner).

PD: Do we plan to link the new documentation @EMaksy is working on in the webpage as well at some point?
PD2: We have the upstream link still in the profile view, in the new analytics enablement button. I guess this will be changed at some point as well


abravosuse (approved):
Tested on PR environment. LGTM.

PR #3778, opened 2025-09-17
Token introspection

nelsonkopliku:
# Description

This is the first of the two PRs introducing the token introspection endpoint.
See:
- https://www.oauth.com/oauth2-servers/token-introspection-endpoint/ 
- https://datatracker.ietf.org/doc/html/rfc7662)

Specification mandates that
> To prevent token scanning attacks, the endpoint MUST also require some form of authorization
>

However it is currently left out to make room for a discussion.
Options are:
- we leave the endpoint unsecured initially
- we do basic auth (web and wanda share some `username/password` pair - or in the form of `client_id/client_secret`)
- we leverage the already shared `ACCESS_TOKEN_ENC_SECRET` or we require `SECRET_KEY_BASE` to be shared for another token exchange for this route

## How was this tested?

Automated tests.

Note: contextually some refactoring has been made.

TODO: wanda needs to call this endpoint at authentication time.

### DIFF
--- a/test/trento/personal_access_tokens_test.exs
+++ b/test/trento/personal_access_tokens_test.exs
@@ -242,17 +242,18 @@ defmodule Trento.PersonalAccessTokensTest do
   end
 
   describe "validating a Personal Access Token" do
-    test "should return false when validating a non existent PAT" do
-      # invalid jti/user_id combination
+    test "should return false when validating a non existent PAT: invalid jti/user_id combination" do
       refute PersonalAccessTokens.valid?(Faker.UUID.v4(), FakerRandom.random_between(3, 100))
+    end
 
-      # non existent PAT for a user
+    test "should return false when validating a non existent PAT: non existent PAT for a user" do
       %User{id: user_id} = insert(:user)
       insert(:personal_access_token, user_id: user_id)
 
       refute PersonalAccessTokens.valid?(Faker.UUID.v4(), user_id)
+    end
 
-      # non matching PAT-user association
+    test "should return false when validating a non existent PAT: non matching PAT-user association" do
       %User{id: user_id} = insert(:user)
       %PersonalAccessToken{jti: pat_jti} = insert(:personal_access_token, user_id: user_id)
       %User{id: other_user_id} = insert(:user)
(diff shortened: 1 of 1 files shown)


test/trento/personal_access_tokens_test.exs
```diff
@@ -283,6 +284,70 @@ defmodule Trento.PersonalAccessTokensTest do
...
+  describe "validating and introspecting a Personal Access Token" do
+    test "should return an error when validating and introspecting a non existent PAT" do
+      # invalid jti/user_id combination
+      assert {:error, :invalid_token} =
+               PersonalAccessTokens.validate_and_introspect(
+                 Faker.UUID.v4(),
+                 FakerRandom.random_between(3, 100)
+               )
+
+      # non existent PAT for a user
```
arbulu89, line 296: question: 
why not individual tests?

arbulu89 (approved):
Looks good to me!
I have just some code structure details, but overall it doesn't change the functionality.
Besides this PR is not yet affected by how the introspection endpoint is guarded, so I'm fine to move forward

lib/trento_web/auth/tokens.ex
```diff
@@ -0,0 +1,114 @@
...
+    jwt_token
+    |> introspect_token()
+    |> case do
+      {:ok, claims} -> Map.put(claims, "active", true)
+      {:error, _} -> %{"active" => false}
+    end
+    |> Map.drop(["typ"])
+  end
+
+  defp validate_token(jwt_token), do: apply_callback(jwt_token, &validate_token/2)
```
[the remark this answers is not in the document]
nelsonkopliku: Right, I changed like this daee4e4b921b8cb618e1208190819bae55c14102. Let me know if that looks better.

test/trento/personal_access_tokens_test.exs
nelsonkopliku: Because they were just detailed different reasons of the main use case 🙈 

Anyway I changed to this fa4a3b66b0a539833c64fbdd14fc31d645fa465a. Let me know if that's ok.

PR #3783, opened 2025-09-19
Updated HTTPoison to 2.2.3

skrech:
# Description

Updates http library HTTPoison from 2.2.1 to 2.2.3 which fixes and problem caused by wrong merging of SSL options when making HTTPS requests.

Manually tested if SUMA integration works.


nelsonkopliku (approved):
Awesome, thanks! 🚀

balanza:
> Actually, `~> 2.2.3` means `>= 2.2.3 and < 2.3.0`... While I think we need: `>= 2.2.3 and < 3.0.0`
> 
> Should I change it?

🤷🏼 it makes no difference, change it

gagandeepb (changes requested):
I think we should put a specific version of httpoison in `mix.exs` so that updating this library is an explicit choice, not something implicitly inferred from dependency resolution. The diff from mix.lock is usually hidden from view in PRs, and have noisy diffs sometimes. But mix.exs is usually human readable/reviewable. We would not want to accidentally update httpoison due to the needs of dependency resolution, especially due to the issues we have observed so far.

skrech:
> I think we should put a specific version of httpoison in `mix.exs` so that updating this library is an explicit choice, not something implicitly inferred from dependency resolution. The diff from mix.lock is usually hidden from view in PRs, and have noisy diffs sometimes. But mix.exs is usually human readable/reviewable. We would not want to accidentally update httpoison due to the needs of dependency resolution, especially due to the issues we have observed so far.

But I already updated the specification to:
`{:httpoison, ">= 2.2.3 and < 3.0.0"}` 

You argue that we need to set the version specifier to `{:httpoison, "== 2.2.3"}`? If so, this seems too strict to me. With the same argument we should pin every dependency like that. 
Actually, I agree with Emanuele that only the lock file is enough. To be maximally precise and maybe signal a peculiar case we also updated `mix.exs`. But this should be enough. I don't see value in disabling deps resolution.

skrech:
I see your point. I still find this tad too strict. 
I even argue that if we had pinned the previous version, and then updated OTP, we would have been in the same position. If on the other hand, we updated deps more frequently and that dep was updated to 2.2.3, we'd never have seen that problem in the first place.
That being said, I'll pin the version.

gagandeepb (approved):
Thanks! LGTM.

PR #3789, opened 2025-09-22
Makes npm version constraints flexible

gagandeepb:
Previously, npm/nodejs was pinned to v22. With this change, we allow a range of versions from 20 to 22 (both ends inclusive).

### DIFF
--- a/packaging/suse/rpm/trento-web.spec
+++ b/packaging/suse/rpm/trento-web.spec
@@ -28,7 +28,7 @@ Source2:        package-lock.json
 Source3:        node_modules.spec.inc
 %include        %{_sourcedir}/node_modules.spec.inc
 Group:          System/Monitoring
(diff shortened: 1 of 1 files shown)


packaging/suse/rpm/trento-web.spec
```diff
@@ -28,7 +28,7 @@ Source2:        package-lock.json
 Source3:        node_modules.spec.inc
 %include        %{_sourcedir}/node_modules.spec.inc
 Group:          System/Monitoring
-BuildRequires:  elixir >= 1.15, elixir-hex, npm22, erlang-rebar3, git-core, local-npm-registry, make, gcc
+BuildRequires:  elixir >= 1.15, elixir-hex, npm >= 20, npm < 23, erlang-rebar3, git-core, local-npm-registry, make, gcc
```
gagandeepb, line 31: Question: can this be `nodejs >= 20, nodejs < 23,...` or does it need to refer to `npm`?

PR #3795, opened 2025-09-25
[TRNT-3845] Add option alias openapi operation

arbulu89:
# Description

The openapi opeartion ID must be unique for all the routes. Even though elixir `open_api_spex` creates unique IDs, the format used to distinguish between entries with multiple occurrences doesn't comply with all the openapi linters.

It add a `(n)` (`n` being the occurrence) as suffix to the operation ID. Like this: 
`TrentoWeb.V1.SettingsController.update_suse_manager_settings (2)`

This doesn't comply the [operation-operationid-valid-in-url](https://quobix.com/vacuum/rules/operations/operation-operationid-valid-in-url) linter entry.

This PR adds the option to alias the operation ID in the router code itself, for each entry. It uses the `metadata` option, and I have added a custom `openapi_operation_id` value that will replace the predefined value.

## How was this tested?

UT and open checks with `./hack/api_docs_check.sh`

### DIFF
--- a/lib/trento_web/openapi/api_spec.ex
+++ b/lib/trento_web/openapi/api_spec.ex
@@ -24,6 +24,8 @@ defmodule TrentoWeb.OpenApi.ApiSpec do
         Info,
         License,
         OpenApi,
+        Operation,
+        PathItem,
         Paths,
         SecurityScheme,
         Server,
@@ -146,8 +148,25 @@ defmodule TrentoWeb.OpenApi.ApiSpec do
         excluded_versions = List.delete(available_versions, version)
         actual_versions = List.delete(available_versions, "unversioned")
 
+        # Get route aliasese from metadata openapi_operation_id entries
+        openapi_operation_id_aliases =
+          router.__routes__()
+          |> Enum.filter(fn
+            %{metadata: %{openapi_operation_id: _op_id}} ->
+              true
+
+            _ ->
+              false
+          end)
+          |> Enum.into(%{}, fn %{path: path, verb: verb, metadata: %{openapi_operation_id: op_id}} ->
+            {create_alias_key(path, verb), op_id}
+          end)
+
         router
         |> Paths.from_router()
+        |> Enum.map(fn path_item ->
+          transform_operation_id(path_item, openapi_operation_id_aliases)
+        end)
         |> Enum.reject(fn {path, _info} ->
           current_version =
             path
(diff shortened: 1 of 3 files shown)


antgamdia (approved):
Thank you! 
As discussed offline, this is a less intrusive change than https://github.com/trento-project/web/pull/3764. The only caveat is that the descriptions are, perhaps, too generic to be consumed by an LLM.
That said, since those endpoints are not going to be exposed yet, I don't think it is a problem.

antgamdia:
The main problem is that 

> TrentoWeb.V1.SettingsController.update_suse_manager_settings (2)

It is not a valid operation ID. We need to generate a different ID that is compliant with the OpenAPI Spec. Otherwise, tools dependent on the spec may (and, indeed, they do) fail.

For instance, on the MCP server side, we cannot ingest the OAS file because it throws some errors because the operation ID is not ok :(

arbulu89:
> I don't understand the underlying problem this PR tries to address. I understand [this linter rule to be about](https://quobix.com/vacuum/rules/operations/operation-operationid-valid-in-url/) about the presence of non-url friendly characters in a url. Can you provide a minimal example of the underlying problem this attempts to resolve? A sync conversation also works.

The `operation_id` of each route must be unique, and it should be URL friendly string (so no chars like spaces and brackets).
When `open_api_spex` tries to generate the `operation_id`, it takes the controller module name and the action to create it. For example:
`TrentoWeb.HealthController.health`.
In this case `health` is a function we have in the controller and it is used only in one route.

If the same function is used in multiple routes. For example, `add_tag`, the `open_api_spex` code adds a suffix to the generated operation ID, to be something like `TrentoWeb.TagsController.add_tag (2)`. And this `(2)` makes the linter raise a warning.

This is how `add_tag` is used (the same function used in more than one route):
```
post "/hosts/:id/tags", TagsController, :add_tag,
        assigns: %{resource_type: :host}
...
post "/clusters/:id/tags", TagsController, :add_tag,
        assigns: %{resource_type: :cluster}
```

gagandeepb (commented):
Thanks for explaining, I understand the problem now. I wonder what are your thoughts about addressing this issue by moving tagging actions to the resource specific Controllers, e.g. `HostController`, `ClusterController`, etc. My suggestion is instead of/in-place of the current solution which tries to tweak the output of the spec emitted, which feels (very instinctively, I admit) a bit brittle to me, especially in face of future `open_api_spex`/`phoenix` updates.

arbulu89:
I'm closing this PR in favor of: https://github.com/trento-project/web/pull/3764

PR #3799, opened 2025-09-25
[TRNT-3845] Publish merged API docs in CI

antgamdia:
# Description

This PR updates the script we have to run the linters and adds a "join-only" option. This approach is now used by CI, so, instead of `mix openapi.spec.json --start-app=false --spec TrentoWeb.OpenApi.V1.ApiSpec` (which only generates the V1 docs), it now delegates the generation to the script. This script yields an `openapi.json` file, combining all the specs in a single file.

Related # TRNT-3845

See https://github.com/trento-project/wanda/pull/649

## How was this tested?

N/A

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -141,8 +141,17 @@ jobs:
           key: erlang-${{ steps.setup-elixir.outputs.otp-version }}-elixir-${{ steps.setup-elixir.outputs.elixir-version }}-${{ hashFiles('mix.lock') }}-${{ env.MIX_ENV }}
       - name: Build docs
         uses: lee-dohm/generate-elixir-docs@v1
-      - name: Generate openapi.json
-        run: mix openapi.spec.json --start-app=false --spec TrentoWeb.OpenApi.V1.ApiSpec
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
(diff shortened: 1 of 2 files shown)


arbulu89:
Hey @antgamdia ,
I have been thinking about this. I have other proposal: https://github.com/trento-project/web/pull/3811
Which reduces the need for external tuning of our specs

nelsonkopliku:
I think both @arbulu89's proposals in https://github.com/trento-project/web/pull/3807 and https://github.com/trento-project/web/pull/3811 are valid with regards to exposing only relevant latest APIs from within a service, let's see what fits better.

However we still might need to merge web and wanda specs at a certain point to give a client (MCP in the specific case) the whole picture, correct?

arbulu89:
> I think both @arbulu89's proposals in #3807 and #3811 are valid with regards to exposing only relevant latest APIs from within a service, let's see what fits better.
> 
> However we still might need to merge web and wanda specs at a certain point to give a client (MCP in the specific case) the whole picture, correct?

At the end we improved the MCP server to accept multiple OAS entries (multiple openapi specs) all together.
So we could pass something like `--oas-path web/api/doc --oas-path wanda/api/doc` with latest/complete. 
This should work and it shouldn't need any update, as the latest/complete should be fixed paths

arbulu89:
@antgamdia I'm closing this. Overlapped by other related work

PR #3800, opened 2025-09-25
[TRNT-3845] Tag endpoints for MCP

antgamdia:
# Description

This PR adds the `MCP` tag in the API docs; besides, it tags some endpoints with it. The idea is to flag those endpoints that are ok to be exposed via LLM.

For now, this is just a proposal; we need to check with more stakeholders before merging.

See https://github.com/trento-project/wanda/pull/650

## How was this tested?

CI

### DIFF
--- a/hack/api_docs_check.sh
+++ b/hack/api_docs_check.sh
@@ -101,6 +101,7 @@ rules:
   api-health: "off"
   api-home: "off"
   common-responses-unauthorized: "off"
+  hosts-https-only-oas3: "off"
   no-numeric-ids: "off"
   no-unknown-error-format: "off"
   path-must-match-api-standards: "off"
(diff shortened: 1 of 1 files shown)


arbulu89 (commented):
I have added some suggestions. Let's put them (I know maybe I said the opposite in the past :see_no_evil: )
Let's add:
- `V1.AbilityController.index`
- `V1.ClusterController.list` (as we added the v1 and v2 in wanda, right?)
- `V1.ClusterController.request_checks_execution`  . I know i told you to remove, but thinking twice, I think it is a safe usage, as at the end, it only runs read only actions in the agents
- `V1.DatabaseController.list_databases`
- `V1.HostController.request_checks_execution` 
- `V1.SettingsController.get_api_key_settings` (maybe it can be useful to get the key from the chat...)
-

PR #3803, opened 2025-09-29
Remove `canvas` from prod deps in ./assets

skrech:
# Description

Canvas is wrongly classified as production dependency, while it's a peer dependency of `@jest/environment-jsdom-abstract`, which is a development dependency. This change marks the dependencies of `canvas` to be marked as development deps as well.

Exposes that the reported VUL-0 in https://bugzilla.suse.com/show_bug.cgi?id=1250516 could be disregarded.


arbulu89 (approved):
Let's maybe write in the description that this issue originally tries to fix a CVE in `tar-fs`, which is part of the dependency tree of `canvas`.
If just need it for tests, yes, we can remove.
Nice finding!

PR #3804, opened 2025-09-29
Fixes nodejs references in CI workflow

gagandeepb:
# Description

Reads from env nodejs version. Uses nodejs version 22 for everything except FE tests (where we use both 20 and 22). Improves cache keys by unifying references in favour of the env var.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -371,8 +371,6 @@ jobs:
       matrix:
         containers: [1, 2, 3, 4, 5, 6, 7, 8]
         include:
-          - elixir: "1.15"
-            otp: "26"
           - elixir: "1.18"
             otp: "27"
     env:


.github/workflows/ci.yaml
```diff
@@ -377,17 +375,14 @@ jobs:
             otp: "26"
           - elixir: "1.18"
             otp: "27"
```
skrech, line 377: This `include` doesn't work as expected.  One might expect that there should be runs with `{containers: 1, {elixir: 1.15, otp: 26}}` and `{containers: 1, {elixir: 1.18, otp: 27}}` but actually only one run would be made for every container because the `includes override themselves. 

If we're not going to have matrixed e2e tests, then let's just leave one object in the include, so it's not confusing.

skrech (changes requested):

skrech (approved):
Sure, you are right. But this was very tightly related to the purpose of the PR and was introduced several days ago. So, I think it's a reasonable addition for the scope of this specific PR. Sorry to "block" the merging, it was more to signal change in my mind, after giving an approval. Anyway, everything is cool, now.

gagandeepb:
I am looking into a followup PR to try to address some of @arbulu89's comments.

PR #3807, opened 2025-09-30
Openapi `latest` schema version

arbulu89:
# Description
Add `latest` openapi version.
This version includes all the latest path/operations removing deprecated items.
I have chosen to follow this, as when we include a new version of an operation, we should deprecate the old one.
This way, clients, like the MCP server, will have an openapi schema with only latest routes and without duplicated entries and schemas.

PD: we can find other ways to remove old operations without using the deprecated flag, but I found it harder to implement and a bit more opaque.

## How was this tested?

UT and some manual tests with swagger

### DIFF
--- a/lib/trento_web/openapi/api_spec.ex
+++ b/lib/trento_web/openapi/api_spec.ex
@@ -16,6 +16,7 @@ defmodule TrentoWeb.OpenApi.ApiSpec do
 
   alias OpenApiSpex.{
     Operation,
+    PathItem,
     Paths
   }
 
@@ -150,17 +151,25 @@ defmodule TrentoWeb.OpenApi.ApiSpec do
   def build_paths_for_version("latest", router) do
     router
     |> Paths.from_router()
-    |> Enum.filter(fn {path, path_item} ->
+    |> Map.new(fn {path, path_item} ->
+      attrs =
+        path_item
+        |> Map.from_struct()
+        |> Map.new(fn
+          {verb, %Operation{deprecated: true}} -> {verb, nil}
+          {verb, op} -> {verb, op}
+        end)
+
+      {path, struct(PathItem, attrs)}
+    end)
+    |> Map.filter(fn {_path, path_item} ->
       path_item
-      |> Map.from_struct()
       |> Map.values()
       |> Enum.any?(fn
-        %Operation{deprecated: true} -> false
         %Operation{} -> true
         _ -> false
       end)
     end)
-    |> Map.new()
   end
 
   def build_paths_for_version(version, router) do


lib/trento_web/openapi/api_spec.ex
```diff
@@ -139,34 +144,46 @@ defmodule TrentoWeb.OpenApi.ApiSpec do
           }
         end
       end
+    end
+  end
 
-      defp build_paths_for_version(version, router) do
-        available_versions = router.available_api_versions()
+  def build_paths_for_version("latest", router) do
```
antgamdia, line 150: suggestion: as the number of functions continues to increase, I think it'd be useful to have some comments explaining what each piece of code is doing.

lib/trento_web/openapi/v1/schema/cluster.ex
```diff
@@ -13,6 +13,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
     OpenApiSpex.schema(
       %{
         title: "ClusterResource",
+        deprecated: true,
```
antgamdia, line 16: question: are we deprecating both `operations` and `schemas`, aren't we?

lib/trento_web/openapi/api_spec.ex
```diff
@@ -139,34 +144,46 @@ defmodule TrentoWeb.OpenApi.ApiSpec do
...
+    router
+    |> Paths.from_router()
+    |> Enum.filter(fn {path, path_item} ->
+      path_item
+      |> Map.from_struct()
+      |> Map.values()
+      |> Enum.any?(fn
+        %Operation{deprecated: true} -> false
```
antgamdia, line 158: question/comment: I guess this is filtering out those operations having a `deprecated: true`.
I see we're deprecating both operations and schemas. Do you know if the linter is preventing us from either of these situations: 1) deprecated operation with non-deprecated schema (and not used elsewhere) or 2) non-deprecated operation with at least one response schema deprecated.

I don't think so, though, it's going to be just us being cautious and careful when we deprecate something.

lib/trento_web/openapi/v1/schema/cluster.ex
arbulu89: Yes, I'm deprecating both.
The truth is that deprecating the schemas is not that essential to have the `latest` schema version, as our open api schema generator library excludes schemas that are not used in the included paths.
But I think it is more informative

lib/trento_web/openapi/api_spec.ex
arbulu89: > 1) deprecated operation with non-deprecated schema (and not used elsewhere)

arbulu89:
This solution still has some limitations, as the schemas would still have the same titles, and it requires a bigger maintenance considerations, as the chances of doing mistakes and architectural decisions are there.

At the end, the schema title can be duplicated in different versions, and if some action in `v1` uses schema `x` of `v1`, and other action in `v2` uses a schema `x` again for version 2, only the latest one would be loaded.
I don't really know how realistic this scenario is in a properly designed API, but practically it is possible.

PR #3811, opened 2025-10-01
Openapi `All` specification

arbulu89:
# Description

Other alternative to the `latest` option in: https://github.com/trento-project/web/pull/3807
In this case, the I have included a new `Complete` specification (the name can be changed) which includes all the paths and schemas.
For this to be correct, all the schema titles must be unique, so I have added the `_V1` kind of suffix there.
At the end, this is `open_api_spex` limitation the way we use it. 
I think it is a good approach, as simply putting the schema in a folder might not be enough.

Besides, this helps to remove logic from the `api_docs_check.sh` script, as we already have a "complete" spec.


arbulu89:
@antgamdia @nelsonkopliku I have removed the `_V1` underscore from tittles and renamed `Complete` to `All`

antgamdia (approved):

gagandeepb (approved):

PR #3816, opened 2025-10-06
CI Improvements

gagandeepb:
# Summary 

- Implements a single source of truth for application stack components. Previously every matrix had its own declaration of versions hard-coded.

- Declared top-level `*_dev` and `*_bc` env vars for application stack components and propagates them/references them everywhere. These env vars are copied/taken from the the current and previous tool-versions file used by asdf.

- `*_dev` refers to the versions developed against and also those available in SLE16 and are read from the `.tool-versions` file. `*_bc` are declared in the the CI file under the `env` section and are put there for backwards compatibility (hence the bc) reasons in SLE15. 

- The output of the `setup-matrix-env` job is propagated wherever needed. Note that env is not accessible in the matrix section of any job and this was the main reason for using a job output as a variable here and not env variables.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,6 +22,12 @@ concurrency:
 
 env:
   MIX_ENV: test
+  elixir_dev: "1.18.4-otp-27"
+  erlang_dev: "27.1.3"
+  nodejs_dev: "22.15.1"
+  elixir_bc: "1.15.7-otp-26"
+  erlang_bc: "26.2.1"
+  nodejs_bc: "20.14.0"
 
 jobs:
   elixir-deps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -22,26 +22,54 @@ concurrency:
...
+        uses: endorama/asdf-parse-tool-versions@v1
+      - name: Compute matrix
+        run: |
+          echo "ELIXIR_DEV=${{ env.ELIXIR_VERSION }}" >> $GITHUB_OUTPUT
+          echo "ERLANG_DEV=${{ env.ERLANG_VERSION }}" >> $GITHUB_OUTPUT
+          echo "NODEJS_DEV=${{ env.NODEJS_VERSION }}" >> $GITHUB_OUTPUT
+          echo "ELIXIR_BC=${{ env.ELIXIR_BC }}" >> $GITHUB_OUTPUT
+          echo "ERLANG_BC=${{ env.ERLANG_BC }}" >> $GITHUB_OUTPUT
+          echo "NODEJS_BC=${{ env.NODEJS_BC }}" >> $GITHUB_OUTPUT
+
```
arbulu89, line 52: super nit:
can we remove this double empty line? :see_no_evil: 
One new line should be enough

arbulu89 (approved):
Really nice!
I like the approach.

I guess other alternative would be to have an additional `.tool-version` file, like `.tool-version.bc` to read the values instead of having them in the CI file, so we could even test things locally, but It doesn't change much

gagandeepb (commented):
@arbulu89: The action/step that reads values from `.tool-versions` does not accept/expose a file name parameter and relies on the default naming of the file. Therefore, I believe your suggestion about reading from `tool-versions.bc` while certainly interesting, is somewhat tricky to implement for now.

PR #3817, opened 2025-10-06
Personal access tokens activity log

arbulu89:
# Description
Add personal access token new API calls into the activity log.
**Note: I don't know how important is to have the token `name` in entries. Adding the token name during deletion is not trivial. We cannot get the name from the database, as then moment we log the thing in the activity log, the token is already deleted. The unique "easy" way to do this would be to return the token name in the delete endpoint response body.**

image

image

image

image

## How was this tested?

UT

### DIFF
--- a/lib/trento/activity_logging/parser/phoenix_conn_parser.ex
+++ b/lib/trento/activity_logging/parser/phoenix_conn_parser.ex
@@ -183,10 +183,15 @@ defmodule Trento.ActivityLog.Logger.Parser.PhoenixConnParser do
   def get_activity_metadata(
         :personal_access_token_deletion,
         %Plug.Conn{
-          params: params
+          params: params,
+          assigns: %{
+            deleted_token: %{
+              name: name
+            }
+          }
         }
       ),
-      do: params
+      do: Map.put(params, :name, name)
 
   def get_activity_metadata(
         :personal_access_token_admin_deletion,
@@ -194,10 +199,15 @@ defmodule Trento.ActivityLog.Logger.Parser.PhoenixConnParser do
           params: %{
             id: user_id,
             jti: jti
+          },
+          assigns: %{
+            deleted_token: %{
+              name: name
+            }
           }
         }
       ) do
-    %{user_id: user_id, jti: jti}
+    %{user_id: user_id, jti: jti, name: name}
   end
 
   def get_activity_metadata(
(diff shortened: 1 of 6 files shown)


arbulu89:
@nelsonkopliku @gagandeepb Doing some tests, I was surprised to see that users don't see their own profile changes in the activity log if they don't have any users ability. I have copied the behaviour, and now the users cannot see their own PAT creation/deletion. But if I don't do that, all users see everything which is even worst.

I think having your own user action entries in the activity log would be desirable, so you can see your profile update changes and PAT actions.
But the code is not really designed for that. If I would want to add a condition in the query, to always include the entries that have my users as actor, I would need to match against the `username` as the activity log stores that as `actor`. But this would bring entries for previous users that had the same username.

Maybe this was an agreed decision but I think it is an area of improvement, and I'm leaving it outside this PR and feature at least.

gagandeepb (approved):
LGTM. We can look into refining the bit about user-abilities that you mention at a later point.

nelsonkopliku:
Note: if we end up revisiting the PATs to be something different than JWTs we'll need to retouch this. But we can merge in the meantime.

assets/js/lib/model/activityLog.js
```diff
@@ -277,6 +281,24 @@ export const ACTIVITY_TYPES_CONFIG = {
     resource: (_entry) => 'Profile',
     allowedTo: userManagement,
   },
+  [PERSONAL_ACCESS_TOKEN_CREATION]: {
+    label: 'Personal Access Token Created',
+    message: (_entry) => `Personal access token was created`,
+    resource: (_entry) => 'Profile',
+    allowedTo: userManagement,
```
nelsonkopliku, line 288: by the way, if we no not mark PATs operations as allowed to user management, we'd need to make sure no sensitive information about the token is in the metadata.

Having these gated this way is a good starting point, and it highlights possible improvements in the activity log system when it comes to own actions and sensitive information.

arbulu89: Let's be coherent of what we do with other profile changes by now.

arbulu89:
> If we want to expose the name of the token during a deletion we could also consider assigning it in the conn, without necessarily returning it in the response.

@nelsonkopliku That's an excellent idea! Let me add the change

PR #4127, opened 2026-03-30
Add basic AI provider/models configuration

nelsonkopliku:
# Description

This PR introduces basic config for the AI integration within Trento.

It will be used during users' AI onboarding, to provide the list of available models for the chosen provider.

`enabled` feature flag added.

We are targeting Gemini, OpenAI and Anthropic models, to begin with 

With regards to which models, I relied on 
- https://rancher.github.io/rancher-ai-product-docs/rancher-ai/latest/en/explanations/models.html
- https://ai.google.dev/gemini-api/docs/models
- https://developers.openai.com/api/docs/models/all
- https://platform.claude.com/docs/en/about-claude/models/overview

There are APIs for the specific provider that enumerate their models, but that kind of integration is out of scope for the moment.

Documentation will follow, eventually.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -277,6 +277,40 @@ config :trento, Trento.Repo, types: Trento.Postgrex.Types
 
 config :trento, correlations: Trento.ActivityLog.Correlations.UnscopedCorrelations
 
+config :trento, :ai,
+  enabled: true,
+  providers: [
+    googleai: [
+      models: [
+        "gemini-2.5-pro",
+        "gemini-2.5-flash",
+        "gemini-2.5-flash-lite",
+        "gemini-3.1-flash-preview",
+        "gemini-3.1-flash-lite-preview",
+        "gemini-3.1-pro-preview"
+      ]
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):

gagandeepb (approved):

antgamdia (commented):
I'm not 100% sure of the approach. I mean, while I fully agree on having a closed list of supported model _providers_, I don't think we should hardcode, even if it is just a default value, any specific _model_. 

Providers usually come with a way to list the models they currently support ([example](https://developers.openai.com/api/reference/resources/models/methods/list)). Why not just have a closed set of model providers, plus a user-provided model (maybe with a default value, but not in the source code, but in the helm chart or rpm env vars)

Example of providers: https://mastra.ai/models/providers 

Besides, I think we need to add any "OpenAI-API-compatible model provider", like Ollama (aka SUSE AI).

That said, no strong opinions. We can always go ahead with this and refine it later if we find better options.

PR #4140, opened 2026-04-01
Basic AI config functions

nelsonkopliku:
# Description

This PR introduces the scaffolding to support users' AI onboarding/configuration.

Create/Update functions added.

What's next:
- retrieving a user's AI configuration
- expose the functionalities via API endpoints (GET,POST,PATCH)
- UI

Trying to keep the PRs as contained as possible.
Half of the changes in this PR are tests 🙈

### DIFF
--- a/lib/trento/ai.ex
+++ b/lib/trento/ai.ex
@@ -1,0 +1,49 @@
+defmodule Trento.AI do
+  @moduledoc """
+  The `Trento.AI` module provides functions to interact with the AI features of the Trento application.
+  """
+
+  alias Trento.AI.Configurations
+
+  @doc """
+  Checks if the AI features are enabled.
+  """
+  @spec enabled?() :: boolean()
+  def enabled?,
+    do:
+      :trento
+      |> Application.get_env(:ai, [])
+      |> Keyword.get(:enabled, false)
+
+  @doc """
+  Creates a user configuration for AI.
+
+  See `Trento.AI.Configurations.create_user_configuration/2` for more details.
+  """
+  def create_user_configuration(user, attrs) do
+    execute_if_enabled(fn -> configurations().create_user_configuration(user, attrs) end)
+  end
+
+  @doc """
+  Updates a user configuration for AI.
+
+  See `Trento.AI.Configurations.update_user_configuration/2` for more details.
+  """
+  def update_user_configuration(user, attrs) do
+    execute_if_enabled(fn -> configurations().update_user_configuration(user, attrs) end)
+  end
+
+  defp execute_if_enabled(function) do
+    if enabled?() do
+      function.()
+    else
+      {:error, :ai_features_disabled}
+    end
+  end
+
+  defp configurations do
+    :trento
+    |> Application.get_env(:ai)
+    |> Keyword.get(:configurations, Configurations)
+  end
+end
--- a/lib/trento/ai.ex
+++ b/lib/trento/ai.ex
@@ -3,6 +3,8 @@ defmodule Trento.AI do
   The `Trento.AI` module provides functions to interact with the AI features of the Trento application.
   """
 
+  alias Trento.AI.ApplicationConfigLoader
+
   alias Trento.AI.Configurations
 
   @doc """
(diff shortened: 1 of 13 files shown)


lib/trento/ai.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Trento.AI do
```
gagandeepb: The flow overall is:

If AI is enabled system wide, it can be enabled on a per user basis; what is not clear to me is how to disable it once enabled at the user level. I wonder if we also need an AI configuration `enabled_at` (either a ts or nil) or similar field?

lib/trento/ai/user_configuration.ex
```diff
@@ -0,0 +1,58 @@
...
+  @primary_key false
+  schema "ai_configurations" do
+    field :model, :string
+    field :provider, Ecto.Enum, values: LlmRegistry.providers()
+    field :api_key, EncryptedBinary, redact: true
+
+    belongs_to :user, Trento.Users.User, primary_key: true
+
+    timestamps(type: :utc_datetime_usec)
+  end
```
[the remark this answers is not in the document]
nelsonkopliku: It's not gonna be an issue adding a polymorphic embed field for provide/model specificities.

We will need very likely to adapt, however at this stage we don't have config variations.

test/trento/ai/llm_registry_test.exs
```diff
@@ -0,0 +1,98 @@
+defmodule Trento.AI.LlmRegistryTest do
```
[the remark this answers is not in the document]
nelsonkopliku: Thanks it is indeed flaky https://github.com/trento-project/web/actions/runs/23891979060/job/69667380721

Let me address it by introducing a behaviour to abstract env reading

priv/repo/migrations/20260326100000_add_ai_user_configurations.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Trento.Repo.Migrations.AddAiUserConfigurationsTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:ai_configurations, primary_key: false) do
+      add :user_id, references(:users, on_delete: :delete_all), primary_key: true
+      add :provider, :string
+      add :model, :string
+      add :api_key, :binary
```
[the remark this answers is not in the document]
nelsonkopliku: We should be covered [here](https://github.com/trento-project/web/pull/4140/changes#diff-5f13493fa889b07cc39742d14c67768ca38d6cfc0ab1b7e802905becffb014a8R17) by

```elixir
field :api_key, EncryptedBinary, redact: true
```
Let me know if you find something I missed.

lib/trento/ai/user_configuration.ex
```diff
@@ -0,0 +1,58 @@
...
+    |> validate_required([:user_id, :model, :api_key])
+    |> validate_change(:model, &validate_model/2)
+    |> unique_constraint(:user_id,
+      name: :ai_configurations_pkey,
+      message: "User already has a configuration"
+    )
+    |> foreign_key_constraint(:user_id, message: "User does not exist")
+  end
+
+  defp maybe_update_provider(attrs) do
```
arbulu89: nit:
```
defp maybe_update_provider(%{model: model} = attrs), do: Map.put(attrs, :provider, LLMRegistry.get_model_provider(model))

defp maybe_update_provider(attrs), do: attrs
```

lib/trento/ai.ex
```diff
@@ -0,0 +1,45 @@
...
+  @spec enabled?() :: boolean()
+  def enabled?,
+    do: Keyword.get(ApplicationConfigLoader.load(), :enabled, false)
+
+  @doc """
+  Creates a user configuration for AI.
+
+  See `Trento.AI.Configurations.create_user_configuration/2` for more details.
+  """
+  def create_user_configuration(user, attrs) do
```
arbulu89: question:
is this not a bit of an overkill?
no user can access this functions as long as the api routes are disabled.
it would only be available from the console, which i don't find it dangerous.
I don't know, it adds a layer of complexity that might not be needed.

arbulu89 (commented):
Hey!
I just did a superficial review.
The core logic looks good.
I'm just wondering of the extra complexity of the `enabled` logic, which protects parts of code that shouldn't be accessible for regular users

lib/trento/ai.ex
nelsonkopliku: Ok, removed 609528672968acf0f2b5dcdfaefaa2f1557aed79

gagandeepb (approved):
Thanks for this. I would be fine if we can revisit the enable/disable of the AI config in this or in a follow up PR.

PR #4143, opened 2026-04-02
Expose AI configuration in profile

nelsonkopliku:
# Description

This PR exposes a user's AI configuration as part of the `/profile` endpoint.

We need to decide wether a user manager can operate on specific users' AI configuration, similarly to what happens with PATs. 

Additionally:
- when a user is deleted, its AI config is deleted
- took the chance to add the missing "when a user is deleted, its PATs are deleted" (it does not change the overall behavior, it just helps in keeping things clean)

Depends on https://github.com/trento-project/web/pull/4140

## How was this tested?

Automated and IRL tests.

### DIFF
--- a/lib/trento_web/openapi/v1/schema/user.ex
+++ b/lib/trento_web/openapi/v1/schema/user.ex
@@ -7,6 +7,40 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
   alias TrentoWeb.OpenApi.V1.Schema.Ability.AbilityCollection
   alias TrentoWeb.OpenApi.V1.Schema.PersonalAccessToken.PersonalAccessTokenCollection
 
+  defmodule AIConfiguration do
+    @moduledoc false
+
+    OpenApiSpex.schema(
+      %{
+        title: "AIConfigurationV1",
+        description: "AI configuration for a user.",
+        type: :object,
+        nullable: true,
+        additionalProperties: false,
+        properties: %{
+          provider: %Schema{
+            type: :string,
+            description: "Chosen AI provider.",
+            example: "googleai",
+            nullable: false
+          },
+          model: %Schema{
+            type: :string,
+            description: "Chosen AI model.",
+            example: "gemini-2.0-flash",
+            nullable: false
+          }
+        },
+        example: %{
+          provider: "googleai",
+          model: "gemini-2.0-flash"
+        },
+        required: [:provider, :model]
+      },
+      struct?: false
+    )
+  end
+
   defmodule UserTOTPEnrollmentPayload do
     @moduledoc false
 
@@ -157,6 +191,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
             nullable: false,
             example: true
           },
+          ai_configuration: AIConfiguration,
           created_at: %OpenApiSpex.Schema{
             type: :string,
             format: :"date-time",
--- a/test/trento/users_test.exs
+++ b/test/trento/users_test.exs
@@ -12,6 +12,8 @@ defmodule Trento.UsersTest do
 
   alias Trento.PersonalAccessTokens.PersonalAccessToken
 
+  alias Trento.AI.UserConfiguration, as: AIUserConfiguration
+
   import Trento.Factory
 
   describe "user profile" do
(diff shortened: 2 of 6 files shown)


test/trento/users_test.exs
```diff
@@ -656,6 +686,52 @@ defmodule Trento.UsersTest do
...
+        Trento.Repo.all(
+          from pat in PersonalAccessToken,
+            where: pat.user_id == ^user_id
+        )
+      end
+
+      %{id: user_id} = user = insert(:user)
+      insert_list(3, :personal_access_token, user_id: user_id)
+
+      assert user_id
```
arbulu89: question/nit:
why this assert? It is 100% expected to this happen, unless our factory code is totally broken, right?
I would say that it simply makes the test more convoluted as you need to scroll up and down to find the function definition.
The same for the next test, I think the anonymous functions usage doesn't look really adding any value

arbulu89 (approved):
Looks good!
I didn't check the code where the `AIConfiguration` is implemented, but the code is simple enough to see that it makes sense.
I understand that you don't include the AI data in the users list as it is not planned to be displayed in the frontend, right?

test/trento/users_test.exs
nelsonkopliku: Yes, not necessary 😄  Removed d79503117cfbd5beedca12707854eec0d00f72f3

nelsonkopliku:
> the code where the `AIConfiguration` is implemented
>
It's here https://github.com/trento-project/web/pull/4140, just in case

> I understand that you don't include the AI data in the users list as it is not planned to be displayed in the frontend, right?
>
Indeed, not right now. 
I don't exclude though that at some point we will need user admins to operate on specific users' AI configuration. Similarly to what happens for PATs.

We will add that when needed.

lib/trento_web/openapi/v1/schema/user.ex
```diff
@@ -7,6 +7,40 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
...
+        type: :object,
+        nullable: true,
+        additionalProperties: false,
+        properties: %{
+          provider: %Schema{
+            type: :string,
+            description: "Chosen AI provider.",
+            example: "googleai",
+            nullable: false
+          },
```
gagandeepb, line 26: Initial support is for 3 specific model providers. What do you think about making this an enum-string?

gagandeepb (approved):

lib/trento_web/openapi/v1/schema/user.ex
nelsonkopliku: I am a bit afraid about the fact that changing the enum would require bumping version.

Anyway, I made a note of it so that we can possibly add when this chain of PRs gets merged.

lib/trento_web/openapi/v1/schema/user.ex
```diff
@@ -7,6 +7,40 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
   alias TrentoWeb.OpenApi.V1.Schema.Ability.AbilityCollection
   alias TrentoWeb.OpenApi.V1.Schema.PersonalAccessToken.PersonalAccessTokenCollection
 
+  defmodule AIConfiguration do
+    @moduledoc false
+
+    OpenApiSpex.schema(
```
antgamdia, line 13: comment: Maybe it's not caused by this change, but wanted to mention just in case:

The API docs linter is reporting some warnings:

```
openapi_V1.hQNGiQ.json:35:11    ▲           schema property `fencing_type` is missing `examples` or `example`                
openapi_V1.hQNGiQ.json:2387:27  ▲           enum is defined as nullable but does not contain a `null` value. Nullable enum...
openapi_V1.hQNGiQ.json:3238:34  ▲           enum is defined as nullable but does not contain a `null` value. Nullable enum...
```

nelsonkopliku: Not related to these changes:

first one is not showing up anymore 👀 

The others seem related to:

https://github.com/trento-project/web/blob/54bd084c152259334fd242f53ab63870b360ad6d/lib/trento_web/openapi/v1/schema/host.ex#L141-L147

https://github.com/trento-project/web/blob/54bd084c152259334fd242f53ab63870b360ad6d/lib/trento_web/openapi/v1/schema/resource_health.ex#L7-L18

https://github.com/trento-project/web/blob/54bd084c152259334fd242f53ab63870b360ad6d/lib/trento_web/openapi/v2/schema/cluster.ex#L763-L772

and it seems to be complaining because the property is nullable but null does not appear in the enum list.

To remove the warning this would work, for instance
```
enum: [nil | ClusterState.values()],
```
or having nil as actually part of the `values`, however I wouldn't change for the moment.

antgamdia: Thank you for the clarification and details!

PR #4147, opened 2026-04-03
Create/update AI Configuration

nelsonkopliku:
# Description

This PR adds the endpoints to create and update own AI configuration.

More than half of the changes are tests.

Comes after https://github.com/trento-project/web/pull/4140 and https://github.com/trento-project/web/pull/4143

Depends on https://github.com/trento-project/web/pull/4143

## How was this tested?

Automated and IRL tests.

### DIFF
--- a/lib/trento_web/controllers/v1/ai_configuration_controller.ex
+++ b/lib/trento_web/controllers/v1/ai_configuration_controller.ex
@@ -1,0 +1,75 @@
+defmodule TrentoWeb.V1.AIConfigurationController do
+  use TrentoWeb, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  alias Trento.Users.User
+
+  alias Trento.AI
+
+  alias Trento.AI.UserConfiguration
+
+  alias TrentoWeb.OpenApi.V1.Schema
+
+  import Plug.Conn
+
+  plug TrentoWeb.Plugs.LoadUserPlug
+
+  plug OpenApiSpex.Plug.CastAndValidate, json_render_error_v2: true
+  action_fallback TrentoWeb.FallbackController
+
+  operation :create_ai_configuration,
+    summary: "Creates User's AI Configuration",
+    description: "Creates a new AI configuration for the currently authenticated user.",
+    tags: ["Profile"],
+    request_body:
+      {"CreateUserAIConfiguration", "application/json", Schema.AI.CreateUserConfigurationRequest},
+    responses: [
+      created:
+        {"User AI Configuration created successfully.", "application/json",
+         Schema.AI.UserConfiguration},
+      unprocessable_entity: Schema.UnprocessableEntity.response(),
+      unauthorized: Schema.Unauthorized.response(),
+      forbidden: Schema.Forbidden.response()
+    ]
+
+  def create_ai_configuration(conn, _) do
+    %User{} = current_user = Pow.Plug.current_user(conn)
+
+    creation_params = OpenApiSpex.body_params(conn)
+
+    with {:ok, %UserConfiguration{} = user_ai_config} <-
+           AI.create_user_configuration(current_user, creation_params) do
+      conn
+      |> put_status(:created)
+      |> render(:ai_configuration, %{ai_configuration: user_ai_config})
+    end
+  end
+
+  operation :update_ai_configuration,
+    summary: "Updates User's AI Configuration",
+    description: "Updates the AI configuration for the currently authenticated user.",
+    tags: ["Profile"],
+    request_body:
+      {"UpdateUserAIConfiguration", "application/json", Schema.AI.UpdateUserConfigurationRequest},
+    responses: [
+      ok:
+        {"User AI Configuration updated successfully.", "application/json",
+         Schema.AI.UserConfiguration},
+      unprocessable_entity: Schema.UnprocessableEntity.response(),
+      unauthorized: Schema.Unauthorized.response(),
+      forbidden: Schema.Forbidden.response()
+    ]
+
+  def update_ai_configuration(conn, _) do
+    %User{} = current_user = Pow.Plug.current_user(conn)
+
+    update_params = OpenApiSpex.body_params(conn)
+
+    with {:ok, %UserConfiguration{} = user_ai_config} <-
+           AI.update_user_configuration(current_user, update_params) do
+      conn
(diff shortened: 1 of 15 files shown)


lib/trento_web/openapi/v1/schema/ai.ex
```diff
@@ -0,0 +1,104 @@
...
+    @moduledoc false
+
+    OpenApiSpex.schema(
+      %{
+        title: "CreateUserConfigurationRequestV1",
+        description: "AI configuration request for a user.",
+        type: :object,
+        additionalProperties: false,
+        properties: %{
+          model: %Schema{
```
arbulu89, line 57: question:
Don't we need the AI `provider` field here?
The same for the `update`.
Or do you extract the `provide` in the backend somehow?
Is it possible to have the same model name for multiple providers at any given time in the future?

lib/trento_web/controllers/v1/ai_configuration_controller.ex
```diff
@@ -0,0 +1,75 @@
...
+  alias TrentoWeb.OpenApi.V1.Schema
+
+  import Plug.Conn
+
+  plug TrentoWeb.Plugs.LoadUserPlug
+
+  plug OpenApiSpex.Plug.CastAndValidate, json_render_error_v2: true
+  action_fallback TrentoWeb.FallbackController
+
+  operation :create_ai_configuration,
```
arbulu89, line 20: dump question:
isn't there any option to delete AI configuration for a user?

test/trento_web/controllers/v1/ai_configuration_controller_test.exs
```diff
@@ -0,0 +1,568 @@
...
+      conn: conn,
+      api_spec: api_spec,
+      admin_user: %{id: user_id}
+    } do
+      %UserConfiguration{
+        model: _initial_model,
+        provider: _initial_provider
+      } = insert(:ai_user_configuration, user_id: user_id)
+
+      new_model = Factory.random_ai_model()
```
arbulu89: question:
this is a pretty unusual way of using the factory.
why not use the usual `build` function?

arbulu89 (approved):
Looks good!
I left some comments as I'm unsure about some of the business logic we pretend to implement.
I recommend letting somebody else who is more aware of things to double-check hehe

lib/trento_web/controllers/v1/ai_configuration_controller.ex
nelsonkopliku: Not yet defined. See this https://github.com/trento-project/web/pull/4140#discussion_r3026810715

lib/trento_web/openapi/v1/schema/ai.ex
nelsonkopliku: The idea is to extract the provider based on the model. See [here](https://github.com/trento-project/web/pull/4140/changes#diff-5f13493fa889b07cc39742d14c67768ca38d6cfc0ab1b7e802905becffb014a8)

> Is it possible to have the same model name for multiple providers at any given time in the future?
>
Good question. Technically it might be possible, although it's not in the immediate scope.

Anyway, in the current scheme of things where the provider/model map is defined as follows

```elixir
config :trento, :ai,
  enabled: true,
  providers: [
    googleai: [
      models: [
        "gemini-2.5-pro",
        "gemini-2.5-flash",
        "gemini-2.5-flash-lite",
        "gemini-3.1-flash-preview",
        "gemini-3.1-flash-lite-preview",
        "gemini-3.1-pro-preview"
      ]
    ],
    openai: [
      models: [
        "o3-mini",
        "o3",
        "gpt-4.1",
        "gpt-4",
        "gpt-5-mini",
        "gpt-5.4"
      ]
    ],
    anthropic: [
      models: [
        "claude-opus-4-6",
        "claude-sonnet-4-6",
        "claude-haiku-4-5"
      ]
    ]
  ]
```
we could have the same model in many providers by for instace prepending the provider to the model name `openai/modelname` and `googleai/modelname`

This could be either at config level or extracted when provided as a request parameter.

I think we can extend when the need arises.

test/trento_web/controllers/v1/ai_configuration_controller_test.exs
nelsonkopliku: Yep, here it is 926d41141868d6b7ee84f009a1dae9fe74e170c5

lib/trento_web/openapi/v1/schema/ai.ex
```diff
@@ -0,0 +1,104 @@
+defmodule TrentoWeb.OpenApi.V1.Schema.AI do
+  @moduledoc false
+
+  require OpenApiSpex
+  alias OpenApiSpex.Schema
+
+  defmodule UserConfiguration do
+    @moduledoc false
+
+    OpenApiSpex.schema(
```
antgamdia, line 10: comment: as in the other PR, just in case, I'd check the API docs linter https://github.com/trento-project/web/actions/runs/24125221517/job/70403937574?pr=4147 to ensure these changes are not causing the linter warnings.

gagandeepb (approved):
Thanks for this.

lib/trento_web/openapi/v1/schema/ai.ex
nelsonkopliku: Ok, I made `provider` to be explicitly set upon creation and update in this commit 700aa78010fb6daa48d01184e4b89c6b2231363b.

It was a bit more tedious than I hoped and made the PR changes rise by more than 50% 🙈 (it is fair to say that it is mainly tests, tho)

Now, in total honesty I found the previous version simpler, and it was easily extendable as well to support the same model over many providers scenario by changing the pattern of the model name (the `openai/modelname` and `googleai/modelname` option).

However, in any case it is very likely that things around this topic will change because we aim to be able to get the model list from providers' APIs, but we're not sure it will be in the very first version.

If you ask me, I would revert the last commit 700aa78010fb6daa48d01184e4b89c6b2231363b and go ahead from there, but I want to hear your opinion as well.

lib/trento_web/openapi/v1/schema/ai.ex
nelsonkopliku: same answer https://github.com/trento-project/web/pull/4143#discussion_r3054323454

PR #4154, opened 2026-04-06
Removed upper-limit on npm builddep in RPM spec

skrech:
# Description

Removed upper-limit on npm builddep in RPM spec. This fixes build issues for SLES16.1 and Factory since they provide npm24.

### DIFF
--- a/packaging/suse/rpm/trento-web.spec
+++ b/packaging/suse/rpm/trento-web.spec
@@ -28,7 +28,7 @@ Source2:        package-lock.json
 Source3:        node_modules.spec.inc
 %include        %{_sourcedir}/node_modules.spec.inc
 Group:          System/Monitoring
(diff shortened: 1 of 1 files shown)


antgamdia (approved):
We could also increase the limit, but we are already doing an unbound `elixir >= x.x.x`; doing the same for NPM does not seem like a big deal.

nelsonkopliku (approved):

gagandeepb (approved):

nelsonkopliku:
@skrech is this still needed?

In main we have https://github.com/trento-project/web/blob/954bb58b87ba417cf08437bc48a35653eb1e0f8f/packaging/suse/rpm/trento-web.spec#L38

skrech:
Yea, closing.

PR #4162, opened 2026-04-09
Add resources links in operation forbidden modal

arbulu89:
# Description

Improvement for current operation forbidden modal messages to include some links to referenced resources.
This way it is easier to navigate to the failed resources from the modal itself.

**I'm adding only the host usage by now, as example. Extending to others is pretty simple**.

In order to achieve this, additional metadata information is sent in the `forbidden` message and interpolated in the frontend using `{0}/{1}` kind of templating syntax. 
So for each `{i}`, we need to send it corresponding metadata.
The metadata is composed by `id`, `label` and `type`. These 3 elements are needed to created the link.

I didn't want to send html tags from the backend in the error message as it is not a good practice and you need to render dangerous html in the frontend.

I guess there are other options, so I'm open for feedback.

An example:
link_in_message

## How was this tested?

UT and manual testing

No need to update

### DIFF
--- a/assets/js/common/OperationModals/OperationForbiddenModal.jsx
+++ b/assets/js/common/OperationModals/OperationForbiddenModal.jsx
@@ -24,7 +24,7 @@ const formatError = (error, metadata, onCancel) => {
     // corresponding metadata by index and create the link
     if (part.match(replacePattern)) {
       const index = parseInt(part.substring(1));
-      const { id, label, type } = get(metadata, index);
+      const { id, label, type } = get(metadata, index) ?? {};
 
       return id ? (
         <Link
--- a/assets/js/common/OperationModals/OperationForbiddenModal.jsx
+++ b/assets/js/common/OperationModals/OperationForbiddenModal.jsx
@@ -1,9 +1,47 @@
 import React from 'react';
-import { noop } from 'lodash';
+import { Link } from 'react-router';
+
+import { get, map, noop, split } from 'lodash';
 
 import Modal from '@common/Modal';
 import Button from '@common/Button';
 import Banner from '@common/Banners';
+
+const replacePattern = /({\d+})/;
+const urlTypes = {
+  cluster: 'clusters',
+  database: 'databases',
+  host: 'hosts',
+  sap_system: 'sap_systems',
+};
+
+const formatError = (error, metadata, onCancel) => {
+  // find {i} entries in the text splitting the whole content
+  const parts = split(error, replacePattern);
+
+  return map(parts, (part) => {
+    // if a part matches with the replacement pattern, find the
+    // corresponding metadata by index and create the link
+    if (part.match(replacePattern)) {
+      const index = parseInt(part.substring(1));
+      const { id, label, type } = get(metadata, index);
+
+      return id ? (
+        <Link
+          className="text-jungle-green-500 hover:opacity-75"
+          onClick={onCancel}
+          key={id}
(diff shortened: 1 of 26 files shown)


lib/trento_web/controllers/error_json.ex
```diff
@@ -44,6 +44,19 @@ defmodule TrentoWeb.ErrorJSON do
...
+      errors:
+        Enum.map(errors, fn %{message: message, metadata: metadata} ->
+          %{
+            title: "Forbidden",
+            detail: message,
+            metadata: metadata
+          }
+        end)
+    }
+  end
```
balanza, line 53: **suggestion:** ‏What if we unify these two cases, so that we can avoid the duplication `:forbidden`/`:operation_forbidden`

```suggestion
  def render("403.json", %{operation_errors: errors}) do
    functor = fn
      %{message: message, metadata: metadata} ->
          %{
            title: "Forbidden",
            detail: message,
            metadata: metadata
          },
       error ->
          %{
            title: "Forbidden",
            detail: error
          }
    end 
    
    %{
      errors: Enum.map(errors, functor)
    }
  end
```

assets/js/common/OperationModals/OperationForbiddenModal.jsx
```diff
@@ -1,10 +1,48 @@
...
+const formatError = (error, metadata, onCancel) => {
+  // find {i} entries in the text splitting the whole content
+  const parts = split(error, replacePattern);
+
+  return map(parts, (part) => {
+    // if a part matches with the replacement pattern, find the
+    // corresponding metadata by index and create the link
+    if (part.match(replacePattern)) {
+      const index = parseInt(part.substring(1));
+      const { id, label, type } = get(metadata, index);
```
balanza: **question:** ‏Should we be more defensive on this? An empty item would produce an unhandled exception.

balanza (approved):
Looks good from my side, great improvement for the users. 

Just a couple of questions:
1. Is it possible not to duplicate the `:forbidden` error logic? (see comments below)
2. Would it make sense the same error introspection for other operation failures?

lib/trento_web/controllers/error_json.ex
arbulu89: of course sure!
I can do that.
Less ramification

assets/js/common/OperationModals/OperationForbiddenModal.jsx
arbulu89: Ok! Let me have a look.

lib/trento_web/controllers/error_json.ex
arbulu89: Done!
I just put the same logic inside the `Enum.map` it is simple enough without a new function declaration

assets/js/common/OperationModals/OperationForbiddenModal.jsx
arbulu89: Done!
Defaulted to empty object.
This way, if the coming value is not what we expect, the interpolation won't happen but we won't crash either.

arbulu89:
Hey @balanza ,
I have now added the other resources (cluster, database, sap system) message and metadata usage.
It is just straightforward and same logic.
I don't think a new review is needed.
I will run some additional real-infra tests though

balanza (approved):

abravosuse (approved):
Test made in rolling environment after deploying web image from PR. All test cases return the expected result.

PR #4164, opened 2026-04-09
[TRNT-4339] Add SearchableSelect

antgamdia:
TBD, will update once all the stacked PRs are ready.

Next PR:  https://github.com/trento-project/web/pull/4165

# Description

Also, include relevant motivation and context for this PR. List any dependencies that are required for this change.

Fixes # (issue)

- [ ] **DONE**

## How was this tested?

- [ ] **DONE**

- [ ] **DONE**


arbulu89 (commented):
hey @antgamdia 
I'm fine using the `react-select` `combobox` option.
Could we maybe simply import the `react-select` here instead of changing `MultiSelect`?
We could export the multi select components to keep same styles.
PD:
I think we have a tech-debt here, where we should only have one `Select` element (instead of `Select`, `MultiSelect` and `SearchableSelect`, where we can make it multi or searchable through props). All of them with `react-select`. But this is something we can do afterwards

balanza:
> I think we have a tech-debt here, where we should only have one `Select` element (instead of `Select`, `MultiSelect` and `SearchableSelect`, where we can make it multi or searchable through props). All of them with `react-select`. But this is something we can do afterwards

I'd rather not add another component, in the end this one does the same thing as `MultiSelect` (which just wraps).

antgamdia:
Thanks for your early input here. Let's not create more components then 😅 
Closing in favor of https://github.com/trento-project/web/pull/4165 and the future work from @arbulu89

PR #4165, opened 2026-04-09
[TRNT-4339] Add timezone selector in UI

antgamdia:
# Description

Next PR: https://github.com/trento-project/web/pull/4169

Related # TRNT-4339

---- 

In this PR, we use an existing UI component to render a list of timezones and pass it as a payload for the user/profile settings. This PR is just about the frontend changes, nothing backend-related here.

The logic is simple: display a list of timezones. However, where does this list come from? 
For instance, both `Intl` or the `@date-fns` wrapper we use have some support for timezones. One could possibly do `Intl.supportedValuesOf("timeZone")` to enumerate the timezones, and it works...
BUT: the list does not point to the latest version from IANA. For instance, in my browser, I just got `Europe/Kiev` but it was changed by the `2023`c release ([see source](https://data.iana.org/time-zones/tzdb-2023c/NEWS)) to `Europe/Kyiv`.

By relying on a 3rd party dependency, we stay 100% aligned with any IANA tz release in the future. Note that this also updates the DST per country, so, in case of some political decisions around DST, the IANA database should quickly reflect that.

Some decisions:

- Only "Edit user" and "user's profile" will have the timezone selector
  - That means that when creating a new user, their timezone can't be selected ---> superseeded by https://github.com/trento-project/web/pull/4200
- Timezone for the default super admin is always `Etc/UTC`, and this is not modifiable (like other decisions)

Current status:

image

Note that for the UTC offset (like GMT+2) the current DST settings are taken into account. For instance, at the time of writing, Europe is in DST, meaning that we `Europe/Madrid` is GMT+2, but once in winter, it will be like GMT+1 again.

## How was this tested?

IRL, unit and e2e tests

N/A

## Additional information 

Note: tests are currently failing because the backend is receiving more fields than it can handle. Subsequent stacked PRs will solve that.

### Other proposals

There are some visual alternatives we can discuss, but they haven't been implemented in this PR:

#### Grouping by zone:

image

#### Adding country (requires yet another dependency)

image


assets/js/lib/timezones/index.js
```diff
@@ -0,0 +1,45 @@
+import tzdata from 'tzdata';
```
balanza, line 1: **suggestion:** ‏Consider set this utilites into `lib/date/index.js`. Also, have a look at https://www.npmjs.com/package/@date-fns/tz, it might be useful as we use `@data-fns` across the codebase`

antgamdia: Thanks for having a look! Yeah, really good point. In fact, this `@date-fns` has been the approach I followed in the next PRs 
This file is totally outdated in this PR 😅

antgamdia: Re-reading it, I see you meant `tzdata` for generating the timezones. Unfortunately, this is still beng used, as I haven't been able to find a better alternative.

For instance, both `Intl` or the `@date-fns` wrapper we use have some support for timezones. Once could possible do `Intl.supportedValuesOf("timeZone")` to enumerate the timezones, and it works...
BUT: the list does not point to the latest version from IANA. For instance, in my browser, I just got `Europe/Kiev` but it was changed by the `2023c` release ([see source](https://data.iana.org/time-zones/tzdb-2023c/NEWS)) to `Europe/Kyiv`.

It is pretty nitpicky, I know, but this way we stay 100% aligned with any IANA tz release in the future. Note that this also updates the DST per country, so, in case of some political decisions around DST, the IANA database should quickly reflect that.

balanza: I didn't dig into the specifics of the library, if it's not needed so be it!

antgamdia:
Personal note: check https://github.com/trento-project/web/pull/4192 once merged

antgamdia:
> Personal note: check #4192 once merged

Done at https://github.com/trento-project/web/pull/4165/commits/fb004367b3fd32582bc0a8e2bcb2e7701e8ec187

assets/package.json
```diff
@@ -97,7 +99,7 @@
     "format:check": "prettier -c .",
     "format": "prettier --write .",
     "pretest": "ls ../priv/static/assets/app.css || npm run tailwind:build",
-    "test": "jest --maxWorkers=2",
+    "test": "jest --maxWorkers=2 --silent",
```
arbulu89: nit:
why this change?
if it is to hide warning/errors I wouldn't do it. Many times they are sensible things to do.
anyway, we shouldn't "hide" this change in this PR.

assets/js/state/user.js
```diff
@@ -48,6 +49,7 @@ export const userSlice = createSlice({
           password_change_requested,
           analytics_enabled,
           analytics_eula_accepted,
+          timezone,
```
arbulu89, line 52: nit:
can we add this value in the UT?

assets/js/pages/Users/UserForm.jsx
```diff
@@ -16,7 +16,7 @@ import {
   errorMessage,
 } from '@lib/forms';
 import { getError } from '@lib/api/validationErrors';
-
+import { DEFAULT_TIMEZONE, generateTimezoneOptions } from '@lib/timezones';
```
arbulu89: storybook:
can we add the new `timezone` field to the `UserForm` storybook file?

arbulu89: refactor:
let's pass the output of `generateTimezoneOptions` as a prop to `UserForm` from above.
`UserForm` doesn't have any external side-effect. This way it is really easy to test without mocking anything (you don't now, but the tests might run real tzdate things, which we don't want for this component)
The same for `ProfileForm`

arbulu89 (commented):
Hey @antgamdia 
I have written some "easy to fix" comments first.
Besides, I have some thoughts on how we save the timezones in this custom cache, and maybe we can do the next.
At the end, the code execution is blocked the first time we run the `generateTimezoneOptions` (unless we do it async), so why not simply have:
```
const timezones = generateTimezoneOptions();
export timezones;
```
Without any caching code, just a simple functional method.
I think the result is basically the same (this happens always at the beginning, not in the first page that needs it).
I don't know, the caching thing looks "premature optimization" for a couple of hundreds element list.

Otherwise, if we want to go fancy, we could do the same in a redux saga, but it looks an overkill.

What do you think?
Both are acceptable and similar anyway.

assets/package.json
antgamdia: True, this is an unintended leftover I added while debugging the CI errors. I found we have several warnings that do not affect the test execution. (like the `act` thing). This reduced the output and it was easier to find the failing test cases.
However, 100% agree, this change does not belong in here. Were we to add it, it deserves its own PR.

arbulu89: I'm having a look into some warnings myself. Let's see how many of them I can fix

antgamdia:
Feedback addressed; thanks a lot!

assets/js/pages/Users/UserForm.jsx
antgamdia: I've created https://github.com/trento-project/web/pull/4239 to do this automatically... it seems we had several properties missing (or worse: some non-existing properties in the stories). It's not ready yet, but at least... it's something.

arbulu89 (approved):
Great!
Really nice!

antgamdia:
Merging to feature branch: [TRNT-4339](https://github.com/trento-project/web/tree/TRNT-4339)

PR #4166, opened 2026-04-10
Ai onboarding e2e

nelsonkopliku:
# Description

Adding e2e tests for AI configuration


test/e2e/cypress/e2e/users.cy.js
```diff
@@ -486,6 +487,217 @@ describe('Users', () => {
...
+        usersPage
+          .apiCreateAIConfiguration(
+            'googleai',
+            'gemini-2.5-pro',
+            'test-api-key'
+          )
+          .then(({ provider, model }) => {
+            // reload() needed to update the profile view with the new configuration generated by the API
+            // we might remove when/if we add rectiveness so that the backend broadcasts the changes
+            cy.reload();
```
vicenteqa: nit: `usersPage.refresh()` should be available, which wraps `cy.reload()`

test/e2e/cypress/e2e/users.cy.js
```diff
@@ -486,6 +487,217 @@ describe('Users', () => {
...
+    describe('Creating AI Configuration', () => {
+      it('should not allow user to create an AI configuration if the api key is not set', () => {
+        const provider = 'Google Gemini';
+        const model = 'gemini-2.5-pro';
+
+        usersPage.clickEditAIConfigurationButton();
+        usersPage.selectAIProvider(provider);
+        usersPage.selectAIModel(model);
+
+        const shouldWaitForRequest = false;
```
vicenteqa, line 540: question, is this `shouldWaitForRequest` needed because in some scenarios the request doesn't happen after some specific actions? If that's the case 👍. Otherwise I don't see any reason to not wait always for the request to happen.

vicenteqa (approved):
just a small nit and a question, it looks good testing-wise :)

test/e2e/cypress/e2e/users.cy.js
nelsonkopliku: That's the case, sometimes the request is not even made.

PR #4168, opened 2026-04-10
Remove unnecessary LLMRegistry.get_model_provider/1 function

nelsonkopliku:
# Description

A test was failing in main https://github.com/trento-project/web/actions/runs/24236988457/job/70762242041#step:7:637

That's because in test environment we have the scenario where the same model is available for many providers.

Using the, now, deleted function was causing flakiness since retrieving the provider for a specified model could have returned different values.


gagandeepb (approved):
Thanks! LGTM, if the CI is green.

PR #4169, opened 2026-04-10
[TRNT-4339] Add timezone in the backend

antgamdia:
# Description

Previous PR: https://github.com/trento-project/web/pull/4165

Next PR: https://github.com/trento-project/web/pull/4174

Related # TRNT-4339

---- 

In this PR, once we have the UI form ready with a proper value for the timezone, we will just receive it in the backend and store it as a user setting. 

Some decisions:

- Only "Edit user" and "user's profile" will have the timezone selector
  - That means that when creating a new user, their timezone can't be selected
- Timezone for the default super admin is always `Etc/UTC`, and this is not modifiable (like other decisions)

Current status:

https://github.com/user-attachments/assets/4bd26b53-f5ac-4522-af2a-992b1b39a7a3

## How was this tested?

IRL, unit and e2e tests

N/A

## Additional information 

Note: tests are currently failing because the e2e tests need some adaptations.


assets/package.json
```diff
@@ -45,6 +45,7 @@
     "tailwindcss": "^3.4.19"
   },
   "dependencies": {
+    "@date-fns/tz": "^1.4.0",
```
gagandeepb: We already have a timezone data/tz-database/DST accounting dependency in the backend/server-side code. Can you motivate why we need this additional npm dependency?

test/trento_web/controllers/v1/users_controller_test.exs
```diff
@@ -381,13 +400,35 @@ defmodule TrentoWeb.V1.UsersControllerTest do
...
+        |> put_req_header("content-type", "application/json")
+        |> put_req_header("if-match", "#{lock_version}")
+        |> patch("/api/v1/users/#{user_id}", %{timezone: "Europe/Madrid"})
+        |> json_response(:ok)
+        |> assert_schema("UserItemV1", api_spec)
+
+      assert resp.timezone == "Europe/Madrid"
+
+      Application.put_env(:trento, :oidc, enabled: false)
+    end
```
gagandeepb: Updating/Adding global mutable state in this manner is a certain recipe for test flakiness, in my understanding. Consider re-implementing these tests in a way that that doesn't call `Application.put_env` in the test body.

arbulu89: This is true, but unfortunately we already do this in the current tests.
@antgamdia I would suggest to simply move the `timezone` test to the already existing `should update the profile with allowed fields when SSO is enabled` test for the `profile` part and leave only first one for the `user`, so there is less chances of collision.

arbulu89 (commented):
Hey @antgamdia !
The backend code looks good.
In order to proceed with a more logical review sequence, could we simply have elixir/backend code in this PR?
We can put other frontend and e2e code in different parts, I'm seeing already changes between different open PRs that can make things more difficult

assets/package.json
antgamdia: It's great that there is already a library for that in the server-side code, however, the file you are referring to is purely frontend code (which I mistakenly added in this PR 😅 ). Refer to the motivation at https://github.com/trento-project/web/pull/4165#discussion_r3086235205

lib/trento_web/openapi/v1/schema/user.ex
```diff
@@ -545,7 +579,8 @@ defmodule TrentoWeb.OpenApi.V1.Schema.User do
           updated_at: "2024-01-15T10:30:00Z",
           analytics_enabled: false,
           totp_enabled_at: "2024-01-15T09:00:00Z",
-          last_login_at: "2024-01-15T09:00:00Z"
+          last_login_at: "2024-01-15T09:00:00Z",
+          timezone: "Europe/Berlin"
```
arbulu89, line 583: super nit:
maybe we can add the `timezone` in the `UserCollection` example as well

test/trento_web/controllers/v1/profile_controller_test.exs
```diff
@@ -145,6 +145,51 @@ defmodule TrentoWeb.V1.ProfileControllerTest do
...
+      conn
+      |> put_req_header("content-type", "application/json")
+      |> patch("/api/v1/profile", %{timezone: "Europe/Berlin"})
+      |> json_response(:ok)
+      |> assert_schema("UserProfileV1", api_spec)
+
+    assert %{id: ^user_id, timezone: "Europe/Berlin"} = resp
+  end
+
+  test "should update timezone when SSO is enabled", %{
```
arbulu89: comment:
as Gagan commented, and to reduce the options for flakyness in test, I would move this test to the already existing test `should update the profile with allowed fields when SSO is enabled`.
At the end, they test exactly the same.

test/trento_web/controllers/v1/users_controller_test.exs
```diff
@@ -358,12 +376,15 @@ defmodule TrentoWeb.V1.UsersControllerTest do
       Application.put_env(:trento, :oidc, enabled: true)
 
       %{id: id, name: name, resource: resource, label: label} = insert(:ability)
-      %{id: user_id, lock_version: lock_version} = insert(:user, locked_at: DateTime.utc_now())
+
+      %{id: user_id, lock_version: lock_version} =
```
arbulu89: comment:
I guess `Mox` is smart enough to add new fields to a factory.
Can we put a default value there in the factory for completeness?
I know that it is filled by the default from the read model, but i think it improves it.

test/trento_web/controllers/v1/users_controller_test.exs
```diff
@@ -358,12 +376,15 @@ defmodule TrentoWeb.V1.UsersControllerTest do
...
+        insert(:user, locked_at: DateTime.utc_now(), timezone: "Europe/Berlin")
 
       valid_params = %{
         fullname: Faker.Person.name(),
         email: Faker.Internet.email(),
         enabled: true,
+        timezone: "Europe/Berlin",
```
arbulu89: comment:
shouldn't we put a new different timezone here to see if the change actually works?
Otherwise we keep the same value.

arbulu89 (approved):
It looks good @antgamdia 
Just some nitpicks.

PD:
@abravosuse This code assumes that the timezone cannot be set during user creation and it will always default to `UTC`. Is this the desired behaviour?
Adding the option to set the timezone during user creation is trivial.

abravosuse:
> PD: @abravosuse This code assumes that the timezone cannot be set during user creation and it will always default to `UTC`. Is this the desired behaviour? Adding the option to set the timezone during user creation is trivial.

I don't think it is necessary, because the user admin -who probably is not even be part of the SAP team- doesn't need to know the timezone that the user is interested in. That said, if it is trivial to add, it wouldn't hurt to have it.

What for me is more important -although this comment probably belongs to the UI PR more than here- is that the format of the timestamp field shows the timezone. It is assertive for the user to see the timezone explicitely.

antgamdia:
I'll go through the comment shortly, but, for now, I've been trying to repro the [CI issue](https://github.com/trento-project/web/actions/runs/24468167122/job/71504090163?pr=4169)... and it does not happen with the latest version of the linter. 

I've bumped it up at: https://github.com/trento-project/web/pull/4194

Edit: confirmed that updating the dependency fixed the issue.

abravosuse:
> I'll go through the comment shortly, but, for now, I've been trying to repro the [CI issue](https://github.com/trento-project/web/actions/runs/24468167122/job/71504090163?pr=4169)... and it does not happen with the latest version of the linter.

@arbulu89 has explained to me that including the timezone in the timestamp fields, even if it's just the timezone initials, might "break"  the UI. He has suggested that we deliver a first iteration without it and, later on, we try with it to evaluate the impact. I am ok with that approach.

arbulu89 (approved):
Really good @antgamdia 
Thank you!
we can add changes to the timezone usage in user creation afterwards if we want

antgamdia:
Thank you both for your comments! 

re:
> That said, if it is trivial to add, it wouldn't hurt to have it.

and

> we can add changes to the timezone usage in user creation afterwards if we want


+1, it is straightforward to add it. I personally would like to have it, but will do it in a follow-up PR eventually.


re:
>  is that the format of the timestamp field shows the timezone. It is assertive for the user to see the timezone explicitely.

and

> He has suggested that we deliver a first iteration without it and, later on, we try with it to evaluate the impact. I am ok with that approach.

Again, this is trivial (like in replacing `dd MMM yyyy` with `dd MMM yyyy xxx`). This is how it looks like.
I'm preparing a follow-up PR for using fewer datetime formats.

image
image

image

antgamdia:
Merging to feature branch: [TRNT-4339](https://github.com/trento-project/web/tree/TRNT-4339)

PR #4170, opened 2026-04-10
[TRNT-1598] Extract versions to env in CI

antgamdia:
# Description

This PR simply extracts some hardcoded versions we had in the pipelines to environment variables. The idea is to simply have, at a glance, all the versions we might want to update. Besides, some of them include a comment to the place where the releases are, to make the bump-up process easier.

Related # TRNT-1598

## How was this tested?

N/A


vicenteqa (approved):
hey @antgamdia, sorry for the delay, I totally forgot about this, it looks good!

PR #4172, opened 2026-04-13
Improve database start/stop operation policies to handle multi-tier setup

arbulu89:
# Description

Now that multi-tier setups are implemented, the `database start/stop` operation must be improved.
- In a multi-tier setup, the 3rd site (the disaster recovery one) is not handled by Pacemaker cluster.
In this case, if we want to start/stop this site, we don't need to check if pacemaker is there or if it is in maintenance mode. This PR filters out database instances that are not part of the requested site in the cluster maintenance check. 
- In a full database stop operation, check if application instances are stopped. It was not checked due this scenario was not even considered in the code
- Replace the usage of Primary/Secondary relation to start/stop. In a multi-tier setup we can have different secondary setups in tier 2 and tier 3. And their relationship is different. 

## How was this tested?

UT and manual testing

No need

### DIFF
--- a/lib/trento/operations/database_policy.ex
+++ b/lib/trento/operations/database_policy.ex
@@ -228,6 +228,13 @@ defmodule Trento.Operations.DatabasePolicy do
        ),
        do: application_instances_stopped(database, nil, nil)
 
+  # full database stop request, check if app instances are stopped
+  defp application_instances_stopped(database, params, _) when not is_map_key(params, :site),
+    do: application_instances_stopped(database, nil, nil)
+
+  defp application_instances_stopped(database, %{site: nil}, _),
+    do: application_instances_stopped(database, nil, nil)
+
   # secondary site
   defp application_instances_stopped(_, _, _), do: :ok
 end
--- a/lib/trento/operations/database_policy.ex
+++ b/lib/trento/operations/database_policy.ex
@@ -74,7 +74,7 @@ defmodule Trento.Operations.DatabasePolicy do
     primary_site = get_primary_site(database)
 
     OperationsHelper.reduce_operation_authorizations([
-      database_instances_cluster_maintenance(database),
+      database_instances_cluster_maintenance(database, params),
       primary_site_started(database, params, primary_site)
     ])
   end
@@ -87,7 +87,7 @@ defmodule Trento.Operations.DatabasePolicy do
     primary_site = get_primary_site(database)
 
     OperationsHelper.reduce_operation_authorizations([
-      database_instances_cluster_maintenance(database),
+      database_instances_cluster_maintenance(database, params),
       secondary_sites_stopped(database, params, primary_site),
       application_instances_stopped(database, params, primary_site)
     ])
@@ -106,11 +106,15 @@ defmodule Trento.Operations.DatabasePolicy do
     end)
   end
 
-  defp database_instances_cluster_maintenance(%DatabaseReadModel{
-         database_instances: database_instances
-       }) do
-    OperationsHelper.reduce_operation_authorizations(
-      database_instances,
+  defp database_instances_cluster_maintenance(
+         %DatabaseReadModel{
+           database_instances: database_instances
+         },
+         params
+       ) do
+    database_instances
+    |> filter_by_site(params)
+    |> OperationsHelper.reduce_operation_authorizations(
       :ok,
       fn database_instance ->
         DatabaseInstanceReadModel.authorize_operation(
--- a/lib/trento/operations/database_policy.ex
+++ b/lib/trento/operations/database_policy.ex
@@ -185,7 +185,7 @@ defmodule Trento.Operations.DatabasePolicy do
          _
        ) do
     database_instances
-    |> Enum.filter(fn %{system_replication_source_site: soure_site} -> site == soure_site end)
+    |> Enum.filter(fn %{system_replication_source_site: source_site} -> site == source_site end)
     |> Enum.all?(fn %{health: health} -> health == :unknown end)
     |> if do
       :ok
--- a/lib/trento/operations/database_policy.ex
+++ b/lib/trento/operations/database_policy.ex
@@ -133,17 +133,32 @@ defmodule Trento.Operations.DatabasePolicy do
   defp primary_site_started(_, %{site: nil}, _), do: :ok
   # primary site
   defp primary_site_started(_, %{site: site}, site), do: :ok
-  # secondary sites, check primary is started
+  # secondary sites, check primary of this site is started.
+  # for that, check site instances with the previous tier number are started.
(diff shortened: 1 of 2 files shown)


abravosuse (approved):
Web image from PR deployed on rolling environment. I can stop/start the unmanaged site regardless the maintenance status of the pacemaker cluster or the HANA resource.

lib/trento/operations/database_policy.ex
```diff
@@ -101,11 +101,15 @@ defmodule Trento.Operations.DatabasePolicy do
...
-    OperationsHelper.reduce_operation_authorizations(
-      database_instances,
+  defp database_instances_cluster_maintenance(
+         %DatabaseReadModel{
+           database_instances: database_instances
+         },
+         params
+       ) do
+    database_instances
+    |> filter_by_site(params)
```
arbulu89, line 116: This is the 1st bullet point.
Filter by requested site to check for cluster state

lib/trento/operations/database_policy.ex
```diff
@@ -124,22 +128,35 @@ defmodule Trento.Operations.DatabasePolicy do
   defp primary_site_started(_, %{site: nil}, _), do: :ok
   # primary site
   defp primary_site_started(_, %{site: site}, site), do: :ok
-  # secondary sites, check primary is started
+  # secondary sites, check primary of this site is started.
```
arbulu89, line 136: 3rd bullet point.
the `primary_site_started` is the more "controversial", as we cannot guarantee right now which is the replication source in a stopped instance.
That's why we check the tier number to see which are "parent/primary" sites.
It is suboptimal, but it is a really rare corner-case, so it is a trade-off we need to accept and acknowledge

lib/trento/operations/database_policy.ex
```diff
@@ -151,16 +168,16 @@ defmodule Trento.Operations.DatabasePolicy do
...
            database_instances: database_instances
          },
-         %{site: primary_site},
-         primary_site
+         %{site: site},
+         _
        ) do
     database_instances
-    |> Enum.reject(fn %{system_replication_site: site} -> site == primary_site end)
+    |> Enum.filter(fn %{system_replication_source_site: soure_site} -> site == soure_site end)
```
arbulu89, line 188: 3rd bullet point.
Check for source site instances instead of primary instance

lib/trento/operations/database_policy.ex
```diff
@@ -224,6 +243,13 @@ defmodule Trento.Operations.DatabasePolicy do
        ),
        do: application_instances_stopped(database, nil, nil)
 
+  # full database stop request, check if app instances are stopped
+  defp application_instances_stopped(database, params, _) when not is_map_key(params, :site),
```
arbulu89, line 247: This is the 2nd bullet point.
Check application instances when full database stop request (no specific site)

abravosuse (approved):
Tested deploying web image from PR in rolling environment. Four issues fixed:

- app instances not considered in full stop
- stop tier 2 secondary when tier 3 is not stopped
- start tier 3 secondary when tier 2 is not started
- check cluster is not considered starting/stopping 3 tier node

test/trento/operations/database_policy_test.exs
```diff
@@ -306,20 +354,40 @@ defmodule Trento.Operations.DatabasePolicyTest do
     end
 
     test "should forbid operation if the request is for the primary site and secondary sites are not stopped" do
+      site1 = "Site1"
+      site2 = "Site2"
+      site3 = "Site3"
+      site4 = "Site3"
```
balanza: **issue:** ‏typo

lib/trento/operations/database_policy.ex
```diff
@@ -162,19 +181,19 @@ defmodule Trento.Operations.DatabasePolicy do
...
            database_instances: database_instances
          },
-         %{site: primary_site},
-         primary_site
+         %{site: site},
+         _
        ) do
     database_instances
-    |> Enum.reject(fn %{system_replication_site: site} -> site == primary_site end)
+    |> Enum.filter(fn %{system_replication_source_site: soure_site} -> site == soure_site end)
```
balanza: **nitpick:** ‏typo `soure_site`

lib/trento/operations/database_policy.ex
```diff
@@ -129,25 +133,40 @@ defmodule Trento.Operations.DatabasePolicy do
...
+                                                  system_replication_site: inst_site
+                                                } ->
+        if site == inst_site do
+          tier
+        end
+      end)
+
     database_instances
-    |> Enum.filter(fn %{system_replication_site: site} -> site == primary_site end)
+    |> Enum.filter(fn %{system_replication_tier: tier} -> tier == requested_site_tier - 1 end)
```
balanza, line 161: **issue:** ‏On an unknown `site`, this function would return `:ok`. I imagine this is considered an impossible scenario given how the function is called. Precisely because it's an impossible condition, wouldn't it be better to return an error?

balanza (commented):
I found some issues that may not be important; the overall PR is ok.

lib/trento/operations/database_policy.ex
arbulu89: More than impossible, if the user requests an operation for a site that doesn't exist (it cannot happen if requested from the frontend) the operation is allowed as it won't have any consequence. Nothing will happen and we will get a known error from wanda that will be shown in the activity log.

And in overall, this "site doesn't exist" is a more generic thing in the database policies, not only attached to this particular function.
I wouldn't add a specific error message in here, as many other code places would still allow it.
If we want we can think of this as a new "feature".
What do you think?

arbulu89: PD: I have double checked and if we wanted to add this new "policy" the main `authorize_operation` would be the place to do it, as it would cover all the operations, and it is pretty simple to implement

balanza: To be honest, I didn't go any deeper than this. My focus was captured from an `authorize_operation` function that, in case of a (somehow) malformed data, allows the operation. I would go for a more pessimistic approach, but it's just an opinion.

I wouldn't enlarge the scope to another feature if not strictly needed.

arbulu89: With "feature" I meant that this thing should not only cover this particular scenario and line of code, that if we wanted to do it, it should cover all the operation and scenarios where the site is not recognized.

balanza (approved):

lib/trento/operations/database_policy.ex
arbulu89: @balanza Let me merge this PR first, and I will open a new one protecting the scenario where a not recognized site comes in a request.
How does that sound?

PR #4173, opened 2026-04-13
Updating mix dependencies for liveview to 1.1

chargio:
Updating mix dependencies. Liveview is not used in the library but it is a component in new modules, so updating instead of deleting it. Ecto and filesystem, as dependencies, get updated too in mix.lock

# Description

The dependencies file did not update the refernece to liveview. Due to the fact that liveview is not used in the current code, it is not properly updated and has not been updated yet.

The new work requires liveview, so I am updating it to the latest version instead of deleting it.

Fixes # (issue)

- [ x ] **DONE**

## How was this tested?

mix test
mix credo
mix dyalizer

As there is no current code using this, no need for more testing.
- [ x ] **DONE**

- [ x ] **DONE**


gagandeepb (approved):
Thanks for this. I have some doubts about some BE flaky tests being visible now due to the minor/patch Ecto update, but we can monitor and address those when the flaky tests job runs next, if needed.

PR #4171, opened 2026-04-13
Implementation of server-side AI assistant capabilities

gagandeepb:
# Description

Please include a summary of the changes and the related issue, if it exists. 
Also, include relevant motivation and context for this PR. List any dependencies that are required for this change.

Fixes # (issue)

## Did you add the right label?

Remember to add the right labels to this PR.
- [ ] **DONE**

## How was this tested?

Describe the tests that have been added/changed for this new behavior.
- [ ] **DONE**

## Did you update the documentation?

Remember to ask yourself if your PR requires changes to the following documentation:

- [User Documentation](https://github.com/trento-project/docs/tree/main/trento/adoc)
- [Developer Documentation](https://github.com/trento-project/docs/tree/main/content)
- [Trento Web Documentation](https://github.com/trento-project/web/tree/main/guides)

- [ ] **DONE**

PR #4186, opened 2026-04-14
Basic assistant UI

nelsonkopliku:
# Description

This PR aims to introduce building blocks for the AI Assistant.

The idea is to leverage open-source libraries to build AI assistants
https://www.assistant-ui.com/
https://github.com/assistant-ui/assistant-ui/tree/main/examples/with-ag-ui

and standards for the client<->agent communication protocol
https://docs.ag-ui.com/

Note that `AIAgentController` is a pure dummy implementation of an AG-UI compatible AI agent over SSE.
This PR will be followed by an implementation of the [AbstractAgent](https://docs.ag-ui.com/sdk/js/client/abstract-agent) to communicate over WebSockets and wire it up with a phoenix channel like in this PR https://github.com/trento-project/web/pull/4171

This PR aims to:
- follow up on this thread on related AI RFC https://github.com/trento-project/docs/pull/164#discussion_r3026724201
- build consensus within the team about the direction

Important note/question: how should we be testing this line of work? 😅 
I mean, I am a bit doubtful about jest tests testing the libraries to be rendedring stuff.
On the other hand I would definitely 
- test in isolation the aforementioned implementation of the `AbstractAgent` that will follow up, for instance
- and other non rendering related components
- add some e2e tests further down the road to cover the overall feature

What actually touches LLM non-determinism, ie the actual responses, is very hard to test in an automated manner.

Open for feedback. I want some degree of agreement around the rough edges that are necessarily going to be there for this work.

### DIFF
--- a/assets/js/common/AIAssistant/AssistantChatProvider.jsx
+++ b/assets/js/common/AIAssistant/AssistantChatProvider.jsx
@@ -1,0 +1,83 @@
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+
+import { HttpAgent } from '@ag-ui/client';
+import { AssistantRuntimeProvider, useAui } from '@assistant-ui/react';
+import { useAgUiRuntime } from '@assistant-ui/react-ag-ui';
+// import { CustomHttpAgent } from './CustomHttpAgent';
+
+export function AssistantChatProvider({ children }) {
+  // AG-UI agent URL pointing to your backend
+  const agentUrl = '/api/v1/ai_agent';
+
+  // // In-memory thread storage for multi-thread support
+  const threadsRef = useRef(new Map());
+  const [currentThreadId, setCurrentThreadId] = useState(() => {
+    const id = crypto.randomUUID();
+    threadsRef.current.set(id, { id, messages: [] });
+    return id;
+  });
+
+  // Create HttpAgent for AG-UI protocol with custom context injection
+  const agent = useMemo(() => {
+    return new HttpAgent({
+      url: agentUrl,
+      threadId: currentThreadId,
+      headers: {
+        Accept: 'text/event-stream',
+        Authorization: `Bearer ${localStorage.getItem('access_token') || ''}`,
+      },
+    });
+  }, [agentUrl, currentThreadId]);
+
+  // Setup thread list adapter for multi-thread support
+  const threadListAdapter = useMemo(
+    () => ({
+      threadId: currentThreadId,
+      onSwitchToNewThread: async () => {
+        const newId = crypto.randomUUID();
+        threadsRef.current.set(newId, { id: newId, messages: [] });
+        setCurrentThreadId(newId);
+      },
+      onSwitchToThread: async (threadId) => {
+        const thread = threadsRef.current.get(threadId);
+        if (!thread) {
+          throw new Error(`Thread ${threadId} not found`);
+        }
+        setCurrentThreadId(threadId);
+        return { messages: thread.messages };
+      },
+    }),
+    [currentThreadId]
+  );
+
+  // Create AG-UI runtime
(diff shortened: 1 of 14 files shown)


assets/js/common/AIAssistant/AssistantChatProvider.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+  // Persist messages to threadsRef when they change
+  useEffect(() => {
+    return runtime.thread.subscribe(() => {
+      threadsRef.current.set(currentThreadId, {
+        id: currentThreadId,
+        messages: runtime.thread.getState().messages,
+      });
+    });
+  }, [runtime, currentThreadId]);
```
nelsonkopliku, line 76: This whole actual functional behavior is going to be revisited. Please consider this as just a starting point.

gagandeepb (commented):
5000 lines of code. 
Can we move some of the generic code to a component library, that is importable via package.json ?

And can you please highlight via a comment the integration point via websockets, since as agreed previously this would be the transport modality/mechanism ?

nelsonkopliku:
> 5000 lines of code. Can we move some of the generic code to a component library, that is importable via package.json ?
> 
> And can you please highlight via a comment the integration point via websockets, since as agreed previously this would be the transport modality/mechanism ?
>

@gagandeepb, 4314 of them is just `package-lock.json`. I cannot do much better. Ignore it.

gagandeepb (commented):
My further concerns are that the protocol assumes some runtime concerns that are already implemented on the backend. This feels like a brand new direction altogether that was not discussed before. Can you please explain why these concerns belong to the UI/front-end/rendering layer ? 
https://docs.ag-ui.com/sdk/js/client/abstract-agent#parameters

gagandeepb (dismissed):
The chat box seems static sized or can it be grown on-demand/on-need ?
Personally, can we please go back to @jagabomb 's discussed/proposed UI proposal but with the communication protocol can be borrowed from wherever.

config/config.exs
```diff
@@ -309,6 +309,10 @@ config :trento, :ai,
     ]
   ]
 
+config :mime, :types, %{
+  "text/event-stream" => ["sse"]
+}
```
nelsonkopliku, line 314: Just a note: besides the fact that this is not needed in case of websocket, to make sse work behind nginx I had to add
```
chunked_transfer_encoding on;
proxy_buffering off;
```

nelsonkopliku:
> My further concerns are that the protocol assumes some runtime concerns that are already implemented on the backend. This feels like a brand new direction altogether that was not discussed before. Can you please explain why these concerns belong to the UI/front-end/rendering layer ? https://docs.ag-ui.com/sdk/js/client/abstract-agent#parameters

@gagandeepb are you referring to this?
```typescript
interface RunAgentParameters {
  runId?: string // Unique ID for this execution run
  tools?: Tool[] // Available tools for the agent
  context?: Context[] // Contextual information
  forwardedProps?: Record<string, any> // Additional properties to forward
}
```
I haven't implemented the websocket based client, yet. However I wouldn't be concerned about that.
I believe, but I will have to check, they would not overlap with backend, and we will use if and when necessary.

For instance I believe we would be able to send in the `context` property UI related context for the AI Agent to use.

gagandeepb (commented):
After a synch chat, I understand the `AbstractAgent` interface better. Many of its initialization parameters are optional, so I believe we should most likely be fine in making the protocol suitable for exchange.

nelsonkopliku:
superseded by 
- https://github.com/trento-project/web/pull/4246
- https://github.com/trento-project/web/pull/4251
- https://github.com/trento-project/web/pull/4245

and https://github.com/trento-project/web/pull/4254

PR #4174, opened 2026-04-13
[TRNT-4339] Add timezone conversion in the frontend

antgamdia:
# Description

This series of PRs aims to make the UI totally timezone-aware, depending on each user's preference. Not a browser preference, but a Trento user setting.

Previous PR: https://github.com/trento-project/web/pull/4169

Next PR: https://github.com/trento-project/web/pull/4187

Related # TRNT-4339

---- 

In this PR, it performs the actual conversion in the frontend, fetching the timezone in the Pages and passing the props to any downstream components. 

~**Done**: update the format of the time in each section; we have many similar formats. Besides, adding the UTC offset has been requested. See https://github.com/trento-project/web/pull/4174#discussion_r3085777403~

Current status:

https://github.com/user-attachments/assets/468d3ac3-af8d-44c6-a974-75768f17885c

https://github.com/user-attachments/assets/69eaa674-6345-43c8-af03-7dee622d830c

https://github.com/user-attachments/assets/c58df174-90b3-46f8-b393-b7b4ce2fb78e

https://github.com/user-attachments/assets/5e28ab3d-59a8-4723-bb35-a5345b2cd87f

https://github.com/user-attachments/assets/059196ed-dfe4-48fc-ae00-402d2510fa64

https://github.com/user-attachments/assets/528d4f31-6819-4cce-b5ba-debd485cc703

https://github.com/user-attachments/assets/9997fc2e-f581-4046-9f46-5574b9ae1e5b

## How was this tested?

IRL, unit and e2e tests

## Did you update the documentation?

N/A

## Additional information 

Note: tests are currently failing because the e2e tests need some adaptations.

PR #4187, opened 2026-04-14
[TRNT-4339] Update unit and e2e tests for timezone awareness

antgamdia:
# Description

Previous PR: https://github.com/trento-project/web/pull/4174

Next PR: https://github.com/trento-project/web/pull/4196

Related # TRNT-4339

---- 

In this PR, we simply update all the frontend unit tests as well as the e2e tests.

## How was this tested?

IRL, unit and e2e tests

N/A

## Additional information 

N/A


assets/js/common/ComposedFilter/ComposedFilter.test.jsx
```diff
@@ -310,4 +315,34 @@ describe('ComposedFilter component', () => {
...
+        prefilled: true,
+        timezone,
+      },
+    ];
+
+    render(
+      <ComposedFilter filters={filters} onChange={mockOnChange} autoApply />
+    );
+
+    await act(() => user.click(screen.getByText('Filter newer than...')));
```
arbulu89: nit:
I don't think you need `act` here. `user.click` already is wrapped somehow in an act
Actually, it creates a warning message which I'm trying to remove.
Can you try removing the `act` (from every place you put actually) and see if the tests pass?

assets/js/common/PatchList/PatchList.test.jsx
```diff
@@ -31,7 +35,11 @@ describe('PatchList', () => {
       expect(screen.getByText(patch.advisory_name)).toBeVisible();
       expect(screen.getByText(patch.advisory_synopsis)).toBeVisible();
       expect(
-        screen.getByText(formatDate(patches[0].update_date, 'd MMM y'))
+        screen.getByText(
+          formatDate(patch.update_date, DATE_DAY_MONTH_YEAR_COMPACT_FORMAT, {
```
arbulu89: super nit:
if we create a helper function to do this, we can use it here as well. To reduce imports.
Applicable to all new code

assets/js/pages/ActivityLogPage/ActivityLogPage.test.jsx
```diff
@@ -1,7 +1,9 @@
 import React, { act } from 'react';
-import { screen } from '@testing-library/react';
+import { fireEvent, screen } from '@testing-library/react';
```
arbulu89: change:
don't use `fireEvent`, it is deprecated (even though we have in our code yet...).
User `userEvent`, it is already imported in the test file.
It must be a way to do the `change` thing

assets/js/pages/HostDetailsPage/HostSummary.test.jsx
```diff
@@ -102,6 +98,27 @@ describe('HostSummary', () => {
     );
   });
 
+  it('should display last boot timestamp using a non-default timezone', () => {
```
arbulu89: super nit:
the other test names are kind of `timestap using the provided timezone`. Just to not have this `non-default` if you want

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.test.jsx
```diff
@@ -5,7 +5,7 @@ import '@testing-library/jest-dom';
 import { faker } from '@faker-js/faker';
 import { noop } from 'lodash';
 
-import { renderWithRouter as render } from '@lib/test-utils';
+import { renderWithRouter, withState } from '@lib/test-utils';
```
arbulu89, line 8: change:
I think i have commented in other PR, but let's not add an state here.
Let's pass the `timezone` as prop from `Page.jsx` and use it here as an easy prop. 
It will look much better.
We don't really need to test the timezone get in the selector, it is kind of trivial

assets/js/pages/Users/Users.test.jsx
```diff
@@ -1,14 +1,15 @@
...
 import { adminUser, userFactory } from '@lib/test-utils/factories/users';
 import { renderWithRouter } from '@lib/test-utils';
 import { userEvent } from '@testing-library/user-event';
 import Users from './Users';
+import { faker } from '@faker-js/faker';
 
 describe('Users', () => {
   it('should render a loading table with a disabled create user button', () => {
-    render(<Users loading />);
+    renderWithRouter(<Users loading />);
```
arbulu89: question:
why this change from `render` to `renderWithRouter`?
If we are not going to navigate, let's use simple `render`, and use the other one if the test needs to navigate somehow.
I think this applies to all the tests you added, at the end, you are not clicking in specific users to go their details view

arbulu89 (commented):
Hey @antgamdia ,
Some few comments.
But it looks good overall

Remember to remove the `act` functions you added in the frontend UTs

PD: it looks like Kiritimati is your next vacation destination XD

assets/js/pages/ActivityLogPage/ActivityLogPage.test.jsx
antgamdia: Good point! It took me longer than expected, and ended up re-implementing some logic in `DateFilter.jsx`. There were some race conditions when using `parseIso`.

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.test.jsx
antgamdia: I think I fixed it in the next PR, but will double-check once merged onto this one.

assets/js/pages/Users/Users.test.jsx
antgamdia: Right! Long story short: I search&replace-ed every occurrence 😓 
Some test cases were failing using plain `render` (`TypeError: Cannot destructure property 'basename' of 'React.useContext(...)' as it is null.`), so I needed to use the `renderWithRouter` (or build another helper, as we don't actually need the routing part).

This is a consequence of using `useSelector` in non-page components... but this is done in the next PR. 
I'll merge and replace those `renderWithRouter` with plain `router`.

assets/js/common/PatchList/PatchList.test.jsx
antgamdia: Done!

assets/js/pages/HostRelevantPatches/HostRelevantPatchesPage.test.jsx
antgamdia: Nope... I missed that, thanks for the heads-up. Fixed now, thank you!

PR #4188, opened 2026-04-15
Detect majority maker nodes

balanza:
~Exclude majority-maker (quorum-only) nodes from cluster check execution. These nodes do not run SAP workloads.~
Majority maker nodes are identified during cluster discovery and marked with a flag in the cluster detail. 

To identify a majority maker node, the resources from `CIB` are considered. 

- The node must have a location constraint (RscLocation) for every SAP resource of kind SAPHana, SAPHanaTopology, SAPHanaController
- Each of those constraints must have score == "-INFINITY" (i.e. the node is explicitly banned from running that resource).
- If the node is banned from only some of the SAP resources, it is not a majority maker


lib/trento_web/openapi/v1/schema/cluster.ex
```diff
@@ -59,6 +59,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
             additionalProperties: %Schema{type: :string}
           },
           virtual_ip: %Schema{type: :string},
+          is_majority_maker: %Schema{type: :boolean},
```
nelsonkopliku, line 62: **thought:** probably it does not harm, but since v1 is deprecated, should we care at all about backporting the field?

test/trento/clusters_test.exs
```diff
@@ -207,6 +207,89 @@ defmodule Trento.ClustersTest do
...
     end
 
+    test "should exclude majority maker hosts from checks execution in a hana scale-up cluster" do
+      mm_hostname = Faker.Internet.domain_word()
+
+      details =
+        build(:hana_cluster_details,
+          nodes: [
+            build(:hana_cluster_node, name: mm_hostname, is_majority_maker: true),
+            build(:hana_cluster_node, is_majority_maker: false)
```
nelsonkopliku: **nit/question:** shouldn't we have 3 nodes here to reflect the hosts inserted afterwards?

test/trento/discovery/policies/cluster_policy_test.exs
```diff
@@ -5820,7 +5821,8 @@ defmodule Trento.Discovery.Policies.ClusterPolicyTest do
                         resources: [],
                         site: nil,
                         status: "Online",
-                        virtual_ip: nil
+                        virtual_ip: nil,
+                        is_majority_maker: true
```
nelsonkopliku, line 5825: **nit:** Let's maybe add at least one explicit `is_majority_maker: false`

nelsonkopliku (approved):
Looks good to me, Emanuele. Just a few minor nits, maybe.

Let's have Alberto test the pr-env image.

arbulu89 (commented):
Hey @balanza 
Code aside I have a more "broad" thought.

When we found this issue, I personally didn't thought about the fix as you did.
After all, we are going to exclude the majority maker from all checks, and ideally, we just wanted to exclude the execution for certain checks only,
There are scenarios and checks where we actually want to check what's going on in the majority maker as any other cluster node (for example, to check installed pacemaker version is the same).

In my head, we would have something similar as the `predicate` we use for operations, where target nodes are excluded based on certain specific node values.

Certainly, the code to identify if the node is a majority maker is valuable and it is needed to do any kind of logic though.

balanza:
> Hey @balanza Code aside I have a more "broad" thought.
> 
> When we found this issue, I personally didn't thought about the fix as you did. After all, we are going to exclude the majority maker from all checks, and ideally, we just wanted to exclude the execution for certain checks only, There are scenarios and checks where we actually want to check what's going on in the majority maker as any other cluster node (for example, to check installed pacemaker version is the same).
> 
> In my head, we would have something similar as the `predicate` we use for operations, where target nodes are excluded based on certain specific node values.
> 
> Certainly, the code to identify if the node is a majority maker is valuable and it is needed to do any kind of logic though.


The filter logic is "too simplistic" (in english: wrong :) ), indeed. Let me skim the filter part out, and we can go ahead with the majority maker detection. Later on, I'll come up with a PR with the filter logic.

test/trento/discovery/policies/cluster_policy_test.exs
balanza: added below

lib/trento/clusters/value_objects/hana_cluster_node.ex
```diff
@@ -23,6 +23,7 @@ defmodule Trento.Clusters.ValueObjects.HanaClusterNode do
     field :nameserver_actual_role, :string
     field :indexserver_actual_role, :string
     field :status, :string
+    field :is_majority_maker, :boolean, default: false
```
arbulu89, line 26: comment:
we need to add this field in `ascs/ers` nodes as well.
they should follow the same logic.
so most probably you need to change the `parse_ascs_ers_cluster_nodes` in the `cluster_policy` code

lib/trento_web/openapi/v1/schema/cluster.ex
arbulu89: +1.
we shouldn't backport the item.
we should maybe remove in the json serialization, we do it with many other elements.
it is not a big deal though

lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -1228,4 +1232,42 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
...
+  end
+
+  defp majority_maker_nodes(_), do: MapSet.new()
+
+  defp banned_from_all_sap_resources?({_node, constraints}, sap_resource_ids) do
+    constrained_resources = MapSet.new(constraints, fn %{resource: r} -> r end)
+    MapSet.subset?(sap_resource_ids, constrained_resources)
+  end
+
+  defp sap_clone_master_ids(%{clones: clones, masters: masters}) do
```
arbulu89, line 1263: nit:
can we put this on top of `banned_from_all_sap_resources`?
it is used before

lib/trento/discovery/policies/cluster_policy.ex
```diff
@@ -1228,4 +1232,42 @@ defmodule Trento.Discovery.Policies.ClusterPolicy do
       if id == resource_id, do: instance_attributes
     end)
   end
+
+  @sap_resource_types ~w(SAPHana SAPHanaTopology SAPHanaController)
+
+  defp majority_maker_nodes(%{constraints: %{rsc_locations: rsc_locations}, resources: resources})
```
arbulu89, line 1238: praise:
really nice!
It looks like pretty AI thing hehe, i wouldn't never think of using `MapSet`

arbulu89 (commented):
Good!
~We just need to apply the same logic to ascs/ers nodes and we are done~

lib/trento/clusters/value_objects/hana_cluster_node.ex
arbulu89: @balanza forget this.
It is not really needed. We have not seen any majority maker usage in ascs/ers clusters so far.
I don't know why I thought this...

arbulu89 (approved):
Ok!
Ready to go from my side.
Feel free to check the nitpicks I commented.

PD: try to test this in a real infra before merging. We are pretty close to a release, so we should carefully test new features. At least to see that nothing old is broken

balanza:
> PD: try to test this in a real infra before merging. We are pretty close to a release, so we should carefully test new features. At least to see that nothing old is broken

I will try to include an event dump in the text fixtures, and write some comprehensive tests on it.

So far, we already have this example in the fixtures https://github.com/trento-project/web/blob/2d98cba4e5a41b2931f98802532a2789a1685015/test/fixtures/discovery/ha_cluster_discovery_hana_scale_out_no_srhooks.json#L1

where the node `hana-s-mm` is tested to be a majority maker https://github.com/trento-project/web/blob/2d98cba4e5a41b2931f98802532a2789a1685015/test/trento/discovery/policies/cluster_policy_test.exs#L5825

PR #4192, opened 2026-04-15
Unify select component

arbulu89:
# Description

Unify `Select` component usage for all scenario.
This way we can remove the existing `MultiSelect` as it is somehow duplicated logic.

Now we will use `react-select` as baseline, which allows many features like, `searchable`, `multi`, `clearable` etc
The code is almost entirely copied from `MultiSelect` even though it looks like we have a ton of changes.
I just needed to change how we use the `renderOption` and some additional classname things.
I have reused most of the previous logic and the end result is 1:1.

PD: the unique thing I couldn't match to the already existing select is that when you open the dropdown the focused value being the selected one (or not at all, if there was not any selected). It will always focus the first element no matter what

## How was this tested?

UT, manual tests and previouslyu existing tests.

No need


assets/js/pages/ActivityLogPage/ActivityLogPage.jsx
```diff
@@ -160,17 +160,22 @@ function RefreshIntervalSelection({ disabled = false, rate, onChange = noop }) {
...
       optionsName="refresh-rate"
       options={refreshRateOptions}
-      value={refreshRate}
+      values={[refreshRate]}
       renderOption={createOptionRenderer(null, (value) => (
         <span className="text-center block">
           {refreshRateOptionsToLabel[value]}
         </span>
       ))}
+      renderControlOption={createOptionRenderer(null, (value) => (
```
arbulu89, line 170: the `renderControlOption` replaces the usage of `selectedItemPrefix`.
This previous item was used to only add a prefix for the top control item but not to the dropdown options.

assets/js/common/Select/Select.jsx
```diff
@@ -1,124 +1,171 @@
...
+  multiValue: () =>
+    'rounded-md bg-green-100 text-green-800 px-1 py-2px space-x-1 mr-1',
+  multiValueLabel: () => 'ml-1',
+  control: ({ isDisabled }) =>
+    classNames(
+      {
+        'text-gray-400 bg-gray-50': isDisabled,
+        'bg-white': !isDisabled,
+      },
+      'relative w-full h-full py-2 pl-3 pr-2 text-left rounded cursor-default border border-gray-300 sm:text-sm'
```
antgamdia, line 81: comment: from memory, I had some styling issues with `cursor-default`. Dropping this comment as a personal reminder to try it out when checking out your changes.

assets/js/common/Select/Select.jsx
```diff
@@ -1,124 +1,171 @@
...
+    ),
+};
+
+// composeOpts compose old style options and default values
+// examples:
+// ['Enabled', 'Disabled'] => [
+// {value: 'Enabled', label: 'Enabled'},
+// {value: 'Disabled', label: 'Disabled'}
+// ]
+const composeOpt = (opt) =>
```
antgamdia, line 101: nit: comment says `composeOpts` but actual function is `composeOpt`  :P

arbulu89: yeah...
I only had `composeOpts` initially and create `composeOpt` afterwards

assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { assign, find } from 'lodash';
 import { getFromConfig } from '@lib/config';
 
-import MultiSelect from '@common/MultiSelect';
+import Select from '@common/Select';
```
EMaksy, line 6: Idea: what about adding a default guard value here?

```
const defaultAbilities = [];
``` 

We could use it in line 108,109

```
function AbilitiesMultiSelect({
  abilities = defaultAbilities,
  userAbilities = defaultAbilities,
```

In line 78 mapAbilities, we use reduce(). 
If abilities are undefined it could crash no ? 

This is  just an edge case, but [] fallback would not harm in isolation.

EMaksy: Or well just directly
```
function AbilitiesMultiSelect({
  abilities = [],
  userAbilities = [],
  ...
``

EMaksy (approved):
Hey @arbulu89 thank you, this lgtm, just left one little idea, besides this lgtm :)

assets/js/common/Select/Select.jsx
arbulu89: changed

assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.jsx
arbulu89: I actually prefer not to put default values there, as it makes the variables like optional.
And the user always must give abilities to this component.
We wouldn't really fix anything and we could hide a bad usage of the component.

EMaksy: Alright no hard opinion here, just personal preference :)

PR #4196, opened 2026-04-16
[TRNT-4339] Unify time/date formats in UI

antgamdia:
# Description

Previous PR: https://github.com/trento-project/web/pull/4187

Next PR: https://github.com/trento-project/web/pull/4197

Related # TRNT-4339

---- 

In this PR, we perform two main changes: 

- 1) remove similar formats to display dates and datetimes;
- ~2) append the timezone to the chosen formats.~ This has been reverted as per PR's feedback.

Besides, two improvements have been added: 
- 1) the date filter now works if the browser's tz does not match the profile's tz; 
- 2) the dropdown tz selector is now using `useMemo` for performance purposes.

Current status:

image

image

image

## How was this tested?

IRL, unit and e2e tests

N/A

## Additional information 

Note: tests are currently failing because the e2e tests need some adaptations.


abravosuse:
@antgamdia as discussed in our call, I think that indicating the timezone is such way might be confusing for the user. If a user reads "12:16:19 +05:30", they might wonder what actual time it's referring to. What is more commonly used to display timestamps is the time zone abbreviation. For example, ET, for the East of the USA without daylight savings. Or CEST, for Central Europe with daylight savings. But since we don't have that mapping (timezone <> abbreviation), I vouch for simply displaying the timestamp, no reference to the timezone whatsoever.

assets/js/common/ComposedFilter/ComposedFilter.test.jsx
```diff
@@ -2,12 +2,12 @@ import React, { useState } from 'react';
...
 import userEvent from '@testing-library/user-event';
 import '@testing-library/jest-dom';
-import { TZDate } from '@date-fns/tz';
 import { parseISO } from 'date-fns';
 import ComposedFilter from '.';
 
-const parseDateTimeLocalToUtc = (dateTimeLocalValue, timezone) =>
-  new Date(TZDate.tz(timezone, parseISO(dateTimeLocalValue)).getTime());
+it('test environment time should always be UTC', () => {
+  expect(new Date().getTimezoneOffset()).toBe(0);
```
arbulu89: question:
what is this test you added everywhere for?
I personally would remove it, it kind of checks for a system level thing for debugging purposes or similar

arbulu89 (commented):
Hey @antgamdia ,
I personally like the change to have less different formats.
It looks cleaner.
Codewise, just some minor comments.
I think we can comment this at some point with @abravosuse

assets/js/common/ComposedFilter/ComposedFilter.test.jsx
antgamdia: I wanted to ensure a consistent timezone configuration during test suite execution. That's why I added the `TZ=utc` in the `npm test` script. Whereas the test can be removed, it ensures that the test system is configured as expected. 

Maybe, we could try to mock it somehow, don't know how easy it would be. But, if we need the test system to be UTC, IMO, we need to keep the test to prevent false failures due to TZ settings.

antgamdia: Edit: removed, as we get rid of the parseISO in test files :)

PR #4197, opened 2026-04-16
[TRNT-4339] Add warning if selected tz does not match browser's

antgamdia:
# Description

Previous PR: https://github.com/trento-project/web/pull/4196

Next PR: https://github.com/trento-project/web/pull/4200

Related # TRNT-4339

---- 

This PR simply adds a warning in the user's profile (not in the edit user page) in case the selected timezone does not match the one currently configured in the browser.

Current status:

image

## How was this tested?

IRL tests.

N/A

## Additional information 

Tests pending, this is just to share the idea and gather early feedback


abravosuse:
@antgamdia It doesn't hurt to warn the user. I suggest a different wording though:
"Warning: Your browser UTC offset is +02:00, but your profile timezone offset is +10:00. The Trento UI will always use your profile timezone to display time stamps, not your browser's."

assets/js/pages/Profile/ProfileForm.jsx
```diff
@@ -117,6 +118,34 @@ function ProfileForm({
   const selectedTimezone =
     timezones.find((opt) => opt.value === timezoneState) || null;
 
+  // Utility for formatting offsets
+  const formatOffset = (mins) =>
```
arbulu89: nit:
maybe we can move this function out of the component, so it doesn't need to re-created in each render.
At the end, it is a totally functional code

assets/js/pages/Profile/ProfileForm.test.jsx
```diff
@@ -476,6 +477,63 @@ describe('ProfileForm', () => {
...
+  it('should display a timezone warning when browser and profile offsets differ', async () => {
+    const { username, fullname, email, abilities } = profileFactory.build();
+    const browserOffsetSpy = jest
+      .spyOn(Date.prototype, 'getTimezoneOffset')
+      .mockReturnValue(-120);
+    const timezone = 'Pacific/Port_Moresby';
+    const expectedWarning =
+      "Warning: Your browser UTC offset is +02:00, but your profile timezone offset is +10:00. The Trento UI will always use your profile timezone to display timestamps, not your browser's.";
+
+    try {
```
arbulu89: nit:
maybe a clearing in a `afterTest` looks cleaner, i think we do this always. Anyway, this is totally fine
```
afterEach(() => {
  jest.restoreAllMocks(); 
});
```

assets/js/pages/Profile/ProfileForm.jsx
```diff
@@ -117,6 +118,34 @@ function ProfileForm({
...
+    `${mins < 0 ? '-' : '+'}${String(Math.trunc(Math.abs(mins) / 60)).padStart(2, '0')}:${String(Math.abs(mins) % 60).padStart(2, '0')}`;
+
+  // State for offsets and warning
+  const [browserOffsetMin, setBrowserOffsetMin] = useState(null);
```
arbulu89: question:
i wonder if we need these `useState` and `useEffect`. At the end, we are doing just simple data calculations.
wouldn't the next be enough:
```
const now = new Date();
const browserOffsetMin = -now.getTimezoneOffset();
  
const profileOffsetMin = selectedTimezone 
    ? tzOffset(selectedTimezone.value, now) 
    : null;

const showTimezoneWarning = 
    selectedTimezone && 
    profileOffsetMin !== null && 
    browserOffsetMin !== profileOffsetMin;
```
we could even extract the logic to a function with only `selectedTimezone` as param

arbulu89 (commented):
Hey @antgamdia ,
I dropped a message on the usage of `useState` and `useEffect`, maybe they are not really needed.
Let me know if the simplified version works fine

assets/js/pages/Profile/ProfileForm.jsx
antgamdia: Thx, yep, I have moved it to the tz library instead.

assets/js/pages/Profile/ProfileForm.test.jsx
antgamdia: Thanks! Yes... makes way more sense to do it that way.

assets/js/pages/Profile/ProfileForm.jsx
antgamdia: True, it was an overkill. I've removed the useXXX and extracted some part of the helper. Thanks!

PR #4210, opened 2026-04-20
Adapt e2e tests for remote instance

vicenteqa:
# Description
- Make Cypress E2E tests work on both local and remote Trento instances.
- Replace hardcoded localhost assumptions with dynamic URL/API-key handling.
- Update Photofinish and agent heartbeat flows to support real remote environments.
- Adds mocks at cypress level for SUMA tests and skips tests that rely on wanda running with demo profile.
- Increased to way more generous timeouts to make tests work against a remote instance.
- Added logic to ignore status code 500 responses on prometheus related endpoints in SLES16
- As the installation in remote instances includes the agent installation, added logic to handle the cleanup of that agent and asserting one more subscription.
- Add retry logic for when restoring host after clean-up, this logic only affects when targeting a remote trento instance and will show a message in the logs.

Fixes https://jira.suse.com/browse/TRNT-4272

- [ ] **DONE**

## How was this tested?

- [ ] **DONE**

- [ ] **DONE**


test/e2e/cypress/e2e/checks_customization.cy.js
```diff
@@ -118,6 +118,13 @@ context('Checks customization', () => {
   });
 
   describe('Execution with customized check values', () => {
+    before(function () {
+      const isWandaRealInstance = !Cypress.env('wandaUrl').includes(
+        'http://localhost:4001'
+      );
+      if (isWandaRealInstance) this.skip();
```
nelsonkopliku: **thought:** relying on the url to determine whether wanda is in demo mode or not looks a bit weak.

At the end wanda running on `http://localhost:4001` can be either a demo or prod artifact, and the same is true for a wanda instance **not** running on that specific host/port combination.

What do you think about using a more explicit env variable like `WANDA_MODE=<something>` and use it in cypress?

test/e2e/cypress/e2e/host_details.cy.js
```diff
@@ -155,7 +155,14 @@ context('Host Details', () => {
   });
 
   describe("Node exporter status should be 'running'", () => {
-    beforeEach(() => hostDetailsPage.visitSelectedHost());
+    beforeEach(() => {
+      const isTrentoRealInstance =
+        !Cypress.config().baseUrl.includes('localhost');
```
nelsonkopliku: **thought:** same consideration about wanda mode.

test/e2e/cypress/e2e/suse_manager_overviews.cy.js
```diff
@@ -4,12 +4,18 @@ context('SUSE Manager overviews', () => {
   before(() => hostDetailsPage.preloadTestData());
 
   beforeEach(() => {
-    hostDetailsPage.clearSUMASettings();
-    hostDetailsPage.saveSUMASettingsForAdmin();
+    const isDevEnvironment =
+      Cypress.config().baseUrl.includes('localhost:4000');
```
nelsonkopliku, line 8: **thought:** same as above

test/e2e/cypress/pageObject/activity_log_po.js
```diff
@@ -56,8 +56,8 @@ export const spyActivityLogRequest = () => {
   );
 };
 
-export const waitForActivityLogRequest = () =>
-  basePage.waitForRequest(activityLogEndpointAlias);
+export const waitForActivityLogRequest = (timeout = 10000) =>
```
nelsonkopliku, line 59: **question:** can this type of timeout be set once for the whole suite in the configuration?

test/e2e/cypress/pageObject/base_po.js
```diff
@@ -242,27 +242,53 @@ export const preloadTestData = ({
...
+  const runPhotofinish = (apiKey) => {
+    photofinishCommand = apiKey
+      ? `${photofinishCommand} "${apiKey}"`
+      : photofinishCommand;
+    cy.log(`Shooting scenario "${scenario}" to: ${baseUrl}`);
+    return cy.exec(photofinishCommand, { timeout: 360000 });
+  };
+
+  if (baseUrl.includes('localhost')) return runPhotofinish();
+  else return getApiKey().then((apiKey) => runPhotofinish(apiKey));
```
nelsonkopliku, line 283: **thought:** what do you think about always running with an API key?

It would possibly simplify the code and would work in any set up: real or not.

test/e2e/cypress/pageObject/hana_cluster_details_po.js
```diff
@@ -103,9 +103,12 @@ export const visit = (clusterId = '') => basePage.visit(`${url}/${clusterId}`);
 
 export const waitForInitialEndpoints = () => {
   basePage.waitForRequest(lastExecutionEndpointAlias);
-  basePage.waitForRequest(catalogEndpointAlias);
+  waitForCatalogEndpoint();
 };
 
+export const waitForCatalogEndpoint = () =>
+  basePage.waitForRequest(catalogEndpointAlias);
```
nelsonkopliku: **nit:** are we going to use this function somewhere else other than in the wrapper above? 👀

nelsonkopliku (commented):
Hey Vicente, thanks for this! 
I believe the ability to run our e2e test suite agains a non totally mocked trento instance will improve the overall test topic.

My main concern is about relying on urls to determine whether we're running on a mocked/dev/demo/prod situation, we might consider alternative approaches.

A couple of more questions:
- when wanda is not in demo mode, we're skipping checks related tests because of course we don't have agents in the picture providing facts gathering, correct?
- SMLM related tests are always intercepted and mocked, right?

test/e2e/cypress/pageObject/base_po.js
vicenteqa: I added the logic because when running against a dev env it isn't required, but I guess it would, of course, work even if retrieved.

test/e2e/cypress/pageObject/hana_cluster_details_po.js
vicenteqa: I'll remove it, I did it because when I started to implement this whole task I was using it somewhere else, then left it there :neutral_face:

vicenteqa:
>when wanda is not in demo mode, we're skipping checks related tests because of course we don't have agents in the picture providing facts gathering, correct?

**Yes**

>SMLM related tests are always intercepted and mocked, right?

Yes, I saw that when running trento in dev, we mock that at `elixir` level to run the tests, I was exploring options of having that mock data somehow shared in a single file but in the end I found it a bit complex to implement.

test/e2e/cypress/pageObject/activity_log_po.js
vicenteqa: do you mean, setting it here
https://github.com/trento-project/web/blob/b0f294882ba2ff84e9487062a9fccb57e9ffbc26/test/e2e/cypress/pageObject/base_po.js#L234
for example?

test/e2e/cypress/e2e/checks_customization.cy.js
```diff
@@ -118,6 +118,10 @@ context('Checks customization', () => {
   });
 
   describe('Execution with customized check values', () => {
+    before(function () {
```
arbulu89, line 121: Just curiosity.
Did you consider using @cypress/grep ?
It is a pretty common way to filter tests execution by filters.
Anyway, I see you use this `env` modes heavily, so I guess you don't only want to filter

in any case, if you still continue with this approach, i would use a constant to save demo and other strings.

test/e2e/cypress/e2e/hana_cluster_details.cy.js
```diff
@@ -323,10 +327,11 @@ context('HANA cluster details', () => {
   });
 
   describe('Cluster with unknown provider', () => {
-    before(() => {
-      hanaClusterDetailsPage.loadScenario('cluster-unknown-provider');
-      hanaClusterDetailsPage.visitAvailableHanaCluster();
-    });
+    before(() =>
```
arbulu89, line 330: question:
doesn't this change look a bit suspicious if you don't change the rest of the providers tests that do pretty much the same?

The same with the next `Cypress.session.clearAllSavedSessions()` addition.
Should be maybe have some other `before` somewhere?
Or groups tests differently.
It looks a bit out of place

test/e2e/cypress/e2e/host_details.cy.js
```diff
@@ -155,7 +155,13 @@ context('Host Details', () => {
   });
 
   describe("Node exporter status should be 'running'", () => {
-    beforeEach(() => hostDetailsPage.visitSelectedHost());
+    beforeEach(() => {
+      const isTrentoRealInstance = Cypress.env('web_mode') === 'real';
```
arbulu89, line 159: nit:
maybe we can hide this logic somewhere else, like in a command or support function

test/e2e/cypress/pageObject/hana_cluster_details_po.js
```diff
@@ -491,26 +495,33 @@ export const notAuthorizedTooltipIsNotDisplayed = () => {
 };
 
 // API
+
 export const interceptGetChecks = () =>
-  cy.intercept('GET', '/api/v1/groups/*/checks?*').as(getChecksEndpointAlias);
+  cy
+    .intercept('GET', `${Cypress.env('wandaUrl')}/api/v1/groups/*/checks?*`)
```
arbulu89, line 501: nit/unsure:
maybe you can store the `Cypress.env('wandaUrl')` above as a constant?
To avoid the repetition

test/e2e/cypress.config.js
```diff
@@ -1,28 +1,36 @@
...
     idp_url: 'http://localhost:8081',
+    wanda_mode: 'demo', //demo: local dev instance / real: real instance installed via rpm
+    web_mode: 'dev', //dev: local dev instance / real: real instance installed via rpm
   },
   e2e: {
     // We've imported your old cypress plugins here.
     // You may want to clean this up later by importing these.
-    setupNodeEvents(on, config) {
+    async setupNodeEvents(on, config) {
+      config.env.wandaUrl = calculateWandaUrl(config);
```
arbulu89, line 30: comment:
As i guess that you will pass the `web_mode` and `wanda_mode` as env, why don't you use that value to choose which url to use for wanda?

```
const calculateWandaUrl = (config) => {
  if (config.env.wandaUrl) return config.env.wandaUrl;
  return config.env.wanda_mode === demo ?
    ? 'http://localhost:4001'
    : `${config.baseUrl}/wanda`;
};
```

arbulu89 (commented):
Hey @vicenteqa ,
I have done a superficial review.
Additionally, why are you putting the `return` in almost every function?

test/e2e/cypress/pageObject/activity_log_po.js
nelsonkopliku: That, or somewhere in `test/e2e/cypress.config.js` if that's an option. Like `defaultCommandTimeout`

test/e2e/cypress/e2e/checks_customization.cy.js
vicenteqa: tbh, I just stuck to this way of skipping tests because we were already using it in other tests in the suite, but I agree that tagging them would be cleaner than this. Perhaps in a different task though?

vicenteqa:
> Hey @vicenteqa , I have done a superficial review. Additionally, why are you putting the `return` in almost every function?

Thanks for taking a look @arbulu89
there are things that I will need to carefully check because I implemented them some time ago, like the `suspicious` thing you mentioned :sweat_smile: 

About the returns thingy, tbh this is something that I am still addressing how it should be done properly. The recommendation says that functions that wrap cypress commands should have a return statement for cypress to properly chain them (now or in the future), BUT, I noticed that I added returns to `forEach` statements, which are completely useless as they always return undefined. I am going to fix that by using `cy.wrap().each()` and I will add return to every wrapper function that doesn't have it. Adding these returns will also help to prevent race conditions. It forces cypress to wait for all internal logic (like API calls or loops) to fully complete before moving to the next step of the test. Without the return, the test might continue executing while our setup or assertions are still running in the background, leading to flakiness.

test/e2e/cypress/e2e/checks_customization.cy.js
arbulu89: sure!

test/e2e/cypress/e2e/hana_cluster_details.cy.js
vicenteqa: Now I recall, the rest of providers-suites don't have this change because they only have one test, so it doesn't really affect. The change in this `describe` group is to have a clean visit for every test in the suite. I can modify the rest of providers-suites to match this pattern or just have it in mind in case we add more tests to the suites in the future.

Regarding `Cypress.session.clearAllSavedSessions()` this was added because the tests take so much time when executing against a remote instance that sometimes the token was expiring at that point, so I added those 2 lines to renew the session. I am thinking that maybe I can create a `refreshLoginToken` method or similar that groups both lines, adding a logic to only execute when `web_mode!=='dev'`

vicenteqa:
@arbulu89 @nelsonkopliku hey guys, I addressed the feedback that you left, let me know if you think that anything else would be better in some other way, additionally as Xabi suggested to me yesterday (very good idea btw), I added a couple of lint rules to make sure we always return something and reduce boilerplate.

test/e2e/cypress/e2e/suse_manager_overviews.cy.js
nelsonkopliku: **question:** what are we going to do about this?

test/e2e/cypress/pageObject/base_po.js
nelsonkopliku: if you strongly prefer keeping it like this, I won't insist. Otherwise it could be
```js
const runPhotofinish = (apiKey) => {
  cy.log(`Shooting scenario "${scenario}" to: ${baseUrl}`);
  return cy.exec(`${photofinishCommand} "${apiKey}"`, { timeout: 360000 });
};

return getApiKey().then((apiKey) => runPhotofinish(apiKey));
```

Up to you.

test/e2e/cypress.config.js
nelsonkopliku: **question:** in which scenarios `config.env.wandaUrl` is not passed? 🤔

nelsonkopliku (commented):
Hey Vicente, looking great! I just left a couple of more questions.

test/e2e/cypress/e2e/checks_customization.cy.js
arbulu89: Can we put this `demo` (`prod`, whatever other choices we have) in constants?
Just to make it easier to find the options

test/e2e/cypress/e2e/hana_cluster_details.cy.js
arbulu89: I still think the `refreshLoginPage` looks really out of place.
We have a pretty big refresh token default value, we shouldn't logout in such a small time.
Maybe we need to have a look into the login and the `cy.session` thing, and understand what happens

test/e2e/cypress/e2e/host_details.cy.js
```diff
@@ -155,7 +155,10 @@ context('Host Details', () => {
   });
 
   describe("Node exporter status should be 'running'", () => {
-    beforeEach(() => hostDetailsPage.visitSelectedHost());
+    beforeEach(() => {
+      hostDetailsPage.interceptNodeExporterStatusMockedForRealInstance();
```
arbulu89: nit:
I would reword this somehow. `Real` terminology is totally missleading. All the instances are real. Maybe use `Prod` instead. 
At the end, I think you are splitting things between `demo` and `prod`

arbulu89 (commented):
Hey @vicenteqa ,
Good improvements but I still have some comments.
- The usage of `Real` wording looks pretty inappropriate. I would try to find something else
- The `refreshLogin` things look something we need to re-visit as it highlight something is wrong with our testing

The PR is really big and difficult to review.
@vicenteqa Can I suggest one thing? Can we open just a PR with the new linting thing fixing the returns? And later on rebase here.
This would help a lot, as we wouldn't overlook so many changes.

PD: nit: the commit history will look pretty massive XD

test/e2e/cypress/pageObject/base_po.js
vicenteqa: honestly, I think it's better to avoid unnecessary steps if possible when targeting `dev` env.

test/e2e/cypress.config.js
vicenteqa: actually it's not passed anywhere, but I set this logic to preserve the value if we decide to explicitly set it.

vicenteqa:
- well that's it, there are many corrections that came after fixing conflicts with the previous merge about the linting rules for cypress. 
- I also added replaced `forEach` in favor of `cy.wrap().each()`
- I also removed functions that were waiting for host restoration and refreshing the login, as this might be "cheating" a bit and we need to investigate what is going on when there are failures.

nelsonkopliku (approved):
Hey @vicenteqa, I think we might want to move forward with this.

My remarks/takeaways remain the following:
- we might run the suite against a different combination of artifact mode (dev, demo, prod)
- the artifact mode, IMHO, shouldn't necessarily imply whether they're running on a given host/port and vice-versa (ie `localhost:4000` does not imply a given mode and the other way around, although sometimes it might)
- certain tests need to be skipped (or mocked by intercepting requests with cypress) based on the mode (for instance checks execution might be skipped or mocked when running in a mode that does not have actual agents running)

As long as these make sense from a general perspective, I am fine with how we are addressing them right now, and we can improve as per the needs.

PR #4211, opened 2026-04-20
Validate operation request site parameter in database operation policy

arbulu89:
# Description

Follow up for: https://github.com/trento-project/web/pull/4172

Validate if the `site` coming in the operation request exists.
Otherwise, fail with a forbidden message.
This scenario won't happen when the operation is requested from the frontend.

## How was this tested?

UT


balanza (approved):
Thanks!

PR #4212, opened 2026-04-21
Group operation abilities by resource

arbulu89:
# Description

Change how operation permissions are grouped in the frontend. `operation:all` is not generic, and doesn't allow tailored permissions.
Having permissions per resource (`cluster`, `database`, `host` and `sap_system`) is enough, as these are the main "admin" groups we have seen in the field.

The grouping is done simply in the frontend.
Besides, I have included the `operation:all` ability which authorizes any operation to the user. 
Without this, we cannot keep the `operation:all` in the frontend (which still comes handy) because it would meant grouping abilities in multiple groups.

operation_grouping

## How was this tested?

UT

@abravosuse do you have some reference about this? we should update in that case

### DIFF
--- a/priv/repo/migrations/20260421073949_add_operation_all_abilities.exs
+++ b/priv/repo/migrations/20260421073949_add_operation_all_abilities.exs
@@ -1,0 +1,11 @@
+defmodule Trento.Repo.Migrations.AddOperationAllAbilities do
+  use Ecto.Migration
+
+  def up do
+    execute "INSERT INTO abilities(name, resource, label, inserted_at, updated_at) VALUES ('operation', 'all', 'Permits running operations in all resources', NOW(), NOW())"
+  end
+
+  def down do
+    execute "DELETE FROM abilities WHERE name = 'operation' AND resource = 'all'"
+  end
+end
--- a/assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.jsx
+++ b/assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { assign, find } from 'lodash';
+import { assign, find, orderBy, startsWith } from 'lodash';
 import { getFromConfig } from '@lib/config';
 
 import Select from '@common/Select';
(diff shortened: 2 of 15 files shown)


priv/repo/migrations/20260421073949_add_operation_all_abilities.exs
```diff
@@ -0,0 +1,11 @@
+defmodule Trento.Repo.Migrations.AddOperationAllAbilities do
+  use Ecto.Migration
+
+  def up do
+    execute "INSERT INTO abilities(name, resource, label, inserted_at, updated_at) VALUES ('operation', 'all', 'Permits running operations in all resources', NOW(), NOW())"
+  end
+
+  def down do
+    execute "DELETE FROM abilities WHERE name = 'operation' AND resource = 'all'"
+  end
```
balanza, line 10: **question:** ‏I understand this is done only for changing the description, right?

balanza (approved):

priv/repo/migrations/20260421073949_add_operation_all_abilities.exs
arbulu89: **TLDR**: No, it is not only for changing the description. It is to have the `operation:all` ability together with the other new groups

Let's see if I understand the context of the question  properly:
1. If you ask why we do up/down here, is because ecto doesn't provide any utility function to add and remove rows in a table, so we need to use `execute`. 

2. If you ask why I simply add this new ability, is because how we do the ability groupings in the frontend. We group all abilities in the `AbilitiesMultiSelect` that are described in the constant `groupedAbilities`. So if we choose the group label, behind the scenes we are choosing all of the children abilities. So having 2 groups that would have the same child elements (`operation:all` and `operation:host`, both would have `reboot:host`, for example) wouldn't work with the current code, as the first group with the shared child would take it, but not the second.

And trying to improve this logic, would make the frontend logic yet more confusing.
The reason why we have the groups only in the frontend was simply a way to avoid having too many abilities in the select (we would have +20 entries more in the select). We could do this differently, adding the logic in the backend, and instead of grouping, simply removing abilities we don't want in the frontend. But well, we chose the 1st option back in time

PR #4213, opened 2026-04-21
Update sles subscription on host deregister

arbulu89:
# Description

Fix sles subscription number in the about page when hosts are deregistered.
The number always shows the total number of sles subscriptions without excluding deregistered hosts.

I had 2 options here:

1.
Update the query to get the SLES subscriptions to filter out dereigstered ones. I implemented this

2.
Actually remove the SLES subscriptions from the database. This is somehow more complex as it requires additional things like: 
- Update aggregate to empty the subscriptions list (no need for new events anyway)
- Remove the subscriptions in the projector
- Add a migration to remove "zombie" subscriptions that we might have now

It is not complex (I did the code and it doesn't require that much code), but I don't think it is worthy, as at the end it doesn't add anything compared to the previous option.

In this scenario, the `SlesSubscriptionsUpdated` happens on discovery (as they were removed before) and in the other not (unless the subscription has changed in the discovery of course).

**We cannot simply add the `on_delete: :delete_all` in the migration as hosts are not really deleted, they are soft-deleted**

@vicenteqa I have removed some few files that honestly don't look really needed. Let me know if you think we should keep them.

What do you think?

## How was this tested?
UT and e2e tests

No need


vicenteqa:
> I have removed some few files that honestly don't look really needed. Let me know if you think we should keep them.

that should be fine :+1:

test/e2e/cypress/pageObject/about_po.js
```diff
@@ -29,9 +30,12 @@ export const expectedGithubUrlIsDisplayed = () => {
...
-export const expectedSlesForSapSubscriptionsAreDisplayed = () => {
-  const subscriptions = getValue('subscriptions');
+export const expectedSlesForSapSubscriptionsAreDisplayed = (subscriptions) => {
   return cy
     .get(amountOfSlesForSapSubscriptionsLabel)
     .should('have.text', `${subscriptions} found`);
 };
+
+export const apiDeregisterHost = () => {
+  basePage.apiDeregisterHost(hostToDeregister);
```
vicenteqa, line 40: nit: We need the `return` statement here to keep the cypress command chain alive. Without it, the function returns undefined instantly, and Cypress loses its ability to synchronize this API call with the subsequent test steps, potentially leading to race conditions. Currently the tests without return might be passing because they execute fast enough, or because the next step takes enough time to execute so the previous step already finished. This is related with I want to fully address in https://github.com/trento-project/web/pull/4210#issuecomment-4289787396

lib/trento/hosts.ex
```diff
@@ -80,7 +80,10 @@ defmodule Trento.Hosts do
   def get_all_sles_subscriptions do
     query =
       from s in SlesSubscriptionReadModel,
+        join: h in HostReadModel,
+        on: s.host_id == h.id,
         where: s.identifier == "SLES_SAP",
+        where: is_nil(h.deregistered_at),
```
vicenteqa, line 86: this is just out of curiosity, as I am trying to better understand elixir code, in this change you are reading the Hosts and making sure that the `deregistered_at` timestamp is `nil` to make sure its only counting the registered ones, is that correct?

nelsonkopliku (approved):
Hey Xabi, looks good and the solution looks simple enough.

Question: at the end if the host gets restored and it has different subscriptions, it would eventually be reflected after the relevant discoveries happen, right?

arbulu89:
> Question: at the end if the host gets restored and it has different subscriptions, it would eventually be reflected after the relevant discoveries happen, right?

Yes, absolutely, with a new event.

vicenteqa (commented):
hey, I see it's already merged. I thought I submitted the comments earlier but looks like I didn't. I will address the `return` thing in #4210

lib/trento/hosts.ex
arbulu89: Yes. Join the subscriptions and hosts tables and filter in only hosts with `deregistered_at` as `nil` (registered)

test/e2e/cypress/pageObject/about_po.js
arbulu89: Now I actually see this, putting the curly brackets was a bad idea (I guess just copy pasted).
In this cases, `export const apiDeregisterHost = () => basePage.apiDeregisterHost(hostToDeregister);` is enough.
This would inherently have a `return`. 
Maybe we can skip putting `return` everywhere if the function is other single function execution. It is a really typical `js` thing. Most probably we don't have the same linting rules here.

This is checked by the `"arrow-body-style": ["error", "as-needed"]` rule (and additionally `prefer-arrow-callback` to standardize the func bodies).
We could add it in the config if we want and reduce a lot of boilerplate.

PR #4220, opened 2026-04-23
Refactor table filter tests

arbulu89:
# Description

Refactor frontend tests using `table` `filterTable` and `clearTable` helpers using appropriate `user-event`.
We were using `fireEvent` which prevents adding new test cases properly.
`fireEvent` is deprecated, but we still have usages of it.
This at least removes some of them.

## How was this tested?

UT

No need

### DIFF
--- a/assets/js/common/Table/Table.test.jsx
+++ b/assets/js/common/Table/Table.test.jsx
@@ -126,6 +126,8 @@ describe('Table component', () => {
 
   describe('filtering', () => {
     it('should filter by the chosen filter option with default filter', async () => {
+      const user = userEvent.setup();
+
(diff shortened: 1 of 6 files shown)


EMaksy (approved):
Hey @arbulu89, LGTM, thanks for taking care of this.
Do you think we should open a follow-up ticket for the remaining fireEvent usages, or are you planning to remove them later as well?

arbulu89:
> Hey @arbulu89, LGTM, thanks for taking care of this. Do you think we should open a follow-up ticket for the remaining fireEvent usages, or are you planning to remove them later as well?

There are too many to remove now hehe
Maybe yes, we should have a tech-debt ticket.

PR #4222, opened 2026-04-23
Improve SAP link tooltip

arbulu89:
# Description

Improve the tooltip we use in several places (specially cluster related views) when the SAP system is references, but not registered in Trento (for example, the Cluster of a ASCS/ERS system is found, but the system itself is not in Trento).

In this case, we can improve the tooltip from `System not currently registered` to `SAP system not currently registered` or `HANA database not currently registered`.

It is just a small cosmethic improvement.
I have taken advantage to improve some tech-debts we had and use properly `APPLICATION_TYPE` and `DATABASE_TYPE` to identify the instances.

**Disclamier:** If this happens with HANA+ASCS/ERS cluster, we cannot get the correct type, so we fall back to the previous message. We could try to improve this scenario, it is not impossible, but I think it is such a extreme case that the effort is not worthy.

not_registered_tooltip

Depends on: https://github.com/trento-project/web/pull/4220

## How was this tested?

UT

No need

### DIFF
--- a/assets/js/common/SapSystemLink/SapSystemLink.test.jsx
+++ b/assets/js/common/SapSystemLink/SapSystemLink.test.jsx
@@ -55,6 +55,10 @@ describe('SapSystemLink', () => {
       systemType: undefined,
       expectedTooltip: 'System currently not registered',
     },
+    {
+      systemType: null,
+      expectedTooltip: 'System currently not registered',
+    },
   ])(
     `renders $systemType tooltip when sapSystemId or systemType are not provided`,
     async ({ systemType, expectedTooltip }) => {
--- a/assets/js/common/SapSystemLink/SapSystemLink.test.jsx
+++ b/assets/js/common/SapSystemLink/SapSystemLink.test.jsx
@@ -17,7 +17,7 @@ describe('SapSystemLink', () => {
       </SapSystemLink>
     );
 
-    const sapSystemLinkElement = screen.getByRole('link', { sid });
+    const sapSystemLinkElement = screen.getByRole('link', { name: sid });
 
     expect(sapSystemLinkElement).toBeInTheDocument();
     expect(sapSystemLinkElement).toHaveAttribute('href', `/databases/${id}`);
@@ -32,7 +32,7 @@ describe('SapSystemLink', () => {
       </SapSystemLink>
     );
 
-    const sapSystemLinkElement = screen.getByRole('link', { sid });
+    const sapSystemLinkElement = screen.getByRole('link', { name: sid });
 
     expect(sapSystemLinkElement).toBeInTheDocument();
     expect(sapSystemLinkElement).toHaveAttribute('href', `/sap_systems/${id}`);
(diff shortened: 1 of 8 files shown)


assets/js/state/selectors/sapSystem.js
```diff
@@ -84,37 +84,26 @@ export const getEnrichedDatabaseDetails = createSelector(
...
+      .map((instance) => ({ ...instance, type: APPLICATION_TYPE }))
       .concat(
         databaseInstances.map((instance) => ({
           ...instance,
-          type: 'databases',
+          type: DATABASE_TYPE,
         }))
       )
 );
+
```
arbulu89, line 102: refactor:
this is a small refactor to remove duped code and reuse already existing logic

assets/js/common/SapSystemLink/SapSystemLink.test.jsx
```diff
@@ -22,18 +23,55 @@ describe('SapSystemLink', () => {
...
 
-    const sidElement = screen.getByText(sid);
+    renderWithRouter(
+      <SapSystemLink systemType={APPLICATION_TYPE} sapSystemId={id}>
+        {sid}
+      </SapSystemLink>
+    );
 
-    await act(async () => user.hover(sidElement));
+    const sapSystemLinkElement = screen.getByRole('link', { sid });
```
EMaksy: There is a small typo here,

I think getByRole should be used like this:

```
screen.getByRole('link', { name: sid })
``` 

The second argument here is the options object, and name is the documented way to filter by accessible name.

See this as reference 
https://testing-library.com/docs/queries/byrole/#api

assets/js/pages/ClusterDetails/ClustersList.jsx
```diff
@@ -30,6 +32,22 @@ import ClusterLink from './ClusterLink';
 const getSapSystemBySID = (instances, sid) =>
   instances.find((instance) => instance.sid === sid);
 
+// getSapSystemType gets the current SAP system type.
+// if the SAP system is not found, it evaluates the
+// cluster type.
+const getSapSystemType = (systemType, clusterType) => {
+  if (systemType) return systemType;
```
EMaksy: Idea:

if (systemType) works, but it relies on truthiness.

If the intent is return the provided system type unless it is missing, I’d make that explicit with if (systemType != null) 

So we have:
```
systemType === undefined -> check clusterType
systemType === null -> check clusterType
```

Maybe something like this 
```js
const getSapSystemType = (systemType, clusterType) => {
  if (systemType != null) return systemType;
  if (clusterType === ASCS_ERS) return APPLICATION_TYPE;
  if (clusterType === HANA_ASCS_ERS) return null;
  return DATABASE_TYPE;
};
```
or 
```js
const getSapSystemType = (systemType, clusterType) => {
  if (systemType != null) return systemType;

  switch (clusterType) {
    case ASCS_ERS:
      return APPLICATION_TYPE;
    case HANA_ASCS_ERS:
      return null;
    default:
      return DATABASE_TYPE;
  }
};
```

EMaksy (commented):
Hey @arbulu89  there is just a small fix in the test to do, and just a comment/idea on one of the helper function, but the rest lgtm 👍

assets/js/common/SapSystemLink/SapSystemLink.test.jsx
arbulu89: Good catch!
Fixed

EMaksy (approved):
Thank you lgtm!

assets/js/pages/ClusterDetails/ClustersList.jsx
arbulu89: I have opted for using `sapSystemData?.type ?? getSapSystemType(clusterType)` and removing the need for the comparison in the function itself. It looks maybe cleaner.

PR #4224, opened 2026-04-23
Fix misused waitFor

balanza:
`waitFor` shouldn't be used when the element is meant to already be i the page. By removing, we optimise both failure time and error messages.


vicenteqa (approved):
👍

PR #4234, opened 2026-04-28
Use clustered SAP system/database instances to create SAP system link

arbulu89:
# Description

Change how we associate the SAP system/database in the clusters list view when we create the SAP system link.
Right now, we match by SID using all the registered instances, but this can lead to bad links, as the SIDs can be duplicated (have 2 systems with the same SID). In that case, we would always link to the first finding.
Now, the clusters are enriched with SAP instances data, so we can only check instances that belong to that cluster to create the link. 

Here an example:
fixed_sap_link

## How was this tested?

UT

No need

### DIFF
--- a/assets/js/state/selectors/cluster.js
+++ b/assets/js/state/selectors/cluster.js
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-License-Identifier: Apache-2.0
 
-import { get, find, uniq, has, includes } from 'lodash';
+import { get, find, uniq, has } from 'lodash';
 import { createSelector } from '@reduxjs/toolkit';
 
 import {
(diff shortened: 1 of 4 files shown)


assets/js/state/selectors/cluster.js
```diff
@@ -129,3 +130,37 @@ export const getClusterSelectedChecks = createSelector(
...
+    (state) => state.hostsList.hosts,
+    (state) => getAllSAPInstances(state),
+  ],
+  (clusters, hosts, allInstances) =>
+    clusters.map((cluster) => {
+      const clusterHostIDs = hosts
+        .filter(({ cluster_id: clusterID }) => cluster.id === clusterID)
+        .map(getHostID);
+
+      const enrichedSapInstances = cluster.sap_instances.map((sapInstance) => {
```
EMaksy, line 154: I'm wondering if cluster.sap_instances is always guaranteed here ? 
Maybe adding a guard could be an idea cluster.sap_instances || [] ? 

Or we have it always not fully sure

EMaksy (commented):
Hey @arbulu89 the pr lgtm, just a little question in one comment

assets/js/state/selectors/cluster.js
arbulu89: Yes, it is guaranteed.
All clusters have `sap_instances` list, even if it is empty

EMaksy (approved):

PR #4239, opened 2026-04-28
Update storybook

antgamdia:
# Description

This PR runs an automated script (not ready as of today, it needs some polishing up, find them at https://github.com/trento-project/web/commit/2af5625dc8b42cacba968988eb14288026f23c2a) to synchronize the component's properties with the ones in the stories. As we have been updating components, some properties no longer exist or some others were never added to the story. 

See https://storybook.js.org/docs/api/arg-types

Next, I've used some LLMs to auto-perform some improvements over all the files.

~For now, I'm just adding the properties... I'll get my friend Gemini to populate the descriptions in a second round~ 😅

~Edit: note to myself: check empty argTypes: {}~

## How was this tested?

Building storybook manually.


assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.stories.jsx
```diff
@@ -0,0 +1,39 @@
+import Component from './AbilitiesMultiSelect';
```
arbulu89: comment:
i don't know if importing `Component` is a standard, but we never do it this way.
we would import it as `import AbilitiesMultiSelect from '.'`
And we would import this after the `storybook/actions` import.
We put our "local" import after the 3rd party library imports

assets/js/common/AdvisoryIcon/AdvisoryIcon.stories.jsx
```diff
@@ -4,39 +4,40 @@ import AdvisoryIcon from '.';
...
-        type: 'radio',
-      },
+      control: { type: 'radio' },
     },
   },
 
   render: (args) => <AdvisoryIcon {...args} />,
 };
 
+export const Default = {
```
arbulu89, line 34: nit:
i don't need we need this, it is a copy from `SecurityAdvisory`.
Unless it is "mandatory" to have `Default`

assets/js/common/AIConfiguration/EditModal/AIConfigurationModal.stories.jsx
```diff
@@ -10,94 +10,108 @@ export default {
...
+    onSave: action('onSave'),
+    onUpdate: action('onUpdate'),
   },
 };
 
-export const Default = {};
-
 export const WithUnmappedProvider = {
   args: {
+    open: true,
```
arbulu89, line 67: nit:
I think we can import `Default` here and in the others to reduce duplications.
```
args: {
    ...Default.args,

File: assets/js/common/AIProviderLabel/AIProviderLabel.stories.jsx [hunk 4]
```diff
@@ -5,12 +5,19 @@ export default {
...
     provider: {
-      control: 'select',
+      description: 'AI provider name',
+      control: { type: 'select' },
       options: ['googleai', 'openai', 'anthropic', 'unmapped_provider'],
     },
   },
 };
 
+export const Default = {
```
arbulu89, line 18: nit:
i would remove this `Default` entry

assets/js/common/AlertingSettingsConfig/AlertingSettingsConfig.stories.jsx
```diff
@@ -8,36 +9,43 @@ export default {
...
+    ariaLabel: '',
+    addClasses: '',
+  },
+};
 
 export const WithFilledInValues = {
   args: {
     settings: alertingSettingsFactory.build(),
+    userAbilities: [],
+    onEditClick: action('onEditClick'),
```
arbulu89, line 49: nit:
we should import `Default` args

assets/js/common/ChartsFeatureWrapper/ChartsFeatureWrapper.stories.jsx
```diff
@@ -0,0 +1,20 @@
...
+
+export default {
+  title: 'Components/ChartsFeatureWrapper',
+  component: Component,
+  argTypes: {
+    children: {
+      description: 'Content or text displayed inside the component',
+      control: { type: 'text' },
+    },
+    chartsEnabled: {
```
arbulu89, line 14: question:
is this not a bool?

assets/js/common/CheckCustomizationModal/CheckCustomizationModal.stories.jsx
```diff
@@ -83,54 +83,76 @@ export default {
...
     },
+    customizationStatus: {
+      description:
+        'Status of the check customization operation (ongoing, failed, or invalid)',
+      control: { type: 'text' },
+    },
+  },
+};
+
+export const Default = {
```
arbulu89, line 142: nit:
i'm not sure we need this. It is the same as the next one

assets/js/common/ClusterTypeLabel/ClusterTypeLabel.stories.jsx
```diff
@@ -0,0 +1,24 @@
+import Component from './ClusterTypeLabel';
+
+export default {
+  title: 'Components/ClusterTypeLabel',
+  component: Component,
+  argTypes: {
+    clusterType: {
+      description: 'Type of the cluster',
+      control: { type: 'text' },
```
arbulu89: comment:
all these 3 should be radio type as they are enums

assets/js/common/DisabledGuard/DisabledGuard.stories.jsx
```diff
@@ -52,6 +63,18 @@ const guardElementWithTooltip = (args) => (
   </div>
 );
 
+export const Default = {
+  args: {
+    userAbilities: [{ name: 'all', resource: 'all' }],
+    permitted: [],
+    withTooltip: false,
+    tooltipMessage: '',
```
arbulu89: nit:
I wouldn't add this, at least not with an empty message

assets/js/common/GenericError/GenericError.stories.jsx
```diff
@@ -6,12 +6,22 @@ export default {
   title: 'Components/GenericError',
   component: GenericError,
   args: { message: undefined },
+  argTypes: {
```
arbulu89, line 11: question:
shouldn't we remove `args` from on top of this?

assets/js/common/HealthIcon/HealthIcon.stories.jsx
```diff
@@ -6,7 +6,28 @@ export default {
...
   argTypes: {
     isLink: {
-      control: 'boolean',
+      control: { type: 'boolean' },
+    },
+    health: {
+      type: 'string',
+      description:
+        'The health status to display (passing, warning, critical, absent, pending, or unknown)',
+      control: { type: 'text' },
```
arbulu89: comment:
this should be radio/enum

assets/js/common/HostInfoBox/HostInfoBox.stories.jsx
```diff
@@ -0,0 +1,24 @@
...
+    },
+    agentVersion: {
+      description: 'The agentVersion prop',
+      control: { type: 'text' },
+    },
+  },
+};
+
+export const Default = {
+  args: { arch: '', provider: '', agentVersion: '' },
```
arbulu89: comment:
let's put something better here

arbulu89 (commented):
Hey @antgamdia,
I went through some of the changes, but reviewing everything would take me the whole day.

Some comments: 
- Have a look into my comment of ordering imports.
- We shouldn't have `Default` entries always, some cases don't make sense
- We should use `Default.args` in many entries to avoid duplication

assets/js/common/AdvisoryIcon/AdvisoryIcon.stories.jsx
antgamdia: It is not mandatory _per se_, but adding it in the `args` seems to be the way to set the default value ([source](https://storybook.js.org/docs/api/arg-types#defaultvalue)). We can either define the ´args´ in the object we define ´argTypes` or just do it in the default.

I chose to create a default for each story because it follows a simple inclusion criterion (_every story has a default_), otherwise, we have to think whether we should add one or not.

I don't see any drawbacks if we keep them, but we can list the components that don't require a Default and add a comment explaining it in each file, so we don't forget in the future.

assets/js/common/AbilitiesMultiSelect/AbilitiesMultiSelect.stories.jsx
antgamdia: I ran the linter and assumed it had dealt with the imports organization automatically 😅  I'll have a closer look then, thanks for pointing it out

assets/js/common/AdvisoryIcon/AdvisoryIcon.stories.jsx
arbulu89: Look, it is a nitpick hehe
If it is a convention, it is a convention, but there are some components that `Default` might not make sense (like this one, where we have 3 different icons with different meanings, and none of them is default).

assets/js/common/ChartsFeatureWrapper/ChartsFeatureWrapper.stories.jsx
antgamdia: yep, it is indeed. The initial bulk addition didn't work that well. I've run some checks to modify the types and reduce the mistmatches... but I can't 100% assure they are all correct :S

assets/js/common/AIConfiguration/EditModal/AIConfigurationModal.stories.jsx
antgamdia: Thanks! Done!  (across all the files, hopefully)

assets/js/common/DisabledGuard/DisabledGuard.stories.jsx
antgamdia: I have updated several files with more meaningful defaults. Thanks!

nelsonkopliku (approved):
As long as Xabi's comments have been addressed, LGTM

PR #4244, opened 2026-04-29
Prune events cron job

balanza:
Periodically prune discovery events using a cron job. The default values are borrowed from the batch job implemented in the helm chart https://github.com/trento-project/helm-charts/blob/main/charts/trento-server/charts/trento-web/templates/prune-events-cronjob.yaml

The job can be configured using env variables:
* `PRUNE_EVENTS_INTERVAL` the crontab for the job
* `PRUNE_EVENTS_OLDER_THAN` the retention threshold of the events, in days

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -157,9 +157,17 @@ config :trento, Trento.Scheduler,
       task: {Trento.ActivityLog, :clear_expired_logs, []},
       run_strategy: {Quantum.RunStrategy.Random, :cluster},
       overlap: false
+    ],
+    prune_discovery_events: [
+      schedule: "0 0 * * *",
+      task: {Trento.Discovery, :prune_discovery_events, [10]},
+      run_strategy: {Quantum.RunStrategy.Random, :cluster},
+      overlap: false
     ]
   ],
   debug_logging: false
+
+config :trento, prune_events_cronjob_schedule: "0 0 * * *"
 
 config :trento, Trento.Infrastructure.Messaging,
   adapter: Trento.Infrastructure.Messaging.Adapter.AMQP
--- a/config/config.exs
+++ b/config/config.exs
@@ -167,8 +167,6 @@ config :trento, Trento.Scheduler,
   ],
   debug_logging: false
 
-config :trento, prune_events_cronjob_schedule: "0 0 * * *"
-
 config :trento, Trento.Infrastructure.Messaging,
   adapter: Trento.Infrastructure.Messaging.Adapter.AMQP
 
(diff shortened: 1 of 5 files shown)


lib/trento/discovery.ex
```diff
@@ -88,6 +88,16 @@ defmodule Trento.Discovery do
     Repo.all(query)
   end
 
+  @doc """
+  Prune all discovery events (regular and discarded) older than the given number of days.
+  """
+  @spec prune_discovery_events(number) :: :ok
+  def prune_discovery_events(days) do
+    prune_events(days)
+    prune_discarded_discovery_events(days)
```
nelsonkopliku, line 97: **thought:** we could consider making these two private. If we do so we also need to revisit tests.

Not a biggie anyway

config/config.exs
```diff
@@ -157,10 +157,18 @@ config :trento, Trento.Scheduler,
...
+    prune_discovery_events: [
+      schedule: "0 0 * * *",
+      task: {Trento.Discovery, :prune_discovery_events, [10]},
+      run_strategy: {Quantum.RunStrategy.Random, :cluster},
+      overlap: false
     ]
   ],
   debug_logging: false
 
+config :trento, prune_events_cronjob_schedule: "0 0 * * *"
```
nelsonkopliku: **question:** it does not seem we are using this. If not necessary we could remove.

config/runtime.exs
```diff
@@ -207,6 +207,12 @@ if config_env() in [:prod, :demo] do
       ],
       hosts_checks_execution: [
         schedule: "*/#{System.get_env("CHECKS_INTERVAL", "5")} * * * *"
+      ],
+      prune_discovery_events: [
+        schedule: System.get_env("PRUNE_EVENTS_INTERVAL", "0 0 * * *"),
```
nelsonkopliku, line 212: **praise:** while it deviates from what we do elsewhere, it gives more control.

nelsonkopliku (commented):
Thanks @balanza!

Besides a couple of minor comments LGTM.

Let's add also documentation for the new env variables and what they drive.

config/runtime.exs
balanza: I thought we needed this, because in the helm chart we already provide the job and it's configurable.

config/config.exs
balanza: done

lib/trento/discovery.ex
balanza: done

nelsonkopliku (approved):

PR #4249, opened 2026-04-30
Fix refresh token flow execution in frontend

arbulu89:
# Description

Then a multiple requests are forbidden at the same time, only the first one is retried.

image

Set `deduplicateRefresh: false` in the `axios-auth-refresh` pacakge usage to retry the forbidden requests.
In reality, this field should be let to `true`, as the library states that this is the way to retry to failed requests, but checking the code and usage, it is just the opposite, strange...
Related all closed issue: https://github.com/Flyrell/axios-auth-refresh/issues/123

## How was this tested?

e2e test and manual testing

No need

### DIFF
--- a/assets/js/lib/network/index.js
+++ b/assets/js/lib/network/index.js
@@ -54,7 +54,9 @@ const refreshAuthLogic = async (failedRequest) => {
 
 // Even though it looks counter intuitive, we need to add `deduplicateRefresh: false`
 // to enable the retry of all the requests that got 401 at the same time.
-// Otherwise, only the first request with 401 response is retried.
+// The functionality of this flag is not properly implemented and it doesn't do what
+// it should. However, setting it to false fixes our problem.
+// Without this change, only the first request with 401 response is retried.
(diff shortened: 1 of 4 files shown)


vicenteqa (approved):
nice fix @arbulu89 !
I also confirm that the fix removes the issue completely:
https://github.com/trento-project/web/actions/runs/25158130286

skrech (approved):
After reading the code of this axios plugin I think that what we have now is enough to fix the problem.
However, my concern is that the workings of `deduplicateRefresh: false` is totally wrong (buggy) right now and once it's fixed upstream we'd encounter a problem. Nevertheless, you've written a test, so it should cover for that.
So, I'm approving! Thanks, @arbulu89!

arbulu89:
> However, my concern is that the workings of deduplicateRefresh: false is totally wrong (buggy) right now and once it's fixed upstream we'd encounter a problem. Nevertheless, you've written a test, so it should cover for that.

You are totally right. Let me amend and additional comment line with this and I will create a new ticket in our backlog to try to have a long-term solution.

PR #4255, opened 2026-04-30
Fix test modal component not wrapped in `act` warning

arbulu89:
# Description

Fix `act not wrapped` error in frontend modal tests. When a modal is rendered, it needs to wait until it is completely mounted.
That's why we need to wrap it in an `act`.
Don't get scared by the added lines, everything is just putting `act` in every modal render.

This is the error:
```
An update to TransitionRootFn inside a test was not wrapped in act(...).
      When testing, code that causes React state updates should be wrapped into act(...):
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
```

Besides this, I have fixed other error output in `ActivityLogsSettingsModal` test

## How was this tested?

UT

No need


antgamdia (approved):
Thanks! It was really annoying to check the logs and have to continuously scroll down due to those warnings :P

PR #4270, opened 2026-05-05
Fix timezone flaky test

nelsonkopliku:
# Description

Fix a flaky test related to `faker.location.timeZone()` not being among the actual timezones that the app uses.

Fixes

This run for instance
https://github.com/trento-project/web/actions/runs/25361655478/job/74362818110#step:5:425


antgamdia (approved):
Thanks!

PR #4274, opened 2026-05-06
Add Filesystem type filter to File System Capacity chart

arbulu89:
# Description

Add Filesystem filters to host details view `File System Capacity` chart.
The next types are enabled on initial page view: `['gpfs', 'nfs', 'xfs', 'nfs4']`
We can change them, remove initial filtering, etc.

filter_by_fstype

Only 

## How was this tested?

I did test manually and with storybook.
Testing with jest is not possible as react-chart is creating a canvas without real dom elements.
I could test the pills, but they don't check anything more than they exist.

No need


arbulu89:
Tested with @abravosuse and it works fine!

nelsonkopliku (approved):
Thanks!

PR #4278, opened 2026-05-08
Fix wanda versioning

balanza:
The versioning introduced in https://github.com/trento-project/web/pull/4155 didn't consider the case that the `check_url` config value can be set to a relative path (which is the case of a k8s installation).

This PR fix the thing so that the origin hostname is prepended if needed.

### DIFF
--- a/lib/trento/infrastructure/component_versions/component_versions.ex
+++ b/lib/trento/infrastructure/component_versions/component_versions.ex
@@ -23,12 +23,12 @@ defmodule Trento.Infrastructure.ComponentVersions do
   }
 
   @impl true
-  def get_versions do
+  def get_versions(origin \\ nil) do
     fetchers = [
       {:postgres, &fetch_postgres_version/0},
       {:rabbitmq, &fetch_rabbitmq_version/0},
       {:prometheus, &fetch_prometheus_version/0},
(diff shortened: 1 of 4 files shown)


nelsonkopliku (approved):

antgamdia (approved):

lib/trento/infrastructure/component_versions/component_versions.ex
```diff
@@ -129,12 +129,10 @@ defmodule Trento.Infrastructure.ComponentVersions do
...
 
-  defp fetch_wanda_info do
-    checks_base_url = Application.fetch_env!(:trento, :checks_service)[:base_url]
+  defp fetch_wanda_info(origin) do
+    url = resolve_checks_url("/api", origin)
 
-    case HTTPoison.get("#{checks_base_url}/api", [{"Accept", "application/json"}],
-           recv_timeout: @timeout
-         ) do
+    case HTTPoison.get(url, [{"Accept", "application/json"}], recv_timeout: @timeout) do
```
nelsonkopliku: **issue:** not working in our pr env.

This url resolves to `http://4278.prenv.trento.suse.com/wanda/api` and when we call it we get a 301 to https.

We might need to instruct the http client to follow redirects `follow_redirect: true`

PR #4287, opened 2026-05-12
Analytics - Mask events data

arbulu89:
# Description

Mask analytics events data.
- When a element is clicked, mask the event `$el_text` and chain so it doesn't show the value in the event. Using the `ph-mask` class implies that the field is not showed in the session recording as well. (`ph-mask` is the default class, we could change it: https://posthog.com/docs/libraries/js/usage)

mask_text

## How was this tested?

UT and manual testing

No need


abravosuse:
@arbulu89 I have deployed the image in rolling and it works.
When clicking on username in users view:
{06F0C90A-CBF7-4D32-8628-7A4687587A41}
When clicking in Profile dropdown menu:
{34F749A9-C867-41BE-B931-85EF20A2A578}

jagabomb (approved):
Nice, it works!

PR #4289, opened 2026-05-12
Display empty swap chart when data is not available

arbulu89:
# Description

Set empty Swap chart when it is not configured.

image

## How was this tested?

UT

No need


arbulu89:
@jagabomb @abravosuse  
Implementation updated to use static image:
image

balanza (approved):

PR #4293, opened 2026-05-13
Fix flaky test on personal access token

balanza:
The flakiness is due to a possible overlap of dates when generating tokens. 

It's safe to assume it's a problem on the test itself, not on the code being tested.


arbulu89 (approved):
Thank you!

PR #4294, opened 2026-05-13
Fix flaky test on user form

balanza:
Root cause:

`userEvent.setup()` implies `{ delay: 0 }`. That means that, when using `user.type` to simulate user typing, a `setTimeout(...., 0)` is created at every character to simulate human input. As per how the event loop works, React re-renders will interleave into the chain of events. On slow machines or on high parallelism, the loop can get slow and hence get the test to fail.

By setting `{ delay: null }`, we avoid the timeout mechanism altogether


antgamdia (approved):

arbulu89 (approved):
Good!

PR #4297, opened 2026-05-14
Load OS CA certificates when calling wanda

nelsonkopliku:
# Description

When we get wanda and checks version, we might be calling it via the reverse proxy, and when there is ssl termination enabled with a self-signed certificate the request would fail because we are not providing any ca certificate to the http client.

If the OS has the relevant ca cert, httpoison does not load it automatically.

This fix makes sure it's loaded and provided to the request.

### DIFF
--- a/lib/trento/infrastructure/component_versions/component_versions.ex
+++ b/lib/trento/infrastructure/component_versions/component_versions.ex
@@ -134,7 +134,8 @@ defmodule Trento.Infrastructure.ComponentVersions do
 
     case HTTPoison.get(url, [{"Accept", "application/json"}],
            recv_timeout: @timeout,
-           follow_redirect: true
+           follow_redirect: true,
+           ssl: [verify: :verify_peer, cacerts: resolve_ca_certs()]
          ) do
       {:ok, %HTTPoison.Response{status_code: 200, body: body}} ->
         case Jason.decode(body) do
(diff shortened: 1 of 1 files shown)


vicenteqa (approved):

antgamdia:
It's been a long time since I set up the environment with SSO, but I had some notes about CAs (TRNT-3885). Sharing here just in case:

- If the IdP is deployed using a custom CA, there is no way to make Trento web rely on this CA to establish the TLS connection.
- Skipping TLS wasn't supported by design, so a user should be able to inject this CA somehow.
- Also, some PROXY settings could be allowed to be set as well, as an additional improvement.

Maybe not relevant, as this is just a change for Wanda... but just in case :S

nelsonkopliku:
> If the IdP is deployed using a custom CA, there is no way to make Trento web rely on this CA to establish the TLS connection.
>
I see, however there's no idp involved in this specific fix

> Skipping TLS wasn't supported by design, so a user should be able to inject this CA somehow.
>
Correct and that's still that way

> Also, some PROXY settings could be allowed to be set as well, as an additional improvement.
> 
Yes, there are several improvement areas around the usage of `HTTPoison`, indeed.
Considering the fact that it does not load system's CA certificates automatically, we might need to revisit how we use that client.

> Maybe not relevant, as this is just a change for Wanda... but just in case :S
>
Thanks anyway for the input.

lib/trento/infrastructure/component_versions/component_versions.ex
```diff
@@ -173,4 +174,12 @@ defmodule Trento.Infrastructure.ComponentVersions do
...
     end
   end
+
+  defp resolve_ca_certs do
+    :public_key.cacerts_get()
+  rescue
+    e ->
+      Logger.error("Failed to load system CA certificates: #{inspect(e)}")
+      []
+  end
```
nelsonkopliku, line 184: this is defensive. I'd rather have the request fail because the ssl verification fails, instead of having it crash, even though a very remote possibility.

PR #4299, opened 2026-05-14
Normalize query string to list

arbulu89:
# Description

Support openAPI styles query string arrays usage, like `?severity=info&severity=debug`.
Without this, phoenix doesn't convert these elements into lists even though they are defined in the openAPI schema, as only bracket notation (`severity[]=info`) is supported.

## How was this tested?

UT and manual testing with swagger

No need

### DIFF
--- a/lib/trento_web/controllers/v1/activity_log_controller.ex
+++ b/lib/trento_web/controllers/v1/activity_log_controller.ex
@@ -10,6 +10,11 @@ defmodule TrentoWeb.V1.ActivityLogController do
   alias Trento.Users.User
   alias TrentoWeb.OpenApi.V1.Schema
 
+  plug TrentoWeb.Plugs.NormalizeListsPlug,
+    list_fields: %{
+      get_activity_log: ["severity", "actor", "type"]
+    }
+
   plug OpenApiSpex.Plug.CastAndValidate, json_render_error_v2: true
   action_fallback TrentoWeb.FallbackController
   plug TrentoWeb.Plugs.LoadUserPlug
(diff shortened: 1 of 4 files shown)


arbulu89:
Customer has tested this and validates it works.
Tested together with:  https://github.com/trento-project/mcp-server/pull/123

antgamdia (approved):
Thanks!

I think this controller is the only one subject to this bug/requirement, no? I don't see other `type: :array` as query params, but wanted to double-check.

arbulu89:
> Thanks!
> 
> I think this controller is the only one subject to this bug/requirement, no? I don't see other `type: :array` as query params, but wanted to double-check.

Yes, these 3 params are the unique query string params with array type.

skrech (approved):
I'm approving but I have to note that this is too much processing for something standard.
IMO, we should strive for the minimal maintenance approach. If phoenix expects bracket notation, then for me the best approach is/was to intercept the generated OpenAPI spec and replace the names there.  The fact that description in open_api_spex doesn't work with phoenix's expectations is a clear bug for me. 
If in the future open_api_spex fixes the issue, or phoenix itself supports different type of array arguments, the changes introduced in these 3 PRs (so far) would look confusing.

arbulu89:
> I'm approving but I have to note that this is too much processing for something standard. IMO, we should strive for the minimal maintenance approach. If phoenix expects bracket notation, then for me the best approach is/was to intercept the generated OpenAPI spec and replace the names there. The fact that description in open_api_spex doesn't work with phoenix's expectations is a clear bug for me. If in the future open_api_spex fixes the issue, or phoenix itself supports different type of array arguments, the changes introduced in these 3 PRs (so far) would look confusing.

I think all of us agree that it is a suboptimal solution.
Anyway, I encourage you to propose your idea and we can compare.
We needed to have something because one fix was urgent for one of our customers. That is solved now, as we provided a test package and they could move forward.

PR #4300, opened 2026-05-14
Update gen_rmq dependency to new GitHub repo

chargio:
We were using a gen_rmq from a personal repo from a team member that is no longer part of the team. To control the repo and its updates, we have forked the code to trnt_gen_rmq. This PR modifies the dependency to point to the new place. This will also allow us to publish it on hex.pm if we want to with the new name.

# Description

Change on the dependency from an external repo to an internal one.

Fixes # (issue)

Nothing

- [ x ] **DONE**

## How was this tested?
There is no change. The code is the same, just stored in a different place.

- [  x ] **DONE**

There is no need to do so.

- [x] **DONE**


arbulu89 (commented):
Can you update the `mix.lock` file with `mix deps.update gen_rmq`?

PR #4326, opened 2026-05-21
Export version in rpm packages to use same value during compilation

arbulu89:
# Description

Use `rpm` package version to build the package.
Otherwise the `VERSION` file content is used, and this doesn't include "rolling" versions.

## How was this tested?

Visually on OBS

No need


skrech (approved):
Cool

balanza (approved):

PR #4328, opened 2026-05-21
[TRNT-4419] Update installation process for demo environment

antgamdia:
# Description

This PR is mainly intended to ~enable the MCP server in the demo environment~ (extracted to https://github.com/trento-project/web/pull/4338); however, the installation process needed some revamping (EOL k8s deps, old cert-manager versions, non OCI installation, pulling old deps, etc), so I've squeezed those improvements here.
Also, some workers were using inconsistent versions of the base image; I've replaced those with the 24.04 used elsewhere.

Some additional details:

* Updated `K3S_VERSION` to `v1.36.1+k3s1` and `CERT_MANAGER_VERSION` to `v1.20.2` in `.github/workflows/ci.yaml` for newer Kubernetes and cert-manager support. Added `HELM_VERSION` (`v4.2.0`) to ensure consistent Helm usage.
* * Changed all workflow jobs to use `ubuntu-24.04` runners instead of `ubuntu-latest` or `ubuntu-22.04` for improved compatibility and security.
* Switched Helm chart installation to use OCI registry (`ghcr.io`) instead of downloading and unzipping charts. Added a step to resolve the latest chart version using ~`skopeo`~ the `--devel` helm flag.
* Updated cert-manager installation to use the official OCI Helm chart and improved resource preparation by checking out the `helm-charts` repository and templating resources with `envsubst`.
* Enhanced the `helm upgrade` command to set additional image tags, enable the `trento-mcp-server`, and use the correct values file location.

Related # TRNT-4419

## How was this tested?

I've used a very similar approach in https://github.com/trento-project/helm-charts/pull/195

N/A

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -445,7 +445,7 @@ jobs:
   coveralls-finish:
     needs: [test-elixir]
     if: needs.detect-changes.outputs.elixir-changed == 'true' || always()
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Coveralls Finished
         uses: coverallsapp/github-action@648a8eb78e6d50909eff900e4ec85cab4524a45b # v2
@@ -688,7 +688,7 @@ jobs:
     name: Regression tests
     needs:
       [setup-matrix-env, elixir-deps, npm-deps, npm-e2e-deps, detect-changes]
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         include:
@@ -793,7 +793,7 @@ jobs:
     name: Integration tests
     needs:
       [setup-matrix-env, elixir-deps, npm-deps, npm-e2e-deps, detect-changes]
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         include:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1165,13 +1165,6 @@ jobs:
       - name: Cleanup previous run
         run: |
           rm -fr calico/
-      - name: Resolve latest chart version
-        id: chart-version
-        run: |
-          command -v skopeo >/dev/null || { echo "ERROR: skopeo not installed"; exit 1; }
-          VERSION=$(skopeo list-tags "docker://ghcr.io/trento-project/helm-charts/trento-server" | jq -r '.Tags[]' | grep -E '^[0-9]+\.[0-9]+\.[0-9]' | sort -rV | head -1)
-          echo "Using chart version: ${VERSION}"
-          echo "value=${VERSION}" >> $GITHUB_OUTPUT
       - name: Start a local k8s cluster
         uses: jupyterhub/action-k3s-helm@502a9ff4d816dad543e359971fe2c3128d0f4c6e # v4
         with:
@@ -1209,7 +1202,7 @@ jobs:
       - name: Install trento-server helm chart from OCI registry
         run: |
           helm upgrade -i trento oci://ghcr.io/trento-project/helm-charts/trento-server --wait \
-            --version ${{ steps.chart-version.outputs.value }} \
+            --devel \
             --namespace ${{ env.TRENTO_NAMESPACE }} \
             --create-namespace \
             --timeout 3m \
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1193,12 +1193,12 @@ jobs:
           repository: trento-project/helm-charts
           ref: rolling
           path: helm-charts
-      - name: Prepare cert-manager resources and values for ingress
+      - name: Prepare valid cluster-issuer and certificate (for cert-manager) and values for ingress
         run: |
           envsubst < helm-charts/hack/cert-manager/cluster-issuer.tpl.yaml > helm-charts/hack/cert-manager/cluster-issuer.yaml
           envsubst < helm-charts/hack/cert-manager/certificate.tpl.yaml > helm-charts/hack/cert-manager/certificate.yaml
           envsubst < helm-charts/hack/cert-manager/override-values.tpl.yaml > helm-charts/hack/cert-manager/override-values.yaml
-      - name: Apply cert-manager resources
+      - name: Apply cluster-issuer and certificate (for cert-manager)
         run: |
           kubectl create namespace ${{ env.TRENTO_NAMESPACE }} || true
           kubectl apply -f helm-charts/hack/cert-manager/cluster-issuer.yaml
@@ -1223,7 +1223,7 @@ jobs:
             --set trento-wanda.checks.image.pullPolicy=Always \
             --set trento-wanda.checks.image.repository="${IMAGE_REPOSITORY}/checks" \
             --set trento-wanda.checks.image.tag="rolling" \
-            --set trento-mcp-server.enabled=true \
+            --set trento-mcp-server.enabled=false \
             --set trento-mcp-server.image.pullPolicy=Always \
             --set trento-mcp-server.image.repository="${IMAGE_REPOSITORY}/mcp-server-trento" \
             --set trento-mcp-server.image.tag="rolling" \
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1216,10 +1216,10 @@ jobs:
             --set trento-web.adminUser.password="${{ secrets.DEMO_PASSWORD }}" \
             --set trento-web.image.pullPolicy=Always \
             --set trento-web.image.repository="${IMAGE_REPOSITORY}/trento-web" \
-            --set trento-web.image.tag="rolling" \
+            --set trento-web.image.tag="demo" \
             --set trento-wanda.image.pullPolicy=Always \
             --set trento-wanda.image.repository="${IMAGE_REPOSITORY}/trento-wanda" \
-            --set trento-wanda.image.tag="rolling" \
+            --set trento-wanda.image.tag="demo" \
             --set trento-wanda.checks.image.pullPolicy=Always \
             --set trento-wanda.checks.image.repository="${IMAGE_REPOSITORY}/checks" \
             --set trento-wanda.checks.image.tag="rolling" \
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -1164,53 +1165,72 @@ jobs:
...
+            --timeout 3m \
             --set trento-web.adminUser.password="${{ secrets.DEMO_PASSWORD }}" \
             --set trento-web.image.pullPolicy=Always \
             --set trento-web.image.repository="${IMAGE_REPOSITORY}/trento-web" \
-            --set trento-web.image.tag="demo" \
+            --set trento-web.image.tag="rolling" \
             --set trento-wanda.image.pullPolicy=Always \
             --set trento-wanda.image.repository="${IMAGE_REPOSITORY}/trento-wanda" \
-            --set trento-wanda.image.tag="demo" \
+            --set trento-wanda.image.tag="rolling" \
```
nelsonkopliku: **question:** What is the need to use `rolling`?

Take into account that `demo` tags are basically rolling built for demo, which enable demo features like checks execution, MLM integration...

Using rolling would mean not being able to demo a checks execution, for instance.

nelsonkopliku (commented):
Thanks @antgamdia!

Same comment as in https://github.com/trento-project/helm-charts/pull/195#pullrequestreview-4355836348 point 3.

Additionally, let's be mindful about the tags we use in demo.

.github/workflows/ci.yaml
arbulu89: Exactly. Let's not change this.
And maybe neither the `prometheus.server.auth.type=none`, it looks totally out of context for this PR

.github/workflows/ci.yaml
```diff
@@ -1164,53 +1165,72 @@ jobs:
...
+      - name: Apply cert-manager resources
         run: |
-          envsubst < helm-charts-rolling/hack/cert-manager/certificate.tpl.yaml > helm-charts-rolling/hack/cert-manager/certificate.yaml
-          envsubst < helm-charts-rolling/hack/cert-manager/cluster-issuer.tpl.yaml > helm-charts-rolling/hack/cert-manager/cluster-issuer.yaml
-          envsubst < helm-charts-rolling/hack/cert-manager/override-values.tpl.yaml > helm-charts-rolling/hack/cert-manager/override-values.yaml
-      - name: Apply cluster-issuer and certificate (for cert-manager)
+          kubectl create namespace ${{ env.TRENTO_NAMESPACE }} || true
+          kubectl apply -f helm-charts/hack/cert-manager/cluster-issuer.yaml
+          kubectl apply -f helm-charts/hack/cert-manager/certificate.yaml
+      - name: Login to GHCR for pulling OCI chart
```
arbulu89, line 1199: question:
I understand the logic of using the helm chart from OCI registry, but it is needed?
I mean, we already download the chart itself and are using the code inside even when we run the chart usage

arbulu89 (commented):
Hey @antgamdia ,
I like the upgrade of the process (I personally would have done it in a different PR, to isolate the update).
I don't know if @skrech had some work for this, as he added just last week the option to have the helm chart in the ghcr OCI registry.

Anyway, I wouldn't touch some of the already existing tags in the helm usage.

.github/workflows/ci.yaml
antgamdia: > question: What is the need to use rolling?

Ouch, sorry! I assumed `demo` and `rolling` were equivalent; since some images only had "rolling" I supposed this new tag was the one to use. I'll revert that.

> And maybe neither the prometheus.server.auth.type=none

I don't think so; if unset, it errors, no? See https://github.com/trento-project/docs/pull/205

antgamdia:
> I like the upgrade of the process (I personally would have done it in a different PR, to isolate the update).

Thanks! Well, there's not much regarding the MCP actually 😅 . Except for the `enabled=true`, most of the changes are because of the upgrade. ~I can extract the actual MCP "on" to a follow-up PR.~ Done: https://github.com/trento-project/web/pull/4338

> (...) he added just last week the option to have the helm chart in the ghcr OCI registry.

Good point, let's see. If so, more than happy to sync @skrech !

.github/workflows/ci.yaml
antgamdia: > (...) is it needed?

Do you mean the `Login to GHCR for pulling OCI chart` part? If so, I'm afraid yes, it is needed:

_You need an access token to publish, install, and delete private, internal, and **public** packages._  [Source](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry)

arbulu89: Not login simply login.
In the previous step you checkout the `helm-chart` project from github, so the code is there.
I was just asking if it is needed to use it from the oci registry again, if it was already downloaded (and actually the checkout code is used to override values)

.github/workflows/ci.yaml
arbulu89: Well, i don't argue that. I argue that it belongs to this PR, a change that is mandatory and based on other changes.

.github/workflows/ci.yaml
antgamdia: Ahhh, I see your point! Thanks for clarifying, Then... it is not _needed_, as it is downloaded as you mentioned, but I would prefer to keep it that way (OCI pull). The reason is twofold:

i) we test the usual real workflow: in case the delivery gets broken somehow, dogfooding should give us some sort of early alert;
 ii) we are using some files (like `hack/cert-manager/cluster-issuer.yaml`) that are purely cert-manager stuff, or even "worse", some part of the values.yaml is living in that helm-charts repo (`override-values.yaml`). Maybe those should be moved to another repo like (`.github` or similar), but I kept them just to maintain the changes in this PR to a minimum. 

So, yes, while it is true that we have the helm-chart in the GH runner's context, I still think we should not use it directly and, instead, we should leverage the OCI artifact to ensure a realistic e2e scenario. Anyways, no strong opinion if you all think otherwise, of course.

.github/workflows/ci.yaml
antgamdia: Yeah, it's a pity. We could point to a fixed tag instead of using the latest to skip this change in this PR and then reverting it, but I don't see the real value. Happy to consider any other alternatives you may suggest, though.

skrech:
Hey @antgamdia.

> Switched Helm chart installation to use OCI registry (ghcr.io) instead of downloading and unzipping charts. Added a step to resolve the latest chart version using skopeo.

Since the last release of Helm Charts, we support GHCR registries. However, my understanding is that if you don't specify `--version` in the upgrade/install helm command, it should automatically select the latest version according to SemVer rules. Could we try if this works as expected, instead of manually sorting with bash `sort` command.

I have nothing against extracting the version string first in a step, but maybe leave the default OCI registry logic handle the versioning, for example:

```
helm show chart oci://ghcr.io/trento-project/helm-charts/trento-server --devel | grep 'version:'
```

antgamdia:
> Since the last release of Helm Charts, we support GHCR registries. However, my understanding is that if you don't specify `--version` in the upgrade/install helm command, it should automatically select the latest version according to SemVer rules. Could we try if this works as expected, instead of manually sorting with bash `sort` command.

You are absolutely right, I simply forgot about the existence of the `--devel`  flag 🤦 . Thanks for the heads up! Fixed at https://github.com/trento-project/web/pull/4328/commits/18e87c14baf964daae0d99b69cd0156ff5877eaa



```console
helm upgrade -i trento oci://ghcr.io/trento-project/helm-charts/trento-server --wait  --devel --dry-run=client
Release "trento" does not exist. Installing it now.
Pulled: ghcr.io/trento-project/helm-charts/trento-server:3.2.0-dev1 <-----  it works :)
```

.github/workflows/ci.yaml
```diff
@@ -1168,36 +1169,43 @@ jobs:
...
-          envsubst < helm-charts-rolling/hack/cert-manager/override-values.tpl.yaml > helm-charts-rolling/hack/cert-manager/override-values.yaml
+          envsubst < helm-charts/hack/cert-manager/cluster-issuer.tpl.yaml > helm-charts/hack/cert-manager/cluster-issuer.yaml
+          envsubst < helm-charts/hack/cert-manager/certificate.tpl.yaml > helm-charts/hack/cert-manager/certificate.yaml
+          envsubst < helm-charts/hack/cert-manager/override-values.tpl.yaml > helm-charts/hack/cert-manager/override-values.yaml
       - name: Apply cluster-issuer and certificate (for cert-manager)
         run: |
-          kubectl apply -f helm-charts-rolling/hack/cert-manager/cluster-issuer.yaml
-          kubectl apply -f helm-charts-rolling/hack/cert-manager/certificate.yaml
-      - name: Install trento-server helm chart
+          kubectl create namespace ${{ env.TRENTO_NAMESPACE }} || true
```
skrech, line 1196: Maybe this is redundant now that the helm upgrade command uses `--create-namespace` flag.

.github/workflows/ci.yaml
skrech: My opinion is that it's better to use the artifact from the OCI registry -- it's the realistic use-case. 

Ultimately, we shouldn't need to checkout at all. However, this is a bit deeper topic -- it's good to have this logic isolated somewhere. It's  CI-related, and ideally should be held in the CI, in my opinion. Yet, having logic spilled around GH workflows is something that I don't like.

For me the best way is to put the content of the `hack` dir as a helper functions inside our `continuous-delivery` container. Containers look like a good way to move around common utils. However this seems outside the scope of this PR. 

If we can take inspiration from git-ops projects for continuous delivery, they prefer putting these settings into separate repositories. However, this seems overkill for now. I think that going with the container is the best option.

skrech (approved):
As far as I'm concerned, everything looks good to me, so approving. Maybe wait for @arbulu89 approval as well, though.

.github/workflows/ci.yaml
antgamdia: Not sure; this happens before the _trento_ helm install step. In this demo environment, AFAIK, we are just upgrading it, so the namespace actually exists... however, in a brand-new environment, I don't think the trento NS is being created elsewehere.

skrech: OK, if we need to put these cert resources in our namespace, then OK. Maybe then the `--create-namespace` in helm is not needed then. 😄

antgamdia:
> As far as I'm concerned, everything looks good to me, so approving. Maybe wait for @arbulu89 approval as well, though.

@nelsonkopliku or @arbulu89 , can I get a +1 as well? Thanks!

arbulu89 (approved):

PR #4342, opened 2026-05-27
Upgrade elixir/OTP stack to 1.19.5/27.3.4.11

nelsonkopliku:
# Description

Upgrade the elixir/OTP stack to 1.19.5/27.3.4.11

Relevant comments added inline.

Note on workflows: 
- ~unnecessary matrices for different elixir versions removed~ not yet removed to check backward compatibility
- consolidated to parsing asdf tool versions and use the parsed result across the workflow
- consolidated deps cache keys so that they're consistent and can be shared among workflows more efficiently

Runs:
- main CI https://github.com/trento-project/web/actions/runs/26568563117
- Backend flaky test analysis https://github.com/trento-project/web/actions/runs/26568655320
- E2E flaky test analysis https://github.com/trento-project/web/actions/runs/26568663207

Changes on Docker base images and rpm specs will follow if and when we agree this is good to move forward.
I am a bit surprised that building the images https://github.com/trento-project/web/actions/runs/26561043386/job/78244024572 succeeds. I guess it's because we do not have an elixir constraint in `mix.exs`. On the other hand it means this is backward compatible with elixir 1.15 as well

Depends on https://github.com/trento-project/contracts/pull/146


lib/trento/infrastructure/commanded/roll_up/roll_up.ex
```diff
@@ -37,6 +32,13 @@ defmodule Trento.Infrastructure.Commanded.RollUp.RollUp do
     end
   end
 
+  defp do_rollup(stream_id, roll_up_event, stream_archive_id, conn) do
```
nelsonkopliku, line 35: Changed because of nesting levels

lib/trento/software_updates/discovery.ex
```diff
@@ -79,11 +79,11 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
        %HostReadModel{id: host_id, fully_qualified_domain_name: fully_qualified_domain_name} ->
          case discover_host_software_updates(host_id, fully_qualified_domain_name, authentication) do
-           {:error, error} ->
-             {:error, host_id, error}
-
            {:ok, _, _, _, _} = success ->
              success
+
+           {:error, reason} ->
+             {:error, host_id, reason}
```
nelsonkopliku, line 86: just changed the order to match success before and error after

lib/trento/software_updates/discovery.ex
```diff
@@ -212,12 +218,12 @@ defmodule Trento.SoftwareUpdates.Discovery do
       {:error, :command_dispatching, dispatching_error, _} ->
         {:error, dispatching_error}
 
-      {:error, _} = error ->
+      {:error, _failed_operation, _failed_value, _changes_so_far} = error ->
```
nelsonkopliku, line 223: Latent issue: when `Repo.transaction` is provided an `Ecto.Multi` it returns a 4-tuple error in case of failure. See [this](https://hexdocs.pm/ecto/Ecto.Repo.html#c:transaction/2-use-with-ecto-multi).
Test added.

As an upside we are now able to track `system_id` in `DiscoveryResult` also on failed discoveries, when the failure does not come from getting the system_id itself.
See tests at `test/trento/software_updates/discovery_test.exs`

test/support/commanded/test_event_handler_with_failure_context.ex
```diff
@@ -11,9 +11,9 @@ defmodule TestEventHandlerWithFailureContext do
     name: __MODULE__
 
   use Trento.Support.EventHandlerFailureContext,
-    max_retry: 1,
+    max_retries: 3,
     retry_after: 1,
-    skip: true,
+    skip: false,
```
nelsonkopliku, line 16: without this change, the test "should retry 3 times before shutting down and call the callback functions" fails by timing out on waiting for retries.

Not sure why it wasn't failing before.

test/trento/infrastructure/alerting/alerting_test.exs
```diff
@@ -48,7 +48,7 @@ defmodule Trento.Infrastructure.Alerting.AlertingTest do
     end
 
     test "Notify api key will be expired soon" do
-      insert(:api_key_settings, expire_at: DateTime.add(DateTime.utc_now(), 28, :day))
+      insert(:api_key_settings, expire_at: DateTime.add(DateTime.utc_now(), 28 * 24 + 1, :hour))
```
nelsonkopliku, line 51: fixes a flakiness on this test.

mix.exs
```diff
@@ -118,13 +117,15 @@ defmodule Trento.MixProject do
...
-      {:x509, "~> 0.8.8"},
+      # {:amqp, "~> 4.1.0"},
+      {:x509, "~> 0.9.0"},
       {:argon2_elixir, "~> 4.0"},
       {:ecto_commons, "~> 0.3.4"},
       {:bodyguard, "~> 2.4"},
       {:nimble_totp, "~> 1.0"},
       {:phoenix_html_helpers, "~> 1.0"},
       {:pow_assent, "~> 0.4.18"},
+      {:assent, "== 0.2.10", override: true},
```
nelsonkopliku, line 129: This was a nasty one 😅 

During the upgrade assent went to 0.2.13, however e2e tests related to SSO started failing like [here](https://github.com/trento-project/web/actions/runs/26518310170/job/78103320972#step:14:86)

The log wasn't useful so I tried SSO locally and got `** (BadMapError) expected a map, got: nil` in `TrentoWeb.SessionController.callback/2`, specifically here

```elixir
params = Map.drop(body_params, ["session_params"])
session_params = Map.get(body_params, "session_params")

conn
|> Conn.put_private(:pow_assent_session_params, session_params)
|> PowAssentPlug.callback_upsert(provider, params, idp_redirect_uri(provider))
...
```

Bottom line is that `session_params` is coming as `nil` in our controller and assent handles that differently in the two versions

v0.2.10 — tolerated nil (oauth2.ex)
https://github.com/pow-auth/assent/blob/v0.2.10/lib/assent/strategies/oauth2.ex#L145-L188
```elixir
# line 188 — CATCH-ALL
# State verification silently skipped. 
# Trento's flow worked.
defp maybe_check_state(_session_params, _params), do: :ok
```

v0.2.13 — crashes on nil (oauth2.ex)
https://github.com/pow-auth/assent/blob/v0.2.13/lib/assent/strategies/oauth2.ex#L238

```elixir
# line 235
defp verify_state(config, session_params, params) do
  case Keyword.get(config, :state, true) do
    false -> :ok
    _true -> verify_state(session_params.state, params)   # line 238 ← CRASH
  end
end
```

Since the `case Keyword.get(config, :state, true) do` would skip state verification if `config.state` is `false` I tried a version where I added `state: false` in the configuration but:
- there might be security implications
- it didn't work
- it's not wise touching auth now

Bottom line:
whether it is correct that `session_params` comes as `nil` from us or whether we have a latent bug that this upgrade manifests, it is not yet clear and left out of scope.
Further changes to auth are out of scope for this upgrade.

EDIT: item tracked in the backlog

lib/trento/release.ex
```diff
@@ -215,14 +213,16 @@ defmodule Trento.Release do
...
-    certfile_content = :public_key.pem_encode([{:Certificate, certificate, :not_encrypted}])
+    cert =
+      X509.Certificate.self_signed(key, "/CN=#{name}",
+        extensions: [
+          subject_alt_name: X509.Certificate.Extension.subject_alt_name(hostnames)
+        ]
+      )
 
-    {keyfile_content, certfile_content}
+    {X509.PrivateKey.to_pem(key), X509.Certificate.to_pem(cert)}
```
nelsonkopliku, line 225: The previous way of doing this would fail on OTP >= 28 at the very first `{certificate, private_key} = Cert.certificate_and_key(2048, name, hostnames)` because of incompatibility https://github.com/phoenixframework/phoenix/issues/6319.

Now, I know we are not upgrading to OTP 28, but on 27 and this might not be strictly necessary, however we would encounter it when eventually moving to 28.

To keep the current code working on OTP 28 we'd need to update also phoenix to 1.8, but that is not going to be a walk in the park either because some other dependencies, like pow, add a constraint on phoenix being < 1.8.

This change would be future compatible and it should not affect other behaviours or existing installations.

mix.exs
antgamdia: >  from us or whether we have a latent bug that this upgrade manifests, it is not yet clear and left out of scope

Thanks for the details! We certainly need more coverage on the SSO scenario 😅 
Ok with leaving it as is, but just to double-check, did you file a ticket not to forget about this?

antgamdia (approved):
LGTM, but Elixir is not my area of expertise. Let's wait until others chime in.

Thanks for taking the time and in performing the upgrade! Next time... [1.20](https://elixirforum.com/t/elixir-v1-20-0-rc-6-released/75448) :P haha

nelsonkopliku:
@skrech since you asked: the [last commit](https://github.com/trento-project/web/pull/4342/changes/56f32d7f3819c4d34c88b9f899859e3e543f2785) re-introduces the backward compatibility matrix against elixir 1.15.7 and OTP 26.2.1

.github/workflows/ci.yaml
```diff
@@ -258,8 +258,8 @@ jobs:
         id: setup-elixir
         uses: erlef/setup-beam@fc68ffb90438ef2936bbb3251622353b3dcb2f93 # v1
         with:
-          version-file: .tool-versions
-          version-type: strict
+          otp-version: ${{ needs.setup-matrix-env.outputs.ERLANG_DEV }}
+          elixir-version: ${{ needs.setup-matrix-env.outputs.ELIXIR_DEV }}
```
skrech, line 263: If version-file worked... maybe it's better to leave it as before, not including additional dep?

skrech: Or... for consistency, let's  leave it like that since or all the other places it's like that.

mix.exs
```diff
@@ -106,7 +105,7 @@ defmodule Trento.MixProject do
       {:tzdata, "~> 1.1.3"},
       {:trento_contracts,
        github: "trento-project/contracts",
-       ref: "c5530804a09a2e50381c9298bb9fb6c1a82957e9",
+       ref: "9256294c5fdc3f242f1575ce010c25890e2599b3",
```
skrech: Should we coordinate merge of this PR with the contracts PR? Which one is safe to go first?

lib/trento/software_updates/discovery.ex
```diff
@@ -128,13 +128,16 @@ defmodule Trento.SoftwareUpdates.Discovery do
   end
 
   def discover_host_software_updates(host_id, fully_qualified_domain_name) do
-    with {:ok, system_id} <- get_system_id(fully_qualified_domain_name),
+    get_system_id_result = get_system_id(fully_qualified_domain_name)
+    resolved_system_id = extract_system_id(get_system_id_result)
```
skrech: The naming here is a bit confusing to me, I don't understand what is the difference between resolved_system_id and system_id. Maybe system_id is redundant now?
Resolved system id should be string representations of the system_id, which is a number?

test/support/factory.ex
```diff
@@ -1148,7 +1148,7 @@ defmodule Trento.Factory do
       advisory_name: String.downcase(Faker.Pokemon.name()),
       advisory_type: Faker.Util.pick(AdvisoryType.values()),
       advisory_status: "stable",
-      id: RandomElixir.random_between(2000, 5000),
+      id: sequence(:patch_id, & &1),
```
skrech, line 1151: So, the range for the random number was chosen randomly? Eg. the actual value is not important as long as it's unique?

test/trento/software_updates/discovery_test.exs
```diff
@@ -106,7 +112,7 @@ defmodule Trento.SoftwareUpdates.DiscoveryTest do
 
           assert %DiscoveryResult{
                    host_id: ^host_id,
-                   system_id: nil,
+                   system_id: ^expected_system_id,
```
skrech, line 115: This should ideally be number, right? Not a string. Do you thing it's better to convert it so we don't slip through a potential bug in the future?

skrech (approved):
Great! Thank you @nelsonkopliku ! Let's leave the matrix in the CI for now, until we get all things sorted on the OBS side.

Btw, I see a lot of the changes are adding pattern matching for the type of a specific variable, it this something related to 1.19 or it's more of a tidying effort?

mix.exs
nelsonkopliku: First the contracts, then web and wanda with the new sha generated by the merge

mix.exs
```diff
@@ -118,13 +117,19 @@ defmodule Trento.MixProject do
       # https://github.com/deadtrickster/ssl_verify_fun.erl/pull/27
       {:ssl_verify_fun, "~> 1.1", manager: :rebar3, override: true},
       {:parallel_stream, "~> 1.1.0"},
-      {:x509, "~> 0.8.8"},
+      # {:amqp, "~> 4.1.0"},
```
arbulu89: question:
is this left over intentionally?
I understand that the dependency is added indirectly, and maybe you were playing with it?

.dialyzer_ignore.exs
```diff
@@ -2,5 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
 [
-  ~r/.*domo_generated_code.*/
+  # keeping for reference
+  # Dialyzer :call_without_opaque warnings for MapSet and Ecto.Multi are known false positives
+  # in Elixir 1.19 / Erlang 28. These occur because Dialyzer manages to "peek" into the
```
arbulu89: question:
what bout Erlang 27?

.dialyzer_ignore.exs
```diff
@@ -2,5 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
 [
-  ~r/.*domo_generated_code.*/
+  # keeping for reference
```
arbulu89: question:
is there any reason to put the comment inside the empty list?
maybe put the comment on top and `[]` ?

test/trento/software_updates/discovery_test.exs
skrech: Nevermind. I see that the system_id is a number in the failure_setup func.

arbulu89: More than a number it should be a UUID.
I don't really see what the test does, but yes, having proper typing would help

lib/trento/software_updates/discovery.ex
nelsonkopliku: Let me know what you think about this https://github.com/trento-project/web/pull/4342/changes/e867d6d97b71099a6b8755f257ba8741f2523ad2

`system_id` coming from  `get_system_id(fully_qualified_domain_name)` is supposedly a number - `pos_integer` - however in the database entity at `Trento.SoftwareUpdates.Discovery.DiscoveryResult` is stored as a string. That is why we do `"#{system_id}"` later.

Now, can it be just an integer in `DiscoveryResult` and the database? Probably yes.

I honestly do not remember if there was some constraints about saving it as a string when we (I) implemented that or if it was just a mistake.

I'd not add further changes to this now.

test/support/factory.ex
nelsonkopliku: exactly

lib/trento/software_updates/discovery.ex
skrech: I like the changes, way more clear now. :)
And yes, I don't have complaints about how it's saved in the DB, no problems there.

lib/trento/software_updates/discovery.ex
```diff
@@ -148,28 +151,31 @@ defmodule Trento.SoftwareUpdates.Discovery do
...
   defp discover_host_software_updates(host_id, fully_qualified_domain_name, _),
     do: discover_host_software_updates(host_id, fully_qualified_domain_name)
 
-  defp finalize_failed_discovery(host_id, {:error, reason}) do
+  defp finalize_failed_discovery(host_id, system_id, {:error, reason}) do
     %DiscoveryResult{}
     |> DiscoveryResult.changeset(%{
       host_id: host_id,
-      system_id: nil,
+      system_id: "#{system_id}",
```
arbulu89, line 180: question:
why do we convert this into a string?
we would have `"nil"`, potentially.
Is it not better to leave it as-is?

arbulu89 (commented):
Looks good!
Some minor questions and comments.
PD:
I guess one of the next steps is to remove the matrix usage in the CI to only test with one elixir/erlang version

.dialyzer_ignore.exs
nelsonkopliku: not an issue on erlang 27. That's why the entries are commented out.

If you prefer we can remove them altogether.

arbulu89: Ok, now I understand the thing.
it is fine to keep the files. I would simply move the main comment to the top of the list (just ocd). But it is fine!

nelsonkopliku: Here we go e325b6454a351bc15470b1838ceca5ab99a8bb5c

.dialyzer_ignore.exs
nelsonkopliku: Ditto https://github.com/trento-project/web/pull/4342#discussion_r3317933056

lib/trento/software_updates/discovery.ex
skrech: I guess the idea is to make a best effort to extract the `system_id` even in case of error.
But your point is valid, however maybe inline if/else would solve problem.

nelsonkopliku: Same as here https://github.com/trento-project/web/pull/4342#discussion_r3317833907

> `system_id` coming from `get_system_id(fully_qualified_domain_name)` is supposedly a number - `pos_integer` - however in the database entity at `Trento.SoftwareUpdates.Discovery.DiscoveryResult` is stored as a string. That is why we do `"#{system_id}"` later.
> 
> Now, can it be just an integer in DiscoveryResult and the database? Probably yes.
> 
> I honestly do not remember if there was some constraints about saving it as a string when we (I) implemented that or if it was just a mistake.
> 
> I'd not add further changes to this now.
>

> we would have "nil", potentially.
>
Ecto casts to `nil` correctly. So if `system_id` is `nil`, when we do `"#{system_id}"` it becomes an empty string `""`, however when provided to the changeset it ends up as `nil`

EDIT: luckily tests cover for that case as well

arbulu89: Maybe we can add a short comment with what you explained

nelsonkopliku: Here we go 3d3b185ee79d36f3ded805b1b3abcf1269e496c3

arbulu89 (approved):
Great @nelsonkopliku 
Thank you

mix.exs
nelsonkopliku: Removed here 733a3ed72fb2396462665e47eb1a53f3ac78c97b

nelsonkopliku:
@trento-project/developers last commit 733a3ed72fb2396462665e47eb1a53f3ac78c97b aligns dependencies between web and wanda and introduces a stricter openapispex casting of numbers.

I've made a tiny adjustment to the relevant spec that bring a formal bc break, but after talking with @arbulu89 we realised that it is not worth bumping api versions but keep the breaking change which actually seems an improvement.

test/support/commanded/test_event_handler_with_failure_context.ex
nelsonkopliku: `max_retry` was a typo.

and `skip: true` was actually a bug fixed in https://github.com/commanded/commanded/commit/0488f8805808be74d0d522ac1db5b8819e970d26 (1.4.5)

PR #4346, opened 2026-05-28
Flip activity log `from_date` and `to_date` logic

arbulu89:
# Description

There is a missmatch between the activity log openAPI params `from_date` and `to_date` description and the real logic.
They work in the reverse logic.
It was reported by a user using the API with AI.

Personally, I think the openAPI spec is the one that is easier to understand (usually, `from` is used from older dates and `to` for the most recent ones).

This PR is just a proposal to invert the internal logic to match with the openAPI spec thing.

Besides, I have thought on changing the frontend filter names, as i think it is more obvious (up to debate):
- `newer than` by `Start date`
- `older than` by `End date`

image

This is just a quick proposal. We could do the opposite and change the openAPI spec to match with current logic.

## How was this tested?
Rewrite existing UT and e2e tests.
And additional manual testing.

@abravosuse I'm not sure if we have any piece of docs on this


abravosuse:
| @abravosuse I'm not sure if we have any piece of docs on this
@arbulu89 the documentation about this filter is vague enough that it doesn't require an update:

` The Activity Log also allows you to filter out entries that are newer and/or older than a specific date and time. Date/time inputs in the UI are interpreted according to the user’s preferred timezone; if no preference is set, UTC is assumed.`

Therefore no need to update the documentation on this.

abravosuse (approved):
I think the changes in the UI make it easier for the user to understand the purpose of the filters. Thank you!

abravosuse:
Hey @arbulu89 now that the PR env is available, I have logged in and done a couple of tests. This is probably something that was not contemplated before and maybe it has to be addressed with a separate ticket/PR. But shoudn't the UI prevent us from setting that a `Start date` that is more recent than the `End date` (or the other way around). And regardless whether you use specific dates or time references (1h ago, 24h ago, 7d ago, 30d ago).

arbulu89:
> Hey @arbulu89 now that the PR env is available, I have logged in and done a couple of tests. This is probably something that was not contemplated before and maybe it has to be addressed with a separate ticket/PR. But shoudn't the UI prevent us from setting that a `Start date` that is more recent than the `End date` (or the other way around). And regardless whether you use specific dates or time references (1h ago, 24h ago, 7d ago, 30d ago).

I guess it is possible to do it.
But I don't consider that part of the bugfix. It is kind of a new feature.
Most likely it is not trivial, as those values can come from the query string as well, where the user can put incorrectly ordered values. And creating dependencies between filters doesn't look easy either.

nelsonkopliku (commented):
Hey @arbulu89, thanks for this!

As long as everything stays consistent, I am not against flipping the logic:
- it reflects what documented in the open api spec
- fixes usage with AI agents
- is perhaps more straightforward

Only one thing about the semantics: I honestly find `newer than` and `older than` just spot on because it does not allow for ambiguity.

`Start` and `End`, IMHO open a bit the door for some ambiguity based on how you look at the thing. 

What is `start date`? from this date and older or from this date and newer? (vice versa for `end date`)

The whole flipping issue comes from the append-only-time-based-ordered nature of the activity log and depending on the perspective the meaning flips 😅 

Bottom line: +1 to align the logic with the openapi spec, however I would consider keeping the newer/older than semantics, if possible.

arbulu89:
> Hey @arbulu89, thanks for this!
> 
> As long as everything stays consistent, I am not against flipping the logic:
> 
>     * it reflects what documented in the open api spec
> 
>     * fixes usage with AI agents
> 
>     * is perhaps more straightforward
> 
> 
> Only one thing about the semantics: I honestly find `newer than` and `older than` just spot on because it does not allow for ambiguity.
> 
> `Start` and `End`, IMHO open a bit the door for some ambiguity based on how you look at the thing.
> 
> What is `start date`? from this date and older or from this date and newer? (vice versa for `end date`)
> 
> The whole flipping issue comes from the append-only-time-based-ordered nature of the activity log and depending on the perspective the meaning flips 😅
> 
> Bottom line: +1 to align the logic with the openapi spec, however I would consider keeping the newer/older than semantics, if possible.

I'm fine with either option.
What do you think @abravosuse and @jagabomb ?

Edit:
We could even rename them to `Filter From date` and `Filter To date`.

abravosuse:
@arbulu89 @nelsonkopliku I am personally fine with both approaches. But considering how date/time filters normally work in an SAP context, I think that the "Start date" / "End date" approach would come more naturally to SAP admins. Have a look, for example, on how these fields are presented in the screen that gives access the system log in a SAP system:

image

nelsonkopliku (approved):
@abravosuse if that works better for you, then ok for me 👍

arbulu89:
@abravosuse Seeing that, what if put put `Filter From date` and `Filter To date`?
Maybe even better, right?

abravosuse:
> @abravosuse Seeing that, what if put put `Filter From date` and `Filter To date`? Maybe even better, right?

Yes, even better.

arbulu89:
Alright, last version:
image

PR #4352, opened 2026-06-01
Split cluster discovered health events

arbulu89:
# Description

Split `ClusterDiscoveredHealthChanged` event into specific events based on the real cause of the health change:
`ClusterReplicationHealthChanged` and `ClusterDistributedHealthChanged`.

I have done some relevant changes:
- Move the calculation of the health to the aggregate. I think this is a product and business logic and I find it more appropriate to be in the aggregate
- Upcasting of events
- In one of the latest commits, I have experimented with the option of putting all the health details in a value object struct. Maybe it helps reducing the number of new fields we need in the aggregate itself and have a more hierarchical structure for the healths. Before this commit, I just simply added the healths in the aggregate itself. This is the commit: https://github.com/trento-project/web/commit/85790f8e3f2abf22c30bbc2941490fede7170fc1

Potential improvements for next PRs:
- We could add additional fields to the new events. For example, for the `ClusterReplicationHealthChanged` event, we could add the `sr_health` and `replication_status` fields. They would be just informative fields showed in the activity log.

Find here the text used for the activity log:
image

## How was this tested?

UT which should cover all the updates, and I did some additional manual testing just in case

No need

### DIFF
--- a/lib/trento/clusters/cluster.ex
+++ b/lib/trento/clusters/cluster.ex
@@ -78,8 +78,9 @@ defmodule Trento.Clusters.Cluster do
 
   alias Trento.Clusters.ValueObjects.{
     AscsErsClusterDetails,
+    AscsErsClusterHealthDetails,
     HanaClusterDetails,
-    HealthDetails,
+    HanaClusterHealthDetails,
     SapInstance
   }
 
@@ -136,7 +137,6 @@ defmodule Trento.Clusters.Cluster do
     field :hosts_number, :integer
     field :provider, Ecto.Enum, values: Provider.values()
     field :health, Ecto.Enum, values: Health.values(), default: Health.unknown()
-    embeds_one :health_details, HealthDetails, defaults_to_struct: true, on_replace: :update
     field :state, Ecto.Enum, values: ClusterState.values()
     field :hosts, {:array, :string}, default: []
     field :offline_hosts, {:array, :string}, default: []
@@ -153,6 +153,17 @@ defmodule Trento.Clusters.Cluster do
         ascs_ers: [module: AscsErsClusterDetails, identify_by_fields: [:sap_systems]]
       ],
       on_replace: :update
+    )
+
+    polymorphic_embeds_one(:health_details,
+      types: [
+        hana_scale_up: HanaClusterHealthDetails,
+        hana_scale_out: HanaClusterHealthDetails,
+        ascs_ers: AscsErsClusterHealthDetails
+      ],
+      on_replace: :update,
+      use_parent_field_for_type: :type,
+      on_type_not_found: :nilify
     )
 
     embeds_many :sap_instances, SapInstance
@@ -179,7 +190,7 @@ defmodule Trento.Clusters.Cluster do
         }
       ) do
     health_details = parse_discovered_health(details)
-    health = compute_discovered_health(type, health_details)
+    health = compute_discovered_health(health_details)
 
     [
       %ClusterRegistered{
@@ -222,7 +233,7 @@ defmodule Trento.Clusters.Cluster do
         hosts_number: nil,
         details: nil,
         health: Health.unknown(),
-        health_details: %HealthDetails{},
+        health_details: nil,
         state: ClusterState.unknown()
       },
       %HostAddedToCluster{
@@ -251,7 +262,7 @@ defmodule Trento.Clusters.Cluster do
         hosts_number: nil,
         details: nil,
         health: Health.unknown(),
-        health_details: %HealthDetails{},
+        health_details: nil,
         state: ClusterState.unknown()
       },
       %HostAddedToCluster{
@@ -366,14 +377,22 @@ defmodule Trento.Clusters.Cluster do
       maybe_emit_cluster_details_updated_event(cluster, command)
     end)
     |> Multi.execute(fn cluster ->
-      maybe_emit_cluster_replication_health_changed_event(cluster, command, %HealthDetails{
-        replication_health: Health.unknown()
-      })
+      maybe_emit_cluster_replication_health_changed_event(
+        cluster,
+        command,
+        %HanaClusterHealthDetails{
+          replication_health: Health.unknown()
+        }
+      )
     end)
     |> Multi.execute(fn cluster ->
-      maybe_emit_cluster_distributed_health_changed_event(cluster, command, %HealthDetails{
-        distributed_health: Health.unknown()
-      })
+      maybe_emit_cluster_distributed_health_changed_event(
+        cluster,
+        command,
+        %AscsErsClusterHealthDetails{
+          distributed_health: Health.unknown()
+        }
+      )
     end)
     |> Multi.execute(&maybe_emit_cluster_health_changed_event/1)
   end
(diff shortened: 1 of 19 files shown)


test/trento/clusters/cluster_test.exs
```diff
@@ -95,13 +100,181 @@ defmodule Trento.ClusterTest do
...
-          discovered_health: :passing,
-          health: :passing,
+          health: Health.unknown(),
+          health_details: %HealthDetails{},
           state: state
         }
       )
     end
 
+    test "should register a cluster with correct health for HANA type clusters" do
```
arbulu89, line 113: comment:
we were not even testing the correct parsing of health before in the policy

lib/trento/clusters/cluster.ex
```diff
@@ -727,10 +821,6 @@ defmodule Trento.Clusters.Cluster do
         hosts_number: hosts_number,
         state: ClusterState.stopped(),
         details: details
-      },
-      %ClusterDiscoveredHealthChanged{
```
arbulu89, line 731: This event emission replaced above with the next snippet:
```
|> Multi.execute(fn cluster ->
      maybe_emit_cluster_replication_health_changed_event(cluster, command, %HealthDetails{
        replication_health: Health.unknown()
      })
    end)
    |> Multi.execute(fn cluster ->
      maybe_emit_cluster_distributed_health_changed_event(cluster, command, %HealthDetails{
        distributed_health: Health.unknown()
      })
```

lib/trento/clusters/events/cluster_distributed_health_changed.ex
```diff
@@ -0,0 +1,18 @@
...
+
+defmodule Trento.Clusters.Events.ClusterDistributedHealthChanged do
+  @moduledoc """
+  This event is emitted when the distribution of ASCS/ERS nodes in a cluster changes.
+  Only applicable for ASCS/ERS clusters.
+  """
+
+  use Trento.Support.Event
+
+  require Trento.Enums.Health, as: Health
```
skrech, line 12: Minor: do we need the `as:`, the name is the same.

lib/trento/clusters/cluster.ex
```diff
@@ -35,16 +35,23 @@ defmodule Trento.Clusters.Cluster do
...
   It is composed by sub-health elements:
 
-  - Discovered health
+  - Replication health (only applicable for HANA clusters)
+  - Distributed health (only applicable for ASCS/ERS clusters)
   - Checks health
 
-  The main cluster health is computed using the values from these two. This means that the cluster health is the
+  The main cluster health is computed using the values from all of them. This means that the cluster health is the
   worst of the two.
```
skrech: Minor: "worst of the two" might need correction.

lib/trento/clusters/events/cluster_replication_health_changed.ex
```diff
@@ -0,0 +1,18 @@
...
+
+defmodule Trento.Clusters.Events.ClusterReplicationHealthChanged do
+  @moduledoc """
+  This event is emitted when the replication health of a cluster changes.
+  Only applicable for HANA clusters.
+  """
+
+  use Trento.Support.Event
+
+  require Trento.Enums.Health, as: Health
```
skrech, line 12: Minor: same as for the distributed health.

lib/trento/clusters/cluster.ex
```diff
@@ -937,33 +992,34 @@ defmodule Trento.Clusters.Cluster do
...
-  defp maybe_add_distributed_health(healths, _, _), do: healths
+  defp add_primary_health(%AscsErsClusterHealthDetails{
+         distributed_health: distributed_health
+       }),
+       do: [distributed_health]
 
   defp maybe_add_checks_health(healths, Health.unknown()), do: healths
   defp maybe_add_checks_health(healths, checks_health), do: [checks_health | healths]
 
   defp parse_discovered_health(%HanaClusterDetails{} = details) do
```
skrech: Note: The operation that we're doing in these functions is not parsing. I think, a better name is "derive_discovered_health".

lib/trento/clusters/cluster.ex
```diff
@@ -916,18 +974,15 @@ defmodule Trento.Clusters.Cluster do
...
+             checks_health: checks_health
+           } = health_details,
+         health: health
        }) do
     new_health =
-      []
-      |> maybe_add_replication_health(replication_health, cluster_type)
-      |> maybe_add_distributed_health(distributed_health, cluster_type)
+      health_details
+      |> add_primary_health()
```
skrech, line 1038: Maybe "extract_primary_health" is better name?

lib/trento/clusters/cluster.ex
```diff
@@ -916,18 +974,15 @@ defmodule Trento.Clusters.Cluster do
...
       |> maybe_add_checks_health(checks_health)
```
skrech, line 1039: Here, leaving it as "add" make sense to me.

lib/trento/clusters/cluster.ex
```diff
@@ -986,16 +1042,15 @@ defmodule Trento.Clusters.Cluster do
 
   defp parse_distributed_health(_), do: Health.unknown()
 
-  defp compute_discovered_health(cluster_type, %HealthDetails{
+  defp compute_discovered_health(%HanaClusterHealthDetails{
```
skrech: I think that `compute_discovered_health` could wrap the `derive_discovered_health` since having two functions with very similar names is a bit confusing, and it seems that to call compute_discovered_health, you always have to call `derive_discovered_health'.

However, I'm not entirely sure on this, so up to you.

skrech (commented):
I'm continuing tomorrow.

lib/trento/clusters/cluster.ex
arbulu89: Already updated

lib/trento/clusters/events/cluster_distributed_health_changed.ex
arbulu89: Yes, we need it.
Otherwise we would we forced to use the full path all the time -> `Trento.Enums.Health.values()` etc.
Or alias on top first and require later.

lib/trento/clusters/cluster.ex
arbulu89: But we need the result of both in the `ClusterRegistered` part, as both are fields needed to emit the `ClusterRegistered` event.

```
health_details = parse_discovered_health(details)
health = compute_discovered_health(health_details)
...
%ClusterRegistered{
        ...
        health: health,
        health_details: health_details,
       ...
      },
```

lib/trento/clusters/cluster.ex
arbulu89: Yes, I think you are right. `derive_` looks more appropriate for this case.
I think we use `parse_` all over the code, maybe without really thinking that much about "semantics".

arbulu89: Updated

lib/trento/clusters/events/cluster_replication_health_changed.ex
arbulu89: Answered above

skrech (approved):
Great work, @arbulu89! Thanks!

PR #4355, opened 2026-06-04
Add database details sap system details view

arbulu89:
# Description

Add database information in the SAP systems view.
The health is updated automatically when it is changed in the backend.
Besides this, the `Type` value in SAP system details view is changed.
Now, instead of being a static `Application Server`, it will use the same type we have in the overview, which is `ABAP`, `JAVA` or `ABAP+JAVA`.

image

## How was this tested?

UT and e2e

### DIFF
--- a/assets/js/common/ListView/ListView.jsx
+++ b/assets/js/common/ListView/ListView.jsx
@@ -21,12 +21,13 @@ function ListView({
       {data.map(
         ({
           title,
+          key,
           content,
           className: contentClassName = '',
           render = (component) => <span>{component}</span>,
         }) => (
           <div
-            key={title}
+            key={key || title}
             className={
               orientation === 'vertical'
                 ? 'grid grid-flow-row'
--- a/assets/js/lib/model/sapSystems.js
+++ b/assets/js/lib/model/sapSystems.js
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-License-Identifier: Apache-2.0
 
+import { uniq, flatMap } from 'lodash';
+
 export const DATABASE_TYPE = 'database';
 export const APPLICATION_TYPE = 'application';
 
@@ -15,3 +17,10 @@ export const isValidEnsaVersion = (ensaVersion) =>
 
 export const getEnsaVersionLabel = (ensaVersion) =>
   ensaVersion === NO_ENSA ? '-' : ensaVersion.toUpperCase();
+
+export const getSapSystemType = (applicationInstances) =>
+  uniq(flatMap(applicationInstances, ({ features }) => features.split('|')))
+    .filter((item) => item === 'J2EE' || item === 'ABAP')
+    .map((item) => (item === 'J2EE' ? 'JAVA' : item))
+    .toSorted()
+    .join('+');
--- a/assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
+++ b/assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
@@ -42,6 +42,7 @@ import {
   APPLICATION_TYPE,
   DATABASE_TYPE,
   getEnsaVersionLabel,
+  getSapSystemType,
 } from '@lib/model/sapSystems';
 import { isSomeHostHeartbeatPassing } from '@lib/model/hosts';
 
@@ -67,8 +68,8 @@ import {
 
 const SR_INACTIVE = 'INACTIVE';
 
-const renderType = (t) =>
-  t === APPLICATION_TYPE ? 'Application server' : 'HANA Database';
+const renderType = (t, system) =>
+  t === APPLICATION_TYPE ? getSapSystemType(system.instances) : 'HANA Database';
 
 const getUniqueHosts = (hosts) =>
   Array.from(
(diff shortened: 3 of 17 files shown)


assets/js/common/ListView/ListView.jsx
```diff
@@ -21,12 +21,13 @@ function ListView({
       {data.map(
         ({
           title,
+          key,
```
arbulu89, line 24: This change is required in order to give the option to set a key that is not a title.
For example, in our case, the application/database icon on the right doesn't have a title.
So we can give other key to avoid the `key must be unique` react warning

assets/js/lib/model/sapSystems.js
```diff
@@ -15,3 +17,10 @@ export const isValidEnsaVersion = (ensaVersion) =>
 
 export const getEnsaVersionLabel = (ensaVersion) =>
   ensaVersion === NO_ENSA ? '-' : ensaVersion.toUpperCase();
+
+export const getSapSystemType = (applicationInstances) =>
```
arbulu89, line 21: Copied here from `SapSystemsOverview` code

assets/js/pages/SapSystemDetails/GenericSystemDetails.jsx
```diff
@@ -304,55 +304,97 @@ export function GenericSystemDetails({
...
-                  },
-                ]
-              : []),
-            {
-              title: '',
-              content: type,
-              render: (content) => (
-                <div className="justify-end float-right">
-                  {content === APPLICATION_TYPE ? (
+        {type === APPLICATION_TYPE ? (
```
arbulu89, line 307: I have changed the logic of how items in the list are created.
Before, we have 1 `ListView` and the items were added conditionally for each type.
Now, I have created 2 `ListView`, so there is only one `if` condition.
Besides, it allows us to change the class name `grid-rows-2`.
Otherwise, the colocation of elements couldn't work without modifying `ListView` itself

abravosuse (approved):
I have updated rolling from main and deployed the image of this PR on it. Then I have played around starting/stopping dbs and watching the transitions from the corresponding SAP system details view. It works as I expected. It's useful to have this information about the db in that view. I like also the change of the Type field to display the stack type. Thank you!

lib/trento/sap_systems/projections/sap_system_projector.ex
```diff
@@ -266,6 +267,13 @@ defmodule Trento.SapSystems.Projections.SapSystemProjector do
     end
   )
 
+  # Do nothing in the projection.
+  # This event is only here to broadcast the new health in the after_update
+  project(
+    %SapSystemDatabaseHealthChanged{},
+    fn multi -> multi end
+  )
+
```
skrech: Why are we using this combination of `project/after _update`? Does Commanded provide more appropriate means to achieve this "effector/reactor" type of behaviour. I think this is a good candidate for Event Handler definition?

arbulu89: Well, the projector is an event handler.
So far, the 99% of our broadcasts happen in projections, that's why I added it there. It is less code, the testing files are there, and I thought it convenient. 
Now, yes, it can go in an event handler with the exact same result. I thought in the event handler option as well, but it looked an "overkill" just for this.
Do you prefer me to change? I don't have a strong opinion

arbulu89: PD:
I could add the event handling in the `SapSystemDatabaseHealthEventHandler` event handler.
It now handles the dispatch of the command that later on causes to emit the event we are talking about.
This way, we don't need to create a new handler of the same topic.
What do you think?

skrech: I'm not entirely familiar with that part of the code, but re-using an event handler sounds good to me if it's possible. 
By reading the documentation, it seems that after_update callback runs in the same DB transaction. However, we have a broadcast event that is not blocking... so... maybe for consistency with the rest of the codebase it OK to make an "exception" and let that logic live is in the same file. So, up to you.

test/e2e/cypress/pageObject/sap_system_details_po.js
```diff
@@ -45,15 +52,31 @@ export const visitNonExistentSapSystem = () =>
 export const validatePageUrl = (systemId = selectedSystem.Id) =>
   basePage.validateUrl(`/sap_systems/${systemId}`);
 
-export const pageTitleHealthIsCorrectlyDisplayed = () =>
-  basePage.pageTitleHealthIsCorrectlyDisplayed(selectedSystem.Health);
+export const pageTitleHealthIsCorrectlyDisplayed = (
+  health = selectedSystem.Health
```
skrech, line 56: Just a comment: I'm personally not fan of putting default values on these kind of functions (assertion functions). If it was me, I'd make the health explicit every time, because calling pageTitleHealthIsCorrectlyDisplayed what is the correct health?

lib/trento/sap_systems/projections/sap_system_projector.ex
arbulu89: Alright!
I have moved the event handling from the project to the event handler.
I'm absolutely fine with this.
It looks a bit more "pure" as we don't have a real projection (and thus the "empty" usage).
I like it.

skrech (approved):
Everything seems great, thanks @arbulu89

test/e2e/cypress/pageObject/sap_system_details_po.js
arbulu89: I get your point. It is a practice we have all over the place, I guess to make the tests more readable and "hide" the details in the page objects.

PR #4362, opened 2026-06-08
Consider SBD health in aggregated cluster health

skrech:
# Description

Made SBD health participate in aggregated health of the cluster. SBD health is only considered when there are actual SBD devices connected to  the nodes. The front-end code is only augmented in its Activity Log area, where a new event is added.

In the process of implementing the SBD health, a re-factoring was made to the health aggregation handling.

Additionally, the test-data validation is tightened, where this was easy to do.

## How was this tested?

Added new BE UT and E2E tests.


lib/trento/clusters/value_objects/ascs_ers_cluster_health_details.ex
```diff
@@ -9,14 +9,15 @@ defmodule Trento.Clusters.ValueObjects.AscsErsClusterHealthDetails do
   aggregate docstring.
   """
 
-  @required_fields []
+  @required_fields [:sbd_health]
```
arbulu89: comment:
We need to be really careful putting `required` fields.
Just be aware that we load the events from the database, and if we load an event that doesn't have this field, it will potentially raise an error.
And having the value with `nil` wouldn't work.
I don't know if upcasting is enough, or the check would fail before this.
Something to be really careful with

lib/trento/clusters/value_objects/sbd_device.ex
```diff
@@ -7,11 +7,17 @@ defmodule Trento.Clusters.ValueObjects.SbdDevice do
   """
 
   @required_fields :all
+  @allowed_statuses ["healthy", "unhealthy"]
```
arbulu89: question:
what about making the `status` a proper enum?
This way, invalid values are discarded.
Again, we need to be careful with this, as this is a breaking change.
I would check the agent to see what values we can send in this field.

test/support/factory.ex
```diff
@@ -737,11 +738,16 @@ defmodule Trento.Factory do
...
-  def sbd_device_factory do
-    %SbdDevice{
+  def sbd_device_factory(attrs) do
+    sbd_device = %{
       device: Faker.File.file_name(),
-      status: Enum.random(["healthy", "unhealthy"])
+      status: "healthy"
     }
+
+    sbd_device
```
arbulu89, line 752: question:
why this changed?
what's the purpose of them?
I think the initial code looks a more standard factory

arbulu89 (commented):
Alright!
some comments about potential "dangers" we should consider.

lib/trento/clusters/events/cluster_registered.ex
```diff
@@ -66,4 +66,10 @@ defmodule Trento.Clusters.Events.ClusterRegistered do
     do: Map.put(params, "health_details", %{"distributed_health" => health})
 
   def upcast(params, _, 3), do: Map.put(params, "health_details", nil)
+
+  # Version 4 -- Adds "sbd_health" to "health_details" key if it exists
+  def upcast(%{"health_details" => %{}} = params, _, 4),
```
arbulu89, line 70: nit:
maybe a more elixir kind of code:

```
def upcast(%{"health_details" => nil} = params, _, 4), do: params
def upcast(params, _, 4), do: 
   put_in(params, ["health_details", "sbd_health"], Health.unknown()) 
def upcast(params, _, 4), do: params
```

lib/trento/clusters/events/cluster_rolled_up.ex
```diff
@@ -72,4 +74,23 @@ defmodule Trento.Clusters.Events.ClusterRolledUp do
...
+  def upcast(
+        %{
+          "snapshot" => %{
+            "health_details" => %{}
+          }
+        } = params,
+        _,
+        3
+      ) do
+    update_in(
```
arbulu89: nit:
why not `put_in(params, ["snapshot", "health_details", "sbd_health"], Health.unknown())` ?

lib/trento/clusters/value_objects/ascs_ers_cluster_health_details.ex
arbulu89: I guess we can add the `distributed_health` in the required fields (and `replication_health` in the other object)

lib/trento/clusters/value_objects/sbd_device.ex
```diff
@@ -10,8 +10,10 @@ defmodule Trento.Clusters.ValueObjects.SbdDevice do
 
   use Trento.Support.Type
 
+  require Trento.Clusters.Enums.SbdDeviceStatus, as: SbdDeviceStatus
+
   deftype do
     field :device, :string
-    field :status, :string
+    field :status, Ecto.Enum, values: SbdDeviceStatus.values()
```
arbulu89, line 17: question:
do we need to do some upcasting after this change?
I think not, but asking just in case

lib/trento_web/openapi/v1/schema/cluster.ex
```diff
@@ -107,7 +107,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.Cluster do
         type: :object,
         properties: %{
           device: %Schema{type: :string},
-          status: %Schema{type: :string}
+          status: %Schema{type: :string, enum: [:healthy, :unhealthy]}
```
arbulu89: nit:
we can use `SbdDevice.values()` from the created enum here as well

arbulu89 (commented):
Looking really good @skrech

lib/trento/clusters/events/cluster_registered.ex
skrech: You meant without the last line, right?

lib/trento/clusters/value_objects/sbd_device.ex
skrech: Hm... I don't think so. It should work out of the box, IMO.

lib/trento/clusters/events/cluster_registered.ex
skrech: Now that I think of it... what is the difference if I'll check for nil or the other way around. Checking for a map gives better protection that later the put would succeed. 

I'll definitely replace the update_in with put_in, though, I agree about that.

lib/trento/clusters/value_objects/ascs_ers_cluster_health_details.ex
skrech: Added distributed_health and replicated_health as required fields. All the test have passed, so it should be safe :)

lib/trento/clusters/value_objects/sbd_device.ex
skrech: Done.

lib/trento/clusters/value_objects/ascs_ers_cluster_health_details.ex
skrech: I'm resolving this -- For now it looks that I've taken care of the upcasts needed (on the ClusterRegistered and ClusterRollUp) and test seem to suggest that it's fine.

lib/trento/clusters/events/cluster_registered.ex
```diff
@@ -66,4 +66,10 @@ defmodule Trento.Clusters.Events.ClusterRegistered do
     do: Map.put(params, "health_details", %{"distributed_health" => health})
 
   def upcast(params, _, 3), do: Map.put(params, "health_details", nil)
+
+  # Version 4 -- Adds "sbd_health" to "health_details" key if it exists
```
arbulu89: nit:
I personally wouldn't put these comments because I think you are describing elixir coding.
The piece of code is so declarative that it is code -> human translation.
It doesn't document any reason or real logic.
My 2 cents

lib/trento/clusters/cluster.ex
```diff
@@ -766,17 +777,16 @@ defmodule Trento.Clusters.Cluster do
...
-       do: replication_health
-
-  defp compute_discovered_health(%AscsErsClusterHealthDetails{
-         distributed_health: distributed_health
-       }),
-       do: distributed_health
+  defp aggregate_health_details(nil), do: Health.unknown()
 
-  defp compute_discovered_health(_), do: Health.unknown()
+  defp aggregate_health_details(health_details) when is_struct(health_details) do
```
arbulu89, line 791: comment:
2 things here:
The first is a nit: I don't think the `is_struct` is needed. We cannot have anything else, and if we had this would raise in a not controlled error.

And the more important thing:
I think we need to re-implement this computation logic.
It will always use sbd health to compute the overall health, and we don't want that.
Remember that sbd is an optional feature, many clusters won't have it, and this means it shouldn't be used for computation.
I think we need to follow the same strategy than with checks, and it it is unknown don't use it in the computation.
Otherwise, all clusters without sbd will have unknown health.

test/trento/clusters/cluster_test.exs
```diff
@@ -54,6 +55,7 @@ defmodule Trento.ClusterTest do
 
   alias Trento.Clusters.Cluster
 
+  require Trento.Clusters.Enums.SbdDeviceStatus, as: SbdDeviceStatus
```
arbulu89: dudep:
Not your code, but I'm seeing that `require Trento.Enums.Health, as: Health` is done twice.
We could remove this one and put the new addition on top, with the other `require` clauses

test/trento/clusters/cluster_test.exs
```diff
@@ -269,15 +277,115 @@ defmodule Trento.ClusterTest do
...
+        expected_health: Health.passing()
+      },
+      %{
+        sbd_status: SbdDeviceStatus.unhealthy(),
+        expected_health: Health.critical()
+      }
+    ]
+    for cluster_conf <- @cluster_confs,
+        health_conf <- @health_confs do
+      @tag :wip
```
arbulu89: question:
i guess eventually we will remove the `wip` tag, right?
maybe add a `TODO` comment to not forget?

test/trento/discovery/policies/cluster_policy_test.exs
```diff
@@ -215,14 +215,14 @@ defmodule Trento.Discovery.Policies.ClusterPolicyTest do
                     }
                   ],
                   sbd_devices: [
-                    %SbdDevice{
+                    SbdDevice.new!(%{
                       device: "/dev/vdc",
-                      status: "healthy"
-                    },
-                    %SbdDevice{
+                      status: :healthy
```
arbulu89: nit:
maybe we could use `SbdStatus.healthy()` instead of the plain atom.

arbulu89 (commented):
Alright!
Only a one important comment about how we aggregate the overall health in the new code.
I think we need to re-evaluate it.
Besides that, only some nits that most of them are minor/unimportant things

lib/trento/clusters/cluster.ex
skrech: Now, that changes a lot of things :D 
Yea, if the sbd is optional, then I need to make it optional indeed. And add a function to drop it if it's missing. Good catch. We'd need a test for that case as well.

lib/trento/clusters/cluster.ex
```diff
@@ -318,7 +319,12 @@ defmodule Trento.Clusters.Cluster do
         cluster_host_status: ClusterHostStatus.online()
       }
     end)
-    |> maybe_update_cluster(command)
+    |> Multi.execute(fn cluster ->
```
arbulu89: comment:
yes, in this case now it is duped and not needed.
I guess before as the code was reused it was needed

lib/trento/clusters/cluster.ex
```diff
@@ -810,23 +812,68 @@ defmodule Trento.Clusters.Cluster do
     end
   end
 
-  defp maybe_update_cluster(
-         multi,
-         %RegisterOnlineClusterHost{host_id: host_id, details: details} = command
-       ) do
+  defp accumulate_cluster_health_events(cluster, health_details) do
+    health_details_emitters = [
```
arbulu89: nit/personal comment:
for me personally, the intention to try to make this with less lines of code makes it more difficult to read and understand.
Anyway, it is just my opinion

lib/trento/clusters/cluster.ex
```diff
@@ -810,23 +812,68 @@ defmodule Trento.Clusters.Cluster do
...
+
+  defp maybe_emit_cluster_health_details_events(
+         %Cluster{state: state},
+         %RegisterOfflineClusterHost{}
+       )
+       when state != ClusterState.stopped(),
+       do: nil
+
+  defp maybe_emit_cluster_health_details_events(
+         %Cluster{state: state, type: cluster_type} = cluster,
```
arbulu89: nit:
let's put `ClusterState.stopped()` in the pattern matching above: `%Cluster{state: ClusterState.stopped(), type: cluster_type}` and not in the guard.

lib/trento/clusters/cluster.ex
```diff
@@ -810,23 +812,68 @@ defmodule Trento.Clusters.Cluster do
...
+         %RegisterOfflineClusterHost{}
+       )
+       when state == ClusterState.stopped() do
+    health_details =
+      case cluster_type do
```
arbulu89: question/nit:
what do you think splitting this call in different functions instead of using `case`?
the `accumulate_cluster_health_events` would be repeated, but i think the more declarative way of writing the code is a win:
```
  defp maybe_emit_cluster_health_details_events(
         %Cluster{state: ClusterState.stopped(), type: ClusterType.ascs_ers()} = cluster,
           %RegisterOfflineClusterHost{}
       ) do

    health_details = %AscsErsClusterHealthDetails{
            sbd_health: Health.unknown(),
            distributed_health: Health.unknown()
          }

    accumulate_cluster_health_events(cluster, health_details)
end

  defp maybe_emit_cluster_health_details_events(
         %Cluster{state: ClusterState.stopped(), type: cluster_type} = cluster,
           %RegisterOfflineClusterHost{}
       ) when cluster_type in [ClusterType.hana_scale_out(), ClusterType.hana_scale_up()] do

    health_details = %HanaClusterHealthDetails{
            sbd_health: Health.unknown(),
            replication_health: Health.unknown()
          }

    accumulate_cluster_health_events(cluster, health_details)
end

defp maybe_emit_cluster_health_details_events(
         %Cluster{state: ClusterState.stopped()} ,
           _
       ), do: nil
```

Otherwise, in fact, we could end up sending `nil` as `health_details` to `accumulate_cluster_health_events`, which doesn't look a good idea (even thought it might be handled)

lib/trento/clusters/cluster.ex
```diff
@@ -810,23 +812,68 @@ defmodule Trento.Clusters.Cluster do
...
+        _ ->
+          nil
+      end
+
+    accumulate_cluster_health_events(cluster, health_details)
+  end
+
+  defp maybe_emit_cluster_health_details_events(
+         %Cluster{cluster_id: cluster_id} = cluster,
+         %RegisterOnlineClusterHost{designated_controller: true, details: details}
```
arbulu89: comment:
this `designated_controller: true` and `when not is_nil(cluster_id)` doesn't smell well (specially the 2nd).
The `RegisterOnlineClusterHost` command shouldn't be reaching this piece of code if it is not `DC`.
Am I missing something?

arbulu89 (commented):
Alright!
Some more comments, most of then nitpicks

lib/trento/clusters/cluster.ex
skrech: What do you think of preserving the `maybe_emit_host_added_to_cluster_event` function, and not the explicit event, eg. remove the previous execute?

arbulu89: I think it is worst.
Because this case is not a `maybe`. the event must be emitted yes or yes.
I would remove the `maybe` entry. It is just a leftover after reorganizing code

lib/trento/clusters/cluster.ex
skrech: Yes, I like the idea very much! 
I don't think I need to handle the last case of ClusterState.stopped() and unknown cluster_type, it would be handled by the catch-all at the end,  what do you think?

arbulu89: You are right. The last entry would have the same result as the last catch-all entry.
So we don't need it.

lib/trento/clusters/cluster.ex
skrech: This is actually a big thing... :)
I think that it's good that we protect `handle_cluster_health_events`. Thus, we make it independent piece of code that we can rely on, without having prior knowledge of where that function was called.
It's kind of secondary level aggregate `execute` function -- it receives the same arguments the cluster and the command and the multi. However, if we want to make `handle_cluster_health_events`  independent, then we have to accept some duplication in the pattern matching in the `execute`  functions.
I'm looking at this like the following:
1. execute functions would pattern-match on their interested events, do some work and then handle processing of health to the health entry point `handle_cluster_health_events`.  Eg. execute functions handle event-specific operations that are not related to health and then act like triggers for `handle_cluster_health_events`.
2. `handle_cluster_health_events` could be executed from any `execute` function, and it would check it's conditions on its own.

lib/trento/clusters/cluster.ex
skrech: For me the purpose of this re-factor was not to have less lines if code, but to detach the command argument from `maybe_emit_*_health_event` functions. I perceive this as tight coupling, where it seems that it's not needed.

arbulu89: I won't counter-argument the whole point. I'm just talking about this specific function implementation.
For me, the usage of this `reduce` and list of function names is less readable than simply running the functions.

skrech: Hm... but how can I do it in different way? You suggest to just call them one after the other? 

I've done it like that so I can handle the output combination into the reducing function (the appending of the result to a list, by the book usage of reduce :D ). I'm open to suggestions.

lib/trento/clusters/cluster.ex
arbulu89: Ok, I not going to oppose to the change.

But......... XD
For me it continues being a smell. 
The code should not enter here if it is not a designated controller. This is a private function, created for our own needs.
I think we shouldn't protect things we don't want to happen. It opens the door for unwanted code slipping away in the catch-all function.
Having less and independent private functions is not the goal in my opinion

lib/trento/clusters/cluster.ex
skrech: OK, will remove the maybe function, then :)

lib/trento/clusters/cluster.ex
arbulu89: What about simply this?
```
    events = [
      maybe_emit_cluster_replication_health_changed_event(cluster, health_details),
      maybe_emit_cluster_distributed_health_changed_event(cluster, health_details)
      # maybe_emit_cluster_sbd_health_changed_event(cluster, health_details)
    ]
    
    Enum.reject(events, &is_nil/1)
```

lib/trento/clusters/cluster.ex
skrech: I'll loosen the pattern matching. Please note that I have some reservations about this for the future. :D

arbulu89: :see_no_evil:

arbulu89 (commented):
Alright!
If the tests pass and the emitted events are the same, I like the refactor!
kudos

arbulu89 (approved):
Image

PR #4363, opened 2026-06-08
Fix loading checks execution detail when catalog is still loading

nelsonkopliku:
# Description

While working on pulling wanda's capabilities as AI tools I encountered this bug: host checks execution result detail page crashes on refresh.

How to reproduce:
- run a checks execution for a host
- navigate to check results page for that host
- navigate to any check row, expand targets and go to CheckResultDetailPage (`/hosts/:hostId/executions/last/:checkID/host/:hostname`)
- refresh the page and an error should be shown

Reference on demo https://demo.trento-project.io/hosts/2b64f956-a49c-4652-ae03-3c24b570eb17/executions/last/3A361F/host/hana1

## How was this tested?

Added new test and tested manually

### DIFF
--- a/assets/js/pages/ExecutionResults/CheckResultDetail/CheckResultDetailPage.jsx
+++ b/assets/js/pages/ExecutionResults/CheckResultDetail/CheckResultDetailPage.jsx
@@ -127,7 +127,12 @@ function CheckResultDetailPage({ targetType }) {
     );
   }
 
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):
Nice!
Maybe we could add it in the 3.1.1 as well as it is a bug.
If we want that, mark it with the milestone

PR #4365, opened 2026-06-09
Replace SAP and DB instances health by status

arbulu89:
# Description

Replace SAP system and database instances `health` field by `status` field. This way, we show real instance status.
The intention behind is to use the status as it is, the running status, which says in which running state the instance is. Previous health values didn't match properly.

The new mapping is simple:
- `passing` -> `green`
- `warning` -> `yellow`
- `critical` -> `red`
- `unknown` -> `gray`

Such a small changes requires a massive code changes, as we need to change the events, aggregate, read models, etc

**The API backward compatibility CI check fails, as it compares strict values. However, this is the "standard" approach to deprecate a reference, and it is in fact backward compatible**

**I'm leaving the frontend part outside. The current code should be backward compatible.**

## How was this tested?

UT. I have done some manual testing with the rollup feature just in case

No need


lib/trento/sap_systems/events/sap_system_rolled_up.ex
```diff
@@ -19,8 +20,32 @@ defmodule Trento.SapSystems.Events.SapSystemRolledUp do
     DatabaseRolledUp
   end
 
-  defevent resource: "sap_system", version: 2 do
+  defevent resource: "sap_system", version: 3 do
     field :sap_system_id, Ecto.UUID
     embeds_one :snapshot, Trento.SapSystems.SapSystem
   end
+
+  def upcast(params, _, 2), do: params
```
arbulu89, line 28: comment:
this `upcast` with version `2` is needed, as version of the previous event was already 2, due to the `supersede` function usage

lib/trento/databases/events/database_instance_registered.ex
```diff
@@ -40,4 +42,7 @@ defmodule Trento.Databases.Events.DatabaseInstanceRegistered do
     do: Map.put(params, "system_replication_tier", nil)
 
   def upcast(params, _, 4), do: params
+
+  def upcast(params, _, 5),
+    do: HealthService.upcast_health_to_status(params)
```
skrech: Do you think that `upcast_health_to_status` function is correctly placed in the HealthService? It seems like highly internal-maintenance function only. Maybe it should be placed somewhere in the `support` dir? I can't make up my mind on this.

lib/trento/databases/events/database_rolled_up.ex
```diff
@@ -44,4 +46,26 @@ defmodule Trento.Databases.Events.DatabaseRolledUp do
...
+        3
+      ) do
+    %{
+      "database_id" => database_id,
+      "snapshot" => %{
+        snapshot
+        | "instances" => Enum.map(instances, &HealthService.upcast_health_to_status/1)
+      }
+    }
+  end
```
skrech, line 64: What do you think about using `Kernel.update_in()` here?

lib/trento/sap_systems/enums/status.ex
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-License-Identifier: Apache-2.0
+
+defmodule Trento.SapSystems.Enums.Status do
```
skrech, line 4: Isn't it better to put this enum in a more common place, like `lib/trento/enums/`?

lib/trento/databases/events/database_instance_registered.ex
arbulu89: Yes, it is not the best place...
At the end, I spent 1 hour writing code and 8 hours thinking where to put the function, and this looked the less worst XD
What about if I put the function in the `upcaster` folder in `trento/databases/events/upcaster` ?
I know it is shared by database and sap systems, but I think it is better than a random folder in a `support` folder

lib/trento/databases/events/database_rolled_up.ex
arbulu89: No strong opinions. Little it changes, but since you asked, I will change XD

lib/trento/sap_systems/enums/status.ex
arbulu89: I didn't want to put it there because it is a really SAP and database specific thing, and the `SapSystems` namespace gives this information.
Otherwise, I would need to change the name to something else, because `Status` could mean anything

lib/trento/databases/events/database_rolled_up.ex
arbulu89: Done!

lib/trento/databases/events/database_instance_registered.ex
arbulu89: Code moved to `trento/lib/databases/upcaster/upcast_helper.ex`

lib/trento/sap_systems/enums/status.ex
skrech: I see your point. I don't have strong opinion either.

skrech (approved):
Looks great, @arbulu89 ! Thank you!

My only gripe is with HealthService and the Status enum that are placed under SAPSystems. It feels a bit counter-intuitive to me to include functions from one "context" into the other one, while the meaning is equally distanced from both. Maybe I'm overly puristic, but I find it more attractive to  put the enum and the health service into context that is sibling to the sapsystems/database one. And they can be named SapStatus and SapHealthService, smth like that.

Anyway, this is just my though, ultimately it's up to you.

arbulu89:
> Looks great, @arbulu89 ! Thank you!
> 
> My only gripe is with HealthService and the Status enum that are placed under SAPSystems. It feels a bit counter-intuitive to me to include functions from one "context" into the other one, while the meaning is equally distanced from both. Maybe I'm overly puristic, but I find it more attractive to put the enum and the health service into context that is sibling to the sapsystems/database one. And they can be named SapStatus and SapHealthService, smth like that.
> 
> Anyway, this is just my though, ultimately it's up to you.

At the end, as HANA is a SAP system by definition, both concepts are tightly interconnected. I think it is fine to have a relationship like this. Specifically in this case, where the `Status` and its values are exactly the same for a HANA and SAP sysmte

PR #4369, opened 2026-06-10
Home dashboard cluster not available icon

arbulu89:
# Description

Update health summary view icons for not available clusters, using the dash icon.

image

## How was this tested?

Update already existing tests

@abravosuse do we need to update docs? In that case, let's create a ticket as reminder

### DIFF
--- a/lib/trento/sap_systems/services/health_summary_service.ex
+++ b/lib/trento/sap_systems/services/health_summary_service.ex
@@ -8,8 +8,6 @@ defmodule Trento.SapSystems.Services.HealthSummaryService do
 
   import Ecto.Query
 
-  require Trento.Enums.Health, as: HealthEnum
-
   alias Trento.Databases.Projections.DatabaseInstanceReadModel
 
   alias Trento.SapSystems.Projections.{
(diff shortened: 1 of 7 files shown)


abravosuse (approved):
Code apart, it looks good to me. Thank you!

test/trento/sap_systems/services/health_summary_service_test.exs
```diff
@@ -138,7 +138,7 @@ defmodule Trento.SapSystems.Services.HealthSummaryServiceTest do
              ] == HealthSummaryService.get_health_summary()
     end
 
-    test "should set as unknown the clusters health when they are not available" do
+    test "should set as nil the clusters health when they are not available" do
```
skrech: Do you think we can improve the naming --> "when there are/is no cluster/s" or something like that.

lib/trento/sap_systems/services/health_summary_service.ex
```diff
@@ -77,7 +75,7 @@ defmodule Trento.SapSystems.Services.HealthSummaryService do
       end)
 
     case cluster_id do
-      nil -> HealthEnum.unknown()
+      nil -> nil
```
skrech, line 78: Up to you: Do you think that a comment here would make sense or it's too obvious?

skrech (approved):
Great Xabi, thank you! Just two small things, that you can ignore:

lib/trento/sap_systems/services/health_summary_service.ex
arbulu89: I personally think it is obvious.
Actually, we should've put maybe put a comment in the old code to explain why we were putting `Unknown` when the cluster didn't exist

test/trento/sap_systems/services/health_summary_service_test.exs
arbulu89: Done

PR #4370, opened 2026-06-11
Prepare release workflows

balanza:
TBD

### DIFF
--- a/.github/workflows/prepare-release.yaml
+++ b/.github/workflows/prepare-release.yaml
@@ -1,0 +1,254 @@
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-License-Identifier: Apache-2.0
+
+name: Prepare Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version_bump:
+        description: "Which part of the semver to bump from the current VERSION"
+        required: true
+        type: choice
+        options:
+          - major
+          - minor
+          - patch
+      source:
+        description: "Where to compose the release from"
+        required: true
+        type: choice
+        default: release
+        options:
+          - release
+          - main
+
+concurrency:
+  group: prepare-release
+  cancel-in-progress: false
+
+jobs:
+  prepare:
+    name: Compose ${{ inputs.version_bump }} release from ${{ inputs.source }}
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+      pull-requests: write
+    env:
+      GH_TOKEN: ${{ secrets.TRENTOBOT_GH_PAT }}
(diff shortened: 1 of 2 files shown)


.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,254 @@
...
+          ref: ${{ inputs.source }}
+          fetch-depth: 0
+          token: ${{ secrets.TRENTOBOT_GH_PAT }}
+
+      - name: Configure git author
+        run: |
+          git config user.name 'trento-bot'
+          git config user.email 'trentobot@users.noreply.github.com'
+
+      - name: Compute current and next version
```
antgamdia, line 53: comment: maybe it is an overkill, but have a look at some dependencies aimed at handling semver comparisons and bump-ups: https://github.com/fsaintjacques/semver-tool

I don't think we are using prerelease versions, but, if so, this regex is not capturing them.

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,254 @@
...
+      - name: Fetch merged PRs for milestone ${{ steps.versions.outputs.next }}
+        if: inputs.source == 'release'
+        run: |
+          set -euo pipefail
+          gh pr list \
+            --state merged \
+            --base main \
+            --search "milestone:${{ steps.versions.outputs.next }}" \
+            --json number,title,headRefName,mergeCommit \
+            --limit 200 > prs.json
```
antgamdia, line 106: comment: I think this limit is safe, but should we add a message in case the returned results actually hit the 200 limit? Otherwise, we would be silently ignoring it, no?

.github/workflows/prepare-release.yaml
```diff
@@ -0,0 +1,254 @@
...
+          count=$(echo "$others" | jq 'length')
+          if [ "$count" -gt 0 ]; then
+            echo "::error::Found ${count} open release-coordinating PR(s) against '${SOURCE}' for a different version. Resolve them before starting a new ${NEXT} release."
+            echo "$others" | jq -r '.[] | "  #\(.number) \(.title) (head: \(.headRefName))"'
+            exit 1
+          fi
+
+      - name: Fetch merged PRs for milestone ${{ steps.versions.outputs.next }}
+        if: inputs.source == 'release'
+        run: |
```
antgamdia, line 99: comment: maybe we should be extracting those lengthy scripts into separate files for maintainability. We can run linters more easily, etc.

antgamdia (commented):
Just dropping in some early comments, but I know this is still a draft, no worries or pressure!

PR #4384, opened 2026-06-11
Generalize CommandedUtils usage

arbulu89:
# Description

Generalize the usage of `CommandedUtils`.
I have moved the function to dispatch multiple commands to the support file. I want to use it in a future implementation.
After that, I have used the utils file everywhere.

## How was this tested?

UT

No need

### DIFF
--- a/lib/trento/support/commanded_utils.ex
+++ b/lib/trento/support/commanded_utils.ex
@@ -5,7 +5,30 @@ defmodule Trento.Support.CommandedUtils do
   @moduledoc false
   alias Trento.ActivityLog
 
-  def dispatch(command), do: commanded().dispatch(command)
+  @type command :: struct
+
+  @spec dispatch(command | [command], Keyword.t()) :: :ok | {:error, any}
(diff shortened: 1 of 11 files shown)


lib/trento/support/commanded_utils.ex
```diff
@@ -5,7 +5,31 @@ defmodule Trento.Support.CommandedUtils do
...
+        {{:error, error}, :ok} ->
+          {:error, [error]}
+
+        {{:error, error}, {:error, errors}} ->
+          {:error, [error | errors]}
+      end
+    end)
+  end
+
+  def dispatch(command, []), do: commanded().dispatch(command)
```
arbulu89, line 31: Note:
I have added this `dispatch(command, [])` match to avoid changing already existing tests that expect the dispatch with one param.
(18 tests I think).
Let me know if it is OK or if I should remove the match and update the test.

arbulu89:
I'm closing this by now, as I'm not using the added functions anymore

PR #4391, opened 2026-06-15
Enforce using erlang26 on SLES15

balanza:
Because the code channel now ships both `erlang26` and `erlang27`, we must be explicit on which package to use so to avoid resolution failures at build time.

```
conflict for providers of erlang27 needed by elixir119, (provider erlang27 conflicts with erlang)
```


arbulu89 (approved):
Alright!
If it is tested in obs/ibs and it works, green light from me!

nelsonkopliku (approved):

skrech:
Great!

PR #4394, opened 2026-06-16
Update SAP instance status frontend design

arbulu89:
# Description

Update SAP and HANA instances new status design in frontend.
Now, instead of using the old health icon, it uses a plain colored circle with a tooltip.
Besides this, I have moved the status column to the 1st column in the details view to be more aligned with other views. Let me know what do you think about this. I can move back to previous column. With this order, adding the color text doesn't look that good

image

image

## How was this tested?

Modify existing UT and e2e

No need


abravosuse (commented):
It looks pretty neat, @arbulu89. Thank you!
I like the unified approach, in terms of icon design and tooltip approach, between the overview and the details view, for both SAP systems and HANA databases. The tooltip is important to support color blind people. I've also noticed that you have updated the SAP instances section in the host details view. Good.

Just one little problem that might be related to the screen size. The tooltip is not aligned with the icon itself in none of the affected views:

Image

Image

Image

Image

Image

arbulu89:
@abravosuse pushed a new commit which should align the icon and tooltip correctly.

abravosuse (approved):
The tooltips and the icons are now aligned in all the relevant views: Host details view, SAP systems overview, HANA databases overview, SAP system details view and HANA database details view. Tested on Chrome, Firefox and Edge. Thank you!

jagabomb:
@arbulu89 I like the alignment of having the status icon on the far left. I think it looks more consistent and the tooltips are working well.

Just a last consideration for consistency, do you think we should also move the **instance** hostname to the far left on the **SAP System Overview** screen next to the **instance status**? @abravosuse What do you think?

**Instance Detail (SAP System Overview)**
Screenshot 2026-06-17 at 11 18 24

**Instance Detail (SAP System Detail)**
Screenshot 2026-06-17 at 11 18 40

abravosuse:
> @arbulu89 I like the alignment of having the status icon on the far left. I think it looks more consistent and the tooltips are working well.
> 
> Just a last consideration for consistency, do you think we should also move the **instance** hostname to the far left on the **SAP System Overview** screen next to the **instance status**? @abravosuse What do you think?
> 
Technically they are different things. The SAP systems overview shows the physical host whereas the SAP system details view shows the virtual hostname used for the installation. I wouldn't touch it as that might make the user think that they are the same thing.

jagabomb (approved):
Nice! LGTM

test/e2e/cypress/pageObject/hana_database_details_po.js
```diff
@@ -153,19 +153,16 @@ const validateHostClass = (hostName, status) => {
...
 const validateHostStatus = (hostName, status) => {
   const hostNameCellSelector = layoutTableHostNameCell(hostName);
-  return cy
-    .get(hostNameCellSelector)
-    .nextAll()
-    .eq(5)
-    .should('have.text', `SAPControl: ${status}`);
+  cy.get(hostNameCellSelector).prev().find('svg').trigger('mouseover');
+
+  return cy.get(`span:contains("${status}")`).should('exist');
```
skrech, line 165: Nit: `.should('exist')` is redundant in Cypress. I personally prefer to omit it, but no big deal if you decide to leave it.

test/e2e/cypress/pageObject/sap_systems_overview_po.js
```diff
@@ -301,16 +301,16 @@ export const javaSystemIsDiscoveredCorrectly = () => {
 export const tableDisplaysExpectedAmountOfSystems = (systemsAmount) =>
   cy.get(sapSystemsTableRows).should('have.length', systemsAmount);
 
-export const eachInstanceHasItsHealthStatusCorrectlyUpdated = () => {
+export const eachInstanceHasItsStatusCorrectlyUpdated = () => {
   const sapSystemsFirstRow = `${sapSystemsTableRows}:eq(0)`;
   const collapsibleCell = `${sapSystemsFirstRow} > td:eq(0)`;
   cy.get(collapsibleCell).click();
 
   return cy.wrap(Object.entries(healthMap)).each(([state, health], index) => {
```
skrech: I see its not from this PR, but do we need this `return` statement?

test/e2e/cypress/pageObject/sap_systems_overview_po.js
```diff
@@ -301,16 +301,16 @@ export const javaSystemIsDiscoveredCorrectly = () => {
...
     basePage.loadScenario(`sap-systems-overview-${state}`);
```
skrech: A general note: I think that this function mixes concerns --> it executes actions (loadScenario) and does assertions. I think that it needs to be split in some way that it's more clear what is the action and what are the assertions in the test. As a whole, loadScenario calls should be obvious from the test code (spec file).

test/trento/databases/projections/database_projector_test.exs
```diff
@@ -497,7 +496,6 @@ defmodule Trento.Databases.Projections.DatabaseProjectorTest do
     adapted_database_instances =
       database_instances
       |> Map.put(:sap_system_id, database_id)
-      # TODO: to remove once frontend is aligned
```
skrech, line 500: I'm a bit confused, aren't we supposed to  remove the deprecated health for instances? Here only the comment is removed. 
Why do we need to put `health` into the `database_restored` event? 

In that sense, shouldn't we remove add_deprecated_health from HealthService? That function shouldn't be needed.

skrech (approved):
Look great! Thanks, @arbulu89 

I'm approving, although I think we should double check on HealthService's `add_deprecated_health` func.

test/e2e/cypress/pageObject/hana_database_details_po.js
arbulu89: I think it is redundant in `test passes of fails` but it is not in how cypress reports the step.
Without the `should` cypress doesn't put this steps as an assertion.

See here.
With `should`:
image

Without `should`:
image

Maybe it is just a formality, but I think it is a plus to have it.

test/e2e/cypress/pageObject/sap_systems_overview_po.js
arbulu89: We need to return the `cypress` chain I guess. Otherwise you cannot chain any `should` or similar functions.
I don't think we need, but well, I guess it is better to have.
There was an effort done by @vicenteqa to put these return values. Otherwise the function would return `undefined`

arbulu89: I guess you are asking because the function that is changed doesn't have the `return`, right?
That case is different, as it is simply an `arrow` function, so the value returned by the function inside is the final returned value, so we can still chain.

test/e2e/cypress/pageObject/sap_systems_overview_po.js
arbulu89: I do agree in general in the concept.
I guess in this case it was just to make it more convenient.
Let me change it.

test/e2e/cypress/pageObject/hana_database_details_po.js
skrech: Fair enough.

test/e2e/cypress/pageObject/sap_systems_overview_po.js
arbulu89: @skrech I have changed some tests there which where loading scenarios in the assertion.
Maybe there are still more in the whole e2e suite code.
Let me know what do you think.

test/trento/databases/projections/database_projector_test.exs
arbulu89: Yes, ideally yes.
And I thought to do that before.
The thing is that we have a generic `database` function in the view code, which adds the deprecated field for the API backward compatibility. And we use the same function in the `restore` event broadcast.

When I added the comment, I thought it would be trivial to remove the field afterwards, but as the function is reused, and I don't want to have more helper functions to add more complexity and code simply to remove this, I thought simply to leave it. It is not harmful to have the field in the broadcast. The redux store database will have the field, and it won't be used. Nothing else.

test/e2e/cypress/pageObject/sap_systems_overview_po.js
skrech: Perfect, thank you very much for this!

test/e2e/cypress/pageObject/sap_systems_overview_po.js
skrech: This is auto-resolved by the test restructuring in the following commit. :) Thanks

PR #4399, opened 2026-06-17
Stale application instances when host hearbeat fails

arbulu89:
# Description

Stale application instances when the host heartbeat fails.
It includes new events to change the aggregate of the SAP system and the corresponding changes in the read model.

~~The most important architectural decision and want that I want to discuss is the usage of the new event handler, `stale_date_event_handler`.~~

~~I decided to implement the dispatch of the stale command in there in order to make things simpler. As far as I see, there is not any scenario where when a host is not sending data, updating the application instances data (or any other component running in the host as `databases/clusters`, so what we have in the read model should be consistent).
The other (more optimal but complex) option would be to use a process manager.~~

The most important architectural decision and want that I want to discuss is the usage of the new event handler or a process manager.

At the end I see these options:
- Create event handler and accept the fact that we depend on the read model being correctly updated
- Create a new process manager (almost the entirely of the code would be duplicated with the `deregistration_process_manager)
- Use the currently existing `deregistration_process_manager` to dispatch the command on `HeartbeatFailed`. This process manager stores in its state the data we need to dispatch the commands. 

I have implemented and tried both. I think putting the stale logic in the already existing `deregistration_process_manager` is the best option. Stale logic is not that different of the deregistration (in fact, we can only deregister stale hosts). 
It solves any eventual consistent issue and we don't create new duped code.

## How was this tested?

UT (we will work on e2e tests once we have the whole "stale" feature finished)

Not yet


lib/trento/infrastructure/commanded/process_managers/deregistration_process_manager.ex
```diff
@@ -196,6 +206,39 @@ defmodule Trento.Infrastructure.Commanded.ProcessManagers.DeregistrationProcessM
...
+          database_instances: []
+        },
+        %HeartbeatFailed{}
+      ) do
+    nil
+  end
+
+  def handle(
+        %DeregistrationProcessManager{
+          application_instances: application_instances
```
skrech, line 222: I guess there is no way to  have application instances with nil cluster_id? 

Maybe it's better to check if there are empty application instances in the first clause of the function, and not check for cluster_id: nil and db instances empty?

test/support/factory.ex
```diff
@@ -596,6 +598,23 @@ defmodule Trento.Factory do
...
+    })
+  end
+
+  def application_instance_data_marked_in_sync_event_factory do
+    ApplicationInstanceDataMarkedInSync.new!(%{
+      sap_system_id: Faker.UUID.v4(),
+      instance_number: "00",
+      host_id: Faker.UUID.v4()
+    })
+  end
```
skrech, line 616: Cool, I'm glad that you use the new! func to create these :)

test/trento/sap_systems/events/sap_system_rolled_up_test.exs
```diff
@@ -35,14 +35,15 @@ defmodule Trento.SapSystems.Events.SapSystemRolledUpTest do
       sap_system_id = Faker.UUID.v4()
       sid = "PRD"
       health = :passing
-      instances = build_list(2, :sap_system_instance, status: Status.green())
+      instances = build_list(2, :sap_system_instance, status: Status.green(), stale_at: nil)
```
skrech, line 38: `stale_at: nil` should be the default, isn't it redundant?

test/trento/sap_systems/projections/sap_system_projector_test.exs
```diff
@@ -446,4 +448,83 @@ defmodule Trento.SapSystems.Projections.SapSystemProjectorTest do
...
+
+  test "should mark application instance data as stale when ApplicationInstanceDataMarkedStale event is received" do
+    stale_at = DateTime.utc_now()
+
+    %{
+      sap_system_id: sap_system_id,
+      instance_number: instance_number,
+      host_id: host_id,
+      sid: sid
+    } = insert(:application_instance, stale_at: nil)
```
skrech, line 460: Again, isn't `nil` the default value?

lib/trento/sap_systems/sap_system.ex
```diff
@@ -188,6 +197,28 @@ defmodule Trento.SapSystems.SapSystem do
...
+    case get_instance(instances, host_id, instance_number) do
+      %Instance{stale_at: nil} ->
+        %ApplicationInstanceDataMarkedStale{
+          sap_system_id: sap_system_id,
+          instance_number: instance_number,
+          host_id: host_id,
+          stale_at: stale_at
+        }
+
+      _ ->
```
skrech, line 217: As discussed offline, I'd prefer to not handle the `nil` case since, as I understand it, we shouldn't be able to end-up here :)

skrech (approved):
Great work, @arbulu89!

Some minor remarks:

lib/trento/infrastructure/commanded/process_managers/deregistration_process_manager.ex
arbulu89: > I guess there is no way to have application instances with nil cluster_id?

Yes, there can be application instances without a cluster_id. It is a pretty common thing.
This initial match is just a simple way to avoid doing any operation if we know the host doesn't have any resource in it.
We could even remove it, but it looks fine to me.
We will touch the 2nd match when we add databases and clusters logic.

test/trento/sap_systems/events/sap_system_rolled_up_test.exs
arbulu89: It is redundant, it was to give some "obvious" context that stale was nil without going to the factory code.
Anyway, i will remove it.

lib/trento/sap_systems/sap_system.ex
arbulu89: I have been thinking about it, and I think handling the scenario where the aggregate receives the command to mark as stale, but the instance is already stale is a good thing to have.
This way, it will never emit an event if the instance was already in this state.
I think it makes the aggregate logic sound and resilient.

That being said, it is true that by now, only the deregistration process manager dispatches this command, it shouldn't dispatch it when the instance is already stale. But the aggregate code doesn't know this, and simply protecting it against future usages of the command doesn't look an overkill.

PR #4409, opened 2026-06-18
Make only SAPSYSTEM and SAPLOCALHOST properties mandatory

arbulu89:
### Description

We have received an scenario where some SAPcontrol properties obtained from `sapcontrol -function GetInstanceProperties` have an empty value. In this case, the `Protected Webmethods` attribute.
The value of this property is optional, as it can be disabled.

For our payload, only `SAPSYSTEM` and `SAPLOCALHOST` properties need a value.
This PR makes them mandatory but leaves the rest as optional, so a blank value is accepted.

### How was this tested?

UT

### Documentation changes

No need


skrech (approved):
All good. Thanks @arbulu89

PR #4411, opened 2026-06-19
Enforce using erlang26 on SLES15

balanza:
Backport of #4391 for the 3.1.2 patch release.

### DIFF
--- a/packaging/suse/rpm/trento-web.spec
+++ b/packaging/suse/rpm/trento-web.spec
@@ -29,7 +29,6 @@ Source2:        package-lock.json
 Source3:        node_modules.spec.inc
 %include        %{_sourcedir}/node_modules.spec.inc
 Group:          System/Monitoring
-BuildRequires:  elixir >= 1.15
 BuildRequires:  elixir-hex
 BuildRequires:  erlang-rebar3
 BuildRequires:  gcc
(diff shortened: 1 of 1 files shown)


nelsonkopliku (approved):
Ok with this, however the [failing pipeline](https://github.com/trento-project/web/actions/runs/27823196066/job/82341283401?pr=4411) is very annoying 😅 

I believe that the very specifc erlang 27.1.3 is causing that problem.

I tried bumping erlang to 27.3.4.13 in https://github.com/trento-project/web/pull/4413 and the issue should disappear from [the CI](https://github.com/trento-project/web/actions/runs/27825073644).

Are we safe?
- yes, the change to `.tool-versions` only affects github actions
- yes on sles15 since we're forcing erlang26
- hopefully yes on sles16 and wherever we'd be using erlang 27 because we do not download deps from the internet (to be careful here because it seems that the pinned `erlang27` is actually at 27.1.3 🫠 )

If you want to have a green build, feel free to consider bumping `.tool-versions` as in https://github.com/trento-project/web/pull/4413 attempt. Then we can close that one.

Question: does wanda need re-releasing?

balanza:
@nelsonkopliku yes the CI failures are annoying. However, the changes in this PR are not affecting CI builds


> Question: does wanda need re-releasing?

No, as there are no patch to be backported

PR #4428, opened 2026-06-23
Table style row with content

arbulu89:
### Description

Update `Table` component `rowClassName` configuration option to accept functions based on content.
This way, table row classes are configurable depending on the row content itself.
I have changed how the coloring `bg-` classes are applied to be white background as table level, not cell level, otherwise we cannot override the row colors.

We can do things like this, where one row styles are different:
image

In this last option the usecase implementation would be:
```
rowClassName: ({ stale_at }) =>
    classNames({
      'bg-gray-100': !!stale_at,
    }),
```

### How was this tested?

UT

### Documentation changes

No

### DIFF
--- a/assets/js/common/Table/CollapsibleTableRow.jsx
+++ b/assets/js/common/Table/CollapsibleTableRow.jsx
@@ -28,7 +28,7 @@ function CollapsibleTableRow({
         {collapsibleDetailRenderer && (
           <td
             className={classNames(
-              'border-b border-gray-200 bg-white',
+              'border-b border-gray-200',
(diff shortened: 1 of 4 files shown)


skrech (approved):
Looks great!
Question: You're just adding this feature now with corresponding storybook and tests, but there is no such usage yet in Trento, correct?

arbulu89:
> Looks great! Question: You're just adding this feature now with corresponding storybook and tests, but there is no such usage yet in Trento, correct?

Correct. This is a preliminary work to enable the style usage of stale rows.
This way, next frontend PRs will only "consume" this new functionality

PR #4437, opened 2026-06-30
Stale database instances when host hearbeat fails

arbulu89:
### Description

Stale database instances when the host heartbeat fails.
It includes new events to change the aggregate of the database and the corresponding changes in the read model.
It replicates the same logic done in SAP systems: https://github.com/trento-project/web/pull/4399

### How was this tested?

UT

### Documentation changes

No (not yet, at least)

### DIFF
--- a/lib/trento/databases/projections/database_projector.ex
+++ b/lib/trento/databases/projections/database_projector.ex
@@ -21,6 +21,8 @@ defmodule Trento.Databases.Projections.DatabaseProjector do
   alias Trento.Databases.Events.{
     DatabaseDeregistered,
     DatabaseHealthChanged,
+    DatabaseInstanceDataMarkedInSync,
+    DatabaseInstanceDataMarkedStale,
     DatabaseInstanceDeregistered,
     DatabaseInstanceMarkedAbsent,
     DatabaseInstanceMarkedPresent,
@@ -121,7 +123,8 @@ defmodule Trento.Databases.Projections.DatabaseProjector do
           system_replication_operation_mode: system_replication_operation_mode,
           system_replication_source_site: system_replication_source_site,
           system_replication_tier: system_replication_tier,
-          status: status
+          status: status,
+          stale_at: nil
         })
 
       Ecto.Multi.insert(multi, :database_instance, database_instance_changeset)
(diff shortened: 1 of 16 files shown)


lib/trento/databases/projections/database_projector.ex
```diff
@@ -231,6 +234,51 @@ defmodule Trento.Databases.Projections.DatabaseProjector do
...
+      changeset =
+        DatabaseInstanceReadModel
+        |> Repo.get_by(
+          database_id: database_id,
+          instance_number: instance_number,
+          host_id: host_id
+        )
+        |> DatabaseInstanceReadModel.changeset(%{
+          stale_at: nil
+        })
```
skrech, line 276: General comment that we might consider in the future:
Although it might seem idiomatic to use changesets to update the read model, I tend to believe this is a good place to actually fallback to the lower-level primitives like `update_all` calls. Since we are certain (because the aggregate validated that) that something needs to be update, it's better to directly send a single SQL request for update, rather than first fetching the record, then update it in memory then write it to the DB. The fetching is redundant in this case but has a performance tax.
Fetching-first in most cases is an artifact from ORM-days and we can improve on this I think. It's good only if we need to detect if there are changes, so we can act on them. But this part is handled by CQRS in our case.

test/trento/databases/database_test.exs
```diff
@@ -2541,6 +2547,222 @@ defmodule Trento.Databases.DatabaseTest do
...
+                       instance_number: ^instance_number,
+                       stale_at: ^stale_at
+                     }
+                   ]
+                 } = state
+        end
+      )
+    end
+
+    test "should mark database instance data as in sync when registering a stale instance" do
```
skrech: I suspect that the description of the test is wrong/confusing. The test is about showing the DB as InSync when returning back from Stale, right?

lib/trento/infrastructure/commanded/process_managers/deregistration_process_manager.ex
skrech, line 221: This clause still looks redundant to me. By inspecting the implementation of the following clause... it covers the case covered by this one too.

skrech (approved):
All good! Thanks @arbulu89 

Small remarks:

test/trento/databases/database_test.exs
arbulu89: Yes, i will rewrite it

lib/trento/databases/projections/database_projector.ex
arbulu89: We had similar discussions in the past. We decided intentionally to go with this approach.

We decided to go this way because it is the way it is documented in `Ecto`, `Ecto.Multi` and `commanded-ecto-projections` projects (yes, I know, it is just simple documentation, it doesn't cover all use-cases).
We already discussed that the select was not needed in most of the cases, but there some cases we need to consider.
For example, the `after_update`. If the `after_update` needs all the fields of the instance that was updated, you cannot simply put and "empty" changeset there, as you would only receive the changed fields in the `after_update`. You can still change the behaviour of this putting `returning: true` in the `update`, that's true.

Bottom line, yes, we knew that we might have been doing not needed `select` queries in the projection, but it was an intentional thing to simply avoid doing mistakes (if you do the `get` thing, you are always going to get the whole struct).

Anyway, it is true that it can be changed. I'm open to have a team discussion. I was as well in favor of not doing the `get` in most of the cases, but since we already do it as well, most of the cases it was just a copy/paste thing in the new additions.

skrech: Yea, let's discuss this asynchronously. For the purposes of this PR, everything seems great, let's merge.

PR #4438, opened 2026-07-01
Fix E2E flaky test: should fail if required fields are missing

vicenteqa:
### Description
The previous selector was too broad and could match the `Create User` button instead of the form submit button. As a result, the second click could trigger navigation again (it was clicking the same button again before loading the form) rather than submitting the form, so the required field errors were not displayed.

Fixes #<issue>
Flaky test on `users.cy.js` suite

### How was this tested?
https://github.com/trento-project/web/actions/runs/28502888711

### Documentation changes

N/A

### DIFF
--- a/test/e2e/cypress/e2e/users.cy.js
+++ b/test/e2e/cypress/e2e/users.cy.js
@@ -20,7 +20,8 @@ describe('Users', () => {
       usersPage.pageTitleIsCorrectlyDisplayed('Create User');
     });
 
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):
Most probably you could still check only for `Create` button doing a more restrictive usage of `contains`, maybe with a regular expression:
https://docs.cypress.io/api/commands/contains

Anyway, if the current change fixes the thing, go ahead!

PR #4439, opened 2026-07-01
Handle exclude predicate

balanza:
Following RFC https://github.com/trento-project/docs/pull/188

This PR extends the check execution flow to pass per-host metadata (starting with is_majority_maker) as target attributes, enabling check policies to reference host properties. It also updates the execution results UI to surface hosts that were excluded by a check's exclude predicate.


lib/trento/clusters.ex
```diff
@@ -326,12 +326,9 @@ defmodule Trento.Clusters do
          selected_checks: selected_checks,
          details: details
        }) do
-    hosts_data =
-      Repo.all(
-        from h in HostReadModel,
-          select: %{host_id: h.id},
-          where: h.cluster_id == ^cluster_id and is_nil(h.deregistered_at)
-      )
+    majority_maker_hostnames = majority_maker_hostnames(details)
```
nelsonkopliku, line 329: **question:** we are propagating majority maker information for:
- `hana_scale_up` clusters (this matching `maybe_request_checks_execution`)
- `hana_scale_out`, `hana_ascs_ers` and `unknown` clusters (the following `maybe_request_checks_execution`)

We are effectively not doing so for `ascs_ers` in the previous `maybe_request_checks_execution`.

I just want to understand if the majority maker bit of information is being sent in all relevant cases.
The case that surprises me a bit is about `unknown` cluster, but if it should be like that, so be it.

assets/js/lib/test-utils/factories/executions.js
```diff
@@ -115,6 +115,15 @@ export const agentCheckErrorFactory = Factory.define(() => ({
   message: faker.hacker.phrase(),
 }));
 
+export const agentCheckExcludedFactory = Factory.define(() => ({
+  agent_id: faker.string.uuid(),
+  status: 'excluded_by_policy',
+  exclude_expression: 'host["is_majority_maker"] == true',
```
nelsonkopliku: **supernit:** just consistency OCD
```suggestion
  exclude_expression: 'host.is_majority_maker == true',
```

assets/js/pages/ExecutionResults/CheckResultDetail/CheckResultDetail.jsx
```diff
@@ -39,6 +40,28 @@ function CheckResultDetail({
...
+          <div>
+            <p className="text-sm text-gray-500">Exclusion predicate:</p>
+            <pre className="mt-1 p-2 bg-gray-100 rounded text-sm overflow-x-auto">
+              {excludeExpression}
+            </pre>
+          </div>
+        )}
+      </div>
+    );
+  }
```
nelsonkopliku, line 68: **thought 1:** 
I think we have two options to improve code organization
1. moving this into its own `ExcludedCheckResultDetail` component. It can very well be an "internal" component in this same file, the test already covers properly
2. moving this into its own `ExcludedCheckResultDetail` component and make it a sibling of `CheckResultDetail` (ie in its own directory with tests and stories) and lift the `isAgentExcluded` check to `CheckResultDetailPage` and have there `if isAgentExcluded then <ExcludedCheckResultDetail /> else <CheckResultDetail />`

I would be fine with 1, as I don't see much reasons to "promote" `ExcludedCheckResultDetail`, yet.

---

**thought 2:** just raising a thought from a UX/UI perspective. Very personal. Feel free to disregard if there is already an agreement on the direction.

The check result detail for a skipped check on a specific host looks, IMHO, very empty and the most interesting new information - the exclusion predicate - seems lost in the page.

Image

Have we considered an alternative to improve this?

Options I see:
- somehow highlight visually the exclusion predicate section
- do not allow to go to a check result for a host that has been excluded from an execution and provide the exclusion predicate for instance in a popover when hovering the "Excluded" row in the check result outline itself

Image

But again, just personal preference, feel free to disregard.

nelsonkopliku (commented):
Thanks @balanza, besides the comments left inline I have some other notes, maybe not strictly related to web changes, but about the overall feature.

About cluster wide checks (aka `expect_same`):
---

it seems that exclusion is properly leading to considering `total_hosts - excluded_host` as the overall hosts to be taken into account for cluster-wide check

This example for instance is a 2 host cluster, where 1 is excluded (might not be realistic, but that is not the point) and the cluster-wide check passes (even though only one host is actually to be considered). Expectable.

Image

We are surfacing the excluded targets in the outline itself and it seems good, however when we go in the check result detail I believe we might want to consider not having the excluded target from the `Gathered Facts` section (`hana2` in this example)

Image


<details>

<summary>Disregarded issue</summary>

About Multiple checks selection
---

I haven't identified yet the root cause, but something very strange is happening mainly when I select more checks for an execution.

The duplicated `hana2` entries in `00081D` and `61451E` should not be there.

I can show you how I get here.

Image

Let me iterate more on wanda code.

</details>

lib/trento/clusters.ex
balanza: All `hana_*` cluster types get the flag, as it makes sense in HANA architecture.

About `unknown`, my rationale was: if the host has been discovered as the majority maker, just pass it through even if we fail to determine the cluster topology; otherwise, `false` is a non-variant for the checks.

balanza:
> thought 2: just raising a thought from a UX/UI perspective. Very personal. Feel free to disregard if there is already an agreement on the direction.
> 
> The check result detail for a skipped check on a specific host looks, IMHO, very empty and the most interesting new information - the exclusion predicate - seems lost in the page.
> 
> Have we considered an alternative to improve this?
>
> Options I see:https://meet.google.com/jcd-dizj-nqt
>
>    * somehow highlight visually the exclusion predicate section
>    * do not allow to go to a check result for a host that has been excluded from an execution and provide the exclusion > predicate for instance in a popover when hovering the "Excluded" row in the check result outline itself

The information is unharmful, I would leave it there. It can be useful to users, and it's not really noisy (you must go inside the host report to see it). 
About enhancing the UI: I would summon @jagabomb to understand if that is at least acceptable, if not enough.

nelsonkopliku:
> The information is unharmful, I would leave it there. It can be useful to users, and it's not really noisy (you must go inside the host report to see it). About enhancing the UI: I would summon @jagabomb to understand if that is at least acceptable, if not enough.
>

Never found the information harmful 🙂, I actually find it useful knowing _why_ a check was excluded from a certain host. I am fine with the current proposal or with what Jurgen might come with.

nelsonkopliku (approved):
@balanza these changes look good to me.

I investigated further about the duplicated entries issue mentioned here https://github.com/trento-project/web/pull/4439#pullrequestreview-4624118123 and it does not seem related to the implementation per se, but it is more about at which level the facts gathering is mocked in demo mode (I used that to test locally).

I filed a fix here https://github.com/trento-project/wanda/pull/732. It basically moves the mocking point so that drift from actual implementation cannot happen.

lib/trento/clusters.ex
arbulu89: We can in theory have a majority maker in `ascs/ers` clusters, but I understand it is not in our best practices, so maybe we are not going to find one. I guess we can postpone its usage as long as it is properly documented

assets/js/pages/ExecutionResults/CheckResultDetail/CheckResultDetail.jsx
```diff
@@ -9,13 +9,33 @@ import {
...
   isTargetHost,
   getExpectSameFacts,
 } from '../checksUtils';
 import ExpectationsResults from './ExpectationsResults';
 import ExpectedValues from './ExpectedValues';
 import GatheredFacts from './GatheredFacts';
 
+function ExcludedCheckResultDetail({ excludeExpression }) {
+  return (
+    <div className="p-4 space-y-2" data-testid="excluded-by-policy">
```
arbulu89, line 22: rename/nit:
maybe we could rename the `excluded-by-policy` to simply `excluded, as this "term" is not used anymore.

assets/js/pages/ExecutionResults/CheckResultDetail/CheckResultDetail.test.jsx
```diff
@@ -277,4 +278,42 @@ describe('CheckResultDetail Component', () => {
       expect(screen.getByText('No facts were gathered')).toBeVisible();
     });
   });
+
+  it('should render an excluded-by-policy detail for a host excluded by the check predicate', () => {
```
arbulu89, line 282: rename:
maybe rename the tests

```suggestion
  it('should render an excluded execution detail for a host excluded by the check predicate', () => {
```

arbulu89 (approved):
Hey @balanza ,
the code looks good.
I think we could remove some `Excluded by policy` leftovers, as the term has been removed from `wanda` to simply `Excluded`.
Besides this, we use `host` as the object to send the new fields. I think `target` would be more appropriate, as it is the name we use in wanda itself. Like `target.is_majority_maker`.
What do you think?

assets/js/pages/ExecutionResults/CheckResultDetail/CheckResultDetail.jsx
chargio: Was this updated before merge?

PR #4441, opened 2026-07-01
Add stale information to instance status icon

arbulu89:
### Description

Add stale information to `InstanceStatus` component. It includes the new clock mini-icon and the tooltip with the date when the data was marked stale.

It is not used yet. That part will come later.

Sneak-peak:
image

### How was this tested?

UT and storybook

### Documentation changes

No


skrech (approved):
Great, nothing to complaint about! Thanks @arbulu89

PR #4443, opened 2026-07-01
Add clear AI configuration endpoint

nelsonkopliku:
### Description

This PR wires the clear ai configuration rest endpoint.

Follow up:
- activity logging
- UI changes
- e2e tests

### How was this tested?

Automated and IRL.

### DIFF
--- a/lib/trento_web/controllers/v1/ai_configuration_controller.ex
+++ b/lib/trento_web/controllers/v1/ai_configuration_controller.ex
@@ -75,4 +75,23 @@ defmodule TrentoWeb.V1.AIConfigurationController do
       |> render(:ai_configuration, %{ai_configuration: user_ai_config})
     end
   end
+
+  operation :clear_ai_configuration,
+    summary: "Clears User's AI Configuration",
+    description: "Removes the AI configuration for the currently authenticated user.",
+    tags: ["Profile"],
+    responses: [
+      no_content: "User AI Configuration cleared successfully.",
+      unauthorized: Schema.Unauthorized.response(),
+      forbidden: Schema.Forbidden.response()
+    ]
+
+  def clear_ai_configuration(conn, _) do
+    with :ok <-
+           conn
+           |> Pow.Plug.current_user()
+           |> AI.clear_user_configuration() do
+      send_resp(conn, :no_content, "")
+    end
+  end
 end
--- a/lib/trento_web/controllers/v1/ai_configuration_controller.ex
+++ b/lib/trento_web/controllers/v1/ai_configuration_controller.ex
@@ -78,7 +78,7 @@ defmodule TrentoWeb.V1.AIConfigurationController do
 
   operation :clear_ai_configuration,
     summary: "Clears User's AI Configuration",
-    description: "Removes the AI configuration for the currently authenticated user.",
+    description: "Clears the AI configuration for the currently authenticated user.",
     tags: ["Profile"],
     responses: [
       no_content: "User AI Configuration cleared successfully.",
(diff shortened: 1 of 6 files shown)


lib/trento_web/controllers/v1/ai_configuration_controller.ex
```diff
@@ -75,4 +75,22 @@ defmodule TrentoWeb.V1.AIConfigurationController do
       |> render(:ai_configuration, %{ai_configuration: user_ai_config})
     end
   end
+
+  operation :clear_ai_configuration,
```
antgamdia, line 79: I see some minor warnings in the API docs linter, but they're not caused by this PR, so ok

```
Location                            Message                                                                          
openapi_V1.pIKJcd.json:35:11    ▲           schema property `fencing_type` is missing `examples` or `example`                
openapi_V1.pIKJcd.json:2555:27  ▲           enum is defined as nullable but does not contain a `null` value. Nullable enum...
openapi_V1.pIKJcd.json:3421:34  ▲           enum is defined as nullable but does not contain a `null` value. Nullable enum...
```

lib/trento_web/controllers/v1/ai_configuration_controller.ex
```diff
@@ -75,4 +75,22 @@ defmodule TrentoWeb.V1.AIConfigurationController do
...
+    summary: "Clears User's AI Configuration",
+    description: "Removes the AI configuration for the currently authenticated user.",
```
antgamdia: nit: has _clear_ the same meaning than _remove_ in this _delete_ operation? I think the resulting situation is the same (= empty fields in the UI), but I'd use "delete" instead of "clear". No strong opinions or reasons, just gut feelings.

test/trento_web/controllers/v1/ai_configuration_controller_test.exs
```diff
@@ -773,4 +773,29 @@ defmodule TrentoWeb.V1.AIConfigurationControllerTest do
...
+
+      assert nil == Trento.Repo.get_by(UserConfiguration, user_id: user_id)
+    end
+
+    test "should be idempotent", %{conn: conn, admin_user: %{id: user_id}} do
+      insert(:ai_user_configuration, user_id: user_id)
+
+      Enum.each(1..4, fn _ ->
+        conn
+        |> delete("/api/v1/profile/ai_configuration")
```
antgamdia, line 796: question: I haven't checked, but I'm assuming we are using snake_case for other endpoints, no?

antgamdia (approved):
Thanks!

test/trento_web/controllers/v1/ai_configuration_controller_test.exs
nelsonkopliku: exactly

lib/trento_web/controllers/v1/ai_configuration_controller.ex
nelsonkopliku: yes those are pre-existing. I guess we checked sometime ago and there was some caveats about them.

lib/trento_web/controllers/v1/ai_configuration_controller.ex
antgamdia: I've seen you're using clearing elsewhere... so maybe I'd remove _remove_ instead?  Some ideas: https://www.merriam-webster.com/thesaurus/clear#thesaurus-entry-2-2

```suggestion
    summary: "Clears User's AI Configuration",
    description: "Empties the AI configuration for the currently authenticated user.",
```

or even just


```suggestion
    summary: "Clears User's AI Configuration",
    description: "Clears the AI configuration for the currently authenticated user.",
```

nelsonkopliku: Good catch. It's inconsistent.

Let me consolidate to "Clear".

Just for context: I personally try to stay away from too CRUD-ish naming as much as I can. I tend to prefer alternative naming. Although sometimes it _is_ just CRUD 🤷

nelsonkopliku: Addressed 4c9ffc5a5f768b9070764b67c551c14ff76b4620

PR #4445, opened 2026-07-02
fix: Firefox datetime-local calendar picker not applying date filter

balanza:
## Problem

When using the Activity Log date filters (`from_date` / `to_date`) in **Firefox**, selecting a date from the native calendar picker does not apply the filter. The date appears chosen in the picker UI but the filter never triggers.

This is a known Firefox quirk with `<input type="datetime-local">`: when the calendar picker is used to select a date, Firefox may only set the date part (`YYYY-MM-DD`) without the time component, leaving the input value incomplete. This caused:

1. **Value is empty** (input in invalid state — no time part): `dateTimeLocalToUtcDate` returned `null` early, the date change was silently dropped.  
2. **Value is just `YYYY-MM-DD`** (missing `T00:00`): `new Date("2024-01-15")` parsed as midnight **UTC**, then `.getHours()`/`.getMinutes()` returned values in the **browser's** local timezone, producing an incorrect UTC timestamp when converted to the user's profile timezone.

## Solution

In `DateFilter.jsx`, before parsing the `datetime-local` value, detect when the value is a date-only string (`YYYY-MM-DD` without time) and normalize it to `YYYY-MM-DDT00:00`. This ensures:

- The value is always in a valid `datetime-local` format  
- `new Date("2024-01-15T00:00")` correctly interprets midnight in the browser's timezone  
- `getHours()/getMinutes()` return `0/0`, correctly mapping to midnight (start of day) in the user's profile timezone

## Related

Closes the Firefox date selection bug described in the issue (date selection with the calendar in Activity Log filters don't work in Firefox).

## Testing

- ✅ All 16 DateFilter tests pass  
- ✅ All 53 ActivityLogPage tests pass  
- ✅ All 14 timezone lib tests pass  
- ✅ All 5 ComposedFilter tests pass

### DIFF
--- a/assets/js/common/DateFilter/DateFilter.test.jsx
+++ b/assets/js/common/DateFilter/DateFilter.test.jsx
@@ -144,6 +144,8 @@ describe('DateFilter component', () => {
     );
     await user.click(screen.getByText('Filter by date...'));
     const input = container.querySelector('input[type="datetime-local"]');
+    await user.click(input);
+    await user.clear(input);
     await user.type(input, datetime);
 
     const expectedDate = parseDateTimeLocalToUtc(datetime, timezone);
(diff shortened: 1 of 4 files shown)


antgamdia (commented):
Nice! It's a pity the selector does not work... while the solution is not perfect, it is ok-ish enough to work around the issue. Would it be possible to add a unit test case covering this?

assets/js/common/DateFilter/DateFilter.test.jsx
```diff
@@ -167,13 +169,58 @@ describe('DateFilter component', () => {
...
+    await user.click(input);
+    await user.clear(input);
     await user.type(input, datetime);
 
     const expectedDate = parseDateTimeLocalToUtc(datetime, timezone);
 
     expect(mockOnChange).toHaveBeenCalledWith(['custom', expectedDate]);
   });
 
+  it('leaves the custom input empty until focused, then fills the current time so a date-only calendar pick still applies the filter (Firefox)', async () => {
```
antgamdia, line 181: Thanks!

antgamdia (approved):
Thanks! Triggering also our friend copilot for a second pair of eyes and +1ing to unblock it.

PR #4446, opened 2026-07-02
Update instances stale state redux

arbulu89:
### Description

Update redux store application and database instances from broadcasted messages.
And i have added the backend events in the activity log list to be displayed properly as well (the severity still needs to be changed).

### How was this tested?

UT

### Documentation changes

No

### DIFF
--- a/assets/js/lib/model/activityLog.js
+++ b/assets/js/lib/model/activityLog.js
@@ -106,6 +106,10 @@ export const APPLICATION_INSTANCE_MARKED_ABSENT =
   'application_instance_marked_absent';
 export const APPLICATION_INSTANCE_MARKED_PRESENT =
   'application_instance_marked_present';
+export const APPLICATION_INSTANCE_DATA_MARKED_STALE =
+  'application_instance_data_marked_stale';
+export const APPLICATION_INSTANCE_DATA_MARKED_IN_SYNC =
+  'application_instance_data_marked_in_sync';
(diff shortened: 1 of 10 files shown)


assets/js/lib/model/activityLog.js
```diff
@@ -667,6 +685,16 @@ export const ACTIVITY_TYPES_CONFIG = {
...
+    label: 'Database Instance Data Marked Stale',
+    message: (_entry) => `Database instance data was marked stale`,
+    resource: databaseResourceType,
+  },
+  [DATABASE_INSTANCE_DATA_MARKED_IN_SYNC]: {
+    label: 'Database Instance Data Marked In Sync',
+    message: (_entry) => `Database instance data was marked in sync`,
+    resource: databaseResourceType,
+  },
   [DATABASE_INSTANCE_REGISTERED]: {
```
skrech, line 698: I guess at some point we need to be more precise on which instance is stale or in sync?

skrech (approved):
All good! Thanks @arbulu89!

One small question apart from the offline-discussed topic about removing the action constants in `@state`, which would be worked on in a separate PR.

assets/js/lib/model/activityLog.js
arbulu89: I think it is not needed. This is simply the message in the activity log box, we still have the rest of the information available:
image

PR #4447, opened 2026-07-02
Replace missing health by status in instances

arbulu89:
### Description

Replace missing `health` usage by `status` in frontend.
There were some leftovers from the main change.
It code actually was working, as health was still sent as backward compatibility.
I have checked, and there shouldn't be more `health` usages in application and database instances

### How was this tested?

UT

### Documentation changes

No


skrech (approved):
All good!

PR #4448, opened 2026-07-02
Use stale value instance status

arbulu89:
### Description

Use `staleAt` and `timezone` in `InstanceStatus` components.

Depends on: https://github.com/trento-project/web/pull/4441

image

### How was this tested?

UT

### Documentation changes

No

### DIFF
--- a/assets/js/common/Table/Table.jsx
+++ b/assets/js/common/Table/Table.jsx
@@ -14,7 +14,7 @@ import EmptyState from './EmptyState';
 import CollapsibleTableRow from './CollapsibleTableRow';
 
 const defaultCellRender = (content) => (
-  <p className="text-gray-900 whitespace-no-wrap">{content}</p>
+  <p className="whitespace-no-wrap">{content}</p>
 );
 
 const renderCells = (columns, item) => (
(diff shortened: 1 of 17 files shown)


assets/js/common/Table/Table.jsx
```diff
@@ -14,7 +14,7 @@ import EmptyState from './EmptyState';
 import CollapsibleTableRow from './CollapsibleTableRow';
 
 const defaultCellRender = (content) => (
-  <p className="text-gray-900 whitespace-no-wrap">{content}</p>
```
arbulu89, line 17: This change is done to give the option to render cells text color dynamically. The default color now is in the table class

assets/js/lib/tables/index.js
```diff
@@ -0,0 +1,4 @@
+// SPDX-FileCopyrightText: SUSE LLC
+// SPDX-License-Identifier: Apache-2.0
+
+export const STALE_ROW = 'bg-gray-100 text-gray-500';
```
skrech, line 4: You've put that into `@lib/tables/index.js` because you consider this not generic enough to be placed into `@common/Table`?

skrech (approved):
Great, thanks @arbulu89 !

assets/js/lib/tables/index.js
arbulu89: Exactly. And because some other table-sh components (they look like table, but don't use the `Table` component) use it.
I had doubts where to put it, and `Table` component was one option any way.
As we have many constants in `lib` I put it there. But it could very well go in `Table`.
Would you put it in `Table` yourself? I wouldn't disagree

PR #4450, opened 2026-07-02
Health Details for Host Aggregate

skrech:
### Description

Re-factoring PR that combines the individual health fields in the Host aggregate into a HealthDetails struct.
No events are changed, except the HostRolledUp event. 

Additionally, a small re-factoring was made to the software_updates_discovery_health emitting code -- it's extracted into a function now to be more consistent with the rest of the health emitting events.

### How was this tested?

UT only. Tests were adapted to fit the new shape of the Host aggregate.

### Documentation changes

No

### DIFF
--- a/lib/trento/hosts/events/host_rolled_up.ex
+++ b/lib/trento/hosts/events/host_rolled_up.ex
@@ -11,8 +11,36 @@ defmodule Trento.Hosts.Events.HostRolledUp do
 
   alias Trento.Hosts.Host
 
-  defevent do
+  defevent version: 2 do
     field :host_id, Ecto.UUID
     embeds_one :snapshot, Host
+  end
+
+  # Handle old HostRolledUp, inherited from the previous aggregate
+  # when checks_health, saptune_health and
+  # software_updates_discovery_health were independent fields. They
+  # are now moved into combined struct located at health_details key.
+  def upcast(
+        %{
+          "snapshot" =>
+            %{
+              "checks_health" => checks_health,
+              "saptune_health" => saptune_health,
+              "software_updates_discovery_health" => software_updates_discovery_health
+            } = snapshot
+        } = params,
+        _,
+        2
+      ) do
+    new_snapshot =
+      snapshot
+      |> Map.put("health_details", %{
+        "checks_health" => checks_health,
+        "saptune_health" => saptune_health,
+        "software_updates_discovery_health" => software_updates_discovery_health
+      })
+      |> Map.drop(["checks_health", "saptune_health", "software_updates_discovery_health"])
+
+    Map.put(params, "snapshot", new_snapshot)
   end
 end
--- a/lib/trento/hosts/host.ex
+++ b/lib/trento/hosts/host.ex
@@ -910,19 +910,6 @@ defmodule Trento.Hosts.Host do
     end
   end
 
-  def maybe_emit_host_software_updates_discovery_health_changed_event(
-        %Host{
-          host_id: host_id,
-          health_details: %HealthDetails{software_updates_discovery_health: current_health}
-        },
-        health
-      )
-      when health != current_health do
-    %SoftwareUpdatesHealthChanged{host_id: host_id, health: health}
-  end
-
-  def maybe_emit_host_software_updates_discovery_health_changed_event(_host, _health), do: nil
-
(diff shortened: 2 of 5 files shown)


lib/trento/hosts/events/host_rolled_up.ex
```diff
@@ -11,8 +11,36 @@ defmodule Trento.Hosts.Events.HostRolledUp do
...
+  defevent version: 2 do
     field :host_id, Ecto.UUID
     embeds_one :snapshot, Host
   end
+
+  # Handle old HostRolledUp, inherited from the previous aggregate
+  # when checks_health, saptune_health and
+  # software_updates_discovery_health were independent fields. They
+  # are now moved into combined struct located at health_details key.
+  def upcast(
```
arbulu89, line 23: comment:
Do we need to put a "default" upcast for version 2?
I mean, when new aggregate comes with the new format, this might not match any entry.
Maybe `def upcast(params, _, 2), do: params`

lib/trento/hosts/host.ex
```diff
@@ -911,17 +910,30 @@ defmodule Trento.Hosts.Host do
...
+        },
+        health
+      )
+      when health != current_health do
+    %SoftwareUpdatesHealthChanged{host_id: host_id, health: health}
+  end
+
+  def maybe_emit_host_software_updates_discovery_health_changed_event(_host, _health), do: nil
+
+  def maybe_emit_software_updates_discovery_lifecycle_events(_host_id, nil, nil), do: []
```
arbulu89, line 913: super nit:
maybe this `maybe_emit_software_updates_discovery_lifecycle_events` should go on top of the other, as it is used before the other

lib/trento/hosts/host.ex
```diff
@@ -911,17 +910,30 @@ defmodule Trento.Hosts.Host do
     end
   end
 
-  def maybe_emit_software_updates_discovery_events(_host_id, nil, nil), do: []
-  def maybe_emit_software_updates_discovery_events(_host_id, fqdn, fqdn), do: []
+  def maybe_emit_host_software_updates_discovery_health_changed_event(
```
arbulu89: nit:
maybe you can do the opposite comparison:

```
# when they are the same, do nothing
def maybe_emit_host_software_updates_discovery_health_changed_event(%Host{
      host_id: host_id,
      health_details: %HealthDetails{software_updates_discovery_health: health}
    },
    health
  ), do: nil

def maybe_emit_host_software_updates_discovery_health_changed_event(%Host{host_id: host_id}, health), do:
    %SoftwareUpdatesHealthChanged{host_id: host_id, health: health}
```

arbulu89 (approved):
Really really nice!
Just some small nitpick remarks, and double check if we need to add other `upcast` function just in case

lib/trento/hosts/events/host_rolled_up.ex
skrech: I can do that, but doesn't it upcasts only if the event is with older version?

lib/trento/hosts/host.ex
skrech: I was thinking about the same... but I don't remember now why I decided to ho with the way it is now... If you think it's safe, I like yours better and I'll change it.

lib/trento/hosts/events/host_rolled_up.ex
arbulu89: Most probably you are right!
(PD: if you want to be sure, we can test it "live". Testing rollup of hosts is simple. We just need to register a host, do whatever we need, deregister the host (this will emit the rollup event), and register back again)

lib/trento/hosts/host.ex
arbulu89: I think it is safe. I guess you went this way because it was the original code you changed from.
PD: anyway, current code is fine as well.

skrech: I updated it, I like it better now. :)

lib/trento/hosts/events/host_rolled_up.ex
skrech: Thank you for testing this out, @arbulu89! It works as it is, resolving.

arbulu89: Tested now, and you assumption was right. The upcast is not executed with the snapshot with the correct version, so we don't need to guard it with a "match all" guard.
So the current code is correct!

PR #4452, opened 2026-07-02
Add e2e tests for AI settings clean up

nelsonkopliku:
### Description

E2E tests for AI Configuration clean up.

### How was this tested?

THIS IS THE TEST.


antgamdia (approved):

balanza (approved):

test/e2e/cypress/e2e/users.cy.js
```diff
@@ -744,6 +746,50 @@ describe('Users', () => {
...
+      });
+
+      it('should clear the AI configuration on confirm', () => {
+        usersPage
+          .apiCreateAIConfiguration(
+            'googleai',
+            'gemini-2.5-pro',
+            'test-api-key'
+          )
+          .then(() => {
```
vicenteqa: I think this  `.then()`  is not needed here since we are not using the returned value. `apiCreateAIConfiguration()`  already returns the Cypress chain, so the following commands will be queued after it

test/e2e/cypress/e2e/users.cy.js
```diff
@@ -744,6 +746,50 @@ describe('Users', () => {
...
+
+    describe('Clearing AI Configuration', () => {
+      it('should keep configuration when clear modal is dismissed', () => {
+        usersPage
+          .apiCreateAIConfiguration(
+            'googleai',
+            'gemini-2.5-pro',
+            'test-api-key'
+          )
+          .then(({ provider, model }) => {
```
vicenteqa: Perhaps here we can remove `.then()` too because the returned values are already constants that can be declared before, wdyt?

test/e2e/cypress/pageObject/users_po.js
```diff
@@ -670,6 +679,34 @@ export const clickModalUpdateAIConfigurationButton = (shouldWait = true) => {
...
+export const clickCancelClearAIConfiguration = () =>
+  cy.get(cancelClearAIConfigurationButton).click();
+
+export const clickConfirmClearAIConfiguration = (shouldWait = true) => {
+  cy.intercept('DELETE', '/api/v1/profile/ai_configuration').as(
+    'clearAIConfiguration'
+  );
+  cy.get(confirmClearAIConfigurationButton).click();
+  if (shouldWait) {
+    basePage.waitForRequest('clearAIConfiguration');
```
vicenteqa: nit: we can extract the `clearAIConfiguration` alias into a constant.

vicenteqa (approved):
lgtm! just some minor stuff :)

test/e2e/cypress/e2e/users.cy.js
nelsonkopliku: Thanks! I addressed here d257bce5bc96f69645f8ac1df4b334af9a2e72ac and adjusted other tests that had the unnecessary `then`

test/e2e/cypress/pageObject/users_po.js
nelsonkopliku: Addressed here d257bce5bc96f69645f8ac1df4b334af9a2e72ac

PR #4466, opened 2026-07-06
Auto-retry failed CI jobs

antgamdia:
### Description

This PR encapsulates some CI jobs inside a "retry" action. This is mainly to avoid some flakiness without someone manually re-triggering those failed CI jobs. 
The number of retries is subject to discussion; maybe just 1 extra retry is enough.

Related #TRNT-4543

### How was this tested?

CI

### Documentation changes

No

### Additional information

This is a prequel PR for enabling auto-merge of Dependabot's PR.


.github/workflows/ci.yaml
```diff
@@ -388,7 +388,14 @@ jobs:
             assets/node_modules
           key: nodejs-${{ steps.setup-node.outputs.node-version }}-${{ hashFiles('assets/package-lock.json') }}
       - name: Run JS tests
-        run: cd assets && npm test
+        uses: nick-fields/retry@ad984534de44a9489a53aefd81eb77f87c70dc60 #v4
```
balanza: I'm a bit afraid to use an action from a random guy on the web 🙃

antgamdia: Yeah, fair point. We can replace it with custom logic. I just went the easiest path, but given the current supply chain situation nowadays... 😅

antgamdia: I've tried to replace the GHA with a custom, simple retry script: https://github.com/trento-project/web/pull/4466/commits/34501f325e531a7fd910a09db5a71d9fac62eb67

.github/workflows/ci.yaml
```diff
@@ -932,15 +965,15 @@ jobs:
...
         env:
           cypress_video: false
-        with:
-          working-directory: test/e2e
-          spec: ${{ matrix.cypress_spec }}
-          wait-on-timeout: 30
-          config: baseUrl=http://localhost:4000
-          expose: photofinish_binary=${{ steps.photofinish-path.outputs.path }},${{ matrix.cypress_expose }}
+          CYPRESS_SPEC: ${{ matrix.cypress_spec }}
+        run: |
```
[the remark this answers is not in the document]
antgamdia: Using `--expose` now (https://docs.cypress.io/api/cypress-api/expose#CLI-Flags), as `--env` was deprecated.

.github/workflows/ci.yaml
```diff
@@ -783,15 +810,15 @@ jobs:
...
         env:
           cypress_video: false
-        with:
-          working-directory: test/e2e
-          spec: ${{ matrix.cypress_spec }}
-          wait-on-timeout: 30
-          config: baseUrl=http://localhost:4000
-          expose: photofinish_binary=${{ steps.photofinish-path.outputs.path }}
+          CYPRESS_SPEC: ${{ matrix.cypress_spec }}
+        run: |
```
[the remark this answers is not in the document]
antgamdia: Using `--expose` now (https://docs.cypress.io/api/cypress-api/expose#CLI-Flags), as `--env` was deprecated.

.github/workflows/ci.yaml
```diff
@@ -670,17 +689,19 @@ jobs:
           photofinish_path="$(command -v photofinish)"
           echo "path=$photofinish_path" >> "$GITHUB_OUTPUT"
       - name: Cypress run
-        uses: cypress-io/github-action@948d67d3074f1bbb6379c8bdbb04e95d2f8e593f # v7.4.0
         env:
           SPLIT: ${{ strategy.job-total }}
           SPLIT_INDEX: ${{ strategy.job-index }}
           SPLIT_FILE: cypress-split-timings.json
           cypress_video: false
```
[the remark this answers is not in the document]
antgamdia: Using `--expose` now (https://docs.cypress.io/api/cypress-api/expose#CLI-Flags), as `--env` was deprecated.

.github/workflows/ci.yaml
```diff
@@ -670,17 +689,19 @@ jobs:
...
-          config: baseUrl=http://localhost:4000
-          expose: photofinish_binary=${{ steps.photofinish-path.outputs.path }}
+        run: |
+          cd test/e2e
+          export SPLIT="${{ strategy.job-total }}"
+          export SPLIT_INDEX="${{ strategy.job-index }}"
+          export SPLIT_FILE="cypress-split-timings.json"
+
+          source ../../.github/scripts/retry.sh
+          retry "npx cypress run --config baseUrl=http://localhost:4000 --expose photofinish_binary=${{ steps.photofinish-path.outputs.path }}" "E2E tests (container ${{ matrix.containers }})" 2 1800
```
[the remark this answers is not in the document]
antgamdia: Not relevant; according to the documentation:

> Video recording is disabled by default, but can be turned on by setting [video](https://docs.cypress.io/app/references/configuration#Videos) to true from within your configuration.

.github/workflows/ci.yaml
```diff
@@ -932,15 +965,15 @@ jobs:
...
-          wait-on-timeout: 30
-          config: baseUrl=http://localhost:4000
-          expose: photofinish_binary=${{ steps.photofinish-path.outputs.path }},${{ matrix.cypress_expose }}
+          CYPRESS_SPEC: ${{ matrix.cypress_spec }}
+        run: |
+          cd test/e2e
+          SPEC_LIST=$(echo "$CYPRESS_SPEC" | xargs | tr ' ' ',')
+
+          source ../../.github/scripts/retry.sh
+          retry "npx cypress run --spec \"$SPEC_LIST\" --config baseUrl=http://localhost:4000 --expose photofinish_binary=${{ steps.photofinish-path.outputs.path }},${{ matrix.cypress_expose }}" "Integration tests (${{ matrix.test }})" 2 1800
```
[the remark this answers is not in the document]
antgamdia: Not relevant; according to the documentation:

> Video recording is disabled by default, but can be turned on by setting [video](https://docs.cypress.io/app/references/configuration#Videos) to true from within your configuration.

.github/workflows/ci.yaml
```diff
@@ -783,15 +810,15 @@ jobs:
...
-          wait-on-timeout: 30
-          config: baseUrl=http://localhost:4000
-          expose: photofinish_binary=${{ steps.photofinish-path.outputs.path }}
+          CYPRESS_SPEC: ${{ matrix.cypress_spec }}
+        run: |
+          cd test/e2e
+          SPEC_LIST=$(echo "$CYPRESS_SPEC" | xargs | tr ' ' ',')
+
+          source ../../.github/scripts/retry.sh
+          retry "npx cypress run --spec \"$SPEC_LIST\" --config baseUrl=http://localhost:4000 --expose photofinish_binary=${{ steps.photofinish-path.outputs.path }}" "Regression tests (${{ matrix.test }})" 2 1800
```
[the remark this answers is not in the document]
antgamdia: Not relevant; according to the documentation:

> Video recording is disabled by default, but can be turned on by setting [video](https://docs.cypress.io/app/references/configuration#Videos) to true from within your configuration.

vicenteqa (commented):
Could we use native test retries where available? Cypress and Jest can retry only failed tests, whereas this wrapper reruns the whole command/suite on failure. For ExUnit/Coveralls this generic retry may still make sense if there is no native equivalent.
Also, if there is a specific need to rerun the whole test suite, for Cypress I’d suggest rerunning the whole job instead, so it starts from a clean environment without leftover test data from the previous attempt. wdyt?

.github/workflows/ci.yaml
```diff
@@ -670,17 +687,18 @@ jobs:
           photofinish_path="$(command -v photofinish)"
           echo "path=$photofinish_path" >> "$GITHUB_OUTPUT"
       - name: Cypress run
-        uses: cypress-io/github-action@948d67d3074f1bbb6379c8bdbb04e95d2f8e593f # v7.4.0
```
vicenteqa: Genuine question: is there a specific reason for replacing `cypress-io/github-action` with a direct `npx cypress run` invocation?

Could we keep using the GitHub Action and configure retries through its `config` parameter instead of changing `cypress.config.js`? That would keep the retry behavior scoped to CI while preserving the existing action setup.

Example from another Trento workflow: https://github.com/trento-project/trento-installation-automation/blob/498ab2b264b2265fe80abe41f8cece873be734af/.github/workflows/trento-installation-base.yml#L345

.github/scripts/retry.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
gagandeepb: In my opinion, we should not unconditionally/blindly retry test suites without accounting for specific features/modalities of retries available to the lang specific test runners/clis.

gagandeepb (commented):
Please consider/evaluate conditional and fractional retries in case of flaky test failures, i.e. only retry the failed tests not the entire suite if the first run has failed for flakiness reasons. Otherwise we risk making CI runs too long and increase the developer feedback loop from CI for very questionable gains.

nelsonkopliku:
+1 to what @vicenteqa and @gagandeepb are suggesting. A too eager retry mechanism might be wasteful in terms of resources - although github's 🙈 - and time

antgamdia:
> Could we keep using the GitHub Action and configure retries through its config parameter instead of changing cypress.config.js? 

I think so. Let me give it a try. Since the implementation has evolved thanks to all the fruitful pieces of feedback, the usage drifted from the initial one. Thanks for the pointer!


> +1 to what .vicenteqa and .gagandeepb are suggesting. A too eager retry mechanism might be wasteful in terms of resources - although github's 🙈 - and time

I hear you. However: what do you do when a Dependabot's PR is failing? Hit `re-run failed jobs` button? This PR is to do exactly the same. 
Context: we want to avoid having a person dedicated to reviewing Dependabot's PR and clicking "rerun" and "merge". What do you suggest then? @gagandeepb , @nelsonkopliku

nelsonkopliku:
@antgamdia maybe the two things do not exclude each other.

What I mean is we can have native test retrial (jest, cypress and elixir) AND retry the pipeline for other possible failures.

At the end when we rerun a pipeline it would rerun the failed jobs only, right?

What do you think?

nelsonkopliku:
@antgamdia any feedback about the hybrid approach? https://github.com/trento-project/web/pull/4466#issuecomment-5055479544

.github/workflows/ci.yaml
```diff
@@ -432,7 +432,9 @@ jobs:
       - name: Run test
         env:
           MIX_ENV: test
-        run: mix coveralls.json --warnings-as-errors --trace
+        run: |
+          source .github/scripts/retry.sh
+          retry "mix coveralls.json --warnings-as-errors --trace" "Backend tests (Elixir ${{ matrix.elixir }})" 2 1800
```
antgamdia: I don't know if there is a built-in mechanism for retrying this, so I just added a simple retry function via scripting.

.github/workflows/ci.yaml
```diff
@@ -939,7 +941,7 @@ jobs:
           working-directory: test/e2e
           spec: ${{ matrix.cypress_spec }}
           wait-on-timeout: 30
-          config: baseUrl=http://localhost:4000
+          config: baseUrl=http://localhost:4000,retries=2
```
antgamdia: I've moved this to a native retry using jest's options

.github/workflows/ci.yaml
```diff
@@ -948,6 +950,32 @@ jobs:
...
+      always() &&
+      fromJSON(github.run_attempt) < 3 &&
+      (
+        needs.test-javascript.result == 'failure' ||
+        needs.test-elixir.result == 'failure' ||
+        needs.test-e2e.result == 'failure' ||
+        needs.test-regression.result == 'failure' ||
+        needs.test-integration.result == 'failure' ||
+        needs.coveralls-finish.result == 'failure'
+      )
```
antgamdia: If everything still fails, run a couple of extra retries.

antgamdia (commented):
Re https://github.com/trento-project/web/pull/4466#issuecomment-5129510144, I have tried to balance the native retries vs GHA retries. See what you think

(_The workflow_dispatch is not working in this PR as it can only target a workflow file that already exists on the repo's default branch_)

.github/workflows/ci.yaml
antgamdia: Sorry, it was a leftover from a previous attempt 😅

.github/scripts/retry.sh
gagandeepb: I don't think this file is strictly needed, in my understanding.

.github/workflows/ci.yaml
gagandeepb: Consider the following sourced from the [mix docs](https://mix.hexdocs.pm/1.19.5/Mix.Tasks.Test.html#module-command-line-options):
There is an option called `--failed`; please look into doing something along the following lines by adding in coverage flags as well:

```
mix test || if [[ $? = 2 ]]; then mix test --failed; else false; fi
```

.github/workflows/ci.yaml
```diff
@@ -948,6 +950,32 @@ jobs:
...
+    steps:
+      - name: Trigger rerun of failed jobs
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+        run: |
+          gh workflow run rerun.yml \
+            --ref "${{ github.head_ref || github.ref_name }}" \
+            -F run_id=${{ github.run_id }}
+
```
gagandeepb: I am not sure why we would need something like this; as in under what sort of circumstances, etc. Can you please explain?

.github/workflows/rerun.yml
```diff
@@ -0,0 +1,25 @@
+# SPDX-FileCopyrightText: SUSE LLC
```
gagandeepb: I am not sure why we would need something like this; as in under what sort of circumstances, etc. Can you please explain?

.github/workflows/ci.yaml
antgamdia: Thanks for the pointer! Honestly, I haven't read much about what `mix` offers. I have applied your suggestion:

.github/workflows/rerun.yml
antgamdia: Right! I've deleted the file.

.github/workflows/ci.yaml
antgamdia: Right! I've deleted the changes.

.github/scripts/retry.sh
antgamdia: Right! I've deleted the file.

antgamdia:
Thanks for the input, you all! I have removed the CI-job-level retry mechanism for now. Let's try just leveraging the native retry that the mix/jest ecosystem uses first. Later, if we still have issues or spurious failures, we can add the "hard" retry one.
My initial idea was just to mimic what we do when we manually hit "retry job" in GHA, but this lower-level approach can be better, I guess.
See what you think!

vicenteqa:
if the CI is happy I'm happy, giving my +1

vicenteqa (approved):

PR #4469, opened 2026-07-07
Improve stale icon positioning in instance status icon

arbulu89:
### Description

Improve the stale icon positioning in the `InstanceStatus` component to work properly with different sizes and screen font size.

### How was this tested?

Manual inspection

### Documentation changes

No

### DIFF
--- a/assets/js/pages/SapSystemDetails/InstanceStatus.jsx
+++ b/assets/js/pages/SapSystemDetails/InstanceStatus.jsx
@@ -9,9 +9,16 @@ import {
 } from 'eos-icons-react';
 import { capitalize } from 'lodash';
 import { formatDateTime } from '@lib/timezones';
+import { getIconSize } from '@lib/icon';
 import Tooltip from '@common/Tooltip';
 
-function InstanceStatus({ status, absent = false, staleAt = null, timezone }) {
+function InstanceStatus({
+  status,
+  size = 20,
+  absent = false,
+  staleAt = null,
+  timezone,
+}) {
   let cssClass;
 
   switch (status) {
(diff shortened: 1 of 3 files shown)


assets/js/pages/SapSystemDetails/InstanceStatus.jsx
```diff
@@ -53,11 +62,11 @@ function InstanceStatus({ status, absent = false, staleAt = null, timezone }) {
         wrap={false}
       >
         <div className="relative">
-          <IconComponent size="20" className={iconClass} />
+          <IconComponent size={convertedSize} className={iconClass} />
           {staleAt && (
             <EOS_SCHEDULE_OUTLINED
-              size="14"
-              className={`${iconClass} absolute -bottom-1 -right-1 bg-white rounded-full`}
+              size={convertedSize * 0.7}
```
skrech, line 68: Isn't it better to just specify the size in percentages? Then we won't need all the mappings and functions for taking the size?

assets/js/pages/SapSystemDetails/InstanceStatus.jsx
```diff
@@ -53,11 +62,11 @@ function InstanceStatus({ status, absent = false, staleAt = null, timezone }) {
...
+              className={`${iconClass} absolute bottom-0 right-0 translate-x-1/4 translate-y-1/4 bg-white rounded-full`}
```
skrech: I suggest to explicitly specify the `overflow: visible` property (although it's the default) to signal the reader that we expect the smaller icon to overflow the rectangle of the parent one.

assets/js/pages/SapSystemDetails/InstanceStatus.jsx
skrech: Hm... maybe this won't be possible with size property...

assets/js/pages/SapSystemDetails/InstanceStatus.jsx
arbulu89: Done

skrech (approved):

PR #4470, opened 2026-07-08
Mark database as stale

arbulu89:
### Description

Mark database data as stale. It is marked stale when at least one instance is stale.
It handles the scenario where database instances are removed, which cause a database data going in sync (stale instance removed).

I have chosen to add this logic in the aggregate itself, instead of doing some basic computation in the views as we need the reactiveness of the events to update the frontend accordingly.

### How was this tested?

UT

### Documentation changes

No

### DIFF
--- a/lib/trento/databases/database.ex
+++ b/lib/trento/databases/database.ex
@@ -310,19 +310,14 @@ defmodule Trento.Databases.Database do
       ) do
     case get_instance(instances, host_id, instance_number) do
       %Instance{stale_at: nil} ->
-        database
-        |> Multi.new()
-        |> Multi.execute(fn _ ->
+        [
           %DatabaseInstanceDataMarkedStale{
             database_id: database_id,
             instance_number: instance_number,
             host_id: host_id,
             stale_at: stale_at
           }
-        end)
-        |> Multi.execute(fn database ->
-          maybe_emit_database_data_marked_stale_event(database, stale_at)
-        end)
+        ] ++ maybe_emit_database_data_marked_stale_event(database, stale_at)
 
       _ ->
         nil
@@ -725,13 +720,15 @@ defmodule Trento.Databases.Database do
          },
          stale_at
        ) do
-    %DatabaseDataMarkedStale{
-      database_id: database_id,
-      stale_at: stale_at
-    }
+    [
+      %DatabaseDataMarkedStale{
+        database_id: database_id,
+        stale_at: stale_at
+      }
+    ]
   end
 
-  defp maybe_emit_database_data_marked_stale_event(_, _), do: nil
+  defp maybe_emit_database_data_marked_stale_event(_, _), do: []
 
   defp maybe_emit_database_data_marked_in_sync_event(%Database{stale_at: nil}), do: nil
 
--- a/lib/trento/databases/database.ex
+++ b/lib/trento/databases/database.ex
@@ -31,6 +31,8 @@ defmodule Trento.Databases.Database do
   }
 
   alias Trento.Databases.Events.{
+    DatabaseDataMarkedInSync,
+    DatabaseDataMarkedStale,
     DatabaseDeregistered,
     DatabaseHealthChanged,
     DatabaseInstanceDataMarkedInSync,
@@ -81,6 +83,7 @@ defmodule Trento.Databases.Database do
     field :sid, :string, default: nil
     field :health, Ecto.Enum, values: Health.values()
     field :rolling_up, :boolean, default: false
+    field :stale_at, :utc_datetime_usec
     field :deregistered_at, :utc_datetime_usec, default: nil
 
     embeds_many :instances, Instance
@@ -261,6 +264,7 @@ defmodule Trento.Databases.Database do
     |> Multi.execute(fn _ ->
       maybe_emit_database_instance_data_marked_in_sync_event(instance, command)
     end)
+    |> Multi.execute(&maybe_emit_database_data_marked_in_sync_event/1)
     |> Multi.execute(&maybe_emit_database_health_changed_event/1)
     |> Multi.execute(fn database ->
       maybe_emit_database_tenants_updated_event(database, tenants)
@@ -297,7 +301,7 @@ defmodule Trento.Databases.Database do
   end
 
   def execute(
-        %Database{database_id: database_id, instances: instances},
+        %Database{database_id: database_id, instances: instances} = database,
         %MarkDatabaseInstanceDataStale{
           instance_number: instance_number,
           host_id: host_id,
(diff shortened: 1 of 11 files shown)


lib/trento_web/controllers/v1/database_json.ex
```diff
@@ -141,6 +141,17 @@ defmodule TrentoWeb.V1.DatabaseJSON do
...
+  def database_stale_changed(%{
+        database: %{
+          id: id,
+          stale_at: stale_at
+        }
+      }),
+      do: %{
+        id: id,
+        stale_at: stale_at
+      }
```
skrech, line 153: Question: Why do we provide argument with `database` key. Seems redundant to me.

lib/trento/databases/database.ex
```diff
@@ -695,6 +713,38 @@ defmodule Trento.Databases.Database do
...
+       ) do
+    [
+      %DatabaseDataMarkedStale{
+        database_id: database_id,
+        stale_at: stale_at
+      }
+    ]
+  end
+
+  defp maybe_emit_database_data_marked_stale_event(_, _), do: []
```
skrech, line 731: nit: maybe return `nil` for consistency?

lib/trento/databases/database.ex
```diff
@@ -695,6 +713,38 @@ defmodule Trento.Databases.Database do
...
+           stale_at: nil
+         },
+         stale_at
+       ) do
+    [
+      %DatabaseDataMarkedStale{
+        database_id: database_id,
+        stale_at: stale_at
+      }
+    ]
```
skrech, line 728: nit: no need for list here, isn't it?

lib/trento/databases/database.ex
```diff
@@ -306,12 +310,14 @@ defmodule Trento.Databases.Database do
...
-          stale_at: stale_at
-        }
+        [
+          %DatabaseInstanceDataMarkedStale{
+            database_id: database_id,
+            instance_number: instance_number,
+            host_id: host_id,
+            stale_at: stale_at
+          }
+        ] ++ maybe_emit_database_data_marked_stale_event(database, stale_at)
```
skrech, line 320: I see now why you returned a list from  `maybe_emit_database_data_marked_stale_event` but I think it's better to handle the list concatination here by creating a new list. I perceive it as more clean.

skrech (commented):
Partial review of the logic. Will review the test tomorrow.

test/support/factory.ex
```diff
@@ -584,6 +586,19 @@ defmodule Trento.Factory do
...
+      database_id: Faker.UUID.v4(),
+      stale_at: DateTime.utc_now()
+    })
+  end
+
+  def database_data_marked_in_sync_event_factory do
+    DatabaseDataMarkedInSync.new!(%{
+      database_id: Faker.UUID.v4()
+    })
+  end
```
skrech, line 610: I'm very happy that you use the `new!` method, but for it to be actually useful without surprises we need to re-create what ExMachina is doing behind the scenes, eg. merge the incoming attributes. 

With the current implementation if you want to override some of the fields in the factory the following would happen:
1. DatabaseDataMarkedInSync struct would be created and validated.
2. But it would be validated with the default values.
3. Then ExMachina would override the values in the already created struct.

In effect, the struct would be verified for consistency before applying the overrides which might lead to inconsistent data. 

IMO, the correct way is to modify the signature of the function to "get the full control":
```
def database_data_marked_in_sync_event_factory(attrs) do
```

Then add:
```
    <constructed_event_MAP_with_default_data>
    |> merge_attributes(attrs)
    |> evaluate_lazy_attributes()
    |> <event_struct>.new!()

```

test/trento/databases/database_test.exs
```diff
@@ -2761,6 +2775,384 @@ defmodule Trento.Databases.DatabaseTest do
...
+          database_id: database_id,
+          instance_number: instance_number_3,
+          host_id: host_id_3,
+          stale_at: stale_at
+        ),
+        build(:database_data_marked_stale_event,
+          database_id: database_id,
+          stale_at: stale_at
+        )
+      ]
```
skrech, line 3046: Isn't the realistic order of the events the following:
1. `:database_instance_data_marked_stale_event` (for instance 1)
2. `:database_data_marked_stale_event` (for the whole db)
3. `:databsae_instance_data_marked_stale_event` (for instance 3)
?

Small nit: If the above is true, maybe we can introduce stale_at2 value that is different that stale_at1 so we are sure stale_at is fixed from the first "mark stale" event.

lib/trento_web/controllers/v1/database_json.ex
arbulu89: Basically because we do it in many other similar entries, nothing else.
I see that we are not 100% consistent in any case.
Do you prefer without it?

lib/trento/databases/database.ex
arbulu89: I did this way intentionally to have the option to concat the result with other events.
Otherwise, I would need to remove `nil` results from the events list where this is used (which is really ugly) or use `Multi`, which is not need in the places where we use.

lib/trento/databases/database.ex
arbulu89: Yes, we need the list. Check that we are concating the result of this with other list.
This is done this way to avoid the usage of `Multi` (which works fine simply with `nil` return`).
As commented above, otherwise, we need to have an ugly `Enum.filter(&is_nil/1)` or similar when we use it.
And this being a private function I preferred to avoid it.

lib/trento/databases/database.ex
arbulu89: Well, I intentionally did the opposite because I thought this being more clean, rather than having (with the reject/filter):
```
Enum.reject([
    %DatabaseInstanceDataMarkedStale{
        database_id: database_id,
        instance_number: instance_number,
        host_id: host_id,
        stale_at: stale_at
    },
    maybe_emit_database_data_marked_stale_event(database, stale_at)
], &is_nil/1)
```

skrech: OK, I get it now... however, I'm thinking if using a Multi is that bad? Maybe we can extract the part that uses `get_instance` into a `maybe_emit_database_instance_data_marked_stale_event` and then use two Multi.executes with `maybe_emit_database_instance_data_marked_stale_event` and `maybe_emit_database_data_marked_stale_event`?

Ultimately, it's up to you since it's not a big deal, after all.

test/support/factory.ex
arbulu89: Changed

lib/trento/databases/database.ex
arbulu89: I think it is fine as it is :slightly_smiling_face: 
I don't think using `Multi` has a new real bad side-effect more than the fact the "message" we send to ourselves. We know that `Multi` is used to update the state before checking the next event, so I think putting it to just have a more consistent code it is not the best trade-off.

Anyway, as you said, it is not a big deal. The code is easy to understand either way. Most probably we do the `Multi` thing in other parts of the code when we don't really need, but well, it is fine. It is fine as long as it works the way we expect hehe

test/trento/databases/database_test.exs
arbulu89: It is true indeed.
About the `stale_at` comment, I will add a different `stale_at` value. It won't change anything, because well, we are not testing even that part, as we are creating the state ourselves in the test, but maybe it is more realistic.

skrech (approved):
Great!

PR #4472, opened 2026-07-08
Augmented HealthIcon with stale indications

skrech:
### Description

HealthIconi is now augmented with stale indication capabilities:

1. When staleAt prop is provided then a small "clock" icon is added to the original health icon, indicating that the status shown is stale. This icon is with the same color as the original icon.
2. Added is a tooltip that shows since when the status is stale. Additionally, the tooltip shows the capitalized health.

Implementation details: The stale functionality (adding the clock icon with the logic when to show it and the tooltip) are extracted into a higher order React component. It takes a reference to a component rendering function and returns another rendering function that adds the stale functionality. This could potentially be used in InstanceStatus component and in future stale indicators.

### How was this tested?

UT

### DIFF
--- a/assets/js/common/HealthIcon/HealthIcon.jsx
+++ b/assets/js/common/HealthIcon/HealthIcon.jsx
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import React from 'react';
+import { capitalize } from 'lodash';
+
 import { computedIconCssClass } from '@lib/icon';
 
 import {
@@ -15,8 +17,17 @@ import {
   EOS_REMOVE_FILLED,
 } from 'eos-icons-react';
 
-import Spinner from '@common/Spinner';
 import classNames from 'classnames';
+import Spinner from '@common/Spinner';
+import StaleIconWrapper from '@common/StaleIconWrapper'
+
+const PassingIconBlank = StaleIconWrapper(EOS_CHECK_CIRCLE_OUTLINED);
+const PassingIconLink = StaleIconWrapper(EOS_CHECK_CIRCLE_FILLED);
+const WarningIconBlank = StaleIconWrapper(EOS_WARNING_OUTLINED);
+const WarningIconLink = StaleIconWrapper(EOS_WARNING_FILLED);
+const CriticalIconBlank = StaleIconWrapper(EOS_ERROR_OUTLINED);
+const CriticalIconLink = StaleIconWrapper(EOS_ERROR_FILLED);
+const UnknownIcon = StaleIconWrapper(EOS_LENS_FILLED);
 
 function HealthIcon({
   health = undefined,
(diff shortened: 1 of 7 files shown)


assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
```diff
@@ -0,0 +1,71 @@
...
+ * message needs to be provided with the `timezone` property. The text
+ * before the staleness info can be configured by `tooltipText`
+ * property.
+ */
+function StaleIconWrapper(WrappedIcon) {
+  function StaleAugmentedHealthIcon({
+    staleAt,
+    timezone,
+    tooltipText,
+    className,
```
arbulu89: nit:
should we default this `className` to empty string or something?
what happens if it doesn't come? 
because we would have `undefined.split(' ')` which would crash

assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
```diff
@@ -0,0 +1,71 @@
...
+    size,
+    ...props
+  }) {
+    const convertedSize = getIconSize(size);
+    const fillColor =
+      className.split(' ').find((cls) => cls.startsWith('fill-')) ??
+      'fill-gray-500';
+    const tooltipContent = (
```
arbulu89, line 36: question:
this question is more for @jagabomb maybe.
Do we always want a tooltip here? Even it the health is not stale?
We added the `status` tooltip in `InstanceStatus` for accessibility issues, since the icon was the same for all the statuses except the color.
Here the icon is different, so I would assume we don't need a tooltip (unless to notify the stale timestamp)

assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
```diff
@@ -0,0 +1,71 @@
...
+            <br />
+            (Stale since {formatDateTime(staleAt, timezone)})
+          </>
+        )}
+      </span>
+    );
+
+    return (
+      <div className="flex items-center mx-1">
+        <Tooltip content={tooltipContent} place="top" isEanbled wrap={false}>
```
arbulu89: typo:
`isEanbled` -> `isEnabled`

assets/js/common/HealthIcon/HealthIcon.jsx
```diff
@@ -17,66 +19,82 @@ import {
...
       return (
         <PassingIcon
-          size={size}
           className={classNames(
             hoverOpacityClass,
             computedIconCssClass('fill-jungle-green-500', centered)
           )}
+          size={size}
+          staleAt={staleAt}
+          timezone={'Etc/UTC'}
```
arbulu89: comment:
 we cannot use `timezone` here this way. The `timezone` must be a new variable for the `HealthIcon` component (maybe with `Etc/UTC` as default).
Once you add it, don't forget to add the prop in storybook

arbulu89 (commented):
Alright!
Good work.
Dropped some nitpicks and a question (maybe more for @jagabomb about the tooltip content).

assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
skrech: You're absolutely right, I was think that all these values should be with default values. I'll add these defaults.

assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
skrech: I didn't know this was the reason. It's OK for me to change this! It's easy.

assets/js/common/HealthIcon/HealthIcon.jsx
skrech: Yea, I intentionally made it like that since it's not used. But OK, will make it a healthicon param.

assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
arbulu89: nit:
With this change the `Stale since` comment is not wrapped in parenthesis, so we could not reuse the `StaleIconWrapper` in `InstanceStatus`, right?
It is just an observation. I don't know what we should do if we want to reuse the component

arbulu89 (approved):
Looks great!
Since you didn't add any test for the `absent` thing, and we are going to remove it, does it makes sense to merge this first: https://github.com/trento-project/web/pull/4473

assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
skrech: What do you mean? The text needs to be in parenthesis? If so and it's that important, I can make it in parenthesis if tootipText is present. Should I do it or I'm getting this wrong.

arbulu89: I'm telling that if we want to reuse this in `InstanceStatus`, this component wraps the stale text in the tooltip in a parenthesis, and I'm not sure how we could make it reusable, without adding extra params, etc.

skrech: I think it's OK now, could you check if you like it? I've tested it with HealthIcon, and everything is perfect. I've temporary modified one healthicon to add tooltipText and then it wrapped the Stale since message in parenthesis. So, I think we cover all of it?

arbulu89: what about this, to make it a one liner.:
```
{staleAt && `${tooltipText && "{"}Stale since ${formatDateTime(staleAt, timezone)}${tooltipText && "}"}`}
```

skrech: To be honest, I prefer it to stay like that since it's more easy to grasp, IMO. In any case, the previous line with the `<br />` tag needs to stay, so I think is better this way.

PR #4492, opened 2026-07-12
Bump brace-expansion to 2.1.2 and 1.1.16

balanza:
Backport of #4476 for the 3.1.3 patch release.

PR #4495, opened 2026-07-13
Bump otp minor version

gagandeepb:
### Description

Motivation: due to certificate mismatch between otp and the hex cdn, the CI fails. See commit description for details around this topic. This is needed by https://github.com/trento-project/web/pull/4492.

### How was this tested?

CI is ~green. We might need to evaluate artifact building and subsequent testing.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
 elixir 1.18.4-otp-27
-erlang 27.1.3
+erlang 27.3.4
 nodejs 24.13.1
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -9847,9 +9847,9 @@
       }
     },
     "node_modules/@typescript-eslint/typescript-estree/node_modules/brace-expansion": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.1.2.tgz",
-      "integrity": "sha512-w5JZcKgdhDOgOwm8H+KgbosopHMuGcl6qbulwjtz3SM7I7P3yW1eAjzMPLrIE+NQ9vjgANKHWeMHnrT0OXW1oA==",
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-2.0.2.tgz",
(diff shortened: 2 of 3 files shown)


arbulu89 (commented):
Hey @gagandeepb ,
I don't have anything against the change, but we need to be really careful with it.
So far, the version shipped in SLES is 27.1.3, so unless that is changed, changing this is not going to have any effect in our supported artifacts.
I think all the changes must go hand by hand.

balanza:
> Hey @gagandeepb , I don't have anything against the change, but we need to be really careful with it. So far, the version shipped in SLES is 27.1.3, so unless that is changed, changing this is not going to have any effect in our supported artifacts. I think all the changes must go hand by hand.

`erlang 27.3.4.11` is already the value on `main` after #4392. See https://github.com/trento-project/web/blob/main/.tool-versions#L2

arbulu89:
> > Hey @gagandeepb , I don't have anything against the change, but we need to be really careful with it. So far, the version shipped in SLES is 27.1.3, so unless that is changed, changing this is not going to have any effect in our supported artifacts. I think all the changes must go hand by hand.
> 
> `erlang 27.3.4.11` is already the value on `main` after #4392. See https://github.com/trento-project/web/blob/main/.tool-versions#L2

Ok, we might already have an issue then :see_no_evil: 
I hope we have plans to update it for our SUSE artifacts, otherwise, it won't matter much what we have in that file.

.tool-versions
```diff
@@ -1,3 +1,3 @@
 elixir 1.18.4-otp-27
```
gagandeepb, line 1: @balanza Not sure if the combination of elixir 1.19 and otp 27.3.4 has been tested on release branch CI. I am under the impression that the build dep available in the build-env is elixir 1.19 and otp 27.3.4. I wonder if the elixir version should be made 1.19 here as well 🤔

gagandeepb (commented):
The other option might be to use a non-official hex mirror with the old version of OTP that is still available in SLES.

gagandeepb:
Closing due to sync conversation about the unavailability of otp 27.3.4 as a build dependency.

PR #4496, opened 2026-07-13
Mark SAP system as stale

arbulu89:
### Description

Add `stale_at` field and transition events to the sap system aggregate.
It follow a similar logic than the database sibling with an exception.
The sap system is stale if the associated database is stale, that's where the `UpdateDatabaseStaleAt` and the event handler come in place.

### How was this tested?

UT

### Documentation changes

No

### DIFF
--- a/lib/trento/activity_logging/activity_catalog.ex
+++ b/lib/trento/activity_logging/activity_catalog.ex
@@ -24,7 +24,8 @@ defmodule Trento.ActivityLog.ActivityCatalog do
 
   @excluded_events [
     Trento.Hosts.Events.HostChecksSelected,
-    Trento.Clusters.Events.ChecksSelected
+    Trento.Clusters.Events.ChecksSelected,
+    Trento.SapSystems.Events.SapSystemDatabaseStaleAtChanged
   ]
 
   @operation_activities [
--- a/lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
+++ b/lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
@@ -14,10 +14,10 @@ defmodule Trento.Infrastructure.Commanded.EventHandlers.SapSystemDatabaseStaleAt
     name: "sap_system_database_stale_at_event_handler"
 
   alias Trento.Databases.Events.{DatabaseDataMarkedInSync, DatabaseDataMarkedStale}
-  alias Trento.Databases.Projections.DatabaseReadModel
   alias Trento.Repo
   alias Trento.SapSystems.Commands.UpdateDatabaseStaleAt
   alias Trento.SapSystems.Events.SapSystemDatabaseStaleAtChanged
+  alias Trento.SapSystems.Projections.SapSystemReadModel
 
   alias TrentoWeb.V1.SapSystemJSON
 
@@ -57,12 +57,11 @@ defmodule Trento.Infrastructure.Commanded.EventHandlers.SapSystemDatabaseStaleAt
          database_stale_at,
          %{correlation_id: correlation_id, causation_id: causation_id}
        ) do
-    %{sap_systems: sap_systems} =
-      Repo.one!(
-        from(d in DatabaseReadModel,
-          where: d.id == ^database_id,
-          preload: [:sap_systems],
-          select: [:id]
+    sap_systems =
+      Repo.all(
+        from(s in SapSystemReadModel,
+          where: s.database_id == ^database_id,
+          select: [:id, :sid]
         )
       )
 
--- a/lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
+++ b/lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
@@ -76,8 +76,7 @@ defmodule Trento.Infrastructure.Commanded.EventHandlers.SapSystemDatabaseStaleAt
           database_stale_at: database_stale_at
         },
         correlation_id: correlation_id,
-        causation_id: causation_id,
-        consistency: :strong
+        causation_id: causation_id
       )
     end
 
--- a/lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
+++ b/lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
@@ -44,8 +44,10 @@ defmodule Trento.Infrastructure.Commanded.EventHandlers.SapSystemDatabaseStaleAt
       "monitoring:sap_systems",
       "sap_system_updated",
       SapSystemJSON.sap_system_database_stale_at_changed(%{
-        id: sap_system_id,
-        database_stale_at: database_stale_at
+        sap_system: %{
+          id: sap_system_id,
+          database_stale_at: database_stale_at
+        }
       })
     )
   end
--- a/lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
+++ b/lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
@@ -1,0 +1,87 @@
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-License-Identifier: Apache-2.0
+
+defmodule Trento.Infrastructure.Commanded.EventHandlers.SapSystemDatabaseStaleAtEventHandler do
+  @moduledoc """
+  This event handler is responsible to forward update database stale_at commands to the SAP systems
+  related to a database that has a new stale_at state.
+
+  Once the resulting event is emitted, the event handler broadcasts its results.
+  """
+
+  use Commanded.Event.Handler,
+    application: Trento.Commanded,
+    name: "sap_system_database_stale_at_event_handler"
+
+  alias Trento.Databases.Events.{DatabaseDataMarkedInSync, DatabaseDataMarkedStale}
+  alias Trento.Databases.Projections.DatabaseReadModel
+  alias Trento.Repo
+  alias Trento.SapSystems.Commands.UpdateDatabaseStaleAt
+  alias Trento.SapSystems.Events.SapSystemDatabaseStaleAtChanged
+
+  alias TrentoWeb.V1.SapSystemJSON
+
+  import Ecto.Query, only: [from: 2]
+
+  require Logger
+
+  def handle(%DatabaseDataMarkedStale{database_id: database_id, stale_at: stale_at}, metadata) do
+    update_sap_systems_database_stale_at(database_id, stale_at, metadata)
+  end
+
+  def handle(%DatabaseDataMarkedInSync{database_id: database_id}, metadata) do
+    update_sap_systems_database_stale_at(database_id, nil, metadata)
+  end
+
+  def handle(
+        %SapSystemDatabaseStaleAtChanged{
+          sap_system_id: sap_system_id,
+          database_stale_at: database_stale_at
+        },
+        _metadata
+      ) do
+    TrentoWeb.Endpoint.broadcast(
+      "monitoring:sap_systems",
+      "sap_system_updated",
+      SapSystemJSON.sap_system_database_stale_at_changed(%{
+        id: sap_system_id,
+        database_stale_at: database_stale_at
+      })
+    )
+  end
+
+  defp update_sap_systems_database_stale_at(
+         database_id,
+         database_stale_at,
+         %{correlation_id: correlation_id, causation_id: causation_id}
+       ) do
+    %{sap_systems: sap_systems} =
+      Repo.one!(
+        from(d in DatabaseReadModel,
+          where: d.id == ^database_id,
+          preload: [:sap_systems],
+          select: [:id]
+        )
+      )
+
+    for %{id: sap_system_id, sid: sid} <- sap_systems do
+      stale_at_message = if database_stale_at, do: database_stale_at, else: "nil (in-sync)"
+      Logger.info("Updating database stale_at of #{sid} SAP system to #{stale_at_message}")
+
+      commanded().dispatch(
+        %UpdateDatabaseStaleAt{
+          sap_system_id: sap_system_id,
+          database_stale_at: database_stale_at
+        },
+        correlation_id: correlation_id,
+        causation_id: causation_id,
+        consistency: :strong
+      )
+    end
+
+    :ok
+  end
+
+  defp commanded,
+    do: Application.fetch_env!(:trento, Trento.Commanded)[:adapter]
+end
--- a/lib/trento/sap_systems/projections/sap_system_projector.ex
+++ b/lib/trento/sap_systems/projections/sap_system_projector.ex
@@ -621,41 +621,30 @@ defmodule Trento.SapSystems.Projections.SapSystemProjector do
 
   @impl true
   def after_update(
-        %SapSystemDataMarkedStale{
-          sap_system_id: sap_system_id,
-          stale_at: stale_at
-        },
+        %SapSystemDataMarkedStale{},
         _,
-        _
+        %{sap_system: %SapSystemReadModel{} = sap_system}
       ) do
     TrentoWeb.Endpoint.broadcast(
       @sap_systems_topic,
       "sap_system_stale_changed",
       SapSystemJSON.sap_system_stale_changed(%{
-        sap_system: %{
-          id: sap_system_id,
-          stale_at: stale_at
-        }
+        sap_system: sap_system
       })
     )
   end
 
   @impl true
   def after_update(
-        %SapSystemDataMarkedInSync{
-          sap_system_id: sap_system_id
-        },
+        %SapSystemDataMarkedInSync{},
         _,
-        _
+        %{sap_system: %SapSystemReadModel{} = sap_system}
       ) do
     TrentoWeb.Endpoint.broadcast(
       @sap_systems_topic,
       "sap_system_stale_changed",
       SapSystemJSON.sap_system_stale_changed(%{
-        sap_system: %{
-          id: sap_system_id,
-          stale_at: nil
-        }
+        sap_system: sap_system
       })
     )
   end
--- a/lib/trento/sap_systems/projections/sap_system_projector.ex
+++ b/lib/trento/sap_systems/projections/sap_system_projector.ex
@@ -20,6 +20,8 @@ defmodule Trento.SapSystems.Projections.SapSystemProjector do
     ApplicationInstanceMoved,
     ApplicationInstanceRegistered,
     ApplicationInstanceStatusChanged,
+    SapSystemDataMarkedInSync,
+    SapSystemDataMarkedStale,
     SapSystemDeregistered,
     SapSystemHealthChanged,
     SapSystemRegistered,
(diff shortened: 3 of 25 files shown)


lib/trento/activity_logging/activity_catalog.ex
```diff
@@ -24,7 +24,8 @@ defmodule Trento.ActivityLog.ActivityCatalog do
 
   @excluded_events [
     Trento.Hosts.Events.HostChecksSelected,
-    Trento.Clusters.Events.ChecksSelected
+    Trento.Clusters.Events.ChecksSelected,
+    Trento.SapSystems.Events.SapSystemDatabaseStaleAtChanged
```
arbulu89, line 28: Adding the `SapSystemDatabaseStaleAtChanged` to the activity log events excluded list.
This event doesn't add real value to the activity log, as it is always associated to the `DatabaseStaleAtChanged` event, which is enough for the activity log.

lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
```diff
@@ -0,0 +1,87 @@
...
+      })
+    )
+  end
+
+  defp update_sap_systems_database_stale_at(
+         database_id,
+         database_stale_at,
+         %{correlation_id: correlation_id, causation_id: causation_id}
+       ) do
+    %{sap_systems: sap_systems} =
```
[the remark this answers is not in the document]
arbulu89: Not an issue. Those values always come in the metadata

lib/trento_web/openapi/v1/schema/sap_system.ex
```diff
@@ -252,6 +270,7 @@ defmodule TrentoWeb.OpenApi.V1.Schema.SAPSystem do
             database_sid: "HA1",
             database_instances: [],
             tags: [],
+            stale_at: "2024-01-15T08:00:00Z",
             inserted_at: "2024-01-15T10:30:00Z",
             updated_at: "2024-01-15T12:30:00Z"
```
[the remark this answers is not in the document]
arbulu89: Changed

lib/trento_web/controllers/v1/sap_system_json.ex
```diff
@@ -78,6 +82,9 @@ defmodule TrentoWeb.V1.SapSystemJSON do
   def sap_system_database_health_changed(%{id: id, database_health: database_health}),
     do: %{id: id, database_health: database_health}
 
+  def sap_system_database_stale_at_changed(%{id: id, database_stale_at: database_stale_at}),
```
skrech: This is something like a follow-up on a previous comment I made on the Database staleness PR --> I asked why we prefix the supplied argument to these functions with the resource name. Now I get it. it's a requirement in Phoenix views to receive a single argument called `assigns` so they can work with the `render/3` function of Phoenix. Now, in `assigns`, it's a convention to give values keyed on the resource they represent. So, as to preserver this convention it's better to prefix the expected Map with `sap_system` as most of the other functions in this file are doing.

lib/trento/sap_systems/projections/sap_system_projector.ex
```diff
@@ -588,6 +619,47 @@ defmodule Trento.SapSystems.Projections.SapSystemProjector do
...
+      @sap_systems_topic,
+      "sap_system_stale_changed",
+      SapSystemJSON.sap_system_stale_changed(%{
+        sap_system: %{
+          id: sap_system_id,
+          stale_at: stale_at
+        }
+      })
+    )
+  end
```
skrech, line 635: To better integrate with the Phoenix JSON views you can get the the whole sap_system from the `changes` argument (the 3rd one) and directly pass it to with key `sap_system`. Its values for `id` and `stale_at` would be the same as the one in the event.

lib/trento/sap_systems/projections/sap_system_projector.ex
```diff
@@ -588,6 +619,47 @@ defmodule Trento.SapSystems.Projections.SapSystemProjector do
...
+      @sap_systems_topic,
+      "sap_system_stale_changed",
+      SapSystemJSON.sap_system_stale_changed(%{
+        sap_system: %{
+          id: sap_system_id,
+          stale_at: nil
+        }
+      })
+    )
+  end
```
skrech, line 650: Same as above.

lib/trento/sap_systems/sap_system.ex
```diff
@@ -127,6 +133,7 @@ defmodule Trento.SapSystems.SapSystem do
     |> Multi.execute(fn sap_system ->
       maybe_emit_sap_system_restored_event(sap_system, instance)
     end)
+    |> Multi.execute(&maybe_emit_sap_system_data_marked_in_sync_event/1)
```
skrech, line 136: Just a question: If the system is restored, is there a potential for `database_stale_at` to be out of sync with the restored state current state? Eg. if the database was stale before deregistrartion, then deregistration of the sap ssytem, then the database comes back in sync, then the sap system is restored. In that case there might be a problem?

lib/trento/sap_systems/sap_system.ex
```diff
@@ -301,6 +312,41 @@ defmodule Trento.SapSystems.SapSystem do
...
+        %UpdateDatabaseStaleAt{database_stale_at: database_stale_at}
+      ) do
+    [
+      %SapSystemDatabaseStaleAtChanged{
+        sap_system_id: sap_system_id,
+        database_stale_at: database_stale_at
+      }
+    ] ++
+      maybe_emit_sap_system_data_marked_stale_event(sap_system, database_stale_at)
+  end
```
skrech, line 346: Ah, I don't like the last two causes... If we add additional guard to `maybe_emit_sap_system_data_marked_stale_event` we can have both `maybe_emit_sap_system_data_marked_stale_event` and `maybe_emit_sap_system_data_marked_in_sync_event` be executed one after the other into a single `execute` clause. Also, we'll have to use a Multi here.

lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
```diff
@@ -0,0 +1,87 @@
...
+      Logger.info("Updating database stale_at of #{sid} SAP system to #{stale_at_message}")
+
+      commanded().dispatch(
+        %UpdateDatabaseStaleAt{
+          sap_system_id: sap_system_id,
+          database_stale_at: database_stale_at
+        },
+        correlation_id: correlation_id,
+        causation_id: causation_id,
+        consistency: :strong
```
skrech: Do we need to strong consistency?

lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
```diff
@@ -0,0 +1,87 @@
...
+         %{correlation_id: correlation_id, causation_id: causation_id}
+       ) do
+    %{sap_systems: sap_systems} =
+      Repo.one!(
+        from(d in DatabaseReadModel,
+          where: d.id == ^database_id,
+          preload: [:sap_systems],
+          select: [:id]
+        )
+      )
```
skrech, line 66: This is fine, but maybe we can directly query the sap systems read model filtering by database_id, should be more performant?

skrech (commented):
Great work @arbulu89! Several comments, though:

PS. I'll review the tests section later.

test/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler_test.exs
```diff
@@ -0,0 +1,138 @@
...
+      "sap_system_updated",
+      %{
+        id: ^sap_system_id,
+        database_stale_at: ^stale_at
+      },
+      1000
+    )
+  end
+
+  test "should broadcast database stale_at change with nil when a SapSystemDatabaseStaleAtChanged event is received with nil" do
```
skrech: Not a biggie, but this test seems redundant. The broadcasting logic is retranslating whatever value it receives from the event, so maybe this test is not needed. Or at least can be parameterized with a matrix (@scenario/unquote)

lib/trento_web/controllers/v1/sap_system_json.ex
arbulu89: Keep in mind that this function doesn't go through the controller `render` function, it is only used in the `broadcast` function. So it would work regardless.
Anyway, I will change it for consistency.

lib/trento/sap_systems/projections/sap_system_projector.ex
arbulu89: alright!

lib/trento_web/controllers/v1/sap_system_json.ex
arbulu89: Done

lib/trento/sap_systems/projections/sap_system_projector.ex
arbulu89: done

lib/trento/sap_systems/projections/sap_system_projector.ex
arbulu89: done

lib/trento/sap_systems/sap_system.ex
arbulu89: This is a good point.
Actually, it would be in sync, but now I see I don't like it this way :sweat_smile: 
Let me explain your scenario and what would happen now:
- Database changes to stale -> We dispatch `UpdateDatabaseStaleAt` and emit `SapSystemDatabaseStaleAtChanged` and `database_stale_at` gets a value
- Deregister SAP system -> We emit `SapSystemDeregistered`. For the user, the SAP system doesn't exist in trento anymore
- Database comes back to sync -> We dispatch `UpdateDatabaseStaleAt` (**the event handler is not filtering out deregistered SAP systems! this is mistake**), and we emit `SapSystemDatabaseStaleAtChanged` for a deregistered SAP system. This shouldn't happen, we shouldn't emit there events for deregistered events
- As a side-effect it would update the `database_stale_at` value, and when the SAP system is restored it would have a correct value. 
So even if it would be in sync, the process is wrong.
I need to think and update this.

lib/trento/sap_systems/sap_system.ex
arbulu89: I had that initially and I changed it to this in fact. I found it was easier to understand and gave less things to do to the `maybe_` functions, that pretended to be really simple things.
i will have a look, but I don't promise I will change it XD

lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
arbulu89: In fact not.
We don't even have any projector with `strong` consistency.
The unique event handler and command needed this consistency is the `roll up` logic.
This was the first created event handler, and I guess we copied it from it.
I will remove it and open a tech-debt ticket to remove from other event handler to not pollute this PR.

arbulu89: As a comment, this didn't have any bad side-effect, as we didn't have any event handler with `strong` consistency, the command `dispatch` didn't stop waiting for anything. Anyway, better remove them to avoid confusion and bad usage.

lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
arbulu89: Good idea

lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
arbulu89: Done

lib/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler.ex
arbulu89: Done

test/trento/infrastructure/commanded/event_handlers/sap_system_database_stale_at_event_handler_test.exs
arbulu89: Ok, let's remove it

lib/trento/sap_systems/sap_system.ex
arbulu89: Alright. I was able to merge both conditions in one without touching the `maybe_` functions, just taking care of using them in the correct order.
Maybe it is less explicit, but it works as fine as before, so I'm happy with it!

lib/trento/sap_systems/sap_system.ex
arbulu89: @skrech I have created this commit to solve the issue:
https://github.com/trento-project/web/pull/4496/changes/f36a2e345633970d85f33dd51fd9fa2ff6d6dce5

With this, we get the database stale_at field for the `RegisterApplicationInstance` command as well. The same way we do with the `database_health`. This way, we can set correct database stale_at value during `SapSystemRegistered` and `SapSystemRestored`.

So, this fixes as well the case of the initial registration together with the restore.
Without this, it could happen (corner-case, but well, it could happen) that we would register SAP system as in sync even though the related database was stale.
Let me know what do you think

lib/trento/sap_systems/sap_system.ex
```diff
@@ -279,6 +290,14 @@ defmodule Trento.SapSystems.SapSystem do
     |> Multi.execute(&maybe_emit_sap_system_health_changed_event/1)
   end
```
arbulu89, line 292: @skrech this here is an important change.
I have moved this condition above commands that change the database health and stale_at values.
We shouldn't update those values of deregistered systems, as well, they don't exist.
with the other code, we should cover changes of this values on restoration.

I prefer to keep them here, as this is our business logic rule.
I could've put this logic in the event handler, but I think this is the correct place to have it.
The event handler will send the commands, and the aggregate will discard them.

I'm talking about the `UpdateDatabaseHealth` and `UpdateDatabaseStaleAt` commands, which shouldn't do anything on deregistered systems.

This is related to: https://github.com/trento-project/web/pull/4496/changes#r3595442166

lib/trento/sap_systems/sap_system.ex
```diff
@@ -321,30 +321,20 @@ defmodule Trento.SapSystems.SapSystem do
...
-      ) do
-    [
       %SapSystemDatabaseStaleAtChanged{
         sap_system_id: sap_system_id,
         database_stale_at: database_stale_at
       }
-    ] ++
+    end)
+    |> Multi.execute(fn sap_system ->
       maybe_emit_sap_system_data_marked_stale_event(sap_system, database_stale_at)
```
skrech, line 343: Hm... If `stale_at` was `nil` and we received `database_stale_at` as `nil`, then  we'll emit `SapSystemDataMarkedStale`, which is an error. But maybe this case can not happen, because if `database_stale_at` was stale, then `stale_at` goes stale as well. So there is no case where `database_stale_at` changes to nil and the value of `stale_at` is nil, correct?

lib/trento/sap_systems/sap_system.ex
skrech: You have left the comment on an ambiguous line, but I suspect you meant 
https://github.com/trento-project/web/commit/f36a2e345633970d85f33dd51fd9fa2ff6d6dce5#diff-6cb7ee790e4b1436838576fa9ddf1a7edea935fc7e4ec764e98020b29989a74fR293
Correct?

lib/trento/sap_systems/sap_system.ex
```diff
@@ -824,12 +829,15 @@ defmodule Trento.SapSystems.SapSystem do
...
-        database_health: database_health
+        database_health: database_health,
+        database_stale_at: database_stale_at
       }
     end
   end
 
-  # Restore a SAP system when the restore command is received, check for the required instances
+  # Restore a SAP system when the restore command is received, check for the required instances.
+  # stale_at is set to nil as the SAP is restored because the associated database was restored,
```
skrech: This should be `database_stale_at`, correct? And maybe we can improve the end of the sentence:

> ... because the associated database was restored, which can not be stale.

test/trento/sap_systems/sap_system_test.exs
```diff
@@ -3474,6 +3482,113 @@ defmodule Trento.SapSystems.SapSystemTest do
         end
       )
     end
+
+    test "should mark SAP system data as in sync when the SAP is restored and database sync state has changed" do
```
skrech: May be we can improve the name of the test: 

> should mark SAP system data as in sync when the SAP system is restored and database sync state had changed in the meantime.

Or something lie that...

skrech (approved):
Great work @arbulu89 ! Approved!

One clarifying question, though: So, am I correct at my understanding that RegisterApplicationInstance command can have its `database_stale_at` values as nil as well as some timestamp? If so this means that we can register SAP systems with stale DBs. 
However, on SAP System restore, we can only have DBs that are in sync?

test/trento/sap_systems/sap_system_test.exs
```diff
@@ -2784,6 +2808,787 @@ defmodule Trento.SapSystems.SapSystemTest do
...
+                       instance_number: ^instance_number,
+                       stale_at: ^instance_stale_at
+                     }
+                   ]
+                 } = state
+        end
+      )
+    end
+
+    test "should not emit any events when database stale_at is the same" do
```
skrech, line 3392: Question: Is this test realistic? Is it possible to receive `UpdateDatabaseStaleAt` with the same value? Is it even possible to receive `UpdateDatabaseStaleAt` with another value, eg receive that DatabaseStaleAt changed once it was marked stale?

lib/trento/sap_systems/sap_system.ex
arbulu89: Yes, I mean it in the movement of the `execute` which returns `{:error, :sap_system_not_registered}` for `deregistered` systems

lib/trento/sap_systems/sap_system.ex
arbulu89: Exactly.
We cannot have a scenario where `database_stale_at` is not `nil` and `stale_at` is nil.
One directly impacts the other.

Anyway, let me add a guard that works as "documentation" and at the if for some reason in the future we use this command in other circumstances. This `maybe_emit_sap_system_data_marked_stale_event` should only emit the event when the coming `stale_at` 2nd param is not nil.

lib/trento/sap_systems/sap_system.ex
arbulu89: Correct!
Let me try to improve the text

test/trento/sap_systems/sap_system_test.exs
arbulu89: I like it!

test/trento/sap_systems/sap_system_test.exs
arbulu89: > Question: Is this test realistic? Is it possible to receive UpdateDatabaseStaleAt with the same value?

In our current architecture based on the event handler only being the sender of this command, no, we cannot have this scenario.
But I think the aggregate should still being resilient against it, just to be consistent by itself. And as we have the code to handle it, well we can have the test itself.

> Is it even possible to receive UpdateDatabaseStaleAt with another value, eg receive that DatabaseStaleAt changed once it was marked stale?

Again, not with the current overall system. But the aggregate would change it properly if the received command has a different `database_stale_at` value (which I think is correct, to update to the new `database_stale_at` value.

arbulu89:
> One clarifying question, though: So, am I correct at my understanding that RegisterApplicationInstance command can have its `database_stale_at` values as nil as well as some timestamp? If so this means that we can register SAP systems with stale DBs. 

Yes, this is correct, and it can really happen. For example, if the system registers a full database first (example: primary and sceonary), and stops the agent in on of the hosts composing this database. If at this point, the SAP system is registered, it should be set as `stale` as the database is stale

> However, on SAP System restore, we can only have DBs that are in sync?

Yes again. When a database is restored, it must always come from an instance that is sending data by definition, so it cannot be stale (at least, the more I try, I cannot find any scenario like that). Which can happen is that the SAP instances themselves being `stale`, but this is covered by the instances `stale_at` field.

PR #4497, opened 2026-07-14
Create SIDs list properly using filtered array IDs

arbulu89:
### Description

Fix a bug on how the `key` for the SIDs list was created.
The code was using the initial array with an index to get the key in `sids[index]`. This is really buggy, as the the `index` is coming from a different list.
Besides that, the `?.id` is totally wrong, as the items in the list don't have the `id` value, so it always defaults to `undefined`.
We need to use the unique ID as done in the next `key`.

Without this, in scenarios where there are duplicated SID values with different `sap_system_id`s, the list values behave totally wrong, creating duplicated lists everytime you paginate or filter. 
For example:
image

### How was this tested?

UT 

### Documentation changes

No

### DIFF
--- a/assets/js/pages/HostsList/HostsList.jsx
+++ b/assets/js/pages/HostsList/HostsList.jsx
@@ -131,19 +131,21 @@ function HostsList() {
         filterFromParams: true,
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
-        render: (sids, { sap_systems }) => {
-          const sidsArray = uniqBy(sap_systems, getInstanceID);
-          return sidsArray.map((instance, index) => (
-            <span key={`${sids[index]}-${instance?.id}`}>
-              <SapSystemLink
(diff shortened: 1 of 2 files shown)


assets/js/pages/HostsList/HostsList.jsx
```diff
@@ -131,19 +131,21 @@ function HostsList() {
         filterFromParams: true,
         filter: (filter, key) => (element) =>
           element[key].some((sid) => filter.includes(sid)),
-        render: (sids, { sap_systems }) => {
-          const sidsArray = uniqBy(sap_systems, getInstanceID);
-          return sidsArray.map((instance, index) => (
-            <span key={`${sids[index]}-${instance?.id}`}>
```
arbulu89, line 137: This `sids[index]` was the error, as it was using other list that is not running the `.map` function and messing up the `key` generation

skrech (approved):
Great, @arbulu89!

gagandeepb (approved):
Thanks for this.

PR #4498, opened 2026-07-14
Detect stale data for Hosts

skrech:
### Description

Include changes on the backend and frontend that allow for marking host data as being stale when heartbeat is lost from the agent. This change is changing how host health used to be calculated -- before losing heartbeat from agents caused the host health to go to critical. Now, heartbeat is not considered when health is calculated, but rather row it marks the last-known health by being stale.

In addition to that, several components (mainly health icons, stale warning banners and Table rows) received proper ARIA roles so they can be easily targeted in tests. 

### How was this tested?

A lot of tests have been modified, I payed extra attention when changing them but whichever tests are modified, they also include re-factoring, so it might be hard to review.


lib/trento/hosts/host.ex
```diff
@@ -422,33 +423,28 @@ defmodule Trento.Hosts.Host do
   end
 
   def execute(
-        %Host{heartbeat: heartbeat} = host,
+        %Host{heartbeat: heartbeat},
```
arbulu89, line 426: So beautiful! Love proper pattern matching hehe

test/trento/hosts/host_test.exs
```diff
@@ -874,167 +791,89 @@ defmodule Trento.Hosts.HostTest do
...
-      )
-    end
-
-    test "should emit a heartbeat failure and health change for a Host previously in passing status" do
-      host_id = Faker.UUID.v4()
-      host_registered_event = build(:host_registered_event, host_id: host_id)
+    @scenarios [
+      %{
+        test_name:
+          "should emit a successful heartbeat for a Host that hasn't received a heartbeat yet",
```
arbulu89, line 797: nitpick/comment:
i think we can improve the test descriptions.
these copy/pasted ones are not really expressing what's going on in my opinion.
for example this one:
`should emit a successful heartbeat event when a host receives a heartbeat the first time`.
and the next:
`should emit a successful heartbeat event when a host that has a failed heartbeat receives a new one`

I would review all heartbeat related ones.
Anyway, just a nitpick, leave as they are if you think they are good enough

arbulu89 (commented):
Nice thing!
About macros usage, It looks fine.
I think if it was me, I would've done it differently (even leave the individual tests as they were, I don't find some duplication bad there necessarily).

lib/trento_web/controllers/v1/host_json.ex
```diff
@@ -61,7 +61,10 @@ defmodule TrentoWeb.V1.HostJSON do
     broadcast_host(%{host: host})
   end
 
-  def heartbeat_result(%{host: %{id: id, hostname: hostname}}), do: %{id: id, hostname: hostname}
+  def heartbeat_result(%{host: %{id: id, hostname: hostname, stale_at: stale_at}}) do
```
arbulu89: nit:
maybe we can use pattern matching:
```
def heartbeat_result(%{host: %{id: id, hostname: hostname, stale_at: nil}}), do:
     %{id: id, hostname: hostname}
def heartbeat_result(%{host: %{id: id, hostname: hostname, stale_at: stale_at}}), do:
     %{id: id, hostname: hostname, stale_at: stale_at}
```

arbulu89 (commented):
nice progress!

lib/trento_web/controllers/v1/host_json.ex
skrech: You're absolutely right! Done!

assets/.storybook/preview.js
```diff
@@ -3,6 +3,9 @@
 
 import '../../priv/static/assets/app.css';
 
+import React from 'react';
```
arbulu89, line 4: nit:
if possible can we put these 2 imports on top of the other import?
Just to have global imports first, and local imports later

assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
```diff
@@ -44,7 +44,12 @@ function StaleIconWrapper(WrappedIcon) {
     );
 
     return (
-      <div className="flex items-center mx-1">
+      <div
+        className="flex items-center mx-1"
+        role="img"
+        aria-label="Health icon"
+        {...(staleAt && { 'data-stale': '' })}
```
arbulu89, line 56: question:
would it make sense to put `staleAt` as data value there?
Maybe it can come handy at some point.

assets/js/pages/HostsList/HostsList.stories.jsx
```diff
@@ -0,0 +1,59 @@
...
+export const WithStaleHost = {
+  parameters: {
+    storeState: {
+      user: admin,
+      hostsList: {
+        hosts: hosts.map((host, idx) =>
+          idx == 1
+            ? {
+                ...host,
+                stale_at: faker.date.past().toISOString(),
```
arbulu89: comment:
I think `faker` `past` function already returns an ISO string:
https://fakerjs.dev/api/date.html#past
```
faker.date.past() // '2021-12-03T05:40:44.408Z'
```

assets/js/pages/HostsList/HostsList.test.jsx
```diff
@@ -157,6 +158,35 @@ describe('HostsLists component', () => {
       renderWithRouter(StatefulHostsList);
       expect(screen.getAllByText(duplicatedSID).length).toBe(1);
     });
+
+    it('should show stale hosts with correct styling', async () => {
+      const staleDate = '2024-01-01T00:00:00.000Z';
```
arbulu89: question:
what about using `faker.date.past()`?

arbulu89 (commented):
Really really great @skrech 
I dropped some nitpicks (ocd alarm, I cannot ignore it XD).
But feel free to ignore them!

assets/js/common/StaleIconWrapper/StaleIconWrapper.jsx
skrech: Hm... This is maybe too much :/ `data-stale` seems enough in my opinion. Having the data inside the attributes seems too much details for an HTML attribute. If the date itself would be involved in styling or visual representation, then it would make more sense to include it, but I don't think we'd have such case.

I want to just note that `date-stale` with empty string would render "bare-bones" attributre, without a value, eg. `<div data-stale>` (no equals sign and value). Other option would be to include "true" "false" strings but I don't like string values for booleans.

test/e2e/cypress/e2e/host_details.cy.js
```diff
@@ -196,6 +201,72 @@ context('Host Details', () => {
...
+        cy.findAllByRole('row')
+          .filter(':not(:has(th))')
+          .should(($rows) =>
+            $rows.each((index, $row) =>
+              expect($row).not.to.have.class('bg-gray-100')
+            )
+          );
+      });
+     });
+  });
```
skrech, line 225: Experiment: This is how Cypress team recommends structuring the tests -- they advise against page objects and encourage more direct approach to DOM queries. 

If we imagined for a moment that most of our elements are structured so they have proper accessibility descriptions (ARIA), then `findByRole` should read easily in pairs of `role, name`. (NOTE: For now, we can't avoid the long selector strings needed to target some elements but it's definitely possible to remediate this in the future).

Thus:
1. We are free to choose the desired parent command - be it `get()`, `findByRole()` or something else. No longer concating and passing around selector strings.
2. We can freely make small variations by just chaining them to the `findByRole()` anchors, thus not having to go through lots of named functions that are effectively non-discoverable (thus almost not reusable).
3. VERY IMPORTANT: We see the assertions in the test itself. This alone is one of my biggest hurdles with the current implementation.
4. We follow more closely the guidelines of the Cypress maintainers. Our code would look more in-line with other Cypress codebases.

And we can go further. Namely, removing the `base_po.js`. It's a file with globally available all-in-one functions (selecting parent actions, hard-coding selectors and assertions). And we should replace it with:
1. Custom Chai assertions -- for example, we can create a `should(has.health)` assertions that checks our health icons. Another example is `should(is.stale)` that can check for every kind of element having html attr of `data-stale`. Now, there are functions in base_po that handle custom assertions in a strange way -- we either have to give a selector and hard-code `get()` (and if we want to use something else, we need to revert to the current alias (`@pageHealthIcon`) hacks), or start passing prepared cypress chains (eg. get().filter() as argument to the custom assertion in the base_po). Both of these options are not optimal, IMO.
2. Custom actions and tasks -- for example, loading scenarios is perfect candidate for a task, etc. As a while, global actions shared across pages should be moved into actions/tasks.

PS. That being said, I've extracted the same thing content into hosts_details_po.js and it's now commented, so I can easily switch back to Cucumber-style functions. Actually, some of the points above are not mutually exclusive with the Cucumber-style -- the main problems are in `base_ps.js`, that very much remind of OOP inheritance problems.

test/e2e/cypress/e2e/host_details.cy.js
```diff
@@ -196,6 +201,72 @@ context('Host Details', () => {
...
+    before(() => {
+      hostDetailsPage.startAgentHeartbeat();
+      hostDetailsPage.visitSelectedHost();
+    });
+
+    beforeEach(() => hostDetailsPage.stopAgentsHeartbeat());
+
+    it('should mark host data as stale when its agent stops reporting', () => {
+      // Page health icon should be stale
+      cy.findByRole('img', { name: hostHealthIconName, timeout: 20000 }).should(
```
arbulu89: I don't think @vicenteqa will like this.
This breaks the PO oriented implementation.
I think for the time being, we need to move this to PO functions (even if you prefer this)

test/e2e/cypress/pageObject/hana_cluster_details_po.js
```diff
@@ -214,11 +213,18 @@ export const validateAvailableHanaClusterCostOptUrl = () =>
 export const expectedClusterNameIsDisplayedInHeader = () =>
   basePage.pageTitleIsCorrectlyDisplayed(availableHanaCluster.name);
 
-export const expectedClusterHealthIsDisplayedInHeader = () =>
-  basePage.pageTitleHealthIsCorrectlyDisplayed(availableHanaCluster.health);
+export const expectedClusterHealthIsDisplayedInHeader = () => {
+  cy.findByRole('img', { name: /cluster health/i }).as('pageHealthIcon');
+  basePage.healthIconIsCorrectlyDisplayed(
```
arbulu89, line 218: todo:
we need to `return` the result from `basePage.healthIconIsCorrectlyDisplayed`, otherwise the CY chain will fail.
The same with some of the next changes when we don't return

arbulu89 (commented):
Ok, looking good!
Only 2 comments:
- The already discussed thing of moving e2e function to PO file. 
- We are missing many `return`s in new functions. If we don't return, we cannot chain cypress functions

test/e2e/cypress/e2e/host_details.cy.js
skrech: Yes, I was expecting this. I re-worked the test cases to use the abstraction functions from the PO. 
Anyway, @vicenteqa when you're back I want to discuss on some of the details of this topic.

test/e2e/cypress/pageObject/hana_cluster_details_po.js
skrech: Since we're asserting in these functions there is no need to `return`. They are supposed to be top-level abstractions that are self-contained -- get whatever elements are needed and assert on them. `return` would only make sense if we intend to additionally chain them. Since the assertions are bundled in, there is no point in further chaining.

test/e2e/cypress/e2e/host_details.cy.js
arbulu89: @vicenteqa :point_up:

test/e2e/cypress/pageObject/host_details_po.js
```diff
@@ -473,6 +490,56 @@ export const saveChecksSelectionButtonIsDisabled = () =>
...
+  cy.findByRole('img', { name: hostHealthIconName }).as('pageHealthIcon');
+  basePage.healthIconIsMarkedStale('@pageHealthIcon');
+};
+
+export const hostHealthIsMarkedInSync = () => {
+  cy.findByRole('img', { name: hostHealthIconName }).as('pageHealthIcon');
+  basePage.healthIconIsMarkedInSync('@pageHealthIcon');
+};
+
+export const hostStaleBannerIsDisplayed = () => {
```
arbulu89: nitpick:
I'm surprised that this doesn't raise as a formatting error to use arrow pointer option:
```
export const hostStaleBannerIsDisplayed = () => 
    cy.findByRole('alert', { name: stalenessBannerName, timeout: 20000 }).should(
    'be.visible'
  );
```

But I guess it comes along the same thing of not putting returns.

test/e2e/cypress/pageObject/host_details_po.js
```diff
@@ -473,6 +490,56 @@ export const saveChecksSelectionButtonIsDisabled = () =>
...
+  cy.findByRole('alert', { name: stalenessBannerName, timeout: 20000 }).should(
+    'be.visible'
+  );
+};
+
+export const hostStaleBannerIsNotDisplayed = () => {
+  cy.findByRole('alert', { name: stalenessBannerName }).should('not.exist');
+};
+
+export const databaseInstanceRowIsMarkedAsStale = () => {
```
arbulu89: nitpick:
the entries in the SAP instances table can be databases and SAP system application instances.
Maybe the function name is missleading.
Maybe rename to `sapInstanceRowIsMarkedAsStale` as this is the header of the table.

test/e2e/cypress/pageObject/host_details_po.js
```diff
@@ -473,6 +490,56 @@ export const saveChecksSelectionButtonIsDisabled = () =>
...
+      .filter(':not(:has(th))')
+      .should(($rows) =>
+        $rows.each((index, $row) => {
+          expect($row).not.to.have.class('bg-gray-100');
+        })
+      );
+  });
+};
+
+export const markHdpDatabaseAsPresent = () => {
```
arbulu89: todo:
can we move this function to the `API` chapter, to be grouped with other `loadScenario` usages

arbulu89 (approved):
Alright!
All looks good.
I'm just writing down here the fact the many cypress functions don't return cypress chain.
I'm not in favor or against it, just a note for @vicenteqa so he is aware.

PD:
dropped some minor nitpick comments

PR #4499, opened 2026-07-14
Update redux with sap system database stale events

arbulu89:
### Description

Listen to SAP system and database stale change events from the backend and update redux state with them.

Depends on: https://github.com/trento-project/web/pull/4496

### How was this tested?

UT

### Documentation changes

No

### DIFF
--- a/assets/js/lib/model/activityLog.js
+++ b/assets/js/lib/model/activityLog.js
@@ -124,6 +124,8 @@ export const SAP_SYSTEM_ROLLED_UP = 'sap_system_rolled_up';
 export const SAP_SYSTEM_ROLL_UP_REQUESTED = 'sap_system_roll_up_requested';
 export const SAP_SYSTEM_TOMBSTONED = 'sap_system_tombstoned';
 export const SAP_SYSTEM_UPDATED = 'sap_system_updated';
+export const SAP_SYSTEM_DATA_MARKED_STALE = 'sap_system_data_marked_stale';
+export const SAP_SYSTEM_DATA_MARKED_IN_SYNC = 'sap_system_data_marked_in_sync';
 
 // Database events
 export const DATABASE_DEREGISTERED = 'database_deregistered';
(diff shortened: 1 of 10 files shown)


gagandeepb (approved):
Thanks, this looks good to me. 
Some questions:
1. A somewhat related question: does this`stale_at` feature require or entail in some way that the correlation of activity log entries needs some augmentation or adjustment  ?
2. For my understanding: what's the difference between `database_stale_at` and `stale_at`?
3. Possibly a more meta question for later: Would it be more informative/interesting to additionally compute/quantify in some way the level of staleness/"sync lag" while a database  is marked as stale?

arbulu89:
> Thanks, this looks good to me. Some questions:
> 
>     1. A somewhat related question: does this`stale_at` feature require or entail in some way that the correlation of activity log entries needs some augmentation or adjustment  ?

No. Luckily enough, we can reuse the deregistration process manager to dispatch the commands to set the `stale_at` field to different aggregates (sap systems, database, clusters when we implement it), as we can use the same correletion id for all of them. So that part is already covered.

>     2. For my understanding: what's the difference between `database_stale_at` and `stale_at`?

The sap system aggregate logic to become stale_at is the next:
- If at least one of the application instances composing the SAP system goes stale, the system goes stale
- If the database associated the SAP system goes stale, the system goes stale

To cover this, the SAP system needs to store if the database is stale or not, so we can cover these 2 scenarios.
So `database_stale_at` stores if the database is stale or not (and when if it is) and `stale_at` stored the SAP system itself is stale or not (and the second depends on the first one)
 
>     3. Possibly a more meta question for later: Would it be more informative/interesting to additionally compute/quantify in some way the level of staleness/"sync lag" while a database  is marked as stale?

It was not part of the feature requirement. The unique requirement was to set the SAP system as stale if the database becomes stale, using the timestamp of this event as the same for the SAP system `stale_at`.
Anyway, the exact timestamp of when the database and any application instance went stale is stored and available for the user, so they can still "compute" that.

PR #4513, opened 2026-07-16
Bump js-yaml 

balanza:
# js-yaml bump

## `assets/package.json`

Added `overrides` to bump the three transitive occurrences:

| Pulled in by | Before | After |
|---|---|---|
| `@istanbuljs/load-nyc-config` | `3.14.2` | `3.15.0` |
| `@eslint/eslintrc` | `4.1.1` | `4.3.0` |
| `fork-ts-checker-webpack-plugin` → `cosmiconfig` | `4.1.1` | `4.3.0` |

```json
"@istanbuljs/load-nyc-config": { "js-yaml": "3.15.0" },
"@eslint/eslintrc": { "js-yaml": "4.3.0" },
"fork-ts-checker-webpack-plugin": { "cosmiconfig": { "js-yaml": "4.3.0" } }
```

## `test/e2e/package.json`

| Pulled in by | Before | After |
|---|---|---|
| `mocha` | `4.1.1` | `4.3.0` |

```json
"mocha": { "js-yaml": "4.3.0" }
```

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -489,12 +489,12 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.29.7",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
-      "integrity": "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw==",
+      "version": "7.29.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.0.tgz",
+      "integrity": "sha512-9NhCeYjq9+3uxgdtp20LSiJXJvN0FeCtNGpJxuMFZ1Kv3cWUNb6DOhJwUvcVCzKGR66cw4njwM6hrJLqgOwbcw==",
       "license": "MIT",
       "dependencies": {
-        "@babel/helper-validator-identifier": "^7.29.7",
+        "@babel/helper-validator-identifier": "^7.28.5",
         "js-tokens": "^4.0.0",
         "picocolors": "^1.1.1"
       },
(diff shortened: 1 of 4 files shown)


arbulu89 (commented):
Hey @balanza ,
I'm not understanding why this is pointed to `release` branch.
Is this a backport?
Why it is not directly fixed in main? Is it not possible?
Because after all, if you need to create a release with this, it will be sent back to `main`.

balanza:
> Hey @balanza , I'm not understanding why this is pointed to `release` branch. Is this a backport? Why it is not directly fixed in main? Is it not possible? Because after all, if you need to create a release with this, it will be sent back to `main`.

This upgrade is needed to fix the vulnerability [CVE-2026-59869](https://www.cve.org/CVERecord?id=CVE-2026-59869).

On `main` it has been fixed already by #4435. Unfortunately, that PR upgrades several other dependencies, including dependencies not used in version 3.1.x (i.e. all the new imported libraries for the UI agent stuff). 

Although it's not correct, I tried to backport all things anyway in #4507, for the sake of having the vulnerability fixed. That resulted in a [build failure](https://build.opensuse.org/package/live_build_log/devel:sap:trento/trento-web/SLES15-Trento-SP3-SP7/x86_64) on OBS. 

Thus, I decided to go the correct way and update only the needed dependency, after referting the previous changes on #4435

arbulu89 (commented):
Just a last question.
If the issue is solved in main, and there we don't have the `overrides`, why do we need them here?
Specially in the `test/e2e`. In fact, checking this, we still use `js-yaml==4.2.0` in `main` for this.
So, what will happen when we release this? what is going to be backported to main?
This code or current code in main (which apparently is fine)?

balanza:
> Just a last question. If the issue is solved in main, and there we don't have the `overrides`, why do we need them here? Specially in the `test/e2e`. In fact, checking this, we still use `js-yaml==4.2.0` in `main` for this. So, what will happen when we release this? what is going to be backported to main? This code or current code in main (which apparently is fine)?

Merging these changes will generate conflicts because of `tests/e2e/` files. We can remove them from the commit as they are not strictly needed.
Even so, the merge will succeed, but the resulting `assets/package-lock.json` file will be corrupted and need re-generation.
_(To reproduce: `git checkout main && git merge bump-js-yaml`)._

PR #4529, opened 2026-07-21
E2E tests for SAP systems and databases stale feature

arbulu89:
### Description

Add e2e tests for the `stale` feature for SAP systems and databases.

Basically:

SAP systems and databases overviews: 
- Instance and sap system row background is grayed

Details view:
- Health icon has stale mini-icon
- Banner is displayed
- Instances and host rows are grayed

### How was this tested?

e2e

### Documentation changes

No

### DIFF
--- a/test/e2e/cypress/e2e/sap_system_details.cy.js
+++ b/test/e2e/cypress/e2e/sap_system_details.cy.js
@@ -100,7 +100,10 @@ context('SAP system details', () => {
   });
 
   describe('Stale data', () => {
-    before(() => sapSystemDetailsPage.startSapSystemAgentsHeartbeat());
+    before(() => {
+      sapSystemDetailsPage.startSapSystemAgentsHeartbeat();
+      sapSystemDetailsPage.visit();
+    });
 
     after(() => sapSystemDetailsPage.stopAgentsHeartbeat());
 
--- a/test/e2e/cypress/e2e/sap_system_details.cy.js
+++ b/test/e2e/cypress/e2e/sap_system_details.cy.js
@@ -99,6 +99,44 @@ context('SAP system details', () => {
     });
   });
 
+  describe('Stale data', () => {
+    before(() => sapSystemDetailsPage.startSapSystemAgentsHeartbeat());
+
+    after(() => sapSystemDetailsPage.stopAgentsHeartbeat());
+
+    it('should mark SAP system data as stale when an agent composing the system stops reporting', () => {
+      sapSystemDetailsPage.stopSapSystemAgentHeartbeat();
+      sapSystemDetailsPage.sapSystemHealthIsMarkedAsStale();
+      sapSystemDetailsPage.sapSystemStaleBannerIsDisplayed();
+      sapSystemDetailsPage.sapSystemInstanceRowIsMarkedAsStale();
+      sapSystemDetailsPage.hostRowIsMarkedAsStale();
+    });
+
+    it('should mark SAP system data as sync when the agent starts reporting data again', () => {
+      sapSystemDetailsPage.startSapSystemAgentHeartbeat();
+      sapSystemDetailsPage.markSapSystemAsPresent();
+      sapSystemDetailsPage.sapSystemHealthIsMarkedInSync();
+      sapSystemDetailsPage.sapSystemStaleBannerIsNotDisplayed();
+      sapSystemDetailsPage.sapSystemInstanceRowIsMarkedInSync();
+      sapSystemDetailsPage.hostRowIsMarkedInSync();
+    });
+
+    it('should mark SAP system data as stale when an agent with a database instance composing the system stops reporting', () => {
+      sapSystemDetailsPage.stopDatabaseAgentHeartbeat();
+      sapSystemDetailsPage.sapSystemHealthIsMarkedAsStale();
+      sapSystemDetailsPage.sapSystemDatabaseHealthIsMarkedAsStale();
+      sapSystemDetailsPage.sapSystemStaleBannerIsDisplayed();
+    });
+
+    it('should mark SAP system data as sync when the agent with a database instance starts reporting data again', () => {
+      sapSystemDetailsPage.startDatabaseAgentHeartbeat();
+      sapSystemDetailsPage.markDatabaseAsPresent();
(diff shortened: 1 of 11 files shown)


test/e2e/cypress/plugins/index.js
```diff
@@ -50,20 +50,24 @@ module.exports = (on, config) => {
 
       agents.forEach((agentId) => {
         heartbeat(agentId);
-        let interval = setInterval(
+        clearInterval(heartbeatsIntervals[agentId]);
+        heartbeatsIntervals[agentId] = setInterval(
           () => heartbeat(agentId),
```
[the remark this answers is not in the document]
arbulu89: It is fine as it is

test/e2e/cypress/pageObject/base_po.js
```diff
@@ -169,6 +169,18 @@ export const removeTagButtonIsDisabled = () =>
...
+  cy.get(element, { timeout }).should('have.class', 'bg-gray-100');
+
+export const elementIsMarkedInSync = (element) =>
+  cy.get(element).should('not.have.class', 'bg-gray-100');
+
+export const healthIconIsMarkedStale = (icon, timeout = 20000) =>
+  cy.get(icon, { timeout }).should('have.length', 2);
+
+export const healthIconIsMarkedInSync = (icon) =>
+  cy.get(icon).should('have.length', 1);
```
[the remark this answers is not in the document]
arbulu89: No need. Default timeout is good enough

test/e2e/cypress/pageObject/base_po.js
```diff
@@ -169,6 +169,18 @@ export const removeTagButtonIsDisabled = () =>
 export const removeTagButtonIsEnabled = () =>
   cy.get(removeEnv1TagButton).should('not.have.class', 'opacity-50');
 
+export const elementIsMarkedStale = (element, timeout = 20000) =>
+  cy.get(element, { timeout }).should('have.class', 'bg-gray-100');
+
+export const elementIsMarkedInSync = (element) =>
+  cy.get(element).should('not.have.class', 'bg-gray-100');
```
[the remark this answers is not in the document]
arbulu89: No need. Default timeout is good enough

arbulu89:
@vicenteqa I'm putting the PR on review. Some tests fail, but the fix is coming in other PR: https://github.com/trento-project/web/pull/4530
In any case, the tests code is definitive, so it is ready to be reviewed

test/e2e/cypress/e2e/sap_system_details.cy.js
```diff
@@ -99,6 +99,44 @@ context('SAP system details', () => {
     });
   });
 
+  describe('Stale data', () => {
+    before(() => sapSystemDetailsPage.startSapSystemAgentsHeartbeat());
```
vicenteqa: Not sure if we want to test the four scenarios consecutively, but in any case I think we should explicitly call  `visit()`  in this block. It could be placed in the  `before`  hook, or in a  `beforeEach`  if we want each test to start from the details page. Otherwise, these tests rely on whichever URL was left by the previous test.

test/e2e/cypress/pageObject/base_po.js
```diff
@@ -169,6 +169,18 @@ export const removeTagButtonIsDisabled = () =>
...
+
+export const healthIconIsMarkedStale = (icon, timeout = 20000) =>
+  cy.get(icon, { timeout }).should('have.length', 2);
+
+export const healthIconIsMarkedInSync = (icon) =>
+  cy.get(icon).should('have.length', 1);
+
```
vicenteqa, line 183: I guess it can take longer for an element to be marked as stale than for it to transition back to sync, right? Is that why the timeout is only increased for the stale assertions? Just out of curiosity as I saw the same pattern in some other page objects too.

test/e2e/cypress/e2e/sap_systems_overview.cy.js
```diff
@@ -193,6 +195,40 @@ context('SAP Systems Overview', () => {
...
+      sapSystemsOverviewPage.hddDatabaseInstanceRowIsMarkedAsStale();
+    });
+
+    it('should mark SAP system data as sync when the agent with a database instance starts reporting data again', () => {
+      sapSystemsOverviewPage.startHddDatabaseAgentHeartbeat();
+      sapSystemsOverviewPage.markHddDatabaseAsPresent();
+      sapSystemsOverviewPage.nwdSystemDataIsMarkedInSync();
+      sapSystemsOverviewPage.hddDatabaseInstanceRowIsMarkedInSync();
+    });
+  });
```
vicenteqa, line 233: I've spent some time thinking about this 😅 because these tests seem to depend to each other’s state. If one test fails after stopping an agent heartbeat, the next one can inherit stale state and pass or fail for the wrong reason. Maybe we should reset the affected heartbeat/data state from both `beforeEach` and `afterEach`, and keep the final `after` only for stopping the heartbeat intervals. Alternatively, if this is intended to be a single flow, we could make it one test that covers the whole sequence.

vicenteqa (commented):
Nice stuff overall 🙂 I just left a few concerns to discuss

test/e2e/cypress/pageObject/base_po.js
arbulu89: Yes, it always takes at least 10 seconds (the time the HeartbeatFailed event needs to be emitted).
The default 10 seconds timeout is not enough. I have added it there and not in the other, because the other one default 10 seconds timeout is always enough.

test/e2e/cypress/e2e/sap_systems_overview.cy.js
arbulu89: Considering that the 2nd test is reverting the 1st test action, I think that it is fine to have a dependency between them. After all, they are part of the same `describe`. I wouldn't try to restore the state in `before/after` states, because after all, we would still depend on the feature working fine.
I created 2 different tests to make them obvious, as they are 2 related but distinct things, and to keep the test description more accurate and not inflate it.
I personally would leave as they are, but if you have strong opinions about unifying the 2 of them in a single test I will do it.

vicenteqa (approved):

PR #4530, opened 2026-07-22
Handle deregistration process manager old state enriching application stale commands

arbulu89:
### Description

Add MarkApplicationInstanceDataStale command to enrichment protocol. Now, it works the same way as the `DeregisterApplicationInstance` command.
This is needed to handle old states of the deregistration process manager (state created when SAP system and database aggregates where the same thing). When the process manager has that old state, it has `application_instances` that are not really SAP systems anymore, they are leftovers, but as the process manager cannot be upcasted they cannot be removed neither.
This code is used to avoid starting a new SAP system aggregate instance for an ID that in reality belongs to a database.

So in essence, this is needed to handle the process manager that was populated with legacy events. I found it running our e2e regression tests: https://github.com/trento-project/web/pull/4529

We had the same issue with the deregistration back in time:
https://github.com/trento-project/web/pull/2485
Find here more detailed information: https://github.com/trento-project/web/pull/2485#discussion_r1549953390

### How was this tested?

Tests from https://github.com/trento-project/web/pull/4529

### Documentation changes

No


lib/trento/infrastructure/commanded/middleware/enrich_legacy_application_instance_commands.ex
```diff
@@ -0,0 +1,41 @@
+# SPDX-FileCopyrightText: SUSE LLC
```
arbulu89, line 1: comment:
This file is a copy/paste from old `enrich_deregister_application_instance.ex` file, only adding the additional command in the `for` entry and changing the function signature.

nelsonkopliku (approved):

gagandeepb (approved):

PR #4531, opened 2026-07-22
Fix Activity Log date filter pagination flake

vicenteqa:
### Description

Fix a flaky Activity Log Cypress test `'should select correct date filter when changing page'` by waiting for Activity Log API requests after filter and pagination actions. When tests finish by asserting on the UI without ensuring that the request has completed, Cypress can consume that response in the following test because we use `testIsolation: false`. This can lead to using the wrong `end_cursor` when asserting pagination changes.

The same wait was also added to similar filter and pagination flows as a harmless safeguard against pending Activity Log requests leaking into subsequent tests.

### How was this tested?
Running a couple of times this: https://github.com/trento-project/web/actions/runs/29900272792
using this branch and only the affected test suite.

### Documentation changes

No

### Additional information

No


antgamdia (approved):
Thanks!

PR #4533, opened 2026-07-22
Fix flaky test for checks execution & increase timeout for restored links availability in UI

vicenteqa:
### Description

The checks result test can sometimes run for more than three minutes before requesting the Wanda `last execution` endpoint. In that case the access token may have expired, because the app config sets `access_token_expiration: 180` seconds in `config/config.exs`. The request can legitimately return `401`, trigger the refresh-token flow, and then be retried by the app, so the test now waits for that retried request before asserting the returned check results. I added the retry method in the `base_po.js` in case we need to apply the same behavior to some other test suite in the future.

The restored host link assertion also gets a targeted 20s timeout because the link can take longer to reappear after the restore scenario finishes. I know this is not the ideal fix but will lower a bit the flakiness on this test.

### How was this tested?
https://github.com/trento-project/web/actions/runs/29999226232

### Documentation changes

No

### Additional information

N/A

### DIFF
--- a/test/e2e/cypress/pageObject/base_po.js
+++ b/test/e2e/cypress/pageObject/base_po.js
@@ -246,10 +246,22 @@ export const apiDeleteAllUsers = () =>
       )
   );
 
-export const waitForRequest = (requestAlias, timeout = 5000) =>
+const hasRefreshToken = () =>
+  cy.window().then((win) => Boolean(win.localStorage.getItem('refresh_token')));
+
+export const waitForRequest = (
+  requestAlias,
+  { timeout = 5000, retryUnauthorized = true } = {}
+) =>
   cy.wait(`@${requestAlias}`, { timeout: timeout }).then((request) => {
-    if (request.response.statusCode !== 401) return request;
-    return cy.wait(`@${requestAlias}`, { timeout: timeout });
+    if (!retryUnauthorized || request.response.statusCode !== 401) {
+      return request;
+    }
+
+    return hasRefreshToken().then((hasToken) => {
+      if (!hasToken) return request;
+      return cy.wait(`@${requestAlias}`, { timeout: timeout });
+    });
   });
 
 export const preloadTestData = ({ isDataLoadedFunc = isTestDataLoaded } = {}) =>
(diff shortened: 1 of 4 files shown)


test/e2e/cypress/pageObject/base_po.js
```diff
@@ -249,6 +249,15 @@ export const apiDeleteAllUsers = () =>
...
   cy.wait(`@${requestAlias}`, { timeout: timeout });
 
+export const waitForRetriedRequestAfterUnauthorized = (
+  requestAlias,
+  timeout = 5000
+) =>
+  waitForRequest(requestAlias, timeout).then((request) => {
+    if (request.response.statusCode !== 401) return request;
+    return waitForRequest(requestAlias, timeout);
+  });
```
nelsonkopliku: **question:** if the server returns 401 consistently and the response time is under the timeout threshold we could incur in in infinite recursion here 🤔 

Yes, probably very edgy, however what do you think about having also a maximum number of retries?

nelsonkopliku (commented):
Thanks @vicenteqa!

Just one note about which I'd like to hear what you think.

For the rest LGTM.

test/e2e/cypress/pageObject/base_po.js
vicenteqa: hmm I don't see recursion here since the function doesn't call itself, it just performs one extra `waitForRequest` if the first response is `401`

vicenteqa: anyway, after an slack conversation with @arbulu89 I will move this logic to cover every potential issue like this one, by moving the logic to the `waitForRequest` function, and have something like this:

```
export const waitForRequest = (requestAlias, timeout = 5000) =>
  cy.wait(`@${requestAlias}`, { timeout: timeout }).then((request) => {
    if (request.response.statusCode !== 401) return request;
    return cy.wait(`@${requestAlias}`, { timeout: timeout });
  });
```

nelsonkopliku: Right, we're not actually calling `waitForRetriedRequestAfterUnauthorized`.

I saw `waitForRequest` inside a `waitForRequest` and my brain went recursion 🙈 

Anyway thanks for looking into it

arbulu89 (approved):
Nice!
I like the code. 
Let's see if it helps with the flaky test. If the result is good, green light from my side

PR #4534, opened 2026-07-23
Disable AI chatbox launcher when AI is not configured

nelsonkopliku:
### Description

This PR makes sure that the AI chatbox trigger is disabled (with a tooltip pointing to the Profile page) when the user has no provider/model configured.

https://github.com/user-attachments/assets/b42abcb4-0f92-4c71-ad65-413cac08ec61

Following up:
- broadcast of the changes to the AI settings for live update of the app's state
- stop a running AI agent when configuration is cleared + notify user in conversation
- notify user about changing model within a running conversation

### How was this tested?

Automated tests, storybooks, IRL.

### DIFF
--- a/assets/js/common/AIAssistant/ModalFrame/ModalFrame.jsx
+++ b/assets/js/common/AIAssistant/ModalFrame/ModalFrame.jsx
@@ -1,29 +1,58 @@
 // SPDX-FileCopyrightText: SUSE LLC
 // SPDX-License-Identifier: Apache-2.0
 
-import React from 'react';
+import React, { forwardRef } from 'react';
 import { Rnd } from 'react-rnd';
+import { Link } from 'react-router';
 import { EOS_CHAT_BUBBLE_OUTLINED } from 'eos-icons-react';
 import { AssistantModalPrimitive } from '@assistant-ui/react';
 
 import Button from '@common/Button';
+import Tooltip from '@common/Tooltip';
 
-const defaultTrigger = (
+// forwardRef + `{...props}` spread are required so the
+// enabled instance works inside `AssistantModalPrimitive.Trigger asChild`
+// (which injects onClick / data-state / aria-expanded + a ref).
+const ChatboxTrigger = forwardRef(({ disabled = false, ...props }, ref) => (
+  // `{...props}` first so injected behavior (onClick / data-state / aria-expanded)
+  // is kept, but the presentational props below win — notably `type="fab"`, which
+  // `Trigger asChild` would otherwise overwrite with the HTML `type="button"`.
   <Button
+    ref={ref}
+    {...props}
     type="fab"
     size="none"
     className="size-full"
-    data-testid="ai-assistant-trigger"
-    aria-label="Open AI Assistant"
+    disabled={disabled}
+    data-testid={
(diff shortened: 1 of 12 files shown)


assets/js/pages/Layout/Layout.jsx
```diff
@@ -103,6 +103,7 @@ function Layout() {
   }, [isCollapsed]);
 
   const { id, username, email } = useSelector(getUserProfile);
+  const aiConfigured = useSelector(hasAIConfiguration);
```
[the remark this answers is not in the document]
nelsonkopliku: Not necessary as the full profile is then loaded in Guard component. The real thing would be avoiding extra requests to profile (one during login and one in Guard), but this is out of scope from here.

assets/js/common/AIAssistant/ModalFrame/ModalFrame.jsx
```diff
@@ -1,29 +1,58 @@
...
+    data-testid={
+      disabled ? 'ai-assistant-trigger-disabled' : 'ai-assistant-trigger'
+    }
+    aria-label={disabled ? 'AI Assistant is disabled' : 'Open AI Assistant'}
   >
     <EOS_CHAT_BUBBLE_OUTLINED className="fill-white" />
   </Button>
+));
+
+ChatboxTrigger.displayName = 'ChatboxTrigger';
```
arbulu89, line 36: dump question:
what is this `displayName` for?

assets/js/lib/test-utils/factories/users.js
```diff
@@ -34,6 +39,7 @@ export const userFactory = Factory.define(() => ({
   last_login_at: formatISO(faker.date.past()),
   created_at: formatISO(faker.date.past()),
   updated_at: formatISO(faker.date.past()),
+  ai_configuration: aiConfigurationFactory.build(),
```
arbulu89: nit:
can we put this on top of `created_at`?
ocd, but `created_at` and `updated_at` look a nice ending for the factory, as they are more metadata entries

arbulu89 (approved):
Looks nice @nelsonkopliku !

assets/js/common/AIAssistant/ModalFrame/ModalFrame.jsx
nelsonkopliku: just a way to name the ref instead of doing

```js
const Component() {}

const RefComponent = forwardRef(Component);
```

gagandeepb (approved):
Thanks for this. I noticed we activate the component with some placeholder credentials ('asdf'). I wonder if we want to validate/verify the AI api credentials successful acceptance as part of a follow up?

nelsonkopliku:
> Thanks for this. I noticed we activate the component with some placeholder credentials ('asdf'). I wonder if we want to validate/verify the AI api credentials successful acceptance as part of a follow up?
>

Current validation in AI configuration setup is about providing a non blank value.
At first interaction with the AI agent if the API key is not valid an error will be returned and shown to the user.

Validating the correctness of the API Key during configuration setup would mean making at least one request to the ai provider, and it is out of scope with regards to the goal of this line of changes.

That upfront validation would be inherent in case we live load the model list from the providers' API, because in order to call that API the API key would be needed first to make the request and a successful result would imply a vaild api key, even before saving the configuration.

PR #4535, opened 2026-07-23
Broadcast AI configuration creation and clean up

nelsonkopliku:
### Description

This PR adds the ability to broadcast creation and clear up of AI configuration so that the UI can react to it.

UI PR at https://github.com/trento-project/web/pull/4556

### How was this tested?

Automated and IRL.

### DIFF
--- a/lib/trento/ai/configurations.ex
+++ b/lib/trento/ai/configurations.ex
@@ -10,6 +10,7 @@ defmodule Trento.AI.Configurations do
 
   alias Trento.Users.User
 
+  alias Trento.AI.Configurations.Events
   alias Trento.AI.UserConfiguration
 
   alias Trento.Repo
@@ -28,6 +29,10 @@ defmodule Trento.AI.Configurations do
     %UserConfiguration{}
     |> UserConfiguration.changeset(Map.put(attrs, :user_id, user_id))
     |> Repo.insert()
+    |> tap(fn
+      {:ok, _} -> Events.broadcast_created(user_id)
+      _ -> :ok
+    end)
   end
 
   def create_user_configuration(%User{}, _), do: {:error, :forbidden}
@@ -70,6 +75,8 @@ defmodule Trento.AI.Configurations do
   def clear_user_configuration(%User{id: user_id, deleted_at: nil, locked_at: nil})
       when not is_nil(user_id) do
     Repo.delete_all(from u in UserConfiguration, where: u.user_id == ^user_id)
+
+    Events.broadcast_cleared(user_id)
 
     :ok
   end
--- a/lib/trento/infrastructure/ai/pub_sub_configuration_events.ex
+++ b/lib/trento/infrastructure/ai/pub_sub_configuration_events.ex
@@ -1,0 +1,25 @@
+# SPDX-FileCopyrightText: SUSE LLC
+# SPDX-License-Identifier: Apache-2.0
+
+defmodule Trento.Infrastructure.AI.PubSubConfigurationEvents do
+  @moduledoc """
+  Default `Trento.AI.Configurations.Events` implementation: per-user AI
+  configuration lifecycle events carried over `Phoenix.PubSub`.
+  """
+
+  @behaviour Trento.AI.Configurations.Events
+
+  @impl true
+  def subscribe(user_id), do: Phoenix.PubSub.subscribe(Trento.PubSub, topic(user_id))
+
+  @impl true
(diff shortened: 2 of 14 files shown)


lib/trento/ai/configurations.ex
```diff
@@ -28,6 +29,10 @@ defmodule Trento.AI.Configurations do
     %UserConfiguration{}
     |> UserConfiguration.changeset(Map.put(attrs, :user_id, user_id))
     |> Repo.insert()
+    |> tap(fn
```
arbulu89, line 32: nice!

lib/trento/infrastructure/ai/pub_sub_configuration_events.ex
```diff
@@ -0,0 +1,25 @@
...
+  @impl true
+  def broadcast_created(user_id), do: broadcast(user_id, {:ai_configuration, :created})
+
+  @impl true
+  def broadcast_cleared(user_id), do: broadcast(user_id, {:ai_configuration, :cleared})
+
+  defp broadcast(user_id, message),
+    do: Phoenix.PubSub.broadcast(Trento.PubSub, topic(user_id), message)
+
+  defp topic(user_id), do: "ai_user_config:#{user_id}"
```
arbulu89: nitpick:
super ocd, but saw a difference between `configuration` above and `config` here and my alarm ringed XD

arbulu89 (approved):
Clean! Good job

lib/trento/infrastructure/ai/pub_sub_configuration_events.ex
nelsonkopliku: let me change it to `ai_user_configuration` 🙈

nelsonkopliku: Here we go 0627d9cb70884c0641fdb98b5a07d999285f9148

lib/trento_web/channels/ai_assistant_channel.ex
```diff
@@ -133,6 +134,18 @@ defmodule TrentoWeb.AIAssistantChannel do
...
+      :ok ->
+        :ok
+
+      error ->
+        Logger.warning("Cannot subscribe to AI configuration events: #{inspect(error)}")
+
+        {:error, :unable_to_subscribe_to_ai_configuration_events}
+    end
+  end
+
```
gagandeepb, line 148: I wonder why does this need to be a PubSub via a lot module indirection? In my opinion this is implemented much simpler as a GenServer call. Then, instead of a handle_info, we have a handle_call callback implementation. PhoenixChannels are a wrapper around GenServers, after all.

nelsonkopliku: The thing is that AI config are changed in their own lifecycle in their dedicated controller.

Here we want to know when something changed, we do not control the action here, but we listen to it and act accordingly.

An async event based implementation with pubsub here is fitting very well IMHO (and as you mention under the hood we're using GenServers).

What alternative do you propose?

gagandeepb: Note, additionally, that in my understanding, PhoenixChannels are also built as an abstraction on top of PubSub, so in addition to what I have mentioned above, I also see this pattern of usage as breaking the abstraction, for gains/reasons I do not yet understand.

lib/trento/ai/configurations.ex
```diff
@@ -71,6 +76,8 @@ defmodule Trento.AI.Configurations do
       when not is_nil(user_id) do
     Repo.delete_all(from u in UserConfiguration, where: u.user_id == ^user_id)
 
+    Events.broadcast_cleared(user_id)
+
```
gagandeepb, line 80: I think these broadcast_created/cleared calls can be moved to the configuration controller? In my understanding, the broadcast should be on a best effort basis, but here it affects the success signal (:ok or :error, etc.) of a database backed operation that the calling side sees.

lib/trento_web/channels/ai_assistant_channel.ex
gagandeepb: Marking as resolved after sync conversation.

gagandeepb (commented):
Just one comment.

lib/trento/ai/configurations.ex
nelsonkopliku: It does not bubble up to the controller, this function returns `:ok` later

gagandeepb (approved):

PR #4536, opened 2026-07-23
React on AI model changes

nelsonkopliku:
### Description

This PR closes the loop about reflecting changes to the ai settings into the AI assistant chatbox.

When the provider model/pair is changed the following happens:
- the channel is notified about the change
- the running agent is updated with the new provider/model pair, if needed
- the user is notified about the change

https://github.com/user-attachments/assets/80786d72-cbc5-42ff-9e4f-ef523c273edc

### How was this tested?

Automated and IRL.

### DIFF
--- a/assets/js/common/AIAssistant/AIAssistant.jsx
+++ b/assets/js/common/AIAssistant/AIAssistant.jsx
@@ -24,6 +24,8 @@ function AssistantUI({
   onNewThread,
   handleClose,
   disabled = false,
+  modelNotice = null,
+  onDismissModelNotice,
 }) {
   const isEmpty = useAuiState((s) => s.thread.isEmpty);
   const isRunning = useAuiState((s) => s.thread.isRunning);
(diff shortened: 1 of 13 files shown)


antgamdia (approved):
Thanks!

Since we already have the hot-swapping logic via user's settings... I wonder if we could even have it built-in within the chat box it self: imagine a user wanting to send a complex analaysis question to a expensive model, but, once it is solved, they want to switch back to a cheaper model.

Just thinking out loud, not for this PR for sure.

nelsonkopliku:
> Thanks!
> 
> Since we already have the hot-swapping logic via user's settings... I wonder if we could even have it built-in within the chat box it self: imagine a user wanting to send a complex analaysis question to a expensive model, but, once it is solved, they want to switch back to a cheaper model.
> 
> Just thinking out loud, not for this PR for sure.
>

Agree. We can change UX if we want. The underlying feature is *almost* the same

PR #4553, opened 2026-07-27
Mark cluster as stale - backend part

arbulu89:
### Description

Implement stale feature for clusters.
In this case, if a host composing a cluster goes stale, the cluster itself goes stale. Simple as that.
We need to be careful on the deregistration side, as deregistering a stale host can put the cluster data in sync (if the remaining hosts are in sync). 

I have created host related and cluster-wide related events. They are needed to have a clean projection and broadcasting.
It is actually possible to only have 2 events, instead of 4, but this would mean adding certain logic in the projection itself (or a different read model configuration, storing each host stale at data).
This way, we follow a similar design used in the SAP system and databases.

### How was this tested?

UT (e2e tests will come at the end)

### Documentation changes

Not yet

### DIFF
--- a/lib/trento/clusters/cluster.ex
+++ b/lib/trento/clusters/cluster.ex
@@ -95,6 +95,7 @@ defmodule Trento.Clusters.Cluster do
   alias Trento.Clusters.Commands.{
     CompleteChecksExecution,
     DeregisterClusterHost,
+    MarkClusterHostStale,
     RegisterOfflineClusterHost,
     RegisterOnlineClusterHost,
     RollUpCluster,
@@ -111,11 +112,15 @@ defmodule Trento.Clusters.Cluster do
   alias Trento.Clusters.Events.{
     ChecksSelected,
     ClusterChecksHealthChanged,
+    ClusterDataMarkedInSync,
+    ClusterDataMarkedStale,
     ClusterDeregistered,
     ClusterDetailsUpdated,
     ClusterDiscoveredHealthChanged,
     ClusterDistributedHealthChanged,
     ClusterHealthChanged,
+    ClusterHostDataMarkedInSync,
+    ClusterHostDataMarkedStale,
     ClusterHostStatusChanged,
     ClusterRegistered,
     ClusterReplicationHealthChanged,
@@ -149,6 +154,7 @@ defmodule Trento.Clusters.Cluster do
     field :state, Ecto.Enum, values: ClusterState.values()
     field :hosts, {:array, :string}, default: []
     field :offline_hosts, {:array, :string}, default: []
+    field :stale_hosts, {:array, :string}, default: []
     field :selected_checks, {:array, :string}, default: []
     field :rolling_up, :boolean, default: false
     field :deregistered_at, :utc_datetime_usec, default: nil
@@ -369,7 +375,8 @@ defmodule Trento.Clusters.Cluster do
           designated_controller: false
         }
       ) do
-    maybe_emit_host_added_to_cluster_event(cluster, host_id, ClusterHostStatus.online())
+    maybe_emit_host_added_to_cluster_event(cluster, host_id, ClusterHostStatus.online()) ++
+      maybe_emit_cluster_host_data_marked_in_sync_event(cluster, host_id)
   end
 
   def execute(
@@ -385,6 +392,9 @@ defmodule Trento.Clusters.Cluster do
     end)
     |> Multi.execute(fn cluster ->
       maybe_emit_cluster_details_updated_event(cluster, command)
+    end)
+    |> Multi.execute(fn cluster ->
+      maybe_emit_cluster_host_data_marked_in_sync_event(cluster, host_id)
     end)
     |> handle_cluster_health_events(command)
   end
@@ -405,6 +415,9 @@ defmodule Trento.Clusters.Cluster do
       maybe_emit_host_added_to_cluster_event(cluster, host_id, ClusterHostStatus.online())
     end)
     |> Multi.execute(fn cluster -> maybe_emit_cluster_details_updated_event(cluster, command) end)
+    |> Multi.execute(fn cluster ->
+      maybe_emit_cluster_host_data_marked_in_sync_event(cluster, host_id)
+    end)
     |> handle_cluster_health_events(command)
   end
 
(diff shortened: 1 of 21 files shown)


lib/trento/clusters/cluster.ex
```diff
@@ -806,31 +868,84 @@ defmodule Trento.Clusters.Cluster do
        ) do
     cond do
       host_id not in hosts ->
-        %HostAddedToCluster{
-          cluster_id: cluster_id,
-          host_id: host_id,
-          cluster_host_status: cluster_host_status
-        }
+        [
```
arbulu89, line 871: comment:
returning as list to be appendable in a non `multi` usage

lib/trento/clusters/cluster.ex
```diff
@@ -446,15 +455,49 @@ defmodule Trento.Clusters.Cluster do
...
+        cluster_id: cluster_id,
+        host_id: host_id,
+        stale_at: stale_at
+      },
+      %ClusterDataMarkedStale{
+        cluster_id: cluster_id,
+        stale_at: stale_at
+      }
+    ]
+  end
```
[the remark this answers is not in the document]
arbulu89: It is a fair point. In the other hand, returning `nil` looks like a "happy" path thing. But sending this command for a non registered host is something the user shouldn't do intentionally, so maybe this should return some sort of `{:error, xxx}`.
But this would require handling this as well.
In any case, as the command is only dispatched by the process manager, and this will only send a valid host, this use-case is not a real thing in our system. So not having it is not an issue.
It looks more a "purity" thing.
For example, we don't handle this scenario for the deregistration logic without any issue so far.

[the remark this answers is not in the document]
arbulu89: @skrech what do you think about this?

lib/trento/clusters/cluster.ex
```diff
@@ -369,7 +375,8 @@ defmodule Trento.Clusters.Cluster do
           designated_controller: false
         }
       ) do
-    maybe_emit_host_added_to_cluster_event(cluster, host_id, ClusterHostStatus.online())
+    maybe_emit_host_added_to_cluster_event(cluster, host_id, ClusterHostStatus.online()) ++
+      maybe_emit_cluster_host_data_marked_in_sync_event(cluster, host_id)
```
skrech, line 379: Small thing but anyway: I find it a bit awkward to concat events like that. This has two problems:
1. Consistency: I find no downsides on using a Multi every time, even if there is no pressing need for it's usage (eg. subsequent events don't depend on the previous state).
2. Again consistency: `maybe_emit_*` functions some times return single event, some times it returns several. It'd be nice to always return a single event, IMO.

lib/trento/clusters/cluster.ex
```diff
@@ -381,7 +388,8 @@ defmodule Trento.Clusters.Cluster do
     cluster
     |> Multi.new()
     |> Multi.execute(fn cluster ->
-      maybe_emit_host_added_to_cluster_event(cluster, host_id, ClusterHostStatus.offline())
+      maybe_emit_host_added_to_cluster_event(cluster, host_id, ClusterHostStatus.offline()) ++
+        maybe_emit_cluster_host_data_marked_in_sync_event(cluster, host_id)
```
skrech, line 392: Hm... the same thing as above, especially here...

lib/trento/clusters/cluster.ex
```diff
@@ -806,31 +868,84 @@ defmodule Trento.Clusters.Cluster do
...
+       )
+       when length(hosts) > 1 do
+    [
+      %ClusterDataMarkedInSync{
+        cluster_id: cluster_id
+      }
+    ]
+  end
+
+  defp maybe_emit_cluster_data_marked_in_sync_event(_cluster, _host_id), do: []
```
skrech, line 947: Looking at the implementation here and given the previous comments about single event per `maybe_emit_*` functions, I think it's possible to re-work the two new functions into handling their responsibility on their own --> cluster_host_data one handles only the host-related events, while the cluster_data one handles the overall cluster. It seems clearer to me. Since it looks like there is an overlap between the two functions right now. The first one handles an event that should be responsibility of the second function, while the second has a special clause to handles the peculiarities already taken care of in the first one. 

PS: Maybe you'd need a `stale_at` field in the aggregate to be able to determine when to fire the cluster-wide in-sync event since the stale_hosts would be cleared after cluster_host_data_in_sync have been processed.

lib/trento/clusters/cluster.ex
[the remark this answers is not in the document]
skrech: This is a non-issue for me because of the de-registration process manager taking care of this.

lib/trento/clusters/cluster.ex
```diff
@@ -736,6 +780,24 @@ defmodule Trento.Clusters.Cluster do
...
+  def apply(
+        %Cluster{stale_hosts: stale_hosts} = cluster,
+        %ClusterHostDataMarkedInSync{host_id: host_id}
+      ) do
+    %Cluster{cluster | stale_hosts: List.delete(stale_hosts, host_id)}
+  end
+
+  def apply(%Cluster{} = cluster, %ClusterDataMarkedStale{}), do: cluster
+
+  def apply(%Cluster{} = cluster, %ClusterDataMarkedInSync{}), do: cluster
```
skrech, line 799: Hm... these two look strange. I think it won't hurt to save the `stale_at` value in the aggregate? Or maybe drop them all-together if this would work?

lib/trento/clusters/cluster.ex
skrech: As already mentioned, I'm against that. :D

lib/trento_web/controllers/v1/cluster_json.ex
```diff
@@ -14,6 +14,7 @@ defmodule TrentoWeb.V1.ClusterJSON do
     |> adapt_v1()
     |> Map.delete(:sap_instances)
     |> Map.delete(:state)
+    |> Map.delete(:stale_at)
```
skrech, line 17: Hm... why do we need this? What do we use the v1 api for?

skrech (commented):
Cool, thanks @arbulu89 !

I've left some comments that are related to the aggregate emitting/handling of the new events. It's a discussion topic, so if you don't like my recommendations, then we can proceed with merging this. Other than that, I have not found any logical errors.

lib/trento/clusters/cluster.ex
arbulu89: I don't know. I'm of a different opinion.
For me, a `Multi` implies that previous events have effect in the new ones. It tells me that they are dependent.
Consistency is important, but for me is secondary to explaining to the code (and to me) why we are emitting the events and if there is correlation between them.

lib/trento/clusters/cluster.ex
arbulu89: We cannot drop them. An `apply` function needs to exist for every emitted event. It is a must.
Now, we don't use the `stale_at` in the aggregate for anything. We could store it, and it could minimally give us the chance to change so pattern matching. But the result would be the same, and we would be storing redundant that.

lib/trento_web/controllers/v1/cluster_json.ex
arbulu89: It is an old deprecated api version.
As we are backward compatible, we keep them both.
But if possible, we try not to add new fields there, to force the user to use the new `v2` version.
That being said, we could drop older versions in major version upgrades.

lib/trento/clusters/cluster.ex
arbulu89: You hit the nail with the last comment hehe
To do that what you propose initially, we would need to include a new stale_at field in the aggregate, otherwise we cannot pattern match. And at the end, it is a redundant value only used for convenience (which is not necessarily bad, but in this case I didn't find it useful).

lib/trento/clusters/cluster.ex
arbulu89: Some docs about adding aggregate fields: https://github.com/commanded/commanded/blob/main/guides/Aggregates.md#effective-aggregate-state--tip
Now, if we added it, we could use it to handle commands, so it is not entirely true that it wouldn't be used, but definitely it is not a mandatory thing

test/trento/clusters/cluster_test.exs
```diff
@@ -2187,6 +2192,255 @@ defmodule Trento.ClusterTest do
...
+            host_id: host_2_id,
+            stale_at: stale_at
+          },
+          # an already stale host does not emit any event
+          %MarkClusterHostStale{
+            cluster_id: cluster_id,
+            host_id: host_1_id,
+            stale_at: stale_at
+          }
+        ],
```
skrech, line 2223: To be honest, if we don't expect this to happen, then it's better not to test it, IMO (only clutters the test). Even if we want to test that, we are better off with another test just for that.

skrech (approved):

test/trento/clusters/cluster_test.exs
arbulu89: Removing that part from the test

PR #4554, opened 2026-07-27
Mark cluster as stale - frontend part

arbulu89:
### Description

Use cluster `stale_at` field in frontend.
- Gray out and apply stale variant to icon in clusters overview list
- Apply stale variant and add banner in cluster details view

Depends on: https://github.com/trento-project/web/pull/4553

### How was this tested?

UT

### Documentation changes

Not yet

### DIFF
--- a/assets/js/pages/ClusterDetails/ClusterDetails.jsx
+++ b/assets/js/pages/ClusterDetails/ClusterDetails.jsx
@@ -28,7 +28,10 @@ import {
   getOperationLabel,
 } from '@lib/operations';
 
+import { formatDateTime } from '@lib/timezones';
+
 import BackButton from '@common/BackButton';
+import Banner from '@common/Banners';
 import Button from '@common/Button';
 import DisabledGuard from '@common/DisabledGuard';
 import OperationsButton from '@common/OperationsButton';
@@ -58,11 +61,13 @@ function ClusterDetails({
   hosts,
   health,
   state,
+  staleAt,
   lastExecution = {},
   operationsEnabled = false,
   runningOperation,
   selectedChecks,
   userAbilities,
+  userTimezone,
   onStartExecution = noop,
   onRequestOperation = noop,
   onCleanForbiddenOperation = noop,
@@ -188,7 +193,12 @@ function ClusterDetails({
       <BackButton url="/clusters">Back to Clusters</BackButton>
       <div className="flex flex-wrap">
         <div className="flex w-1/2 h-auto overflow-hidden overflow-ellipsis break-words">
-          <DetailsViewHeader className="whitespace-normal" health={health}>
+          <DetailsViewHeader
+            className="whitespace-normal"
+            health={health}
+            staleAt={staleAt}
+            timezone={userTimezone}
+          >
             Pacemaker Cluster Details:{' '}
             <span className="font-bold">{clusterName}</span>
           </DetailsViewHeader>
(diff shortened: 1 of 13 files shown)


assets/js/pages/ClusterDetails/ClusterDetails.jsx
```diff
@@ -258,6 +268,13 @@ function ClusterDetails({
           <ClusterStatePill state={clusterState} />
         </div>
       </div>
+      {staleAt && (
+        <Banner type="warning" truncate={false}>
+          An agent in one of the cluster hosts is not reporting since{' '}
```
skrech: Why is the whitespace interpolation needed?

assets/js/pages/ClusterDetails/ClusterDetails.test.jsx
```diff
@@ -54,6 +54,40 @@ describe('ClusterDetails ClusterDetails component', () => {
...
+    expect(
+      screen.getByText(
+        /An agent in one of the cluster hosts is not reporting since 15 Jun 2026, 06:30:00/
+      )
+    ).toBeInTheDocument();
+
+    const header = screen.getByRole('heading', {
+      name: `Pacemaker Cluster Details: ${name}`,
+    });
+    expect(within(header).getAllByTestId('eos-svg-component')).toHaveLength(2);
```
skrech, line 88: Comment: Good enough way of catching stale icons. But we should improve that in the future. We should mark our icons with some HTML attributes to designate `passing-stale` or `passing` state (and so on). Eg. mark our Icons with `role="img" aria-label="Health Indicator" data-element-state="passing-stale"` and then get it with `getByRole('img', '/health indicator/i')` and `expect(healthIcon).toHaveAttribute('data-element-state', 'passing-stale')`

skrech (approved):
Looks great, thank you @arbulu89

nelsonkopliku (approved):

assets/js/pages/ClusterDetails/ClusterDetails.jsx
arbulu89: It is added by the js formatter itself when it cuts the line to avoid a long line.
That being said, if you remove it, the formatter doesn't complain and the result is the same.
I will remove it.

assets/js/pages/ClusterDetails/ClusterDetails.test.jsx
arbulu89: Yes, we can do that. In the past, when we started with react, there were strong opinions about not using `data-testid` etc, to test things, so I tried to avoid it.
That being said, I don't have any reason to dislike that and wouldn't be against it.

assets/js/pages/ClusterDetails/ClusterDetails.jsx
arbulu89: aaa, no, actually we need it. Otherwise it doesn't add the space before the formatted date (I guess because it is inside an interpolation).
PD: The js formatter is still smarter than me XD

arbulu89: image

assets/js/pages/ClusterDetails/ClusterDetails.test.jsx
skrech: Yea, I also prefer to use `getByRole()` if possible (eg. not useing `data-tesetid` and `getByTestId()`). Here the `role` and `aria-label` are used to target the element by `getByRole()` function, while `data-element-type` is an optional attribute that could only easy the assertion, eg not having to match for `eos-svn-component` and css classes.

arbulu89: Ok ok, yes it could be a good improvement then

PR #4557, opened 2026-07-28
Remove GCP scenario from demo

arbulu89:
### Description
Due to some bad conflict resolution `gcp` scenario was added to demo and we don't want that.
This PR removes it back again.

### How was this tested?

No need

### Documentation changes

Not


nelsonkopliku (approved):
Thanks!

PR #4558, opened 2026-07-28
Add debug severity filter to related events link

arbulu89:
### Description

Add `debug` severity to the filter used when `Related events` link is clicked in the activity log entry modal.
Without this, if you click the link in a `debug` event, it will show the related events, but not the clicked one, which is really confusing.
As this feature is a "debugging" tool, it makes sense to add all different severities.
The other option would be to only add `debug` if the event itself is debug, but this would make the usage inconsistent.

### How was this tested?

No need

### Documentation changes

No


gagandeepb (approved):
Thanks for this.

abravosuse (approved):
Thank you!

PR #4559, opened 2026-07-28
Change stale and in sync events severity to debug

arbulu89:
### Description

Change stale and in sync events severity to debug in the activity log.
These events are really verbose and fill up the activity log with noise, specially every time a system is switched off and all of agents stop reporting.

At the end, they don't report any real change in the infrastructure layer. It is more a trento thing, so `debug` looks more appropriate.

Depends on: https://github.com/trento-project/web/pull/4553

### How was this tested?

No need

### Documentation changes

Not yet

### DIFF
--- a/lib/trento/activity_logging/severity_level.ex
+++ b/lib/trento/activity_logging/severity_level.ex
@@ -112,6 +112,8 @@ defmodule Trento.ActivityLog.SeverityLevel do
     "cluster_rolled_up" => :debug,
     "cluster_rollup_requested" => :debug,
     "cluster_tombstoned" => :debug,
+    "cluster_data_marked_stale" => :debug,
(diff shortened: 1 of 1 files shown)


lib/trento/activity_logging/severity_level.ex
```diff
@@ -156,8 +161,17 @@ defmodule Trento.ActivityLog.SeverityLevel do
       },
       condition: :map_value_to_severity
     },
-    "database_instance_deregistered" => :warning,
-    "database_instance_health_changed" => %{
+    "database_instance_marked_absent" => :warning,
+    "database_instance_marked_present" => :info,
```
arbulu89, line 165: comment:
simply changed the order of some entries, to group `database_instance` and `database_` events together

nelsonkopliku (approved):

PR #4561, opened 2026-07-29
Show stale healths in dashboard

arbulu89:
### Description

Display stale information in home view.
We simply need to send the information in the API call and use the stale variant in the icons. If at least one component for the SAP system is stale, the row becomes gray.

I have added e2e tests in the PR as well, as they are pretty small.

Example:
image

image

Depends on: https://github.com/trento-project/web/pull/4498
Depends on: https://github.com/trento-project/web/pull/4554

### How was this tested?

UT and e2e

### Documentation changes

Not yet


assets/js/state/sagas/index.js
```diff
@@ -87,6 +91,30 @@ import { watchOperationEvents } from '@state/sagas/operations';
 
 const RESET_STATE = 'RESET_STATE';
 
+const HEALTH_SUMMARY_REFRESH_EVENTS = [
```
arbulu89, line 93: comment:
Moved the health summary refresh events to this list, so we can iterate and use `all` option to apply 
the `debounce`

lib/trento/sap_systems/services/health_summary_service.ex
```diff
@@ -29,9 +29,11 @@ defmodule Trento.SapSystems.Services.HealthSummaryService do
     |> where([s], is_nil(s.deregistered_at))
     |> order_by(asc: :sid)
     |> Repo.all()
-    |> Repo.preload(application_instances: :host)
-    |> Repo.preload(database_instances: :host)
-    |> Repo.preload(:database)
+    |> Repo.preload([
```
arbulu89, line 32: comment:
I think this is a genuine improvement and a hidden bugfix.
Aside being more efficient when there are a lot of SAP systems (as we don't need to query 2 clusters per SAP system, 1 por app instances other for db instances).
It takes care of the deregistered clusters, as the `HostReadModel` already filters out the preloading the deregistered clusters with the `where: [deregistered_at: nil]` clause.
This way, if a host has a deregistered cluster, it won't be sent in the reponse, as it was now.

assets/js/pages/HealthSummary/HomeHealthSummary.jsx
```diff
@@ -15,6 +18,21 @@ import { getCounters, isMostRelevantPrio } from './summarySelection';
...
+    databaseClusterStaleAt,
+    hostsStaleAt,
+  }) =>
+    classNames({
+      [STALE_ROW]:
+        !!applicationStaleAt ||
+        !!applicationClusterStaleAt ||
+        !!databaseStaleAt ||
+        !!databaseClusterStaleAt ||
+        !!hostsStaleAt,
```
skrech: Maybe `some(Boolean)` would be better?

assets/js/pages/HealthSummary/HomeHealthSummary.stories.jsx
```diff
@@ -94,3 +102,10 @@ export const UnClustered = {
     sapSystemsHealth: unClusteredSummary,
   },
 };
+
+export const Stale = {
+  args: {
+    ...Random.args,
+    sapSystemsHealth: staleSummary,
+  },
+};
```
skrech, line 111: The only thing that Random.args is providing is the `loading` key. Maybe it's better to explicitly set it and not use spread operator?

assets/js/state/sagas/index.js
```diff
@@ -145,86 +171,10 @@ function* watchResetState() {
...
-    loadSapSystemsHealthSummary
-  );
-  yield debounce(
-    debounceDuration,
-    SAP_SYSTEM_RESTORED,
-    loadSapSystemsHealthSummary
+  yield all(
+    HEALTH_SUMMARY_REFRESH_EVENTS.map((event) =>
+      debounce(debounceDuration, event, loadSapSystemsHealthSummary)
+    )
```
skrech: As I understand this, now we are in the case where we debounce every event independently. Eg. receiving a `CLUSTER_HEALTH_CHANGED` within the `DATABASE_HEALTH_CHANGED` debounce period would make `loadSapSystemsHealthSummary` to be executed twice. Only if two `DATABASE_HEALTH_CHANGED` event are received within the debounce period, a single `loadSapSystemsHealthSummary` would be executed.

I think we can easily put all the events arrray into the `debounce` function. This would solve our problem, I think?

arbulu89: You are right on what you say. But this implies a behaviour change from what we had before.
Before (and with the current code), each event had an independent window to refresh the frontend. If we put all entries in one debounce, all events will affect the debounce window, and most probably it might take longer to refresh (as we usually get many of these notifications at the same time).

I'm fine doing what you say, as it would still be eventual consistent, but it is more than a cosmetic change.

assets/js/state/sagas/index.js
arbulu89: More info about this: https://github.com/trento-project/web/pull/4561/changes#r3750333570

assets/js/state/sagas/index.js
arbulu89: @skrech I have updated one e2e test adding a bigger timeout to avoid flakyness.

lib/trento/sap_systems/services/health_summary_service.ex
```diff
@@ -29,9 +29,11 @@ defmodule Trento.SapSystems.Services.HealthSummaryService do
...
+      :database,
+      {:application_instances, host: :cluster},
+      {:database_instances, host: :cluster}
+    ])
```
skrech, line 36: Sorry, I'm having trouble understanding this. It mixes the complex preload syntax, with the elixir quirks... and by trying to decipher the syntax I exhausted my brain capacity to actually understand what this change is doing.

arbulu89: It simply preloads `application_instances` of this SAP system, the host for each application instance, and cluster of this host.

arbulu89: Related: https://github.com/trento-project/web/pull/4561#discussion_r3675011630

lib/trento/sap_systems/services/health_summary_service.ex
```diff
@@ -42,51 +44,60 @@ defmodule Trento.SapSystems.Services.HealthSummaryService do
...
          database_instances: database_instances,
-         database: %{health: database_health, sid: database_sid},
+         database: %{health: database_health, sid: database_sid, stale_at: database_stale_at},
          database_id: database_id
        }) do
-    all_instances = application_instances ++ database_instances
+    hosts =
+      application_instances
+      |> Enum.concat(database_instances)
+      |> Enum.flat_map(fn %{host: host} -> List.wrap(host) end)
```
skrech: Why the flat_map/wrap calls here? We're expecting only structs to come from the instances list?

assets/js/state/sagas/index.js
skrech: I understand that this would change the behaviour, and I'm fine introducing it into another PR. 
However, I really think that the correct operation should be putting all the events into the debounce. Thus it would combine all the requests from the first one to the last one into a single `loadSapSystemsHealthSummary`. 
Actually, if the events share a single debounce, they'd finish deterministically -- independent of the concrete event received, the debounce would obey the interval counted from receiving the first interesting event. With the current implementation it would enuqeue several requests (one per type) and this would potentially slower establishing the final state (multiple requests need to finish and apply their changes)

arbulu89: But the debounce timer would restart everytime we receive an event in the list, so it would most probably make longer the time to wait until refresh.
Anyway, I'm with you. I think it is even better even if this constraint, as we reduce the number of times we request information from the backend, and the final result is exactly identical.

lib/trento/sap_systems/services/health_summary_service.ex
arbulu89: It is to avoid a typical `map+filter`, to filter out `nil` hosts:
```
|> Enum.map(fn %{host: host} -> host end)
|> Enum.reject(&is_nil/1)
```
We can have `nil` hosts in the preload if the host is deregistered (or simply for transient state as host and sap read models are populated in different threads).

skrech: Ah... OK, then. You have an affection towards concatenating lists :D 
I'd prefer the reject route since it's more clear that the item can be nil, but it's a preference thing, so up to you.

arbulu89: I hear you!
Let's be more explicit, I don't think there are any efficiency drawbacks

arbulu89: Changed

assets/js/state/sagas/index.js
skrech: Ah, you're right. What we need is `throttle` :) This should fit perfectly.

test/e2e/cypress/e2e/home.cy.js
```diff
@@ -11,7 +11,48 @@ context('Homepage', () => {
     homePage.validateUrl();
   });
 
+  describe('Stale data', () => {
+    before(() => {
+      homePage.startNwpSystemAgentsHeartbeat();
+    });
+
+    after(() => homePage.stopAgentsHeartbeat());
```
skrech, line 19: It's always a good practice to not rely on anything executing after the tests. Next set of tests should prepare their necessary conditions on their own. Eg. it shouldn't matter in what state you're leaving the system. If any tests have dependency on this, they should arrange for that themselves.

test/e2e/cypress/e2e/home.cy.js
```diff
@@ -11,7 +11,48 @@ context('Homepage', () => {
     homePage.validateUrl();
   });
 
+  describe('Stale data', () => {
+    before(() => {
+      homePage.startNwpSystemAgentsHeartbeat();
```
skrech, line 16: It looks to me that this should be beforeEach hook? Otherwise we rely on shared state between test cases.

test/e2e/cypress/e2e/home.cy.js
```diff
@@ -11,7 +11,48 @@ context('Homepage', () => {
...
+      homePage.nwpSystemRowIsMarkedStale();
+      homePage.nwpApplicationInstancesHealthIsMarkedAsStale();
+      homePage.nwpHostsHealthIsMarkedAsStale();
+    });
+
+    it('should mark the application cluster health as stale when a clustered application instance agent stops reporting', () => {
+      homePage.stopNwpClusteredApplicationInstanceAgentHeartbeat();
+      homePage.nwpSystemRowIsMarkedStale();
+      homePage.nwpApplicationClusterHealthIsMarkedAsStale();
+    });
```
skrech, line 33: Shouldn't hosts always be stale if you're stopping heartbeats?

test/e2e/cypress/e2e/home.cy.js
```diff
@@ -11,7 +11,48 @@ context('Homepage', () => {
...
+  });
+
   describe('Deregistration', () => {
+    before(() => {
+      homePage.visit();
+      homePage.validateUrl();
+    });
+
+    after(() => homePage.restoreNwpSystemData());
```
skrech: I'd remove this because of the aforementioned concerns that it's not good to have after hooks. It might server as some extra protection, but I personally prefer to not give ourselves  aids that make us rely on cleanup actions in the test suite.

test/e2e/cypress/e2e/home.cy.js
arbulu89: This is not for the next tests, it is to stop the heartbeat at all from cypress.
Otherwise, the `task` continues working and we have a zombie process sending heartbeats until cypress is closed (and this happens often if you run cypress with a GUI)

assets/js/state/sagas/index.js
arbulu89: I have doubts about how `throttle` will behave.
It is true that in our system the events that dependent between each other happen really quickly (for example, putting one host stale will put SAP system or database running on the host stale). We would broadcast the event to the frontend, and with throttle it would fetch the data at this point. If at this point, all the read models were not updated, we would get an initial "eventual" state, that wouldn't really show the current scenario, maybe for some milliseconds.
With `debounce` we wouldn't have this intermediate states, we would always have the latest state (even if it takes longer).
I'm not sure if what I say would really happen, as it depends on timings, but in theory it can happen.
What do you think?

test/e2e/cypress/e2e/home.cy.js
arbulu89: PD: otherwise, we would need to stop the agents heartbeat en each `before` clause, which for me is worst.

test/e2e/cypress/e2e/home.cy.js
arbulu89: > Otherwise we rely on shared state between test cases.

I personally find this totally acceptable in tests that are within the same `describe`. Otherwise, all the setting and teardown of scenario becomes really really cumbersome.

For example, in this case, putting the `startNwpSystemAgentsHeartbeat` in `before` or `beforeEach` tests a different thing, because they are different behaviours (the tests would pass the same though).

test/e2e/cypress/e2e/home.cy.js
arbulu89: Yes, it is already tested in the first test case. Do we want to add it in all tests?

test/e2e/cypress/e2e/home.cy.js
arbulu89: I find it valuable and a good practice, but well, as you command

test/e2e/cypress/e2e/home.cy.js
arbulu89: Anyway, I will change it...

skrech: I guess we're reaching a difference in testing approach. I'm always striving to isolate the state between tests, even if they are into the same describe block (and thus running sequentially). Especially in e2e test with complex state management like ours, it might be acceptable to re-use state across test... but I personally would do it if the reverting "transaction" is really heavy.

test/e2e/cypress/e2e/home.cy.js
arbulu89: Added

test/e2e/cypress/e2e/home.cy.js
skrech: Hm... this is very similar to the state we're re-using across the whole test-suite, in my opinion. I personally am maybe fine with leaving these heartbeats to juggle the whole time.

However, it's not that big of a deal to stop them after the describe block. Maybebetter to stop them after the whole spec has executed... 

Again, for me it's the same if thing as with state changes in every test -- it's probably better to revert to the old well-known initial state. But then, it only ever matters if the following test depends on that. If not, I'll just leave it hanging there... unless it's something that burns CPU cycles like crazy. :D 

Up to you.

skrech (approved):
Thank you very much for this @arbulu89!

Had some nice discussion along the way, it was fruitful journey for me :)

test/e2e/cypress/e2e/home.cy.js
arbulu89: At least it bothers me personally. If I run the test suite, and afterward I want to play with my local dev environment, I cannot have a normal behaviour, because I'm still receiving "zombie" heartbeats from a process that should've be finished.
For me it is not optional.
Now, whether we do it here, or anywhere else, I don't mind.

In any case, I don't think we have any general consensus among:
- How to setup/teardown tests
- Tests isolation
- What really should be a test, a group of tests etc
- Do we make exceptions for certain tests?

Maybe, we should have some guidelines to avoid recurring discussions about difference of opinion.

skrech: Yes, you're right. Just leave it in this after hook. The more I think about it the more it make sense to stop the heartbeats as soon as possible. Because they are constantly running, and not a static state.
It's just when I see 'after' my alarm goes away if it's really needed. :D

PR #4563, opened 2026-07-30
Cancel AI assistant in-flight run

nelsonkopliku:
Clicking "New chat" mid-stream swapped threadID and cleared the messages, but nothing settled the run: AgUiThreadRuntimeCore kept isRunningFlag set and WebSocketAIAgent kept the old subscriber alive until the server sent RUN_FINISHED. The new thread therefore rendered an optimistic empty assistant message with a spinner and hid the send button, so the user could not prompt until the abandoned answer finished.

Frontend:
  - AssistantChatProvider clears the thread with `import({messages: []})` and then cancels a run that is still in flight. `import()` rather than `reset()` because the external-store runtime's `reset()` leaves its own message repository untouched; clear-before-cancel because `cancelRun()` re-applies a pre-cancel snapshot a macrotask later.
  - The runtime gets an `onCancel` that routes to the transport, so the abort signal fires before the subscriber settles and the abandoned run stays silent instead of surfacing as a failure.
  - WebSocketAIAgent grows `cancelActiveRun()` (pushes `cancel_run`, completes the subscriber) and ignores RUN_STARTED/FINISHED/ERROR whose runId belongs to a run that is no longer active.

Backend:
  - AIAssistantChannel handles `cancel_run`: cancels the agent's current run and clears `:loading`. That flag is channel-scoped, so without clearing it the first prompt of the new chat would hit the double-send guard and get dropped with no reply at all.
  - Trento.AI.Agent.cancel/1 delegates to Sagents.AgentServer.cancel/1 rather than stop/1: stopping goes through AgentServer.terminate/2, which waits up to 25s for a running task. A dead agent name would exit the caller, so the call is wrapped in `catch :exit`.

### Description

Fixes #<issue>

### Documentation changes

Yes / No


nelsonkopliku:
Splitted into:
- https://github.com/trento-project/web/pull/4648
- https://github.com/trento-project/web/pull/4649
- https://github.com/trento-project/web/pull/4650
- https://github.com/trento-project/web/pull/4651

PR #4569, opened 2026-08-03
Alert heartbeat failed event with an email

arbulu89:
### Description

Alert heartbeat failed event with an email.
When the event is emitted, the event handler catches it and sends an email.
We need to read from the Host read model to get other information like clusters, app and db instances.

I have moved the email to a layout template, so we can reuse it without copying most of the things.

image

### How was this tested?

UT and e2e tests

### Documentation changes

No


lib/trento/infrastructure/alerting/emails/email_templates/layouts/base_email.html.heex
```diff
@@ -0,0 +1,420 @@
+<!--
```
arbulu89, line 1: comment:
copied from the already existing 2 emails, to create this reusable template

lib/trento/infrastructure/alerting/emails/email_alert.ex
```diff
@@ -7,6 +7,10 @@ defmodule Trento.Infrastructure.Alerting.Emails.EmailAlert do
   import Swoosh.Email
   use TrentoWeb, :html
 
+  alias Trento.Infrastructure.Alerting.Emails.EmailLayout
```
skrech, line 10: Question: This is needed because the embedded template needs it?

lib/trento/infrastructure/alerting/emails/email_alert.ex
```diff
@@ -53,6 +57,37 @@ defmodule Trento.Infrastructure.Alerting.Emails.EmailAlert do
...
+    body =
+      %{
+        host_id: host_id,
+        hostname: hostname,
+        failed_at: Calendar.strftime(failed_at, "%Y-%m-%d %H:%M:%S UTC"),
+        cluster: cluster,
+        application_instances: application_instances,
+        database_instances: database_instances
+      }
+      |> heartbeat_failed()
```
skrech: The names of the template and parent function has the same name but different arity so it works... it's a bit confusing, maybe? However, not a big deal.

skrech (approved):
Thank you @arbulu89.

It's very good idea to move the common parts into a Layout!

lib/trento/infrastructure/alerting/emails/email_alert.ex
arbulu89: Because `heex` layout templates are compiled into this module with the `embed_templates "email_templates/layouts/*"`.
Without this alias, we couldn't render `EmailLayout.base_email` in the other templates embedded in this template.
We could perfectly have the `base_email.html.heex` template have in the `email_templates` and avoid the alias, but I think it is nice to have this contextual separation

lib/trento/infrastructure/alerting/emails/email_alert.ex
arbulu89: You are right.
Let me rename the template file (and thus the function) to `heartbeat_failure`. This way all the templates file names are more consistent and we don't have name collision.

arbulu89: Done

PR #4572, opened 2026-08-03
Delete events in batch

balanza:
Unbounded deletes can result in timed-out operations as full table scans are required. 

This PR modifies the deletion algorithm so that the delete query is limited and the prune job is executed in batches. No user-facing changes are introduced.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -13,6 +13,12 @@ import Config
 config :trento,
   ecto_repos: [Trento.Repo]
 
+# Number of rows deleted per individual statement when pruning discovery
+# events. Pruning is performed in batches so that, when a large backlog of
+# historical events has to be pruned, no single DELETE statement runs long
+# enough to exceed the default Ecto checkout timeout.
+config :trento, prune_batch_size: 1_000
+
 # Configures the endpoint
 config :trento, TrentoWeb.Endpoint,
   url: [host: "localhost"],
--- a/lib/trento/discovery.ex
+++ b/lib/trento/discovery.ex
@@ -32,6 +32,20 @@ defmodule Trento.Discovery do
 
   @type command :: struct
 
+  # Number of rows deleted per individual statement. Pruning is performed in
+  # batches so that, when a large backlog of historical events has to be pruned,
+  # no single DELETE statement runs long enough to exceed the default
+  # DBConnection/Ecto checkout timeout (15s) and get cancelled (ERROR 57014).
+  @default_prune_batch_size 1_000
+
+  # Upper bound on the number of delete batches a single prune run may perform.
+  # It caps the bounded iteration in delete_in_batches/3 so pruning never runs
+  # unbounded; in practice the iteration halts earlier, as soon as a batch
+  # deletes fewer rows than the batch size (meaning no more matching rows are
+  # left). The bound is deliberately high (10_000 batches * 1_000 rows = 10M
+  # rows per run) to stay well above any realistic backlog.
+  @default_max_prune_batches 10_000
+
   @doc """
   Transform a discovery in a list of commands event by using the appropriate policy.
   Store the event in the discovery events log for auditing purposes and dispatch the commands.
@@ -105,24 +119,71 @@ defmodule Trento.Discovery do
   defp prune_events(days) do
     end_datetime = Timex.shift(DateTime.utc_now(), days: -days)
 
-    {events_number, nil} =
-      DiscoveryEvent
-      |> where([d], d.inserted_at <= ^end_datetime)
-      |> Repo.delete_all()
-
-    events_number
+    delete_in_batches(
+      DiscoveryEvent,
+      dynamic([d], d.inserted_at <= ^end_datetime),
+      prune_batch_size()
+    )
   end
 
   @spec prune_discarded_discovery_events(number) :: non_neg_integer()
   defp prune_discarded_discovery_events(days) do
     end_datetime = Timex.shift(DateTime.utc_now(), days: -days)
 
-    {discarded_events_number, nil} =
-      DiscardedDiscoveryEvent
-      |> where([d], d.inserted_at <= ^end_datetime)
-      |> Repo.delete_all()
+    delete_in_batches(
+      DiscardedDiscoveryEvent,
+      dynamic([d], d.inserted_at <= ^end_datetime),
+      prune_batch_size()
+    )
+  end
 
-    discarded_events_number
+  @spec delete_in_batches(
+          module(),
+          Ecto.Query.dynamic_expr(),
+          non_neg_integer()
+        ) ::
+          non_neg_integer()
+  defp delete_in_batches(schema, condition, batch_size) do
+    max_batches = max_prune_batches()
+
+    Enum.reduce_while(1..max_batches, 0, fn batch, acc ->
+      ids_subquery =
+        from d in schema,
+          where: ^condition,
+          select: d.id,
+          limit: ^batch_size
+
+      delete_query =
+        from(d in schema,
+          where: d.id in subquery(ids_subquery)
+        )
+
+      {deleted, _} = Repo.delete_all(delete_query)
+
+      next = acc + deleted
+
+      Logger.info(
+        "Pruned #{deleted} #{inspect(schema)} events in this batch " <>
+          "(#{next} deleted so far)"
+      )
+
+      cond do
+        deleted < batch_size ->
+          {:halt, next}
+
+        batch >= max_batches ->
+          Logger.warning(
+            "Pruning of #{inspect(schema)} reached the max batches limit " <>
+              "(#{max_batches}); some events may still be pending and will be " <>
+              "pruned on the next run"
+          )
+
+          {:halt, next}
+
+        true ->
+          {:cont, next}
+      end
+    end)
   end
 
   @doc """
@@ -259,4 +320,10 @@ defmodule Trento.Discovery do
 
   defp commanded,
     do: Application.fetch_env!(:trento, Trento.Commanded)[:adapter]
+
+  defp prune_batch_size,
+    do: Application.get_env(:trento, :prune_batch_size, @default_prune_batch_size)
+
+  defp max_prune_batches,
+    do: Application.get_env(:trento, :max_prune_batches, @default_max_prune_batches)
 end
--- a/test/trento/discovery_test.exs
+++ b/test/trento/discovery_test.exs
@@ -115,6 +115,107 @@ defmodule Trento.DiscoveryTest do
     assert 0 == DiscardedDiscoveryEvent |> Trento.Repo.all() |> length()
   end
 
+  describe "prune discovery events in batches" do
+    setup do
+      original_batch_size = Application.get_env(:trento, :prune_batch_size)
+      Application.put_env(:trento, :prune_batch_size, 3)
+
+      on_exit(fn ->
+        if original_batch_size do
+          Application.put_env(:trento, :prune_batch_size, original_batch_size)
+        else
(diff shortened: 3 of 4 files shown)


test/trento/discovery_test.exs
```diff
@@ -115,6 +115,65 @@ defmodule Trento.DiscoveryTest do
     assert 0 == DiscardedDiscoveryEvent |> Trento.Repo.all() |> length()
   end
 
+  describe "prune discovery events in batches" do
+    setup do
+      original_batch_size = Application.get_env(:trento, :prune_batch_size)
```
nelsonkopliku, line 120: **nit:** `original_batch_size` is going to be `nil`, as long as we don't actually add something in config.

I don't mind eitherway. Just raising my hand about too careful setup compared to actual state of things 😄 

Up to you wether you want to add an actual entry in config or not.

test/trento/discovery_test.exs
```diff
@@ -115,6 +115,65 @@ defmodule Trento.DiscoveryTest do
...
+          Application.put_env(:trento, :prune_batch_size, original_batch_size)
+        else
+          Application.delete_env(:trento, :prune_batch_size)
+        end
+      end)
+
+      :ok
+    end
+
+    test "should delete all outdated discovery events even when they exceed the batch size" do
```
nelsonkopliku: **suggestion:** looks like the following might be reflecting the actual test more realistically.

```suggestion
    test "should batch delete outdated discovery events" do
```

That said, we might be missing a test about events being less than the batch size, hence all of them are actually deleted.

test/trento/discovery_test.exs
```diff
@@ -115,6 +115,65 @@ defmodule Trento.DiscoveryTest do
...
+          inserted_at: DateTime.utc_now()
+        )
+
+      assert :ok == Discovery.prune_discovery_events(10)
+
+      remaining = DiscoveryEvent |> Trento.Repo.all() |> length()
+      assert length(newer_events) == remaining
+    end
+
+    test "should delete all outdated discarded discovery events even when they exceed the batch size" do
```
nelsonkopliku: **(ditto) suggestion:** looks like the following might be reflecting the actual test more realistically.

```suggestion
    test "should batch delete outdated discarded discovery events" do
```

That said, we might be missing a test about events being less than the batch size, hence all of them are actually deleted.

nelsonkopliku (commented):
Thank you, Emanuele! Very valuable.

Just a tiny comment.

test/trento/discovery_test.exs
balanza: Yep, I added a default size of 1000

lib/trento/discovery.ex
```diff
@@ -101,28 +101,52 @@ defmodule Trento.Discovery do
...
+        select: d.id,
+        limit: ^batch_size
 
-    discarded_events_number
+    delete_query =
+      from(d in schema,
+        where: d.id in subquery(ids_subquery)
+      )
+
+    {deleted, _} = Repo.delete_all(delete_query)
```
gagandeepb: I wonder if instead of configuring and controlling the batch size from application code for maintainance actions like pruning, we instead allow a configurable `timeout` for the Repo action: 
https://ecto.hexdocs.pm/Ecto.Repo.html#module-shared-options

I think with the configurable timeout, we let Postgres do the heavy lifting of scheduling/planning the query, the Trento admin just needs to wait a bit longer for larger queries to execute fully.

lib/trento/discovery.ex
```diff
@@ -101,28 +101,52 @@ defmodule Trento.Discovery do
...
   @spec prune_discarded_discovery_events(number) :: non_neg_integer()
   defp prune_discarded_discovery_events(days) do
     end_datetime = Timex.shift(DateTime.utc_now(), days: -days)
 
-    {discarded_events_number, nil} =
-      DiscardedDiscoveryEvent
-      |> where([d], d.inserted_at <= ^end_datetime)
-      |> Repo.delete_all()
+    delete_in_batches(DiscardedDiscoveryEvent, dynamic([d], d.inserted_at <= ^end_datetime))
+  end
```
gagandeepb: Additionally, can some one confirm we that we have an index on this timestamp column `inserted_at`? I don't see an index at least in my local Postgres instance. 🙈

gagandeepb (commented):
I really think we should evaluate/think through the questions raised here, before we proceed with a certain/specific line of solution to this problem.

lib/trento/discovery.ex
```diff
@@ -101,28 +101,52 @@ defmodule Trento.Discovery do
...
+
+  defp prune_batch_size,
+    do: Application.get_env(:trento, :prune_batch_size, @default_prune_batch_size)
+
   @spec prune_events(number) :: non_neg_integer()
   defp prune_events(days) do
     end_datetime = Timex.shift(DateTime.utc_now(), days: -days)
 
-    {events_number, nil} =
-      DiscoveryEvent
```
gagandeepb, line 109: Same question as here: https://github.com/trento-project/web/pull/4572/changes#r3704939861 
No index on `inserted_at` from what I see.

nelsonkopliku (approved):
I am fine with the solution, and I would be fine with the timeout based alternative as well, it would be a one-liner only, possibly.

The discriminating thing could be that batching would very unlikely require any intervention by the user (1000 rows deletion would stay within the 15s default timeout, unless under heavy load), whereas a timeout based solution might still need the user to change the timeout based on the specific case.

What if we also change the default timeout to 30/60 seconds, besides batching? 😅 

We might be even better covered, no?

lib/trento/discovery.ex
arbulu89: No, we don't have an index.

arbulu89 (commented):
I don't have strong opinions between using configurable timeout vs batches.
But what @gagandeepb said about having an index for `inserted_at` makes total sense.
I think we should add an index for `discovery_events` and `discarded_discovery_events`

config/config.exs
```diff
@@ -13,6 +13,12 @@ import Config
 config :trento,
   ecto_repos: [Trento.Repo]
 
+# Number of rows deleted per individual statement when pruning discovery
+# events. Pruning is performed in batches so that, when a large backlog of
+# historical events has to be pruned, no single DELETE statement runs long
+# enough to exceed the default Ecto checkout timeout. Tunable at runtime.
```
arbulu89: comment:
by the way, I think we need to remove the `Tunable at runtime` part, as well, it is not tunable.
At least in the current code.
I don't think we should make it, at the end, i don't think it is something the user should play with and it would require additional documentation

lib/trento/discovery.ex
```diff
@@ -101,28 +101,52 @@ defmodule Trento.Discovery do
     :ok
   end
 
+  # Number of rows deleted per individual statement. Pruning is performed in
+  # batches so that, when a large backlog of historical events has to be pruned,
+  # no single DELETE statement runs long enough to exceed the default
+  # DBConnection/Ecto checkout timeout (15s) and get cancelled (ERROR 57014).
+  # Read at runtime so it can be tuned via config without recompiling.
```
arbulu89: comment:
same here with `runtime` mention. It is not configurable right now on runtime

lib/trento/discovery.ex
```diff
@@ -101,28 +101,52 @@ defmodule Trento.Discovery do
...
+  @default_prune_batch_size 1_000
```
arbulu89: nit:
could we put the constant in top of the code, maybe after `@type command :: struct`

and `defp prune_batch_size` function at the bottom, after `commanded` function, to keep things grouped for context and similarity

lib/trento/discovery.ex
```diff
@@ -101,28 +107,44 @@ defmodule Trento.Discovery do
     :ok
   end
 
+
```
arbulu89: nit:
can we avoid adding these empty new lines here?
They look out of place

lib/trento/discovery.ex
```diff
@@ -101,28 +107,44 @@ defmodule Trento.Discovery do
...
-    {discarded_events_number, nil} =
-      DiscardedDiscoveryEvent
-      |> where([d], d.inserted_at <= ^end_datetime)
-      |> Repo.delete_all()
+    delete_in_batches(DiscardedDiscoveryEvent, dynamic([d], d.inserted_at <= ^end_datetime))
+  end
+
+  @spec delete_in_batches(module(), Ecto.Query.dynamic_expr()) :: non_neg_integer()
+  defp delete_in_batches(schema, condition, acc \\ 0) do
+    batch_size = prune_batch_size()
```
arbulu89: nit:
maybe we could pass the `batch_size` as param, to avoid calling the function everytime, as it is a recursive function

arbulu89 (approved):
Thank you!
Some last nitpick comments. Ignore if you see they don't make sense.
I'm fine either way

lib/trento/discovery.ex
```diff
@@ -105,24 +111,38 @@ defmodule Trento.Discovery do
...
   @spec prune_discarded_discovery_events(number) :: non_neg_integer()
   defp prune_discarded_discovery_events(days) do
     end_datetime = Timex.shift(DateTime.utc_now(), days: -days)
 
-    {discarded_events_number, nil} =
-      DiscardedDiscoveryEvent
-      |> where([d], d.inserted_at <= ^end_datetime)
-      |> Repo.delete_all()
+    delete_in_batches(DiscardedDiscoveryEvent, dynamic([d], d.inserted_at <= ^end_datetime))
+  end
```
gagandeepb, line 138: Consider adding some info/warn level logging here, so that the admin/operator gets some feedback about the state of deletion as it progresses.

lib/trento/discovery.ex
balanza: I have no strong opinions on that. 

The rationale behind the current implementation:
The timeout happening on the client side is the symptom of the core problem, which is: the _delete_ query can get heavy on a densely populated event table. Extending the timeout does not relieve the burden from the database. 
Afaik, a delete query works by appending records to the WAL. Being the `discovery_events` table a write-intensive one, I do not know if a burst of deleting log records can flood the WAL and degrade the data ingestion performance. We might need to collect metrics for this; it's just an educated guess so far. In that regard, we should also consider that we already introduced a possible performance burden on insert by adding the index on the table.

Another rationale:
Deletes are executed on dedicated transactions. That means that all the deletion records are committed altogether once the transaction is done. If any problem occurs during the transaction, the whole job is undone. Being the operation periodical, this would lead to a cumulative size of the recordset to be deleted. I thought that splitting the operation into batches would be a heuristically more resilient solution.


All that is, of course, subject to change if we can perform some data-based evaluation. So far, the rationales have come from heuristics.

lib/trento/discovery.ex
```diff
@@ -105,24 +111,38 @@ defmodule Trento.Discovery do
...
+        where: d.id in subquery(ids_subquery)
+      )
+
+    {deleted, _} = Repo.delete_all(delete_query)
+
+    if deleted == batch_size do
+      delete_in_batches(schema, condition, acc + deleted)
+    else
+      acc + deleted
+    end
```
gagandeepb: I get the sense that this function re-implements a standard library function [(this one)](https://elixir.hexdocs.pm/Enum.html#reduce_while/3) but with unbounded and explicit recursion. It should be possible to bound the recursion with a high enough end state number.

gagandeepb (commented):
Just two comments.

nelsonkopliku (approved):

gagandeepb (approved):
Thanks for this!

arbulu89:
@balanza can we merge this at some point?

PR #4573, opened 2026-08-03
Remove unnecessary strong consistency dispatch

arbulu89:
### Description

Remove unnecessary `strong` consistency dispatch in event handlers. None of them really need strong consistency.
First, because we don't have other event handler with this consistency (except the `rollup`) one. And second, because we don't really need to wait until the projection is done. They are async.

Due the first thing (no projection or event handler with `strong` consistency), this is no-op, as it was being executed as eventual consistency in any case.

It just cleans it up.

### How was this tested?

No need 

### Documentation changes

No


skrech (approved):
Thank you, @arbulu89 !

PR #4574, opened 2026-08-04
Add default version for dependabot

antgamdia:
### Description

Dependabot's Elixir file-fetcher only downloads the manifest files it recognizes into its sandboxed dependabot_tmp_dir. It doesn't know to also fetch a VERSION file just because mix.exs happens to read it at config-evaluation time via File.read!. 
So even though VERSION is committed and present in your normal checkout, it's simply never copied into Dependabot's temp working dir.

```console
"message": "** (File.Error) could not read file \"dependabot_tmp_dir/VERSION\": no such file or directory\n    (elixir 1.19.5) lib/file.ex:435: File.read!/1\n    mix.exs:236: Trento.MixProject.get_version_from_file/0\n    mix.exs:14: Trento.MixProject.project/0\n    (mix 1.19.5) lib/mix/project.ex:1093: Mix.Project.get_project_config/1\n    (mix 1.19.5) lib/mix/project.ex:299: Mix.Project.push_config/2\n    (mix 1.19.5) lib/mix/project.ex:262: Mix.Project.push/3\n\n{\"error\":\"\"}" |                                                                                                                                                                                                                                                                                                                                                                                                                                                     
```

This PR is to fall back to a placeholder version instead of raising an error when VERSION isn't present.

Fixes #TRNT-4622

### How was this tested?

N/A

### Documentation changes

No

### Additional information

- [ ] https://github.com/trento-project/web/pull/4574
- [ ] https://github.com/trento-project/wanda/pull/738

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -241,13 +241,13 @@ defmodule Trento.MixProject do
     end
   end
 
-# Dependabot's hex file fetcher only pulls mix.exs, mix.lock, subapp mixfiles and paths
-# referenced by Code.eval_file/Code.require_file.
-# get_version_from_file/0 reads VERSION,so the reference below is what makes Dependabot fetch it.
-# Do not delete.
-#
-# Code.eval_file("VERSION")
-defp get_version_from_file do
-  __DIR__ |> Path.join("VERSION") |> File.read!() |> String.trim()
-end
+  # Dependabot's hex file fetcher only pulls mix.exs, mix.lock, subapp mixfiles and paths
+  # referenced by Code.eval_file/Code.require_file.
+  # get_version_from_file/0 reads VERSION,so the reference below is what makes Dependabot fetch it.
+  # Do not delete.
+  #
+  # Code.eval_file("VERSION")
+  defp get_version_from_file do
+    __DIR__ |> Path.join("VERSION") |> File.read!() |> String.trim()
+  end
 end
--- a/mix.exs
+++ b/mix.exs
@@ -241,10 +241,13 @@ defmodule Trento.MixProject do
     end
   end
 
-  defp get_version_from_file do
-    case __DIR__ |> Path.join("VERSION") |> File.read() do
-      {:ok, version} -> String.trim(version)
-      {:error, _} -> "0.0.0"
-    end
-  end
+# Dependabot's hex file fetcher only pulls mix.exs, mix.lock, subapp mixfiles and paths
+# referenced by Code.eval_file/Code.require_file.
+# get_version_from_file/0 reads VERSION,so the reference below is what makes Dependabot fetch it.
+# Do not delete.
+#
+# Code.eval_file("VERSION")
+defp get_version_from_file do
+  __DIR__ |> Path.join("VERSION") |> File.read!() |> String.trim()
+end
 end
--- a/mix.exs
+++ b/mix.exs
@@ -243,7 +243,7 @@ defmodule Trento.MixProject do
 
   # Dependabot's hex file fetcher only pulls mix.exs, mix.lock, subapp mixfiles and paths
   # referenced by Code.eval_file/Code.require_file.
-  # get_version_from_file/0 reads VERSION,so the reference below is what makes Dependabot fetch it.
+  # get_version_from_file/0 reads VERSION, so the reference below is what makes Dependabot fetch it.
   # Do not delete.
   #
   # Code.eval_file("VERSION")
--- a/mix.exs
+++ b/mix.exs
@@ -242,6 +242,9 @@ defmodule Trento.MixProject do
   end
 
   defp get_version_from_file do
-    __DIR__ |> Path.join("VERSION") |> File.read!() |> String.trim()
+    case __DIR__ |> Path.join("VERSION") |> File.read() do
+      {:ok, version} -> String.trim(version)
+      {:error, _} -> "0.0.0"
+    end
   end
 end


arbulu89 (commented):
question:
is this not a dangerous change, that could lead us to somehow slip the version 0.0.0?
before this, we didn't have the possibility to have an unintetional version somehow.
just asking, i don't know if there is any other solution though.
i might have a look

arbulu89:
Hey @antgamdia ,
I have looked further, and the end, there is not anything we can do in dependabot part to checkout the file or set and environment variable, so the proposed idea is the valid now.

I have a code change suggestion, to document the logic, to make the function names more consistent with what they do, and protect `prod` compilation for bad usage. It would look like this:

```
# Returns the application version using the following logic:
# 1. VERSION environment variable
# 2. VERSION file content
# 3. Dummy version for dev environment, used by 3rd party tools evaluating
#    the project configuration without a full checkout, like Dependabot
defp get_version do
  version_file = Path.join(__DIR__, "VERSION")
  version_env = System.get_env("VERSION", "")

  case {version_env, File.exists?(version_file)} do
    {"", false} -> get_dummy_version(Mix.env())
    {"", true} -> version_file |> File.read!() |> String.trim()
    {version, _} -> version
  end
end

defp get_dummy_version(:dev), do: "0.0.0"

defp get_dummy_version(_),
  do: raise("VERSION is not available. Set the VERSION env variable or add a VERSION file")
```

I think it covers better what we do and protects "prod".

What do you think?
I'm interesting in @gagandeepb opinion as well

gagandeepb (commented):
I am struggling to understand the end-to-end reasoning/root cause behind the observed behavior (i.e. no dependabot updates for mix deps on current main) and how it relates to this patch. I wonder if @antgamdia can please explain/help me understand, in some degree of detail:

1. Why the updates are not being fetched with current main branch?
2. More interestingly, and why+how this patch addresses the root/underlying issue from 1. 

cc: https://github.com/trento-project/wanda/pull/738

gagandeepb (commented):
.... further more can some more light be shed on uniqueness requirements for the version string? Does this version string need to be unique for every invocation of the underlying GitHub Action for instance? Is it enough to read the current version string or does it need to have some main branch HEAD commit tracking as well, in order to differentiate between duplicate runs, assuming these are version linked ?

nelsonkopliku (commented):
wow, 4 years since last elixir dependabot update.

Shame on us not noticing this 🤦 

Alternatively to @arbulu89 suggestion we have the option to have dependabot use the actual `VERSION` file.

It is a matter of "tagging" `VERSION` as a support file for dependabot.

The trick would be a `Code.eval_file("VERSION")` in a comment

```elixir
# Dependabot's hex file fetcher only pulls mix.exs, mix.lock, subapp mixfiles and paths
# referenced by Code.eval_file/Code.require_file. get_version_from_file/0 reads VERSION,
# so the reference below is what makes Dependabot fetch it. Do not delete.
# Code.eval_file("VERSION")
defp get_version_from_file do
  __DIR__ |> Path.join("VERSION") |> File.read!() |> String.trim()
end
```

See https://github.com/dependabot/dependabot-core/blob/main/hex/lib/dependabot/hex/file_fetcher.rb#L17C7-L18

Yes a bit magic-ish, so I have no strong opinions. The important thing is that we fix it 😅

arbulu89:
> wow, 4 years since last elixir dependabot update.
> 
> Shame on us not noticing this 🤦
> 
> Alternatively to @arbulu89 suggestion we have the option to have dependabot use the actual `VERSION` file.
> 
> It is a matter of "tagging" `VERSION` as a support file for dependabot.
> 
> The trick would be a `Code.eval_file("VERSION")` in a comment
> 
> ```elixir
> # Dependabot's hex file fetcher only pulls mix.exs, mix.lock, subapp mixfiles and paths
> # referenced by Code.eval_file/Code.require_file. get_version_from_file/0 reads VERSION,
> # so the reference below is what makes Dependabot fetch it. Do not delete.
> # Code.eval_file("VERSION")
> defp get_version_from_file do
>   __DIR__ |> Path.join("VERSION") |> File.read!() |> String.trim()
> end
> ```
> 
> See https://github.com/dependabot/dependabot-core/blob/main/hex/lib/dependabot/hex/file_fetcher.rb#L17C7-L18
> 
> Yes a bit magic-ish, so I have no strong opinions. The important thing is that we fix it 😅

@nelsonkopliku If that works, I like it. Less invasive and it just fixes but it is broken and nothing else, in fact using the correct version

gagandeepb (commented):
Repeated invocations of the dependabot job on `main` should ideally not refer to the same static version. The static version read maybe X.Y.Z for a week or more but the main head is continuing to move. Consider doing a `git describe main`, instead of a static version read.  Example: https://github.com/dependabot/dependabot-core/blob/main/hex/spec/dependabot/hex/version_spec.rb#L20

antgamdia:
> Repeated invocations of the dependabot job on `main` should ideally not refer to the same static version. The static version read maybe X.Y.Z for a week or more but the main head is continuing to move. Consider doing a `git describe main`, instead of a static version read. Example: https://github.com/dependabot/dependabot-core/blob/main/hex/spec/dependabot/hex/version_spec.rb#L20

FWIW, this is something we are doing in our golang-based project (agent, mcp server): https://github.com/trento-project/agent/pull/644/changes#diff-8aa632a6f4ad2a3a7fef52661816aa3977f23c0673ef29633c01116fee6745c6R21
I like the idea of relying on the VCS information rather than a static file. However, let's file a separate ticket for that, so that we can unblock dependabot's PRs.


Thanks, Nelson, for the suggestion. Will apply it shortly in both Web and Wanda repos.

nelsonkopliku (approved):
If it works I am happy.

If we want to have dependabot use the dynamic `3.1.5+git.171.1786013781.5f8279a3ea` (for example) we need to revisit what we had [here](https://github.com/trento-project/web/pull/3509/changes#diff-dfa6f4ed74c90e5d4fda283d547d366586e690387289bcfd473e3fa5f9ace308L242) and be mindful of not breaking the release automation.

I don't mind eitherway.

mix.exs
```diff
@@ -241,6 +241,12 @@ defmodule Trento.MixProject do
     end
   end
 
+  # Dependabot's hex file fetcher only pulls mix.exs, mix.lock, subapp mixfiles and paths
+  # referenced by Code.eval_file/Code.require_file.
+  # get_version_from_file/0 reads VERSION,so the reference below is what makes Dependabot fetch it.
```
arbulu89: nitpick: I would put a space in `VERSION,so` after the comma

arbulu89 (approved):
Thank you!
Yes, if this works (we will know 100% when it is merged) I think it is the less invasive logic and it is well documented

mix.exs
antgamdia: Sorry! Me too 😅 . 
Typo fixed, thanks!

gagandeepb (approved):

antgamdia:
@nelsonkopliku , it still fails, but the reason now seems to be different 😆 

```console
+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|                                                                                                                                   Dependencies failed to update                                                                                                                                    |
+-------------------+--------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| Dependency        | Error Type                     | Error Details                                                                                                                                                                                                                                 |
+-------------------+--------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| assent            | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| tzdata            | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| ex_machina        | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| phoenix_live_view | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| phoenix           | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| postgrex          | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| httpoison         | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| pegasus           | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| faker             | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| plug_cowboy       | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| sagents           | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| flop              | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| cloak_ecto        | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| ecto              | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| floki             | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| process_tree      | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| credo             | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| swoosh            | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
| joken             | dependency_file_not_resolvable | {                                                                                                                                                                                                                                             |
|                   |                                |   "message": "Failed to update lockfile: ** (Mix) Command \"git --git-dir=.git checkout --force --quiet a7fd3ec6c755baff382fe79770b076a88b5a81e9\" failed with reason: fatal: unable to read tree (a7fd3ec6c755baff382fe79770b076a88b5a81e9)" |
|                   |                                | }                                                                                                                                                                                                                                             |
+-------------------+--------------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```

I think it is complaining about your fork:

https://github.com/trento-project/web/blob/124bdede5556898cee1de792e5d3b546419be9f0/mix.exs#L63-L66

Since your PR is already merged, I think it time to switch to upstream's again.

PR #4580, opened 2026-08-05
Remove axios

balanza:
Remove `axios` from the project, use the native browser `fetch` for http requests


nelsonkopliku:
@trento-project/developers closing for now. We will possibly re-consider at a later point.

PR #4581, opened 2026-08-05
Disable `New Chat` during a run

nelsonkopliku:
### Description

This PR makes sure that during a run the "New Chat" button is disabled in chat header.

This prevents seeing a "Thinking" spinner on a new chat until the previous run actually finishes.

I am opening this because the more complete cancellation feature at https://github.com/trento-project/web/pull/4563 requires some more love and it is currently lower prio.

Current solution prevents weird situations at least.

### How was this tested?

Automated and IRL


antgamdia (approved):
Thanks! This will prevent impatient users from clicking it and losing the chat context 😅

PR #4585, opened 2026-08-06
Bump langchain ref

nelsonkopliku:
### Description

Previous ref was gone. Whoops


antgamdia (approved):
+1 but can't we use upstream's instead?

arbulu89 (approved):

nelsonkopliku:
> +1 but can't we use upstream's instead?
>

Not yet unfortunately. Mainly because it is not elixir115 compatible

gagandeepb (approved):

PR #4594, opened 2026-08-10
Fix Liz z-index

nelsonkopliku:
### Description

Tiny fix to Liz's z-index so that chatbox is above the main page content but below other full-page modals like settings.

Before

image

After

image

### How was this tested?

IRL


gagandeepb (approved):
Thanks for this.

PR #4597, opened 2026-08-10
Checks execution for ASCS/ERS clusters when SAP system is not registered

arbulu89:
### Description

Fix for ASCS/ERS clusters checks exectution when SAP system is not registered.
I know users should have fully database and SAP systems registered, but trento show this clusters, and the checks execution is available. I think it is good to give the chance to run the correct checks for these cases.

When a SAP system handled by the pacemaker cluster is not registered in Trento, the targets information we create is invalid, as it doesn't create any target.
This resolves not sending any valid check and wanda failing with the next error:

```
[error] Unable to handle message: #GenRMQ.Message<[attributes: %{priority: :undefined, timestamp: 1786372500208, type: :undefined, exchange: "trento.checks", persistent: false, user_id: :undefined, routing_key: "executions", headers: [], consumer_tag: "wanda_checks", content_type: "application/x-protobuf", delivery_tag: 9, redelivered: false, app_id: "gen_rmq", content_encoding: :undefined, correlation_id: :undefined, reply_to: :undefined, expiration: :undefined, message_id: :undefined, cluster_id: :undefined}, payload: <<10, 36, 102, 54, 50, 48, 54, 99, 54, 48, 45, 54, 52, 57, 99, 45, 52, 102, 49, 97, 45, 57, 55, 56, 99, 45, 101, 53, 57, 55, 50, 50, 101, 101, 102, 102, 53, 99, 18, 29, 103, 105, 116, 104, 117, 98, 46, 99, 111, 109, 47, 116, 114, 101, 110, 116, 111, 45, 112, 114, 111, 106, 101, 99, 116, 47, 119, 101, 98, 26, 3, 49, 46, 48, 34, 35, 84, 114, 101, ...>>]>. Reason: :no_checks_selected

```

This causes `web` staying with a spinner saying the checks are being executed forever.

As an additional improvement, the catalog won't show for this cluster checks which need `ensa_version` metadata information, so they cannot select the check itself.

### How was this tested?

UT and some manual testing

### Documentation changes

No

### DIFF
--- a/assets/js/state/selectors/cluster.test.js
+++ b/assets/js/state/selectors/cluster.test.js
@@ -29,6 +29,7 @@ import {
   getFilesystemType,
   getEnsaVersion,
   MIXED_VERSIONS,
+  UNKNOWN_ENSA_VERSION,
 } from './cluster';
 
 describe('Cluster selector', () => {
@@ -276,6 +277,31 @@ describe('Cluster selector', () => {
     };
 
     expect(getEnsaVersion(state, clusterID)).toEqual('ensa1');
+  });
+
+  it('should return an unknown ENSA version if the SAP system is not registered', () => {
+    const clusterID = faker.string.uuid();
+    const cluster = clusterFactory.build({ id: clusterID });
+
+    const state = {
+      hostsList: {
+        hosts: hostFactory.buildList(2, { cluster_id: clusterID }),
+      },
+      databasesList: {
+        databases: [],
+        databaseInstances: [],
+      },
+      sapSystemsList: {
+        sapSystems: [],
+        applicationInstances: [],
+        databaseInstances: [],
+      },
+      clustersList: {
+        clusters: [cluster],
+      },
+    };
+
+    expect(getEnsaVersion(state, clusterID)).toEqual(UNKNOWN_ENSA_VERSION);
   });
 
   it('should return mixed ENSA versions', () => {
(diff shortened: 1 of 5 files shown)


lib/trento/clusters.ex
```diff
@@ -274,38 +274,22 @@ defmodule Trento.Clusters do
...
-
-    clustered_sap_instances =
-      sap_instances
-      |> Enum.map(& &1.instance_number)
-      |> Enum.uniq()
-
     hosts_data =
       Repo.all(
         from h in HostReadModel,
-          join: a in ApplicationInstanceReadModel,
```
arbulu89, line 291: comment:
this `join` is the issue.
If no `ApplicationInstance` is available (which can happen if the SAP system is not fully registered), it would return an empty list, so we wouldn't send any valid target.

arbulu89:
Hey @skrech ,
I have seen that the we have received an answer in the bugzilla ticket that the fix worked.
I think we could review and merge this PR.
What do you think?

assets/js/state/selectors/cluster.test.js
```diff
@@ -278,6 +279,31 @@ describe('Cluster selector', () => {
...
+      },
+      sapSystemsList: {
+        sapSystems: [],
+        applicationInstances: [],
+        databaseInstances: [],
+      },
+      clustersList: {
+        clusters: [cluster],
+      },
+    };
```
skrech, line 302: Question: Shouldn't there be a factory for this?

skrech (approved):

assets/js/state/selectors/cluster.test.js
arbulu89: The cluster data is in fact created by a factory.

PR #4598, opened 2026-08-10
See all available and usable models for a given provider

gagandeepb:
### Description

Enables loading latest available models  from the dependency `llm_db` from the three providers of interest. Allows only chat models, which have tool calling and reasoning enabled to be made available/select-able from the profile view. Filters out all 'preview' class models from 'google'. 

### How was this tested?

Automated tests and env/IRL tests.


config/prod.exs
```diff
@@ -15,4 +15,9 @@ config :trento, Trento.SoftwareUpdates.Discovery,
 config :trento,
   operations_enabled: true
 
+config :llm_db,
+  filter: %{
+    deny: %{google: ["*-preview*"]}
+  }
+
```
gagandeepb, line 22: I can consider moving this to `config.exs` as the default filter unless over-ridden otherwise.

nelsonkopliku: Before doing so I have a question: why are we filtering out all `preview` models?

test/trento/support/struct_helper_test.exs
```diff
@@ -26,7 +26,7 @@ defmodule Trento.Support.StructHelperTest do
       }
 
       assert %{
-               "not_existing_atom" => "some-value",
+               :not_existing_atom => "some-value",
```
nelsonkopliku: **issue:** while this change makes the tests pass it breaks the initial point of the test itself.

Funnily enough llmdb lib is the cause of it starting to fail as it introduces an actual `:not_existing_atom` which actually exists [here](https://github.com/agentjido/llmdb/blob/c791970cec967a246af83cd9ea18ed4290b671a4/lib/llm_db/config.ex#L225).

Let's rename to something like this (name up to you)
```suggestion
               "not_existing_atomzz" => "some-value",
```
and also the initialization a few lines above

```elixir
initial_map = %{
   # ...
  "not_existing_atomzz" => "some-value",
   # ...
}
```

config/dev.exs
```diff
@@ -255,6 +255,12 @@ config :samly, Samly.Provider,
     }
   ]
 
+config :llm_db,
+  filter: %{
+    allow: %{openai: "gpt-5.4", google: :all, anthropic: :all},
```
nelsonkopliku: **suggestion:** let's allow all openai models in dev.

```suggestion
    allow: %{openai: :all, google: :all, anthropic: :all},
```

lib/trento/ai/llm_registry.ex
```diff
@@ -6,28 +6,42 @@ defmodule Trento.AI.LLMRegistry do
...
+      end
+
+    models = LLMDB.models(provider)
+
+    models
+    |> Enum.filter(fn m ->
+      ModelHelpers.streaming_tool_calls?(m) && ModelHelpers.reasoning_enabled?(m) &&
+        ModelHelpers.chat?(m)
+    end)
+    |> Enum.map(& &1.id)
```
nelsonkopliku: **suggestion:** I found that LLMDB has native querying/filtering capabilities https://llm-db.hexdocs.pm/LLMDB.html#candidates/1

I think we can leverage that and save the extra `ReqLLM` dependency

```elixir
defmodule Trento.AI.LLMRegistry do
  #... 

  # Only chat models able to call tools while streaming, and to reason, are usable by the agent.
  @required_capabilities [
    chat: true,
    reasoning: true,
    tools: true,
    streaming_tool_calls: true
  ]

  #... 

  @doc """
  Returns the list of models for a given provider or all models if `:all` is passed.
  """
  @spec get_provider_models(atom() | :all) :: [bitstring()]
  def get_provider_models(:all), do: Enum.flat_map(@providers, &get_provider_models/1)

  def get_provider_models(provider) when provider in @providers do
    scope = Map.get(%{googleai: :google}, provider, provider) # this to be removed when renaming

    [scope: scope, require: @required_capabilities]
    |> LLMDB.candidates()
    |> Enum.map(fn {_provider, model_id} -> model_id end)
    |> Enum.sort()
  end

  def get_provider_models(_), do: []
  
  # rest...
end
```

test/trento/ai/llm_registry_test.exs
nelsonkopliku, line 1: **suggestion:** Let's not leave this module totally untested.

I think we can do something along these lines https://github.com/trento-project/web/commit/42a390f7c6a5900b13040d3d4881b27efdca6657#diff-f194a85498c7697f931155d7a57bd6ec3943fda5d31437a553a1a000833f2c42

nelsonkopliku (commented):
Thanks Gagan, I like the direction.

I left my comments inline and you can find [at this branch](https://github.com/trento-project/web/commit/42a390f7c6a5900b13040d3d4881b27efdca6657) the working suggested changes.

Happy to answer any question

test/trento/support/struct_helper_test.exs
gagandeepb: Good catch, I will address this.

config/prod.exs
gagandeepb: Discussed sync. This ensures that only fully released models are available for real usage, since we cannot be sure about the general/long term availability of 'preview' class models.

config/dev.exs
nelsonkopliku: Hey @gagandeepb, as discussed in order to not have this limitation here we need to adjust the e2e tests so that they assume less about the actual models in the page, as they might change.

- https://github.com/trento-project/web/commit/80b7b9db329430677e66c25b21ea2b43189c3836 this is a general improvement of how we select items in a dropdown. Current implementation would have troubles with `gpt-5` and `gpt-5.4`. The proposed commit is not strictly necessary but it shouldn't harm 
- https://github.com/trento-project/web/commit/e778770759fb4136e1bebea0fcdb78934b50508b this is the actual change. It makes sure to not assume any actual model in the page but rather works on picking the first in the list and then any other randomically.

This should fix your issues with e2e tests 😉

gagandeepb: Thanks for this. Would you be open to sending these changes as a separate PR for general E2E test improvements, even though motivated by this PR. I can rebase my branch on main thereafter, after review/merge?

nelsonkopliku: Sure, here we go https://github.com/trento-project/web/pull/4654

test/trento/ai/llm_registry_test.exs
gagandeepb: Re-added the module but with a different/newer set of assumptions of the functions' contract with callees.

config/prod.exs
antgamdia: Question: don't recall right now, but is this setting modifiable by end users easily?? If not, I don't really think we should impose any limitations around the usable models. We might have scenarios where those preview models are cheaper (because they are not subject to the usual EULA/SLAs) and users want to take the risk in favor of saving some money.

So, I personally would either remove the limitation ot ensure it is easily overrideable by admins.

gagandeepb: @antgamdia Cost/usage-based-pricing related assumptions/steps are out of scope for this line of work. We can reconsider if/when we take these into account. We can also to reconsider if/when we get actual user requests wanting to use preview class models. Until then we are in speculative premature optimization territory, in my opinion.

config/dev.exs
gagandeepb: Rebased and addressed.

config/dev.exs
```diff
@@ -255,6 +255,11 @@ config :samly, Samly.Provider,
     }
   ]
 
+config :llm_db,
+  filter: %{
+    deny: %{google: ["*-preview*", "*-image*"]}
+  }
```
nelsonkopliku: **suggestion:** let's not filter out anything from dev.

mix.exs
```diff
@@ -57,6 +62,7 @@ defmodule Trento.MixProject do
   # Type `mix help deps` for examples and options.
   defp deps do
     [
+      {:llm_db, "~> 2026.7", runtime: false},
```
nelsonkopliku, line 65: **question:** why are we installing the dependency as `runtime: false` but then in releases we are actually loading it?

test/trento/support/struct_helper_test.exs
```diff
@@ -12,9 +12,16 @@ defmodule Trento.Support.StructHelperTest do
       date = Date.utc_today()
       naive_datetime = NaiveDateTime.utc_now()
 
+      # We need to make the atom unique for testing purposes since
+      # there has been a case where the pre-existing atom was acquired
+      # /allocated by a newly added dependency thereby causing this test
+      # to fail.
+      ref = make_ref()
+      some_unique_atom = "not_existing_atom_#{:erlang.term_to_binary(ref)}"
+
```
nelsonkopliku, line 21: **nit:** something like inline "not_existing_trento_atom" would have been sufficient 😅 

Or even a uuid instead of a ref.

lib/trento/ai/llm_registry.ex
```diff
@@ -6,53 +6,39 @@ defmodule Trento.AI.LLMRegistry do
...
-
-  @doc """
-  Checks if a given provider is supported.
-  """
-  @spec provider_supported?(atom()) :: boolean()
-  def provider_supported?(provider) when is_atom(provider), do: provider in providers()
-
-  def provider_supported?(_), do: false
-
-  defp get_ai_providers_config, do: Keyword.get(ApplicationConfigLoader.load(), :providers, [])
```
nelsonkopliku, line 57: We are lucky that for now the only place we need this logic is `Trento.AI.UserConfiguration` so moving the burden of this logic is safe for now.

The moment this logic is needed somewhere else we'll have to move around again.

I can live with this, but it looks to me an escamotage to not write some tests 😅

nelsonkopliku (approved):
Looks good.

I think we just need to remove the models filtering from dev.

The rest is something I can live with.

mix.exs
gagandeepb: The dep was previously (now deprecated functionality) started as part of the application tree. But now it is (moving towards) a pure library. With the highlighted change, we instruct to not start the dep (using the deprecated interface) but only to to load it in the release so we have the non-deprecated interface available to us for our needs.

test/trento/support/struct_helper_test.exs
gagandeepb: Refs have a much lower chance of being duplicated/double allocated than a statically defined string (the situation that this is addressing) :sweat_smile: . UUIDs could have been another option, but I ended up choosing a native function :sweat_smile:

lib/trento/ai/llm_registry.ex
gagandeepb: I just wanted a lower public interface surface area, since the logic you mention is trivially implemented as a combination of just two interface functions. :sweat_smile: It would be overkill for the sake of callee convenience to define recombinations of public functions as an expanded public interface. But we can revisit this at some point, if the need arises due to some more reasons.

config/dev.exs
gagandeepb: Done [here](https://github.com/trento-project/web/pull/4598/commits/ec318251b9090ef447f9f78ad31c9353230ecf9d).

PR #4599, opened 2026-08-10
Rename s/googleai/google

gagandeepb:
### Description

s/googleai/google since the latter is a more canonical name in the model/provider scheme of things.

### How was this tested?

Automated tests.

### DIFF
--- a/lib/trento/ai/llm_registry.ex
+++ b/lib/trento/ai/llm_registry.ex
@@ -6,7 +6,7 @@ defmodule Trento.AI.LLMRegistry do
   This module is responsible for managing the registry of available LLM providers and their models.
   """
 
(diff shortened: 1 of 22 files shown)


lib/trento/ai/llm_registry.ex
```diff
@@ -21,19 +21,10 @@ defmodule Trento.AI.LLMRegistry do
...
   def get_provider_models(provider) when provider in @providers do
-    provider =
-      case provider == :googleai do
-        true ->
-          :google
-
-        _ ->
-          provider
-      end
-
```
gagandeepb, line 36: This is not necessary any more after the rename.

nelsonkopliku (approved):
Thanks 🚀

PR #4615, opened 2026-08-13
Adds explicit system env variable

gagandeepb:
Explicit system env. variable for ipv4 vs ipv6 support. Exposed as operator configuration.

### Description

There are cases when the networking stack in the runtime/deployment environment is either ipv4 or ipv6. Existing detection logic based dynamic config does not correctly detect enabled/disabled status of ipv6 as evidenced in some support cases. The ultimate source of truth here is choice of the deployer of the trento-web artifact, and from their deployment (e.g. cloud) environment constraints/choices available to them.

Default networking stack chosen is ipv4, so if nothing is provided, ipv4 is chosen. ipv6 can be opted into, via explicit choice of the deploying person. In the author's understanding the default networking stack in most cloud VM environments is ipv4, but the choice of the default can be evolved/changed if there is more data from cases to support that. 

Note: In the author's understanding, there is no consideration/support envisioned for dual networking ipv4+ipv6 stack, hence the choice between networking stacks is exclusive, i.e. one or other must be chosen, and the choice made entails that the non-chosen networking stack will not be made use of by the trento-web application.

### How was this tested?

The test package/PTF will be tested by the case reporter, after this PR is reviewed by all involved.

### Documentation changes

Docs changes will be needed and will be done subsequently.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -89,7 +89,7 @@ if config_env() in [:prod, :demo] do
   # upon start up.
   # Exposed as a deployment environment variable.
   # If the environment variable is not set it defaults to ipv4.
-  # Accepts either ipv4 and ipv6 as possible values. (Case-insenstive)
+  # Accepts either ipv4 and ipv6 as possible values. (Case-insensitive)
   ip =
     case System.get_env("IPV4_OR_IPV6", "IPV4") |> String.downcase() do
       "ipv4" -> {0, 0, 0, 0}
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -93,8 +93,8 @@ if config_env() in [:prod, :demo] do
   ip =
     case System.get_env("IPV4_OR_IPV6", "IPV4") |> String.downcase() do
       "ipv4" -> {0, 0, 0, 0}
-       "ipv6" -> {0, 0, 0, 0, 0, 0, 0, 0}
-        _ -> raise "Invalid value set for environment variable IPV4_OR_IPV6"
+      "ipv6" -> {0, 0, 0, 0, 0, 0, 0, 0}
+      _ -> raise "Invalid value set for environment variable IPV4_OR_IPV6"
     end
 
   config :trento, TrentoWeb.Endpoint,
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -84,13 +84,17 @@ if config_env() in [:prod, :demo] do
       For example: yourdomain.example.com
       """
 
-  # Returns an IP address to bind the HTTP server, based on IPv6 support.
+  # Returns an IP address to bind the HTTP server, based on explicit operator choice
   # This prevents startup failures on systems without IPv6 where the app would crash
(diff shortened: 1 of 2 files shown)


nelsonkopliku (commented):
Hey @gagandeepb I am not necessarily against this solution.
I am aware that giving access to a specific env gives more control to an operator, however we need to consider that adding an extra env variable means also:
- documenting it
- making sure it is settable via systemd/ansible/helm installation

What do you think about the solution proposed by Samuel?
```elixir
ip =
    case :gen_tcp.listen(0, [:inet6]) do
      {:ok, socket} ->
        :gen_tcp.close(socket)
        {0, 0, 0, 0, 0, 0, 0, 0}

      {:error, _reason} ->
        {0, 0, 0, 0}
    end
```

Yes, it is more _implicit_, as the previous solution was, but it seems to be working (to be tested both on ipv4 and ipv6). Additionally the surface we need to act on is contained.

gagandeepb (commented):
@nelsonkopliku : I don't like including the dynamic behavior entailed by the solution you refer to since it involves listening on a socket inside the `runtime.exs` context which goes against the way this file should be used (reading config from app/system environment). Happy to talk sync.

nelsonkopliku:
> @nelsonkopliku : I don't like including the dynamic behavior entailed by the solution you refer to since it involves listening on a socket inside the `runtime.exs` context which goes against the way this file is should be used (reading config from app/system environment). Happy to talk sync.
>

Don' get me wrong: I think you are right about that aspect.

I just want to make sure @abravosuse is fine with the extra environment variable and what it brings with it.
If he's fine I am as well, and in that case we need to address the correlated points (doc, ansible, helm).

@arbulu89 thoughts?

chargio:
Is the current behavior to attach the server to a port in all available addresses (0:port) or am I missing something?

arbulu89:
At the end, it all falls of the configuration options we want to give the user.
1. If we want to give more, we could expose `BIND_IP` (or whatever name). This could be flexible, but it opens the door for bad usage
2. Or the behaviour implemented in this PR

I'm in favor of having user configuration options, because each user has different needs, and implicit auto detection might not be enough for some of them.

If it was for me, I would go with option 1 giving `::` as default value (it would be the closer default we have now) but @gagandeepb told me that this can create some potential configuration issues with nginx and k3s ingress (anyway, if the user does incorrect things). So, i'm hesitant of if any of our users would really need more control over the binded IP address.

nelsonkopliku (approved):
Thanks Gagan!

If nobody has objections about adding an extra environment variable I am also fine.

Let's possibly adjust the following:
- https://github.com/trento-project/web/blob/main/guides/Development/environment-variables.adoc
- https://github.com/trento-project/web/blob/main/packaging/suse/rpm/systemd/trento-web.example

And let's prepare the changes needed to:
- docs
- ansible
- helm chart

gagandeepb (commented):
@nelsonkopliku What  milestone should this PR belong to ?

nelsonkopliku:
I would say it could safely go with next patch **3.1.6**. WDYT? In that case backport it by labelling with `backport`

PR #4636, opened 2026-08-17
Run npm audit

nelsonkopliku:
### Description

Running a local `npm audit` highlight some vulnerabilities.

Ran a `npm audit fix --force` to have 0 vulnerabilities from that command.

Should also address some dependabot alerts with regards to `js-yaml`.

### DIFF
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -89,7 +89,7 @@
         "intersection-observer": "^0.12.2",
         "jest": "^30.3.0",
         "jest-environment-jsdom": "^30.3.0",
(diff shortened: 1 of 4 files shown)


antgamdia (approved):

gagandeepb (commented):
FTR: This PR was open for an hour or less. 😅

gagandeepb (commented):
Can we not have/propagate such an aggressive turn around time expectation for PR reviews requests? Or are we fine turning these into "approval requests" and not really have any expectations of an actual review from those we expect reviews from?

PR #4642, opened 2026-08-18
JSconfig and move codebase to ESM

skrech:
### Description

Adding `jsconfig.json` file in the root of `assets` and making it the (almost) single source of truth for aliases definition in the project. That change additionally led to an effort to unify our bundlers since we were using a mixture of them (and their configuration files). Ultimately, this led to unifying the import convention for our front-end to use ES Modules (ESM) and marking that into our `package.json` as `type: "module"`.

`tsconfig.json`/`jsconfig.json` are the de-facto standards for configuring editors/IDEs that would work with LSP (language server protocol) on TS/JS codebases. Although they allow for lot of additional configuration, their support for defining module resolver aliases is widely used and adopted by most bundlers. For example, ESBuild reads them out of the box, while plugins are available for every other bundler or supporting tool.

By trying to adapt every bundler that we use to read these files, an idea came to mind to try concentrate on a single bundler. This would ease the maintenance of the project, lower the dependency and lower the risk of errors caused by discrepancies between different bundlers. So the current situation was:
1. ESBuild for production and dev-server builds.
2. Webpack with Babel for Storybook.
3. Only Babel for running test with Jest.
4. Additionally, ESLint import plugin was using Nodejs common import resolver.

For every one of these tools we were configuring the same aliases over and over. So, the story begins as follows:

ESBuild has a builtin support for `jsconfig.json`. So, this part was simple. Just delete the aliases from `build.js` config and that's it.

For Storybook internet resources were pointing that it could use either Webpack or Vite-based "builders" (bundlers for storybooks). And that the Vite builder is using ESBuild (+RollUp) under the hood. So, I started porting the storybook config to Vite. Vite is a hot topic now, seems like the new go-to bundler for new projects and it's all over the startup guides for every prominent framework. In a sense, showing the way new developments are headed. However... in their newest version, v8, they switched away from esbuild+rollup combo to replace it with a rust-based tool called Rolldown. So, here are the first indications that my plan might be harder than I though. JS ecosystem is constantly changing. Anyway, I still think that playing along with these prominent tools would pay off in time. After the changes, staring up storybook is noticeably faster. And there was a plugin for Vite that handles jsconfig.json so we removed the repetition of defining aliases for storybook specifically. Vite is a bit more strict on using JSX syntax  outside of .jsx files, so some changes needed to be made here and there to comply with that.

While writing my Vite config file, I wanted to use ESM syntax. However, if I were to do so, I had to name it with .mjs extension since Vite (as with all the tools that have their config in the `assets` dir) would be executed in Nodejs runtime, and by default it falls back to CommonJS for .js files. But ESM (in my humble opinion) is where JS unification effort seems to be headed (browsers support ESM for a long time, and Node.js supports ESM as all the other major backend runtimes). Every file in our code-base is already using ESM, except our config files for various supporting tools like bundlers, linters, test runners, etc. So, I decided to pull the trigger and mark our project ESM-exclusive by adding `type: "module"` in `package.json`. This signals Nodejs that if it encounter a `.js` file, it should treat is as ESM file (as opposed to CommonJS). So, this meant that I should replace the `require` calls (and some other minor stuff) into `import` calls in all the .js files in our `assests` directory. If we have to work with CommonJS files for some reason, they just need to be renamed with `.cjs` extension to work properly (and we have such an example, Jest... but we'll reach it later) but we expect to have none of them.  

So, next logical step was to proceed to our test runner and make it use jsconfig.json as the rest of the tools. As a reminder, Jest uses Babel directly (no Webpack) to transpile the code before being processed by the test runner. Internet resources showed that there is a plugin for Jest that can swap out the Babel transpiler for ESBuild. So, I did use that plugin. However, it was filled with problems and the fact that it was not updated for 5 years made me not road any longer... I'll elaborate more on why modern transpilers are not working well with Jest. So, I decieded to just add jsconfig.json support into Jest, or Babel more precisely. There was a plugin for Babel that can enhance it's module resolver to also handle jsconfig.json. Although the plugin was intercepting the jsconfig.json and discovered aliases it had one major problem -- it couldn't handle `jest.mock` calls because of their special hoisting behaviour and the design of the tanspile pipeline used in Jest. I went to into very deep research why the babel plugin for jsconfig.json can't replace aliases usages inside jest.mock() calls. And the long story short is: Babel changed how these mocks are represented in code which broke the plugin detection mechanism and some precedence AST parsing mis-matches prevented any kind of workarounds. So, for Jest I just moved the babel config file inside Jest's own config file since Babel was used only for Jest. Jest is the sole exception that needs its own definition of aliases in our project. At later point we might try to tackle the problem again, but this time concentrate on finding a way to configre jest moduleNameMapper directly, and not target the underlying transpiler. 

However, there is more to the Jest situation -- Jest uses some advanced feature from Nodejs runtime to support test isolation. It's a form of virtualization framework that is not fully ported to work well with ESM modules (its status for ESM is 'experimental'). So, the Jest config is the only exception on another front -- it's the only CommonJS file in our repo, so I had to change the filename to .cjs since marking our project with `type: "module"` don't allow CommonJS files to have .js extension. Additionally, I removed all the comments in that config file to be more discoverable. 

So, now we had unified our aliases for LSP (editor/ide) use, production/dev envs and storybook. This was enough for me to wrap up this PR. But... then ESLint's import plugin was not able to handle some of the cutting-edge packaging features used in some of the new JS packages needed to support the above changes (most notably `vite`, but eslint couldn't resolve it's own package it was just ignored...). This is a known issues in the `eslint-plugin-import`. Actually this plugin is the main reason that we can't update to ESLint 10 -- it and several other plugins from the same author are almost abandoned and the community has moved on with alternatives. Additionally, I found out that ESLint has added all our aliases as Ignored, so we were not checking import semantics for wherever we have aliases in our imports. 

Then, I decided to tackle this final problem. I updated the ESLint resolver  plugin to `eslint-plugin-resolve-x` and it now uses another plugin called `eslint-import-resolver-typescript` (let's not get frighten by its name, it's the same resolver that standard `eslint-plugin-resolve-x` uses, just pre-configured to handle TS sytnax as well, and adds support for tsocnfig/jsconfig) that automatically reads our jsconfig.json so it successfully lints our imports everywhere now.

Since I was looking into this part of the code I noticed that our ESLint config is not following the new standard for ESLint configuration (but was rather auto-migrated from older version by a automated tool). So I overhauled it fully -- split the base part for JS from all the rest. Added special section for JSX handling. Then modified slightly the DevDeps and Tests linting configs from before. The file now uses globals package to define expected global variables in the specified environments, not long list of manually enumerated strings.

By the time I did that, I decided that it's a good occasion to actually finish our long awaiting task of upgrading to ESLint 10: https://github.com/trento-project/web/pull/4264 To do that, I just needed to replace jsx-a11y with jsx-a11y-x plugin and replace the `eslint-plugin-react'. The first one was straight forwards.

Replacing eslint-plugin-react was more complex: I had to double check every react customization that we had and compare with its status in the new plugin called `@eslint-react`. Good news was that most of the rules we have disabled, were missing from `@eslint-react` (eslint-react separates the bug-inducing rules from the opinionated style ones, which it provides as copy-paste recipes if you need them, but are not included in the main plugin). I adapted everything and now even disabled some new rules since they bumped the reported warnings for our project to 200. However, we need to attend these, they looked important. I just preserved the old ones + plus some new ones that were not that much -- so from 42warnings, we have about 80 now.

After `@eslint-react` there was nothing else stopping us from updating to ESLint 10. So I did. And now everything should be complete. But... `@eslint-react` dropped support for  Nodejs 20 in it's later versions... So, I had to revert some version behind and change names of some rules... The config still preserves the new names in comments, so it would be easier upgrade once we drop support for Node20... which we should prioritize.

And now... everything is DONE! Almost... E2E test are failing :D The changes I'm doing to the tooling is not affecting the E2E test suite. I look at it as a separate project... and it is, has it's own package.json. Cypress uses a bundler under the hood as well. It's again the same story -- it can be either Vite or Webpack. Since e2e suite is older project it uses webpack. And webpack enforces strict module import restrictions when accessing files from outside its root that are under a project marked as being ESM -- it requires specifying the file extension since this is the default behaviour of Nodejs when working with ESM. And we access files from our test helpers that we share with the unit-tests. So, a rules to loosen up that limitation (fullySpecified:false) needed to be added to the e2e webpack config, to allow accessing these files. 

So, now everything works, has a single source of truth for aliases (jsconfig.json) and the project is prepared for the modern JS ecosystem.

And final remarks -- 
- Marking the project as ESM-exclusive exposed an setup shortcoming from the very deep past of the project. Phoenix project scaffolding uses vendored JS dependency to provide you with working front-end without messing with Node and `npm_modules`. However, we use full-blown SPA frontend using NPM, so there is no longer need to support these vendored deps. For example, the vendored `topbar` was so old that it didn't even supported ESM. So, I had to remove the vendored dep and install it via npm. However, we have to do the same with `phoenix` and drop `phoenix_html` and `phoenix_live_view` at all. I left that exercise for some other time.
- Also, some unnecessary packages were removed that only polluted our package.json, but were never used or were just transitive deps.

Further improvements:
- The only part that I didn't manage to migrate to ESM was Jest. Since Jest is so tightly coupled to Nodejs, community effort to port it to different modern bundlers has either failed (esbuild) or is not even made in the first place -- vite team made their own test runner that is supposed to be drop-in replacement for Jest called `vitest`. It uses way more well integrated transpilation pipeline. Again, it's well adopted, well documented, modern and Fast! I think we should consider migrating to Vitest for our unit tests.
- If we decipher the switch from esbuild to rolldown from a major bundler in the JS ecosystem as a harbinger call -- then we should expect less contribution to esbuild. Because of the aforementioned _non_-limitations from Phoenix, nothing stops us from choosing our own bundler, we're not limited to esbuild from Phoenix. It's worth unifying our bundling tooling into a single one controlled from  a single config file. Vite is a good candidate. 
- We should consider porting our e2e suite to ESM as well, and potentially use Vite there as a bundler as well. People report low Cypress startup times when using the more modern bundler. 

### How was this tested?
Run all tests UT and E2E manually, manually checked storybook, manually inspected all linters and formatters work.


gagandeepb (commented):
Thanks for this. While I am still evaluating most of this PR, I tried running `npm install` and was informed that `npm warn deprecated tsconfck@3.1.6: unmaintained`. Can you please examine and let me know how is this dependency being used in this PR and to what extent are we impacted by this lack of upstream maintenance?

skrech:
Interesting catch @gagandeepb !

Yea, this is a bit tricky -- This is a dependency for the `vite-tsconfig-paths` plugin. https://github.com/trento-project/web/pull/4642/changes#diff-8027b9fdc37a9be1226dc8f7308b4c321edbef2c76c86ace94eaa14ba5fa5c6fR14

Now, you can see the comment in the code as well. I'm not entirely happy with this as well. 
The thing is that Vite has builtin support for `tsconfig.json`... but not for `jsconfig.json`. This is very strange since most other tooling work with both of the files. But the underlying Rolldown library decided not to implement that. So, I added the plugin to workaround that. Apparently it pulls down some deprecated libs (they are derecated in May 2026, so quite soon).

These files are supposed to be absolutely the same with  the sole difference that when something is reading `jsconfig.json` it should automatically add `allowJs: true` option. However, there were two reasons I didn't do it:
1. I didn't like a file called tsconfig.json when we're not using typescript. Just that stylistic reason, which is not a blocker at all.
2. In my preliminary tests when I was using tsconfig.json with `allowJs: true` my LSP did not work. But this might have been a temporary hiccup. Now that the PR is ready, I will try locally if it works and report back.

skrech:
Just double checked the react linter rules. To achieve feature parity with the old linter we need to add the recipes for:
no-string-refs: https://eslint-react.xyz/docs/migrating-from-eslint-plugin-react#no-string-refs
no-duplicate-props: https://eslint-react.xyz/docs/migrating-from-eslint-plugin-react#jsx-no-duplicate-props
function-component-definition: https://eslint-react.xyz/docs/migrating-from-eslint-plugin-react#function-component-definition (the new recipe enforces arrow function, needs to be adapted to function _declarations_, we can probably compare the one in recipes section to the one in quick start guide: https://eslint-react.xyz/docs/packages/kit#quick-start)

assets/js/app.js
```diff
@@ -23,13 +23,16 @@
 //
 
 // Include phoenix_html to handle method=PUT/DELETE in forms and buttons.
-// Phoenix stuff, safe to ignore the import/no-unresolved
-/* eslint-disable import/no-unresolved */
+/* eslint-disable import-x/no-unresolved */
+// Disabling the linter for these 3 imports because they are generated
+// from the back-end and expect a preparation step to be made which
+// complicates CI and adversely affects linting in isolation.
```
nelsonkopliku, line 29: The complications you are referring to are about having the javascript related jobs needing also the mix deps?

nelsonkopliku (approved):
Hey Kris, thanks a lot for this and for sharing the whole story 😄 

Couple of notes:
- I am not against bundler unification, as long as everything keeps working and we do not end up in  forcing something to somewhere it does not belong
- I am fine with whichever new cool bundler we want to introduce or move towards. We've been around enough to know it's not gonna be the last 🙈 
- if these changes unlock other ones (ESLint) that's very positive
- with regards to node20: we need to upgrade the stack as it's EOLed

That said great job! Let's fix the conflicts and we can move forward I believe.

assets/js/app.js
skrech: Yes, exactly. And the fact that there is a matrix of them, so potentially we need to test against node20 with both version of elixir, then against node22 against both version of elixir...

antgamdia (approved):
Thanks a lot for the effort on this! Unifying our frontend setup around a single bundler makes total sense and will definitely help reduce our overall CVE surface area. Thanks for detailed write-up and context as well.

All the changes LGTM, but I haven't had the chance to checkout the changes locally and play around with them. 

Regarding the pending items:

Jest Migration: I haven't tried Vitest yet, but exploring it during upcoming spare/learning time sounds like a great next step. Maybe it is more resilient to flakiness? 

E2E / Cypress: Honestly, I'd love to get rid of Cypress altogether. We could move the suite to Playwright, which supports ESM seamlessly out of the box. Vicente might also be interested :P

Thanks again!

skrech:
Just a note: eventual migration to Vitest might not be that straight forward, though. The most problems would be around mocking setup logic. Vitest won't be a drop-in replacement in that regard... :/

gagandeepb (approved):
Thanks for this. I have seen a sync demo of the editor integration and it is indeed much better than the status quo, in my understanding.

arbulu89 (approved):
Really neat!
I like it.
I have not so far the habit of using editor LSP, but who knows, maybe once we have them properly setup I will start hehe

Real kudos for reducing dependencies, that's always welcomed!

skrech:
I'm delaying this until 3.1.6 is out.

PR #4648, opened 2026-08-19
AI cancel run core

nelsonkopliku:
### Description

This PR wires up the ability to cancel an ongoing LLM request processing.

Cancelling stops current running LLM task and keeps the agent running, ready for another prompt, whereas stopping means actually bringing down the agent process.

Notes:
- previously implemented stop feature now attempts a best effort cancellation first. This means that for instance when clearing ai settings the teardown of the agent is quicker.
- added a `FakeChatModel` test support utility
- the above allows for more integration oriented testing of the processes stack adding up to what unit tests cannot cover by mocking

Fixes #<issue>

### How was this tested?

Automated and IRL with following PRs in the stack of this feature.

### DIFF
--- a/lib/trento/ai/agent.ex
+++ b/lib/trento/ai/agent.ex
@@ -82,11 +82,39 @@ defmodule Trento.AI.Agent do
   @doc """
   Stops the running agent for `agent_id`, terminating any in-flight run.
 
-  Best-effort — returns the supervisor's result verbatim (`{:error, term()}`
-  when no agent is running for the id).
+  It cancels in-flight runs first, then stops the agent process.
+
+  This is faster than only stopping the agent, because `Sagents.AgentServer.terminate/2`
+  waits for the running task to finish (up to 25s) before the process exits.
+
+  Best-effort — the cancel result is discarded and the supervisor's result is
+  returned verbatim (`{:error, term()}` when no agent is running for the id).
   """
   @spec stop(String.t()) :: :ok | {:error, term()}
-  def stop(agent_id), do: AgentSupervisor.stop_agent(agent_id)
+  def stop(agent_id) do
+    cancel(agent_id)
+    AgentSupervisor.stop_agent(agent_id)
+  end
+
+  @doc """
+  Cancels the in-flight run for `agent_id`, keeping the agent process and its
+  conversation state alive.
+
+  Best-effort — returns `{:error, reason}` when nothing is running for the id.
+  """
+  @spec cancel(String.t()) :: :ok | {:error, term()}
+  def cancel(agent_id) do
+    AgentServer.cancel(agent_id)
+  catch
+    # `AgentServer.cancel/1` is a `GenServer.call` on a via-registry name, so it
+    # exits rather than returning an error when:
+    #   - nothing is registered for `agent_id` (`:noproc`) — never started, or already stopped
+    #   - the reply outlives the 5s default (`:timeout`) — the server answers only after a 2s task grace
+    #   - the server dies mid-call — a concurrent `stop/1` parks in `terminate/2` for up to 25s
+    # None of these leaves a run worth reporting on, and the exit signal must not
+    # take the caller down with it.
+    :exit, reason -> {:error, reason}
+  end
 
   defp maybe_refresh_agent(agent_id, maybe_new_agent, refresh_when) do
     with {:ok, current_agent} <- AgentServer.get_agent(agent_id),
--- a/test/support/ai/ai_case.ex
+++ b/test/support/ai/ai_case.ex
@@ -62,15 +62,19 @@ defmodule Trento.AI.AICase do
   end
 
   @doc """
-  Returns the pid of the running agent process for `agent_id`, failing the test
-  when nothing is registered.
+  Returns the pid of the running agent process for `agent_id`, raising when
+  nothing is registered.
 
   The registry is the one piece of sagents with no `Trento.AI` port in front of
-  it, so it is reached here once instead of at every assertion site.
+  it, so it is reached here once instead of at every call site.
   """
   def agent_pid!(agent_id) do
-    assert {:ok, pid} = Sagents.AgentSupervisor.get_pid(agent_id)
+    case Sagents.AgentSupervisor.get_pid(agent_id) do
+      {:ok, pid} ->
+        pid
 
-    pid
+      {:error, reason} ->
+        raise "no agent process registered for #{inspect(agent_id)}: #{inspect(reason)}"
+    end
   end
 end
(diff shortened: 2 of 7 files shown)


lib/trento/ai/agent.ex
```diff
@@ -82,11 +82,39 @@ defmodule Trento.AI.Agent do
   @doc """
   Stops the running agent for `agent_id`, terminating any in-flight run.
 
-  Best-effort — returns the supervisor's result verbatim (`{:error, term()}`
-  when no agent is running for the id).
+  It cancels in-flight runs first, then stops the agent process.
+
+  This is faster than only stopping the agent, because `Sagents.AgentServer.terminate/2`
```
antgamdia, line 87: Question: currently we are doing read only operations, but could this strategy be problematic if we ever consider other actions or tool invocations that might require some cleanup?

Anyway, fine as is right now. We are far from that point hehe

antgamdia (approved):

lib/trento/ai/agent.ex
nelsonkopliku: That's actually a good point. Let me add a comment about this in the code. Thanks!

lib/trento/ai/agent.ex
```diff
@@ -82,11 +82,39 @@ defmodule Trento.AI.Agent do
...
+  catch
+    # `AgentServer.cancel/1` is a `GenServer.call` on a via-registry name, so it
+    # exits rather than returning an error when:
+    #   - nothing is registered for `agent_id` (`:noproc`) — never started, or already stopped
+    #   - the reply outlives the 5s default (`:timeout`) — the server answers only after a 2s task grace
+    #   - the server dies mid-call — a concurrent `stop/1` parks in `terminate/2` for up to 25s
+    # None of these leaves a run worth reporting on, and the exit signal must not
+    # take the caller down with it.
+    :exit, reason -> {:error, reason}
+  end
```
gagandeepb, line 117: Questions:
1. What prevents concurrent/multiple client-side calls to `cancel`? 
2. Regarding 1., should there be some sort of transactional deduplication of the intent to cancel a particular `agent_id`?
3. I think we can get rid of the catch block (which usually should be a rare occurrence as a general thing) with better explicit pre-condition checking (Registry lookups and Process liveness checking) and thinking of possibly addressing 1 and 2.

gagandeepb (commented):
A few questions. I can also talk sync.

lib/trento/ai/agent.ex
gagandeepb: Resolved after talking sync.

test/support/ai/ai_case.ex
```diff
@@ -18,4 +33,44 @@ defmodule Trento.AI.AICase do
...
+  when nothing is registered.
+
+  The registry is the one piece of sagents with no `Trento.AI` port in front of
+  it, so it is reached here once instead of at every assertion site.
+  """
+  def agent_pid!(agent_id) do
+    assert {:ok, pid} = Sagents.AgentSupervisor.get_pid(agent_id)
+
+    pid
+  end
```
gagandeepb, line 79: Why do we `assert` here in this function? Seems contrary to my usual set of expectations about helper functions/code being separate from test code.

test/trento/ai/agent_test.exs
```diff
@@ -284,23 +318,200 @@ defmodule Trento.AI.AgentTest do
...
+        start_running_agent(block_for: :timer.minutes(1))
+
+      server_ref = Process.monitor(pid)
+      task_ref = Process.monitor(task_pid)
+
+      stopping = Task.async(fn -> TrentoAIAgent.stop(agent_id) end)
+
+      # Nobody releases the pending model, so the run can only end by being killed.
+      # `Sagents.AgentServer.terminate/2` waits up to 25s for a running task,
+      # so a `stop` that did not cancel first would wait for 25s.
```
gagandeepb: Can we make s/Nobody/Nothing, to not ascribe human-like agency to computer process things ? 😅

test/trento/ai/agent_test.exs
```diff
@@ -284,23 +318,200 @@ defmodule Trento.AI.AgentTest do
...
+    assert :ok = TrentoAIAgent.run(agent, initial_user_prompt)
+    pid = agent_pid!(agent_id)
+
+    assert_receive {:agent, {:status_changed, :running, nil}}, @integration_timeout
+    assert_receive {:llm_called, task_pid}, @integration_timeout
+
+    on_exit(fn -> TrentoAIAgent.stop(agent_id) end)
+
+    %{agent: agent, agent_id: agent_id, pid: pid, task_pid: task_pid}
   end
```
gagandeepb: Can we please consider breaking apart helper functions from assert oriented tests ?

nelsonkopliku: Addressed here 636242a6c531c9c0b4b51e7a52cd433639e49747

test/support/ai/ai_case.ex
nelsonkopliku: That's fair.
The idea was to assert by default setup helper state and leave to users of the helpers the burden to assert what is most relevant for them.

That said, I have addressed it here 636242a6c531c9c0b4b51e7a52cd433639e49747

test/trento/ai/agent_test.exs
```diff
@@ -284,23 +318,205 @@ defmodule Trento.AI.AgentTest do
...
+        reply: Keyword.get(opts, :reply, defaults.reply)
+    }
+
+    agent = TrentoAIAgent.new!(agent_id: agent_id, model: model, scope: build(:user))
+
+    :ok = TrentoAIAgent.run(agent, Keyword.get(opts, :initial_user_prompt, "hello"))
+    on_exit(fn -> TrentoAIAgent.stop(agent_id) end)
+
+    %{agent: agent, agent_id: agent_id, pid: agent_pid!(agent_id)}
+  end
```
gagandeepb: I wonder if the options/args for this function can be unified, and multiple invocations of the `start_running_agent` can be removed in favor of a single one in the setup block (that is something that runs before every test) ?

gagandeepb (commented):
One minor test code org comment.

test/trento/ai/agent_test.exs
nelsonkopliku: a single invocation of `:start_running_agent` as a setup block can't be unconditional as different tests need different setup.

I went for a middle ground here 5ff5ec6f7a213530bac28878e2ac181a4d973858

gagandeepb (approved):
Thanks for addressing! The tests look much nicer with the last set of changes 🙂

test/trento/ai/agent_test.exs
gagandeepb: thanks!

PR #4649, opened 2026-08-19
Handle cancellation and thread abandonment in channel

nelsonkopliku:
### Description

This PR wires the cancellation/stopping primitives from https://github.com/trento-project/web/pull/4648 in the ai assistant channel.

Now a client attached to the socket can issue:
- `cancel_run` to cancel the ongoing run processing which leaves the agent running and ready to handle other prompts
- `abandon_thread` to stop the agent process (internally it also cancels ongoing runs being processed)

These will have UI counterparts on:
- a stop button in the chatbox triggering a cancellation
- abandonment is wired to "New Chat" button in the chatbox header

Additional notes:
- improvements to `FakeChatModel` so that it can be instructed on responding behaviour
- `LLMBuilder` made a behavior so we can have it return a`FakeChatModel` in relevant tests
- some redundant ai assistant channel tests were collapsed

### How was this tested?

Automated and IRL

### DIFF
--- a/test/support/ai/ai_case.ex
+++ b/test/support/ai/ai_case.ex
@@ -6,7 +6,7 @@ defmodule Trento.AI.AICase do
 
   use ExUnit.CaseTemplate
 
-  alias Trento.AI.ApplicationConfigLoader
+  alias Trento.AI.{ApplicationConfigLoader, FakeChatModel, LLMBuilder}
   alias Trento.Infrastructure.AI.{SagentsAgentServer, SagentsDynamicSupervisor}
 
   # Booting a real sagents process tree and reaching the model call is slower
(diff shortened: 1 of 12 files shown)


test/support/ai/ai_case.ex
```diff
@@ -22,45 +22,68 @@ defmodule Trento.AI.AICase do
...
+    ai = Application.get_env(:trento, :ai)
+
+    Application.put_env(:trento, :ai, Keyword.put(ai, :llm_builder_adapter, LLMBuilder.Mock))
+
+    on_exit(fn -> Application.put_env(:trento, :ai, ai) end)
+
+    Mox.stub(LLMBuilder.Mock, :build_for_user, fn _user_id -> {:ok, model} end)
+
+    :ok
+  end
```
gagandeepb: I wonder when are these functions called? I ask since my concern is that `Application.put_env` introduces mutable state of the global sort in the context of test processes/execution. It  can also lead to state intended for one test process being read by other test processes, leading to non-determinism.

nelsonkopliku: Thanks for raising, the concern is valid.

In this commit ed1e257d8886254b3f4a6c0060a3ae3e341f0ff4 I made sure that we leverage `ApplicationConfigLoader` and its mocks/stubs to avoid global state change of `Application.put_env`.

EDIT: I meant _this_ commit b70301a1fb3c2b75c1443b471b78e239fa373cc4

That should put us in a safe spot.

Additionally made `async` some more tests.

gagandeepb (commented):
Thanks for addressing the last comment. While the newer Mox based approach is much nicer, I wonder if we need to add a [verify_on_exit!](https://mox.hexdocs.pm/Mox.html#verify_on_exit!/1) some places in the test modules, to ensure/verify that all the set up expectations have been called?

nelsonkopliku:
> Thanks for addressing the last comment. While the newer Mox based approach is much nicer, I wonder if we need to add a [verify_on_exit!](https://mox.hexdocs.pm/Mox.html#verify_on_exit!/1) some places in the test modules, to ensure/verify that all the set up expectations have been called?
>

@gagandeepb, `:verify_on_exit!` is where it should be with regards to what related to this PR.

I did a research and found out that `TrentoWeb.AI.ControllerToolTest` is the only place where we miss it.
Added here 6857d5f95d3141b4cadda15989719e1e07b86e3b

gagandeepb (commented):
Question: how are we introducing ag_ui events and related payloads for handling thread cancellation and abandonment? As far as I remember, ag_ui has no primitive for cancellation, but curious to hear your thoughts.

nelsonkopliku:
> Question: how are we introducing ag_ui events and related payloads for handling thread cancellation and abandonment? As far as I remember, ag_ui has no primitive for cancellation, but curious to hear your thoughts.
>
We aren't introducing any indeed. AG-UI does not have such primitives.

Additionally:
- declinating a `RunFinished` to cancelled/abandoned
- Custom/Raw events

do not fit in the feature either. We could send extra fields not contemplated by the protocol, but they would be discarded on the ui receiving side or make it really tricky and cumbersome to make things fit with how the libs actually work.

This is the most balanced approach i could find to make things work in the sanest way.

gagandeepb (approved):

PR #4653, opened 2026-08-19
Modified health and status changed events in activity log

skrech:
### Description

Modified the severity level for the Health and Status changed events. 

Abstracted the common conditions into elixir constants at the top of the file.

### How was this tested?

By CI.

### DIFF
--- a/lib/trento/activity_logging/severity_level.ex
+++ b/lib/trento/activity_logging/severity_level.ex
@@ -4,6 +4,38 @@
 defmodule Trento.ActivityLog.SeverityLevel do
   @moduledoc false
 
+  @health_condition_standard %{
+    type: :kv,
+    key_suffix: "health",
+    values: %{
+      "warning" => :warning,
+      "critical" => :critical,
+      "*" => :info
+    },
+    condition: :map_value_to_severity
+  }
+  @health_condition_warn_if_unknown %{
+    type: :kv,
+    key_suffix: "health",
+    values: %{
+      "warning" => :warning,
+      "critical" => :critical,
+      "unknown" => :warning,
+      "*" => :info
+    },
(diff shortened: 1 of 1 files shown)


lib/trento/activity_logging/severity_level.ex
gagandeepb, line 145: Would it be possible/interesting to map this one as well to a module constant, in the interests of consistency ?

lib/trento/activity_logging/severity_level.ex
```diff
@@ -4,6 +4,38 @@
...
+    type: :kv,
+    key_suffix: "status",
+    values: %{
+      "yellow" => :warning,
+      "red" => :critical,
+      "*" => :info
+    },
+    condition: :map_value_to_severity
+  }
+
```
gagandeepb, line 38: Thanks for this addition, it is good for reducing the repetition that would have happened eventually.

gagandeepb (approved):
Just two implementation-related comments from my side. Would be interesting to hear especially from Alberto's side if it lines up with his expectations.

nelsonkopliku (approved):
Thanks!

lib/trento/activity_logging/severity_level.ex
skrech: Hm... I'm not sure about that, actually. 

The idea was to combine repeatable health/status checks, and give them names so it's easier to grasp when scanning the code. If a "condition" is used only once, maybe it's not necessary to extract it at the top. This often leads to creating specific names/constants for every specific case, which I find as an anti-pattern. 

What do you think?

gagandeepb: That works too. I only wanted to understand if there had been some thinking on this already.

PR #4655, opened 2026-08-20
Use uuid instead of crypto.randomUUID

nelsonkopliku:
### Description

[`crypto.randomUUID`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) is only available in https or localhost.

This breaks testing installation that do not have https.

While a production installation behind https will not have this issue it is a relief for the mentioned dev/test/evaluation cases.

Added a linter rule about not using that.

### How was this tested?

Automated tests and IRL.


antgamdia (approved):
Thanks for the quick fix!!

PR #4664, opened 2026-08-24
Ensure AI configuration changes in relevant test

nelsonkopliku:
### Description

Tweaking uniqueness of data to make sure the test conditions are the ones that should be, removing flakyness.

See [this run](https://github.com/trento-project/web/actions/runs/32509073986) from some days ago.

### DIFF
--- a/test/trento/ai/configurations_test.exs
+++ b/test/trento/ai/configurations_test.exs
@@ -7,7 +7,7 @@ defmodule Trento.Ai.ConfigurationsTest do
 
   alias Trento.Users.User
 
-  alias Trento.AI.{Configurations, UserConfiguration}
+  alias Trento.AI.{Configurations, LLMRegistry, UserConfiguration}
 
   alias Trento.AI.Configurations.Events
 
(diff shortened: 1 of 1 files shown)


test/trento/ai/configurations_test.exs
```diff
@@ -594,10 +594,15 @@ defmodule Trento.Ai.ConfigurationsTest do
     test "should broadcast that the configuration was updated when the provider/model changed" do
       %User{id: user_id} = user = insert(:user)
 
-      provider1 = build(:random_ai_provider)
-      provider2 = build(:random_ai_provider)
-      model1 = build(:random_ai_model, %{provider: provider1})
-      model2 = build(:random_ai_model, %{provider: provider2})
```
nelsonkopliku, line 600: the issue here is that we have 1/3 of probability of having `provider1` and `provider2` being actually the same.

Then there is also chances that `model1` and `model2` are the same.

Since the test is about actually changing the configuration we better make sure the collision does not happen.

gagandeepb (commented):
Thanks for looking into this. Looks fine to me. I [triggered this run to](https://github.com/trento-project/web/actions/runs/32702455442) verify the fix for flakiness.
I will approve after the run completes with nothing unexpected/relevant.

gagandeepb (approved):
Thanks for this. LGTM.

PR #4665, opened 2026-08-27
Liz dynamic sizing

nelsonkopliku:
### Description

This PR makes Liz positioning and sizing dynamic based on the browser/viewport size.

- always opens right above the trigger (currently if the page is small enough Liz is far too far away and the drag handle hidden, making it unusable)
- user size/position is kept within resizing, still within the boundaries of the viewport
- on small enough screens Liz takes the whole screen

Here's just an example of a basic improvement, for further testing use the PR env https://4665.prenv.trento.suse.com/ by playing with resizing the page.

Before
image

After
image

### How was this tested?

Automated and IRL


jagabomb (approved):
Nice! LGTM

PR #4703, opened 2026-09-02
Remove ranch override dependency

skrech:
### Description

Remove the `ranch` dependency from our top--level dependencies. It was there because its version was overriden, but it's no longer needed. 

Additionally, the version is downgraded from 2.2.1 to 2.2.0, since this is the specified requirement from `rabbit_common`.

### How was this tested?

No testing.


nelsonkopliku (approved):
Looks good.

gagandeepb (commented):
Thanks for this. Did you get a chance to look through the changelog/git history of the library to see if there is anything of interest/concern?

skrech:
@gagandeepb you mean the downgrade from 2.2.1 to 2.2.1? I did but there is no changelog... only git commits... I found nothing of intereest, to be honest. But didn't delved deeper than that.

PR #4730, opened 2026-09-03
Get mix deps in API BC check in case of cache miss, CI

skrech:
### Description

CI is failing for _target_ API BC checks when there is a cache miss. This could happen when the current PR is not rebased on the latest commit of the target branch.


nelsonkopliku (approved):
Thanks

PR #4744, opened 2026-09-03
Fix conflicting erlang27 packages

nelsonkopliku:
### Description

With the release of the erlang27 family of packages now we have this conflict.

> have choice for erlang-providers needed by erlang-rebar3: erlang-providers erlang27-providers
>

This change fixes it so we build against the erlang26 family. 

This has to stay like this until we release 3.2.0 targeting erlang27/elixir119.

To be backported into 3.1.6

### How was this tested?

OBS


antgamdia (approved):

skrech (approved):
Great!

PR #4758, opened 2026-09-08
Test email alerting setttings configuration with a test email

arbulu89:
### Description

Add backend code to enable email alerting settings configuration testing sending a test email.

It adds a new `/api/v1/settings/alerting/test` endpoint which tries to send a email to the configured settings. It only works if alerting is enabled and the user has the correct `user:alerting_settings` abilities.

**The api call is "sync" intentionally, so the execution is blocking. Without this, the logic wouldn't make sense. Besides, the email alert is delivered without any `Task.async` usage, so it will last as long as the email delivery lasts. I'm not sure if we should add something like this to add a "custom" timeout to the "test", but so far, I have not found it needed.**

The api usage is recorded in the activity log in successful and failed deliveries. As the errors have different nature, and they are quite few of them, initially at least, I'm storing the plain error returned by `Swoosh`.

We could try to prettify, but Alberto was fine leaving as they are, as they are SMTP errors at the end.
We could have the next errors:

```
{
  "reason": "Test email delivery failed: {:send, {:permanent_failure, ~c\"localhost\", \"530 5.7.0 Authentication required\\r\\n\"}}",
  "result": "failure"
}
```

```
{
  "reason": "Test email delivery failed: {:retries_exceeded, {:network_failure, ~c\"smtp.google.com\", {:error, :timeout}}}",
  "result": "failure"
}
```

```
{
  "reason": "Test email delivery failed: {:retries_exceeded, {:network_failure, ~c\"demo.trento-project.io/\", {:error, :nxdomain}}}",
  "result": "failure"
}
```

```
{
  "reason": "Test email delivery failed: {:retries_exceeded, {:network_failure, ~c\"localhost\", {:error, :econnrefused}}}",
  "result": "failure"
}
```

_______

Before having the frontend published, the mail looks like this:

image

### How was this tested?

Unit test (and I have done some manual testing with Mailpit as well). I will add e2e tests once the frontend part is implemented.

### Documentation changes

Yes. It will need documentation changes.


lib/trento/activity_logging/parser/phoenix_conn_parser.ex
```diff
@@ -149,11 +149,24 @@ defmodule Trento.ActivityLog.Logger.Parser.PhoenixConnParser do
       )
       when action in [
              :saving_alerting_settings,
-             :changing_alerting_setting
+             :changing_alerting_settings
```
arbulu89, line 152: fix:
Fixing a typo. Otherwise the `changing_alerting_settings` action metadata was not being recorded

lib/trento/infrastructure/alerting/alerting.ex
```diff
@@ -190,11 +201,35 @@ defmodule Trento.Infrastructure.Alerting.Alerting do
 
       {:error, reason} ->
         Logger.error(
-          "Failed to send alert notification with subject \"#{subject}\": #{inspect(reason)}",
-          error: reason
+          "Failed to send alert notification with subject \"#{subject}\"",
+          error: inspect(reason)
         )
```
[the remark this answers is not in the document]
arbulu89: The thing is that putting `error: reason` doesn't work, as it must be used as string.
So, we leave only the interpolated string or follow the current approach.

lib/trento_web/controllers/v1/settings_controller.ex
```diff
@@ -352,6 +353,31 @@ defmodule TrentoWeb.V1.SettingsController do
...
+    responses: [
+      ok: "The test email was successfully delivered.",
+      unauthorized: Schema.Unauthorized.response(),
+      forbidden: Schema.Forbidden.response(),
+      not_found: Schema.NotFound.response(),
+      unprocessable_entity:
+        {"The test email could not be delivered.", "application/json", Schema.UnprocessableEntity}
+    ]
+
+  ai_tool :settings_test_alerting_settings, display_text: "Test alerting settings"
```
gagandeepb, line 370: I am not sure about MCP/tool exposure for this feature. Any scenarios where you see this getting triggered/called?

test/trento/infrastructure/alerting/alerting_test.exs
```diff
@@ -189,6 +189,54 @@ defmodule Trento.Infrastructure.Alerting.AlertingTest do
...
+      relay_ip_address = Faker.Internet.ip_v4_address()
+
+      Application.put_env(
+        :trento,
+        :alerting,
+        enabled: true,
+        smtp_server: "smtp://#{relay_ip_address}"
+      )
+
+      Application.put_env(:trento, Trento.Mailer, adapter: Swoosh.Adapters.SMTP)
```
gagandeepb: Putting things in Application env from inside tests that are running with `async: true` will make the tests very flaky/non-deterministic in my understanding 🙈

lib/trento_web/controllers/v1/settings_controller.ex
arbulu89: Good question. I was doubtful as well. I added it because we have the AI tool for the `Test Multi-Linux-Manager connection` route, and this looked a similar thing.
It doesn't really look something to be used in normal cases in any case.

What do you think @abravosuse ? Would testing the email delivery with Liz be something we want?

test/trento/infrastructure/alerting/alerting_test.exs
arbulu89: You are right. This kind of usage is all over the place from before.
Let me have a look to see if I can change it. In reality, our code supports to have this information in the database, which would make everything easier to test.
I will have a look to see if I can change it (maybe I will open other PR).

arbulu89: Hey @gagandeepb ,
At the end the changes were not that complex.
I have added a new commit with:
- Make the test file `sync`. It will always have shared data (env variables, database data, etc). So flakyness would come sooner or later.
- Use database data instead of env variables. At the end, if the env variables data is empty, the code defaults to db data which is way better for testing

I think it looks way better

test/trento/infrastructure/alerting/alerting_test.exs
```diff
@@ -3,7 +3,7 @@
 
 defmodule Trento.Infrastructure.Alerting.AlertingTest do
   @moduledoc false
-  use ExUnit.Case, async: true
+  use ExUnit.Case
```
gagandeepb, line 6: Thanks for addressing this. I understand the default async value is false. (At least for me, explicit is usually better than implicit.)

gagandeepb: We can also use [`group`](https://ex-unit.hexdocs.pm/1.19.6/ExUnit.Case.html) and/or [parameterize](https://ex-unit.hexdocs.pm/1.19.6/ExUnit.Case.html#module-parameterized-tests) options on Elixir 1.18+ to create test concurrency groups that run sync together. Up to you if you want to introduce that now or we introduce this later in a test refactoring PR (across multiple modules wherever needed).

gagandeepb: The test options would allow us to group all the tests with the same alerting config together as a sync group.

gagandeepb (approved):

lib/trento/infrastructure/alerting/emails/email_templates/test_email_body.html.heex
```diff
@@ -0,0 +1,16 @@
...
+<EmailLayout.base_email>
+  <p>Hi there,</p>
+  <p></p>
+  <div>
+    <b>This is a test email sent by Trento.</b>
+  </div>
+  <p>
+    Receiving this message means that Trento is able to deliver email alerts using
+    the configured SMTP settings. No action is required.
+  </p>
```
antgamdia, line 14: suggestion: Since it is a test email, maybe we could include more info that helps user debug their settings? For instance: when it was scheduled to be sent, what is the server, which are the from/to addresses, etc. Sth like below. Not critical, just an idea.

```suggestion
  <p>
    Receiving this message means that Trento is able to deliver email alerts using
    the configured SMTP settings. No action is required.
  </p>
  <div>
    <p>
      Sent at: <b>{@sent_at}</b>
      <br /> SMTP server: <b>{@smtp_server}:{@smtp_port}</b>
      <br /> From: <b>{@sender}</b>
      <br /> To: <b>{@recipient}</b>
    </p>
  </div>
```

lib/trento_web/controllers/v1/settings_controller.ex
antgamdia: Not sure either... at some point we could implement toolsets, so that people can just enable what they need more granularly. For a general use-case, I don't think we need to expose those operations via MCP, but who knows. Not critical; it just consumes some input tokens, but it's not significant.

antgamdia (approved):
Thanks for the implementation! +1ing. We can always improve or add more details in the test email (if needed) in another PR.

test/trento/infrastructure/alerting/alerting_test.exs
arbulu89: Be aware that our code needs to work in elixir 1.15, where this new utilities are not present.

lib/trento/infrastructure/alerting/emails/email_templates/test_email_body.html.heex
arbulu89: Ok! Nothing against it. Anyway, I want @abravosuse feedback on the email body itself + this kind of improvements

test/trento/infrastructure/alerting/alerting_test.exs
gagandeepb: I see, right. We can defer this for until the time we drop support for Elixir 1.15.

lib/trento/infrastructure/alerting/emails/email_templates/test_email_body.html.heex
abravosuse: I am not sure. I think it's already redundant, to be honest. If a message is delivered successfully, the user is aware of all that additional information.

PR #4763, opened 2026-09-09
Add Test email button in the frontend

arbulu89:
### Description

Add `Test email` button to the frontend.

Pretty simple. It uses the backend API call and puts a toast with the result.
The button clicking is blocking, so it can be clicked only once as it is disabled once it is being executed.

image

Depends on: https://github.com/trento-project/web/pull/4758

### How was this tested?

UT

### Documentation changes

Yes


assets/js/pages/SettingsPage/hooks.test.jsx
```diff
@@ -398,8 +398,9 @@ describe('useAlertingSettings', () => {
     it('sets empty settings and activates error flag', async () => {
       axiosMock.onGet('/api/v1/settings/alerting').reply(422);
 
+      const [hookWrapper, _] = hookWrapperWithState();
```
arbulu89, line 401: comment:
we need to update all this tests as adding the `useDispatch` makes the test depend on the redux state

assets/js/lib/model/activityLog.js
```diff
@@ -293,6 +294,14 @@ export const ACTIVITY_TYPES_CONFIG = {
     message: (_entry) => 'Alerting Settings were changed',
     resource: alertingSettingsResourceType,
   },
+  [TESTING_ALERTING_SETTINGS]: {
+    label: 'Test Email Requested',
+    message: ({ metadata }) =>
+      metadata?.reason
+        ? 'Test email delivery failed'
+        : 'Test email delivery succeeded',
```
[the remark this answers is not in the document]
arbulu89: Updated

assets/js/pages/SettingsPage/hooks.test.jsx
```diff
@@ -491,5 +497,74 @@ describe('useAlertingSettings', () => {
...
+      const [hookWrapper, store] = hookWrapperWithState();
+      const { result } = await act(() =>
+        renderHook(() => useAlertingSettings(), { wrapper: hookWrapper })
+      );
+
+      axiosMock.onPost('/api/v1/settings/alerting/test').reply(200, {});
+
+      await act(() => {
+        result.current.sendTestEmail();
+      });
```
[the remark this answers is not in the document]
arbulu89: Updated

antgamdia (approved):
Thanks!

assets/js/lib/api/alertingSettings.js
```diff
@@ -12,3 +12,5 @@ export const saveSettings = (settings) =>
 
 export const updateSettings = (settings) =>
   networkClient.patch(alertingPath, settings);
+
+export const sendTestEmail = () => networkClient.post(`${alertingPath}/test`);
```
antgamdia, line 16: suggestion; perhaps just a personal preference, but I'd rather not hardcode the endpoint directly on the request.

arbulu89: mmm, how would you do it?
We have hardcoded all endpoint urls in all of our frontend api functions

antgamdia: We could do it at different levels. For example, I'd pretty much advocate for a fully API-contract-first approach (like https://github.com/openapistack/openapi-client-axios or many similar projects out there). Ideally, we should not have to write any endpoint to consume our API if we own the API contract. 
IMO, this only leads to typos, version mismatches, errors in types, etc.

That said, any of those changes are completely out of the scope of a simple button PR :P
Maybe just extracting `${alertingPath}/test` into a different constant could make the trick. No real benefit, as it is only used once, but just a personal preference.

arbulu89: Ok, it looks like a more "long-term" improvement.
Let's leave it as it is by now. Moving to other constant simply adds a new constant hehe

PR #4766, opened 2026-09-09
Add e2e tests for Test email feature

arbulu89:
### Description

Add e2e test to `Test email` feature.

It only tests the happy path as our test setup is hardcoded to use "env" based values to configure alerting and we would need to change those to configure incorrect alerting.

Depends on: https://github.com/trento-project/web/pull/4758
Depends on: https://github.com/trento-project/web/pull/4763

### How was this tested?

e2e

### Documentation changes

No


vicenteqa (approved):
lgtm!

PR #4776, opened 2026-09-15
Add yaml to devDependencies

nelsonkopliku:
### Description

Currently builds are failing [in OBS](https://build.opensuse.org/package/live_build_log/devel:sap:trento:factory/trento-web/SLES15-Trento-SP3-SP7/x86_64).

It fails resolving peer dependencies, specifically `yaml` because vite wants `yaml@^2.4.2` while some other dep wants `yaml@1.10.3`.

This PR does two things:
- made yaml an explicit devDependency via `npm i --save-dev yaml`
- regenerates the lock file with npm 10.8.2 otherwise pipeline `npm ci` fails on the legacy version (note that lock file generated with node 10.8.2 is forward compatible while when generated with a newer version is not backward compatible)

Alternatively we could have added to spec file `--legacy-peer-deps` 

```diff
-local-npm-registry %{_sourcedir} install --include=dev --ignore-scripts
+local-npm-registry %{_sourcedir} install --include=dev --ignore-scripts --legacy-peer-deps
```

however it is not recommended in the official docs https://docs.npmjs.com/cli/v11/using-npm/config#legacy-peer-deps and in our case would result in some dependencies not being installed and possibly some other inconsistencies.

That is why I opted to make yaml dep explicit.


arbulu89 (approved):
man, this `yaml` dependency is a pain...
I think already 3rd or 4th time it breaks the dependency chain...

PR #4794, opened 2026-09-18
pin packageManager to npm@10.8.2

nelsonkopliku:
### Description

Dependabot PRs are failing on `npm ci` on node@20.20.0/npm@10.8.2 https://github.com/trento-project/web/actions/runs/35222660087/job/105207024030?pr=4784 because it removes `node_modules/tailwindcss/node_modules/yaml@1.10.3` from lock file. Related to https://github.com/trento-project/web/pull/4776

This PR sets packageManager to npm@10.8.2 via a `npm pkg set packageManager=npm@10.8.2` in `./assets`.

This is going to tell dependabot to use that specific npm version that would generate forward-compatible lock file.
That works because under the hood dependabot uses [`corepack`](https://github.com/nodejs/corepack) to resolve npm.

Caveats and suggestions:
- locally use `npm ci` instead of `npm install` because the latter would use latest npm, break lock file and pipeline
- when in need to actually install a new dependency use either `corepack npm install ...` or `npx npm@10.8.2 install ...`

Unfortunately this extra care is needed until we drop node 20


antgamdia (approved):

nelsonkopliku:
For completeness, the alternative to this PR is manually fix all dependabot PRs 😄

gagandeepb (approved):

PR #4796, opened 2026-09-18
Drop the stale tailwindcss yaml override

nelsonkopliku:
### Description

After looking closer, it turns out that unfortunately `"packageManager": "npm@10.8.2"` does not affect the actual npm version dependabot uses.

However removing the yaml override for tailwind seems to be cleaning up things

`"tailwindcss": "^3.4.19"` in our package json
wants `"postcss-load-config": "^4.0.2 || ^5.0 || ^6.0",` see [here](https://github.com/tailwindlabs/tailwindcss/blob/v3.4.19/package.json#L90)
and we get postcss-load-config version 6.0.1 that has `"yaml": "^2.4.2"` as a peer dep. see [here](https://github.com/postcss/postcss-load-config/blob/6.0.1/package.json#L62)

Running `npx npm@10.8.2 ci` and `npm ci` generates the same identical lock file


assets/package.json
```diff
@@ -114,6 +111,5 @@
     "pretest": "ls ../priv/static/assets/app.css || npm run tailwind:build",
     "test": "TZ=UTC jest",
     "chromatic": "npx chromatic --exit-zero-on-changes"
-  },
-  "packageManager": "npm@10.8.2"
```
antgamdia, line 118: Is it intended?

antgamdia (approved):

assets/package.json
nelsonkopliku: Yes, I added it here https://github.com/trento-project/web/pull/4794, but eventually did not fix the issue with dependabot 😅

PR #4806, opened 2026-09-21
Add persistent filters to clusters and checks execution results views

arbulu89:
### Description

Continue adding persistent filters in views. In this case, `Clusters` view and `Checks results` view.
This last one already had persistent filters implemented with `redux`, but they were not aligned with the query string usage.
I have refactored it to remove the redux part and use the new implementation. With this new logic, we store the filters for the page, not for cluster/host. So if you put a `passing` filter (for example) for one cluster checks executions results and then go to other cluster execution results, it will keep the same filter (@abravosuse agreed with this).
I have done a frontend components small refactor for the `ExecutionResults` component. With the new logic, some components like the `CheckResultsFilter` are not needed.

### How was this tested?

UT and e2e

### Documentation changes

Not now


test/e2e/cypress/pageObject/base_po.js
```diff
@@ -44,7 +44,8 @@ export const addTagButtons = 'span span:contains("Add Tag")';
 const usernameMenu = `span[class="flex items-center"]:contains("${plainUser.username}")`;
 
 // UI Interactions
-export const visit = (url = '/') => cy.visit(url);
+export const visit = (url = '/', params = '', options = {}) =>
```
arbulu89, line 47: comment:
unfortunately we cannot use `cy.visit` `qs` param to add query string values:
https://docs.cypress.io/api/commands/visit#Arguments
We need to be able to set multiple values for the same key (example: `health=passing&health=warning`, but this format is not supported by `cy.visit`, which would always create something like `health=passing%20warning`, which is not valid for our frontend

test/e2e/cypress/e2e/clusters_overview.cy.js
```diff
@@ -62,6 +63,69 @@ context('Clusters Overview', () => {
     });
   });
 
+  describe('Persistent filters', () => {
+    const clusterName = clustersOverviewPage.hanaCluster1.name;
+    const anotherClusterName = 'netweaver_cluster';
+
+    it('should update the URL with filter params when a filter is selected', () => {
+      clustersOverviewPage.selectNameFilter(clusterName);
+      cy.url().should('contain', `name=${clusterName}`);
```
vicenteqa, line 72: nit: I see this was already done this way before, but what about having ``name=${clusterName}`` in some `const` and having a function in `basePO` like:
```
export const urlContains = (expectedText) => cy.url().should('contain', expectedText);
```
also saying this because I saw we use the same pattern in other suites

vicenteqa (commented):
LGTM test-wise 👍

test/e2e/cypress/e2e/clusters_overview.cy.js
arbulu89: I can definitely add the function.
I think adding the const would simply add "empty" additional functions to the PO files though.
Let me finish merging all these as they are by now, as I already merged the `cy.url().should...` in a previous PR, so I can fix them all in a final PR.

nelsonkopliku (approved):

